PR #1, opened 2022-11-08
Remove extra pointer from module list parameter

FranklandJack:
* Remove an extra `*` from the list of modules passed to the `zerProgramCreate` API via the `phModules` parameter. This argument should be a list of modules of size `count` and hence only requires one level of *derived-ness*. This also makes it consistent with the name of the parameter `phModules` i.e. pointer to a module handle, not pointer to a pointer to a module handle.


kbenzie (approved):
LGTM

PR #2, opened 2022-11-08
Create zer_api.h

smaslov-intel:
Want to add a pre-generated version of the API header to not install dependencies and generate it at users build time


FranklandJack:
Should we have removed `include` from the list of directories in `.gitignore`?: https://github.com/oneapi-src/unified-runtime/blob/ebab58ef78a2a4005d4f9c169b21a4557f8cfe7c/.gitignore#L57

zackwaters:
> Should we have removed `include` from the list of directories in `.gitignore`?:
> 
> https://github.com/oneapi-src/unified-runtime/blob/ebab58ef78a2a4005d4f9c169b21a4557f8cfe7c/.gitignore#L57

Thanks. Fixed.

PR #3, opened 2022-11-09
Include stdbool.h in API headers

FranklandJack:
* `ur_api.h` makes use of `bool`. `ur_api.h` is a C header and in C the `bool` type is defined in `stdbool.h`, hence we should include `ur_api.h` in the API header.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -17,6 +17,7 @@
 // standard headers
 #include <stdint.h>
 #include <stddef.h>
+#include <stdbool.h>
 
 #if defined(__cplusplus)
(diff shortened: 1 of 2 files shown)


bashbaug:
Alternatively, perhaps we should rethink whether we use `bool` in the unified runtime.  Many APIs use a different explicitly sized type instead, e.g. `cl_bool`, `VkBool32`, `ze_bool_t`, to avoid potentially different sizes for `bool`.

FranklandJack:
Closing in favor of https://github.com/oneapi-src/unified-runtime/issues/42.

PR #12, opened 2022-11-09
Add pointer to urSelectBinary out param

FranklandJack:
* The `SelectedBinary` out parameter of the `urSelectBinary` was missing a pointer in its type.

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -1796,9 +1796,9 @@ else:
 ###############################################################################
 ## @brief Function-pointer for urDeviceSelectBinary
 if __use_win_types:
-    _urDeviceSelectBinary_t = WINFUNCTYPE( ur_result_t, ur_device_handle_t, c_ulong, POINTER(c_ubyte*), c_ulong )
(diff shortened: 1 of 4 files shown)


AerialMantis:
I think this change makes sense, the same change is in the PI implementation.

Though there are some other differences between the PI implementation and the current UR spec, the PI implementation has the second and third parameters the other way around, and for the array of binaries the PI implementation uses a struct type; `pi_device_binary *` instead of `const uint8_t**`. Maybe the second point is part of a bigger change.

I also don't see an equivalent to the device partition callback in the current PI implementation, maybe this is another gap we need to explore.

kbenzie:
Should we change the argument order to match PI then do you think @AerialMantis?

AerialMantis:
We could either change the order in UR to match PI or change it in PI, I don't have a preference, but it's probably easier to change in the spec.

kbenzie:
Agreed that changing the spec is easier.

FranklandJack:
I've reversed the order of those arguments.

kbenzie (approved):

PR #23, opened 2022-11-15
Rename API namespace to ur/UR

### LINKED ISSUE #6: Fix "zer" naming
The fact that Unified Runtime is also referred to as "level zero runtime" and uses the `zer.*` prefix for entry points and objects in the API is very confusing considering it is not level zero and sits above it in the stack. This naming convention should be replaced with `ur.*`.

kbenzie:
This patch replaces the zer/ZER namespace prefix with ur/UR to match the Unified Runtime naming of the project.

Fixes #6.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.2.0 FATAL_ERROR)
-project(level_zero)
+project(unified_runtime)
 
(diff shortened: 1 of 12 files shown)


FranklandJack (approved):
LGTM.

jatinx:
LGTM.

veselypeta (approved):
LGTM

PR #24, opened 2022-11-16
Update Jinja2 and include sphinx-book-theme

kbenzie:
When running Sphinx on Ubuntu 20.04 the version of Jinja2 in the requirements file generated an exception on the use of the `PosixPath` class. This patch bumps to a version which fixed this bug.

The Sphinx build uses the `sphinx-book-theme` but the package was not in the requirements file.


FranklandJack (approved):
LGTM.

PR #25, opened 2022-11-17
Replace GetReference with Retain

FranklandJack:
* UR uses the API `ur{object}GetReference` to increment the reference
  count on an object and `ur{object}Release` to decrement the reference
  count. This is inconsistent. We should make this consistent. To match
  OpenCL (which uses Retain/Release) replace the
  `ur{object}GetReference` APIs with `ur{object}Retain`.


AerialMantis (approved):
LGTM.

kbenzie (approved):
++++1

Issue #11, opened 2022-11-09
Remove nullptr references


FranklandJack:
The API spec has references to `nullptr` (e.g. in [error conditions](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/parse_specs.py#L667)) which is a C++ specific feature. As a C API Unified Runtime probably shouldn't make reference to `nullptr`, we should replace these references with `NULL`.

PR #26, opened 2022-11-18
Replace nullptr references with NULL

FranklandJack:
Unified Runtime is a C API. There is no `nullptr` in C but the spec and
headers reference the `nullptr` value. Replace these reference with
`NULL` which is the correct null pointer value in C.

Fixes https://github.com/oneapi-src/unified-runtime/issues/11.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -336,8 +336,8 @@ typedef struct _ur_device_partition_property_value_t ur_device_partition_propert
 ///     - ::UR_RESULT_ERROR_UNINITIALIZED
 ///     - ::UR_RESULT_ERROR_DEVICE_LOST
 ///     - ::UR_RESULT_ERROR_INVALID_NULL_POINTER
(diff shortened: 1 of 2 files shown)


FranklandJack:
Fixes https://github.com/oneapi-src/unified-runtime/issues/11.

include/zer_api.h
```diff
@@ -6820,4 +6820,4 @@ typedef struct _zer_callbacks_t
 } // extern "C"
 #endif
 
-#endif // _ZER_API_H
+#endif // _ZER_API_H
```
kbenzie, line 6823: Is this removing the trailing new line?

FranklandJack: It is yes.

kbenzie: Would be nice if the generator script was consistent...

AerialMantis (approved):
LGTM.

kbenzie (approved):

PR #28, opened 2022-11-18
Rename zer_image_type_t to zer_mem_type_t

kbenzie:
`zer_image_type_t` refers to the type a `zer_mem_handle_t` represents and is not specific to an image object (which is represented by `zer_mem_handle_t`). This patch aligns with the OpenCL type this was inspired by.


FranklandJack (approved):
LGTM.

PR #29, opened 2022-11-18
Add dummy files to fix FetchContent issues

pvchupin:
If other project (e.g. DPC++) is trying to fetch unified_runtime repo it fails due to cmake errors. In DPC++ we currently need includes only so actuall content of these files doesn't matter today.

### DIFF
--- a/source/zer_api.cpp
+++ b/source/zer_api.cpp
@@ -1,0 +1,1 @@
+// Placeholder to fix FetchContent issues for projects from outside
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -1,0 +1,1 @@
+# Placeholder to fix FetchContent issues for projects from outside


FranklandJack:
Is there no way to fix this upstream in the DPC++ repo?

If the files just need to exist but don't actually need to have anything in them it seems like this could be a bug in the upstream CMake, or the upstream repo could create these files on the fly as part of the build?

Adding files to this repo to fix the build of upstream repos seems like reversing the dependency relationship to me.

kbenzie:
@pvchupin if you could provide a link to where this is causing issues I'd appreciate that.

The `CMakeLists.txt` as they are now don't really follow best practices for use with `FetchContent` so I've created #30 to address that.

pvchupin:
@FranklandJack, @kbenzie, to reproduce that you need another project (or just have another CMakeLists.txt) to do a FetchContent for this project. Like here: https://github.com/intel/llvm/pull/7293/files#diff-1bf2ed05c0065126d4e2150f0eda4dbd561e5c3625452e96356aadafa1312e6fR9-R20

This needs to be fixed in unified-runtime project, because currently you makefiles are corrupted. In particular there are 2 problems to fix:
* third-party directory is included https://github.com/oneapi-src/unified-runtime/blob/main/CMakeLists.txt#L50 but no CMakeLists.txt in it.
* zer_api.cpp listed in sources https://github.com/oneapi-src/unified-runtime/blob/main/source/CMakeLists.txt#L5 but no such file in repo.

This PR adds missing files (assuming they will appear at some point in the future). Alternatively corresponding lines could be removed.

kbenzie:
Thanks @pvchupin. I think I have a good idea of how properly resolve this moving forwards but think we can merge this now to unblock work on DPC++.

kbenzie (approved):

Issue #16, opened 2022-11-09
Add get-last-error functionality


FranklandJack:
The PI plugin layer has an API for querying the last error from a plugin implementation/adapter e.g. https://github.com/intel/llvm/blob/sycl/sycl/plugins/opencl/pi_opencl.cpp#L87.  To get parity with PI we will need to add something like this to unified runtime.

Initially we could just have an API that matches that in PI i.e.
```
ur_result_t zerGetLastError(char **message);
```
where the returned value is the last reported error code and `message` is an out parameter which is set to point at some adapter specific string containing a detailed message about the context of the error value.

This approach isn't ideal, since the memory returned is owned by the adapter so any caller would need to make a copy of the returned string if they want it to remain consistent, concurrent accesses may also be problematic and in the above declaration there is no way to get the size of the memory without assuming the string is null terminated, so these things would all need to be documented in the spec. In general though this is a tricky problem to solve with a C API.

FranklandJack:
Yeah I think that is a good idea, I didn't want to deviate too much from the `piGetLastError` signature, but I did think about this and it would solve some of the above issues.

So something like:
```
 zer_result_t zerGetLastError(zer_result_t* out_result, char *message, size *message_size);
```
With intended usage like:
```
// some Unified Runtime code which might result in an error...
zer_result_t previous_error = ZER_RESULT_SUCCESS;
size_t message_length = 0;
zer_result_t result_of_call_to_getLastError = zerGetLastError(&previous_error, NULL, &message_length);
if (ZER_RESULT_SUCCESS != result_of_call_to_getLastError) {
 // error out
}

if (ZER_RESULT_SUCCESS != previous_error) {
char * message = malloc(message_length);
  // check allocation succeeded.
  result_of_call_to_getLastError = zerGetLastError(NULL, message, NULL);
  if (ZER_RESULT_SUCCESS != result_of_call_to_getLastError) {
    // error out
  }
  // do stuff with error message.
}
}
```
Is that the sort of thing you had in mind?

In OpenCL APIs that may return a variable sized output typically also take a `param_value_size` argument e.g. clGetDeviceInfo:
```
cl_int clGetDeviceInfo(
    cl_device_id device,
    cl_device_info param_name,
    size_t param_value_size,
    void* param_value,
    size_t* param_value_size_ret);
```
[...]

PR #31, opened 2022-11-21
Add GetLastError API

FranklandJack:
Fixes https://github.com/oneapi-src/unified-runtime/issues/16.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -217,6 +217,8 @@ typedef enum _ur_result_t
     UR_RESULT_INVALID_HOST_PTR = 0x78000022,        ///< Invalid host pointer
     UR_RESULT_INVALID_USM_SIZE = 0x78000023,        ///< Invalid USM size
     UR_RESULT_OBJECT_ALLOCATION_FAILURE = 0x78000024,   ///< Objection allocation failure
+    UR_RESULT_ADAPTER_SPECIFIC = 0x78000025,        ///< An adapter specific warning/error has been reported and can be
+                                                    ///< retrieved via the urGetLastResult entry point.
     UR_RESULT_ERROR_UNKNOWN = 0x7ffffffe,           ///< Unknown or internal error
     UR_RESULT_FORCE_UINT32 = 0x7fffffff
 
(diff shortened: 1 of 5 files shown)


FranklandJack:
I think there is still some discussion to be had over the naming, but following discussions on #16 I've updated the entry point to match the semantics agreed upon there. @AidanBeltonS if you could take a look at this and confirm it matches your expectations that would be greatly appreciated.

include/ur_api.h
```diff
@@ -1961,6 +1963,36 @@ urTearDown(
...
+/// 
+/// @returns
+///     - ::UR_RESULT_SUCCESS
+///     - ::UR_RESULT_ERROR_UNINITIALIZED
+///     - ::UR_RESULT_ERROR_DEVICE_LOST
+///     - ::UR_RESULT_ERROR_INVALID_NULL_POINTER
+///         + `NULL == pMessage`
+UR_APIEXPORT ur_result_t UR_APICALL
+urGetLastResult(
+    const char* pMessage                            ///< [out] optional pointer to a string containing adapter specific result
```
kbenzie, line 1992: In PI this is `char**`, I don't think this will work. Also why is it optional?

include/ur_api.h
```diff
@@ -1961,6 +1963,36 @@ urTearDown(
...
+///     - The string returned via the pMessage is a NULL terminated C style
+///       string.
+///     - The string returned via the pMessage is thread local.
+///     - The entry point will return UR_RESULT_SUCCESS if the result being
+///       reported is to be considered a warning. Any other result code returned
+///       indicates that the adapter specific result is an error.
+///     - The memory in the string returned via the pMessage is owned by the
+///       adapter.
+///     - The application may call this function from simultaneous threads.
+///     - The implementation of this function should be lock-free.
```
kbenzie, line 1982: I think this should talk about using thread local storage explicitly in addition to these general statements about thread safety.

include/ur_api.h
FranklandJack: Added extra `*`. Removed optional.

AidanBeltonS (approved):
LGTM!

scripts/core/misc.yml
```diff
@@ -22,3 +22,29 @@ params:
...
+       reported is to be considered a warning. Any other result code returned
+       indicates that the adapter specific result is an error."
+    -  "The memory in the string returned via the pMessage is owned by the
+       adapter."
+    - "The application may call this function from simultaneous
+       threads."
+    - "The implementation of this function should be lock-free."
+params:
+    - type: const char**
+      name: pMessage
```
kbenzie, line 48: I think this should be `ppMessage` by convention.

kbenzie (approved):

scripts/core/misc.yml
FranklandJack: Ah! Good catch, forgot to update this, fixed now.

PR #37, opened 2022-11-21
Adding function pointer type for C/C++ and Python

jatinx:
Adds infra for generating function pointer typedef for C and Python bindings.

### DIFF
--- a/scripts/YaML.md
+++ b/scripts/YaML.md
@@ -74,14 +74,14 @@ value: "0"
 
 ```cpp
 /// @brief A brief description...
-#define ZE_NAME 0
+#define UR_NAME 0
 ```
 </td>
 <td>
 
 ```python
 ## @brief A brief description...
-ZE_NAME = 0
+UR_NAME = 0
 ```
 </td></tr>
 <tr><td>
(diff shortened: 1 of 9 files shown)


scripts/YaML.md
```diff
@@ -190,6 +190,49 @@ class zer_name_t(c_void_p):
...
+    - type: void*
+      name: pParams
+      desc: "Brief description of param"
+```
+</td>
+<td>
+
+```c
+/// @brief A brief description...
+typedef void(zer_callback_t)(void* pParams);
```
FranklandJack, line 224: Can we use `$x` here and below so we don't have to update the code when we change the `zer` prefix to `ur`? Or will that not work inside this code snippet?

include/zer_api.h
```diff
@@ -86,6 +86,12 @@ extern "C" {
 /// @brief compiler-independent type
 typedef uint8_t zer_bool_t;
 
+///////////////////////////////////////////////////////////////////////////////
+/// @brief callback function with user data
+typedef void (zer_func_callback_t)(
```
FranklandJack, line 91: Should we update `zerModuleCreate` to take this type instead of the `void **` it currently accepts?

scripts/YaML.md
jatinx: It wont work here, this code is a snippet.

include/zer_api.h
jatinx: Done

kbenzie (approved):
LGTM

PR #43, opened 2022-11-22
[ur] Add zerQueueFlush

### LINKED ISSUE #41: Add urQueueFush
PI contains a API function PIQueueFlush which is not yet been added to UnifiedRuntime. We should add this to have feature parity with PI.

https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/pi.h#L1128

There is an equivalent OpenCL function: https://registry.khronos.org/OpenCL/sdk/1.0/docs/man/xhtml/clFlush.html

veselypeta:
Add `zerQueueFlush` to Unified Runtime.

I've based the semantics from OpenCL, I'm not sure if we need something different for other runtimes.

Closes #41

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -1752,6 +1752,13 @@ if __use_win_types:
 else:
     _urQueueFinish_t = CFUNCTYPE( ur_result_t, ur_queue_handle_t )
 
+###############################################################################
+## @brief Function-pointer for urQueueFlush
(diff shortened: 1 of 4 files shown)


AerialMantis (approved):
LGTM

kbenzie (approved):

FranklandJack (approved):

PR #48, opened 2022-11-23
Add memory order/memory scope and more context query enums

jatinx:
This fixes https://github.com/oneapi-src/unified-runtime/issues/75 and https://github.com/oneapi-src/unified-runtime/issues/55


jatinx:
Will raise different PRs for context and device.

Issue #62, opened 2022-11-28
Add urMemGetInfo entry point


FranklandJack:
# Issue

PI supports a [`piMemGetInfo`](https://github.com/intel/llvm/blob/4f7787c8d12df6bfb8381b412bbc187d0b6aca40/sycl/include/sycl/detail/pi.h#L1164) entry point which [has usage](https://github.com/intel/llvm/search?q=piMemGetInfo) in the DPC++ runtime.

Unified Runtime currently doesn't have a `urMemGetInfo` entry point.

# Task

Add a `urMemGetInfo` entry point to the Unified Runtime spec.

PR #64, opened 2022-11-28
[ur] Stop using [in,out] pointers for queries

### LINKED ISSUE #52: Redesign entry points taking in/out parameter
# Issue
Some Unified Runtime entry points are taking in/out parameters as pointers and treating the argument as an in or out parameter depending on the value at the address held in the pointer. Whilst this approach does allow the spec to reduce the number of arguments passed to these APIs by one, the semantics is confusing especially when coming from PI and OpenCL where the equivalent APIs have dedicated out parameters. It is also error prone if the caller accidentally leaves the value uninitialized which could result in indeterminate behavior at runtime.

Note that this pattern seems to have been adopted inconsistently throughout the UR spec, some entry points use it and some use the PI/OpenCL style (see below)

# Example

[`urDeviceGet`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L2673) has the following documentation:
```
urDeviceGet(
    ur_platform_handle_t hPlatform,                 ///< [in] handle of the platform instance
    ur_device_type_t DevicesType,                   ///< [in] the type of the devices.
    uint32_t* pCount,                               ///< [in,out] pointer to the number of devices.
                                                    ///< If count is zero, then the call shall update the value with the total
                                                    ///< number of devices available.
                                                    ///< If count is greater than the number of devices available, then the
                                                    ///< call shall update the value with the correct number of devices available.
    ur_device_handle_t* phDevices                   ///< [out][optional][range(0, *pCount)] array of handle of devices.
                                                    ///< If count is less than the number of devices available, then platform
                                                    ///< shall only retrieve that number of devices.
    );
```
where the `pCount` is treated as an out parameter if the value at that address is zero and otherwise is the size of the value the array that will be passed via `phDevices` parameter.

Compare this to the [`piDevicesGet`](https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/pi.h#L989) entry point:
```
__SYCL_EXPORT pi_result piDevicesGet(pi_platform platform,
                                     pi_device_type device_type,
                                     pi_uint32 num_entries, pi_device *devices,
                                     pi_uint32 *num_devices);
```
This has dedicated `num_entries` representing the size of the array passed in `devices` and a `num_devices` parameter which will return the number of devices in the platform of type `device_type` to the caller.

This is also the style used in OpenCL with [`clGetDeviceIDs`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clGetDeviceIDs):
```
cl_int clGetDeviceIDs(
    cl_platform_id platform,
    cl_device_type device_type,
    cl_uint num_entries,
    cl_device_id* devices,
    cl_uint* num_devices);

```

Other APIs such as [`urPlatformGetInfo`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3696) unconditionally treat parameters as in/out updating the value at the address passed after first reading it:
```
urPlatformGetInfo(
    ur_platform_handle_t hPlatform,                 ///< [in] handle of the platform
    ur_platform_info_t PlatformInfoType,            ///< [in] type of the info to retrieve
    size_t* pSize,                                  ///< [in,out] pointer to the number of bytes needed to return info queried.
                                                    ///< the call shall update it with the real number of bytes needed to
                                                    ///< return the info
    void* pPlatformInfo                             ///< [out][optional] array of bytes holding the info.
                                                    ///< if *pSize is not equal to the real number of bytes needed to return
                                                    ///< the info then the ::UR_RESULT_ERROR_INVALID_SIZE error is returned and
                                                    ///< pPlatformInfo is not used.
    );
```
In this case the value at `pSize` is unconditionally updated by the entry point however its value as an in parameter is only used in the case that `pPlatformInfo` is non-null.

# Task 

Replace the in/out parameters with dedicated in and out parameters for the following entry points:
* [`urPlatformGet`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3645)
* [`urPlatformGetInfo`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3696)
* [`urDeviceGet`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L2673)
* [`urDeviceGetInfo`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L2822)
* [`urDevicePartition`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L2931)
* [`urProgramGetInfo`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3980)
* [`urKernelGetInfo`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3219)
* [`urKernelGetGroupInfo`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3250)
* [`urKernelGetSubGroupInfo`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3273)

veselypeta:
Resolve #52

### DIFF
--- a/scripts/core/context.yml
+++ b/scripts/core/context.yml
@@ -106,16 +106,20 @@ params:
     - type: $x_context_info_t
       name: ContextInfoType
       desc: "[in] type of the info to retrieve"
-    - type: "size_t*"
-      name: pSize
+    - type: "size_t"
+      name: propSize
       desc: |
-            [in,out] pointer to the number of bytes needed to return info queried.
-            the call shall update it with the real number of bytes needed to return the info
+            [in] the number of bytes of memory pointed to by pContextInfo.
     - type: "void*"
       name: pContextInfo
       desc: |
             [out][optional] array of bytes holding the info.
-            if *pSize is not equal to the real number of bytes needed to return the info then the $X_RESULT_ERROR_INVALID_SIZE error is returned and pContextInfo is not used.
+            if propSize is not equal to or greater than the real number of bytes needed to return 
+            the info then the $X_RESULT_ERROR_INVALID_SIZE error is returned and pContextInfo is not used.
+    - type: "size_t*"
+      name: pPropSizeRet
+      desc: |
+            [out][optional] pointer to the actual size in bytes of data queried by ContextInfoType.
 --- #--------------------------------------------------------------------------
 type: function
 desc: "Return platform native context handle."
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -50,20 +50,24 @@ params:
       name: hPlatform
       desc: "[in] handle of the platform instance"
     - type: "$x_device_type_t"
-      name: DevicesType
+      name: DeviceType
       desc: |
             [in] the type of the devices.
-    - type: "uint32_t*"
-      name: pCount
+    - type: "uint32_t"
+      name: NumEntries
       desc: |
-            [in,out] pointer to the number of devices.
-            If count is zero, then the call shall update the value with the total number of devices available.
-            If count is greater than the number of devices available, then the call shall update the value with the correct number of devices available.
+            [in] the number of devices to be added to phDevices.
+            If phDevices in not NULL then NumEntries should be greater than zero.
     - type: "$x_device_handle_t*"
       name: phDevices
       desc: |
-            [out][optional][range(0, *pCount)] array of handle of devices.
-            If count is less than the number of devices available, then platform shall only retrieve that number of devices.
+            [out][optional][range(0, NumEntries)] array of handle of devices.
+            If NumEntries is less than the number of devices available, then platform shall only retrieve that number of devices.
+    - type: "uint32_t*"
+      name: pNumDevices
+      desc: |
+            [out][optional] pointer to the number of devices.
+            pNumDevices will be updated with the total number of devices available.
 --- #--------------------------------------------------------------------------
 type: enum
 desc: "Supported device info"
@@ -277,17 +281,20 @@ params:
     - type: $x_device_info_t
       name: infoType
       desc: "[in] type of the info to retrieve"
-    - type: "size_t*"
-      name: pSize
+    - type: "size_t"
+      name: propSize
       desc: |
-            [in,out] pointer to the number of bytes needed to return info queried.
-            The call shall update it with the real number of bytes needed to return the info
+            [in] the number of bytes pointed to by pDeviceInfo.
     - type: "void*"
       name: pDeviceInfo
       desc: |
             [out][optional] array of bytes holding the info.
-            If *pSize input is not 0 and not equal to the real number of bytes needed to return the info
+            If propSize is not equal to or greater than the real number of bytes needed to return the info
             then the $X_RESULT_ERROR_INVALID_SIZE error is returned and pDeviceInfo is not used.
+    - type: "size_t*"
+      name: pPropSizeRet
+      desc: |
+            [out][optional] pointer to the actual size in bytes of the queried infoType.
 --- #--------------------------------------------------------------------------
 type: function
 desc: "Makes a reference of the device handle indicating it's in use until paired $xDeviceRelease is called"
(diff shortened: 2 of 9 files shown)


kbenzie (changes requested):
Relates to discussion in #62.

Restore the use of `size_t` for the get info entry points as they relate to the user allocating memory where `size_t` makes sense.

scripts/core/context.yml
```diff
@@ -106,16 +106,19 @@ params:
...
       desc: |
-            [in,out] pointer to the number of bytes needed to return info queried.
-            the call shall update it with the real number of bytes needed to return the info
+            [in] the number of bytes of memory pointed to by pContextInfo.
     - type: "void*"
       name: pContextInfo
       desc: |
             [out][optional] array of bytes holding the info.
-            if *pSize is not equal to the real number of bytes needed to return the info then the $X_RESULT_ERROR_INVALID_SIZE error is returned and pContextInfo is not used.
+            if Size is not equal to the real number of bytes needed to return the info then the $X_RESULT_ERROR_INVALID_SIZE error is returned and pContextInfo is not used.
```
FranklandJack, line 117: Is this to restrictive? Couldn't size be greater than the value returned via `pSizeRet`? In OpenCL the wording is something like:
```
... returns `CL_INVALID_VALUE` if size in bytes specified by `param_value_size` is < size of return type.
```
The entry point is going to write `*pSizeRet` bytes into `pContextInfo` so as long as the memory at `pContextInfo` is at least `*pSizeRet` bytes we should be good right?

scripts/core/device.yml
```diff
@@ -277,17 +281,20 @@ params:
...
-            [in,out] pointer to the number of bytes needed to return info queried.
-            The call shall update it with the real number of bytes needed to return the info
+            [in] the number of bytes pointed to by pDeviceInfo.
     - type: "void*"
       name: pDeviceInfo
       desc: |
-            [out][optional] array of bytes holding the info.
-            If *pSize input is not 0 and not equal to the real number of bytes needed to return the info
+            [out][optional][range(0, Size)] array of bytes holding the info.
+            If Size is not equal to the real number of bytes needed to return the info
```
FranklandJack, line 292: Same comment here as above.

scripts/core/kernel.yml
```diff
@@ -128,18 +128,20 @@ params:
...
-            the call shall update the value with actual number of bytes of the Kernel property.            
+            [in] the size of the Kernel property value.           
     - type: "void*"
       name: propValue
       desc: |
-            [in,out][optional][range(0, *propSize)] value of the Kernel property.
-            If *propSize is less than the number of bytes for the Kernel property,
-            only the first *propSize bytes will be returned.
+            [in,out][optional][range(0, propSize)] value of the Kernel property.
+            If propSize is less than the number of bytes for the Kernel property,
```
FranklandJack, line 139: Should we align on naming here? In some places input size parameters are named `Size` but then in entry points like this one we have `propSize`.

FranklandJack: This behaviour is also slightly different to other entry points which will fail if there isn't enough memory in `propValue` to hold the queried property.

scripts/core/program.yml
```diff
@@ -169,18 +169,20 @@ params:
...
-            the call shall update the value with actual number of bytes of the Program property.
+            [in] the size of the Program property.
     - type: "void*"
       name: propValue
       desc: |
-            [in,out][optional][range(0, *propSize)] value of the Program property.
-            If *propSize is less than the number of bytes for the Program property,
-            only the first *propSize bytes will be returned.
+            [in,out][optional][range(0, propSize)] value of the Program property.
+            If propSize is less than the number of bytes for the Program property,
```
FranklandJack, line 180: Here too, this behaviour differs to that of other info queries e.g. https://github.com/oneapi-src/unified-runtime/pull/64/files#diff-d87d905cff6868fb3e0d3aadb38e002f331e481878eaa7ffae36bcfb468ada93R89 which fails if there isn't enough memory, rather than returning `min(propSize, pPropSizeRet)` bytes.

scripts/core/kernel.yml
```diff
@@ -158,12 +160,15 @@ params:
     - type: $x_kernel_group_info_t
       name: propName
       desc: "[in] name of the work Group property to query"
-    - type: "size_t"
+    - type: "uint32_t"
```
FranklandJack, line 163: Should this be `uint32_t` or `size_t`, it's a memory size in bytes right?

scripts/core/kernel.yml
```diff
@@ -158,12 +160,15 @@ params:
...
       name: propSize
       desc: "[in] size of the Kernel Work Group property value"
     - type: "void*"
       name: propValue
-      desc: "[in,out][range(0, propSize)] value of the Kernel Work Group property."
+      desc: "[in,out][optional][range(0, propSize)] value of the Kernel Work Group property."
+    - type: "uint32_t*"
```
FranklandJack, line 169: Same here, should these be `uint_32_t` or `size_t`?

FranklandJack (approved):
LGTM % a few comments.

scripts/core/context.yml
veselypeta: 👍

scripts/core/device.yml
veselypeta: 👍

scripts/core/kernel.yml
veselypeta: I think I prefer `propSize` over just `Size`. I'll update if no objections.

As you your second point. I think I'll update this to be like openCL, fail if `!size>=paramValue`

scripts/core/kernel.yml
veselypeta: 👍

scripts/core/kernel.yml
veselypeta: 👍

scripts/core/program.yml
veselypeta: I think you're right 👍

kbenzie (approved):
LGTM

PR #65, opened 2022-11-29
Add range to phModules

kswiecicki:
This fixes a regression introduced in https://github.com/oneapi-src/unified-runtime/commit/46a649a448e3d7097662633c5d11e5086f0cd92a that broke loader scripts, range parameter is now required to properly parse phModules.


pbalcer (approved):

FranklandJack:
Thanks for the contribution :grin: I think you will need to regenerate the headers so they are up to date with your changes in order to get CI to pass.

kswiecicki:
> Thanks for the contribution 😁 I think you will need to regenerate the headers so they are up to date with your changes in order to get CI to pass.

Done, thanks for the advice.

FranklandJack (approved):

PR #66, opened 2022-11-29
[ur] Values for sizeof allocations should use size_t

### LINKED ISSUE #61: Entry points should not use `size_t` parameters.
Currently the UR spec uses a mix of `size_t` and `uint32_t` params.

For example `urDevicePartition` uses `uint32_t` https://spec.oneapi.io/unified-runtime/latest/core/api.html#zerdevicegetinfo.
while, `urDeviceInfo` uses `size_t` https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/device.yml#L280

We should make make these all `uint32_t` as `size_t` is platform dependent.

veselypeta:
Closes #61

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -1121,9 +1121,9 @@ else:
 ###############################################################################
 ## @brief Function-pointer for urProgramCreateWithBinary
 if __use_win_types:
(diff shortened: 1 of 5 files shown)


FranklandJack (approved):
LGTM.

kbenzie (approved):
LGTM

Issue #59, opened 2022-11-28
Add piEventSetCallback entry point


FranklandJack:
# Issue

PI supports a [`piEventSetCallback`](https://github.com/intel/llvm/blob/4f7787c8d12df6bfb8381b412bbc187d0b6aca40/sycl/include/sycl/detail/pi.h#L1431) entry point. Despite the fact this entry point has [no uses](https://github.com/intel/llvm/search?q=piEventSetCallback) in the DPC++ runtime it is quite a useful API for compute runtimes to support allowing the callback to trigger work when a command or set of commands completes.

Unified Runtime currently has no `urEventSetCallback` entry point.

# Task

Add `urEventSetCallback` to the Unified Runtime spec.

kbenzie:
@jatinx is looking at this.

PR #70, opened 2022-11-29
Add urEventSetCallback

jatinx:
Mimics piEventSetCallback.

Adds enum which adds event commands execution status from: https://registry.khronos.org/OpenCL/specs/2.2/html/OpenCL_API.html#event-objects

Fixes: https://github.com/oneapi-src/unified-runtime/issues/59

Also fixes a minor bug which adds extra space at end of python function pointer types


jatinx:
Also fixed python generator, which now uses param names instead of generated names like `var1`, `var2`etc.

kbenzie (approved):
LGTM

FranklandJack (approved):
LGTM.

PR #72, opened 2022-11-30
Don't use reserved identifiers

### LINKED ISSUE #18: Stop using _<name>  reserved identifiers
Unified Runtime appears to have followed the example of OpenCL and used `_<name>` in typedefs of the opaque API objects e.g. [`_ur_module_handle_t`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L111).

This is actually illegal in both the C and C++ specifications as all identifiers with the _  prefix are reserved for standard library implementation and internals.

We should rename the structs in this pattern.

kbenzie:
Update the mako templates to stop using identifiers starting with `_` which are reserved for standard library use in both C and C++.

Fixes #18

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -997,7 +997,7 @@ else:
 
 ###############################################################################
 ## @brief Table of Platform functions pointers
-class _ur_platform_dditable_t(Structure):
+class ur_platform_dditable_t(Structure):
     _fields_ = [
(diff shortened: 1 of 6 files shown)


jatinx:
LGTM

veselypeta (approved):

PR #78, opened 2022-12-01
[ur] Add ur_mem_info_t enum

### LINKED ISSUE #77: Missing _ur_mem_info_t enumerations.
Currently in UR we have an enumeration [`_ur_mem_info_t`](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/usm.yml#L22) However this relates to USM allocation. We should rename this to `_ur_mem_alloc_info_t`, which will be used in the corresponding query [`urMemGetMemAllocInfo`](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/usm.yml#L142). Similarly to how it is named in PI [`_pi_mem_alloc_info`](https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/pi.h#L1631)

For the `urMemGetInfo` entry point we should define `_ur_mem_info_t` enumeration in order to complete(#62). Equivalent in [PI](https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/pi.h#L938).

PI only has two enumerations:
* PI_MEM_CONTEXT
* PI_MEM_SIZE

OpenCL has several more [options](https://registry.khronos.org/OpenCL/sdk/1.0/docs/man/xhtml/clGetMemObjectInfo.html). TBD - how many of these we should add.

Both queries appear to be needed as they are both used by the DPC++ runtime.
piGetMemInfo - https://github.com/intel/llvm/blob/sycl/sycl/source/detail/sycl_mem_obj_t.cpp/#L49
piextUSMGetMemAllocInfo - https://github.com/intel/llvm/blob/sycl/sycl/source/detail/usm/usm_impl.cpp/#L534

veselypeta:
Closes #77

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -1474,6 +1474,13 @@ else:
     _urMemCreateWithNativeHandle_t = CFUNCTYPE( ur_result_t, ur_platform_handle_t, ur_native_handle_t, POINTER(ur_mem_handle_t) )
 
 ###############################################################################
+## @brief Function-pointer for urMemGetInfo
+if __use_win_types:
+    _urMemGetInfo_t = WINFUNCTYPE( ur_result_t, ur_mem_handle_t, ur_mem_info_t, c_size_t, c_void_p, POINTER(c_size_t) )
+else:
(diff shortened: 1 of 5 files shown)


FranklandJack (approved):
LGTM.

kbenzie (approved):
LGTM

PR #79, opened 2022-12-01
cmake: Support install, find_package, & FetchContent

### LINKED ISSUE #30: Update CMakeLists.txt to better support FetchContent
As noted in #29 downstream projects want to use `FetchContent` to pull in the unified-runtime into their builds. Currently the `CMakeLists.txt` do not follow best practices for use with `FetchContent`, e.g. no targets are export.

We should properly support `FetchContent`, and include usage examples in `README.md`. The main use case, for this initial support, should be to provide a target which only provides the include path for the header files with all other parts of the build disabled.

kbenzie:
This patch adds files to the `install` target along with CMake package configuration files. The goal of these changes is to properly support `FetchContent`, via `find_package`, to ease the integration of unified-runtime into other projects.

The `install` target will now install the following files to the directory specified by the `CMAKE_INSTALL_PREFIX` option.

```
.
├── include
│   ├── ur_api.h
│   ├── ur_ddi.h
│   └── ur.py
└── lib
    └── cmake
        └── unified-runtime
            ├── unified-runtime-config.cmake
            ├── unified-runtime-config-version.cmake
            └── unified-runtime-targets.cmake
```

It is now possible to use `find_package` with this install directory in a dependent projects `CMakeLists.txt`, as follows:

```cmake
find_package(unified-runtime 0.5.0 REQUIRED)

add_executable(example example.cpp)
target_link_libraries(example PUBLIC unified-runtime::headers)
```

> Note: The `CMAKE_PREFIX_PATH` must point to the install directory when
> `CAMKE_INSTALL_PREFIX` is not on a global search path.

The README has been updated to include a new *Integration* section which utilized `FetchContent`, repeated here for brevity.

```cmake
include(FetchContent)

FetchContent_Declare(
    unified-runtime
    GIT_REPOSITORY https://github.com/oneapi-src/unified-runtime.git
    GIT_TAG main  # This will pull the latest changes from the main branch.
)
FetchContent_MakeAvailable(unified-runtime)

add_executable(example example.cpp)
target_link_libraries(example PUBLIC unified-runtime::headers)
```

Fixes #30

### DIFF
--- a/source/ur_api.cpp
+++ b/source/ur_api.cpp
@@ -1,0 +1,3595 @@
+/*
+ *
+ * Copyright (C) 2020 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * @file ur_api.cpp
+ * @version v0.5-r0.5
+ *
+ */
+#include "ur_api.h"
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Creates a context with the given devices.
+/// 
+/// @details
+///     - All devices should be from the same platform.
+///     - Context is used for resource sharing between all the devices
+///       associated with it.
+///     - Context also serves for resource isolation such that resources do not
+///       cross context boundaries.
+///     - The returned context is a reference and must be released with a
+///       subsequent call to ::urContextRelease.
+///     - The application may call this function from simultaneous threads.
+///     - The implementation of this function must be thread-safe.
+/// 
+/// @remarks
+///   _Analogues_
+///     - **clCreateContext**
+/// 
+/// @returns
+///     - ::UR_RESULT_SUCCESS
+///     - ::UR_RESULT_ERROR_UNINITIALIZED
+///     - ::UR_RESULT_ERROR_DEVICE_LOST
+///     - ::UR_RESULT_ERROR_INVALID_NULL_POINTER
+///         + `NULL == phDevices`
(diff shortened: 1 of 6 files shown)


kbenzie:
This also removes the `source` directory from `.gitignore` and addes the `source/ur_api.cpp` test file which was previously accidentally dropped.

source/ur_api.cpp
```diff
@@ -0,0 +1,3560 @@
...
+ *
+ * Copyright (C) 2020 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * @file ur_api.cpp
+ * @version v0.5-r0.5
+ *
+ */
+#include "ur_api.h"
```
pbalcer, line 11: I'm assuming this is here just to make sure the headers and function definitions compile? If so, I propose we delete it since we will soon be adding a null adapter implementation that also serves that purpose and is a better template for actual adapter implementations.

kbenzie: Sounds reasonable, yeah. I only noticed it was missing when testing the `FetchContent` support and the CMake for this file is unconditional. The `null` adapter sounds like a more robust way of testing the header compiles 👍

FranklandJack (approved):
LGTM.

veselypeta (approved):

PR #84, opened 2022-12-05
add platform argument to GetLastResult

pbalcer:
This makes it possible to have a simple passthrough implementation for this method in the loader. Otherwise we'd have to have a copy of the last error in the loader itself.


kbenzie (approved):
LGTM

PR #86, opened 2022-12-05
Fix for fptr_typedef in function pointer generation on python

jatinx:
Generate a `c_void_p` instead of actual function definition on python side.

This is done because python for `CFUNCTYPE` expects a pythonic class present in ctypes. There is none for a custom function pointer.


scripts/templates/helper.py
```diff
@@ -695,6 +695,14 @@ def make_param_lines(namespace, tags, obj, py=False, decl=False, meta=None, form
         name = _get_param_name(namespace, tags, item)
         if py:
             tname = get_ctype_name(namespace, tags, item)
+            # Handle fptr_typedef
+            if "fptr_typedef" in meta:
+                fptr_types = list(meta['fptr_typedef'].keys())
+                fptr_names = list()
```
FranklandJack, line 701:
```suggestion
                fptr_names = []
```

scripts/templates/helper.py
```diff
@@ -695,6 +695,14 @@ def make_param_lines(namespace, tags, obj, py=False, decl=False, meta=None, form
...
+                for fptr_type in fptr_types:
+                    fptr_names.append(subt(namespace, tags, fptr_type))
```
FranklandJack, line 703: Do we really need to build a list here? Can't we just do the substitution then check if the result is `tname` and if so reassign `tname` to `c_void_p` then execute the loop e.g.

```
for fptr_type in fptr_types:
  if tname == subt(namespace, tags, fptr_type):
    tname == "c_void_p"
    break
```

scripts/templates/helper.py
```diff
@@ -695,6 +695,14 @@ def make_param_lines(namespace, tags, obj, py=False, decl=False, meta=None, form
         name = _get_param_name(namespace, tags, item)
         if py:
             tname = get_ctype_name(namespace, tags, item)
+            # Handle fptr_typedef
+            if "fptr_typedef" in meta:
+                fptr_types = list(meta['fptr_typedef'].keys())
```
FranklandJack, line 700: Could you clarify what is happening here.

For each parameter `item` we construct a list of all function pointer types that the function takes as arguments, we then make the substitution to the c typename for each of these function pointer arguments, then we check if the substituted typename for `item` is equal to any of those substituted function pointer types and if it is we set it to `c_void_p`?

scripts/templates/helper.py
jatinx: Done

scripts/templates/helper.py
```diff
@@ -695,6 +695,16 @@ def make_param_lines(namespace, tags, obj, py=False, decl=False, meta=None, form
         name = _get_param_name(namespace, tags, item)
         if py:
             tname = get_ctype_name(namespace, tags, item)
+            # Handle fptr_typedef
+            # On Python side, passing a function pointer to a CFUNCTYPE is a bit awkward
+            # So solve this, if we encounter a function pointer type, we relpace it with
+            # c_void_p - a generic void pointer
+            if "fptr_typedef" in meta:
+                fptr_types = list(meta['fptr_typedef'].keys())
```
FranklandJack, line 703: Couldn't we construct this list outside the for loop this is inside of?

jatinx: done

FranklandJack (approved):
LGTM.

PR #92, opened 2022-12-07
Fix `urPlatformGet` & `urDeviceGet` to return `UR_RESULT_ERROR_INVALID_SIZE`

### LINKED ISSUE #85: `urPlatformGet` & `urDeviceGet` should return valid error code.
`urPlatformGet` currently says that "If phPlatforms is not NULL, then NumEntries but be greater than zero.", However we do not specify what should be returned in this case.

We should specify that `urPlatformGet` and `urDeviceGet` will return `UR_RESULT_ERROR_INVALID_SIZE` or something similar.

veselypeta:
Closes #85

### DIFF
--- a/scripts/core/platform.yml
+++ b/scripts/core/platform.yml
@@ -26,7 +26,8 @@ params:
       name: NumEntries
       desc: |
             [in] the number of platforms to be added to phPlatforms. 
-            If phPlatforms is not NULL, then NumEntries but be greater than zero.
+            If phPlatforms is not NULL, then NumEntries should be greater than zero, otherwise $X_RESULT_ERROR_INVALID_SIZE,
+            will be returned.
     - type: "$x_platform_handle_t*"
       name: phPlatforms
       desc: |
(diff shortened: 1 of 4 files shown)


scripts/core/platform.yml
```diff
@@ -26,7 +26,8 @@ params:
       name: NumEntries
       desc: |
             [in] the number of platforms to be added to phPlatforms. 
-            If phPlatforms is not NULL, then NumEntries but be greater than zero.
+            If phPlatforms is not NULL, then NumEntries but be greater than zero, otherwise $X_RESULT_ERROR_INVALID_SIZE,
```
FranklandJack, line 29:
```suggestion
            If phPlatforms is not NULL, then NumEntries should be greater than zero, otherwise $X_RESULT_ERROR_INVALID_SIZE,
```

FranklandJack (approved):
LGTM % one comment.

kbenzie (approved):
LGTM

PR #93, opened 2022-12-07
add UR loader, null adapter and basic example

pbalcer:
The components this patch implements are mostly derived from the ones implemented by level-zero.

This patch otherwise makes only the minimal amount of changes required for loader:
 - adds new CI that builds the loader
 - updates cmake scripts to build and install the loader
 - adds generate and check-generated custom cmake targets
 - adds a tiny hello-world example to verify loader functionality

Co-authored-by: Brandon Yates <brandon.yates@intel.com>
Co-authored-by: Lisanna Dettwyler <lisanna.dettwyler@intel.com>
Co-authored-by: Patryk Kaminski <patryk.kaminski@intel.com>
Co-authored-by: Krzysztof Swiecicki <krzysztof.swiecicki@intel.com>


.github/workflows/cmake.yml
```diff
@@ -0,0 +1,32 @@
...
+      run: pip install -r third_party/requirements.txt
+
+    - name: Configure CMake
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}}
+
+    - name: Generate source from spec
+      run: cmake --build ${{github.workspace}}/build --target check-generated --config ${{env.BUILD_TYPE}}
+
+    - name: Build
+      run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}}
```
PatKamin, line 32: For me this --config param doesn't change anything. What is relevant here for setting up debug or release build is the -DCMAKE_BUILD_TYPE argument which you added in the previous step.

.gitignore
```diff
@@ -53,6 +53,9 @@
 # Debug files
 scripts/**/*.json
 
+# Python cache
+scripts/templates/__pycache__/
```
PatKamin, line 57: This dir can be generated in `scripts/`, too. Maybe generalize this to sth like `scripts/**/__pycache__/`?

README.md
```diff
@@ -48,22 +52,31 @@ Tools can be acquired via instructions in [third_party](/third_party/README.md).
...
-make
+$ mkdir build
+$ cd build
+$ cmake {path_to_source_dir}
+$ make
+~~~~
+
+**General**:
+
+If you've made modifications to the specification, you can also run a custom `generate` target prior to building.
```
PatKamin, line 77: I think it's better to move this section to be one of the first ones, so this will be something that the user will try using first instead of doing this the hard way (by figuring out which script should be run and with which params, see Source Code Generation section and the linked README).

.github/workflows/cmake.yml
kbenzie: I believe `--config` is important when using the Visual Studio project generator. This will be defaulting to the Unix Makefiles generation though.

pbalcer:
No rush - I know it's a huge PR, but I think it's the bare minimum that makes sense. Thankfully most of it is automatically generated...

.github/workflows/cmake.yml
pbalcer: I used a template :) I'll remove that config part. Windows is probably going to need its own separate job anyway.

source/common/ur_singleton.h
```diff
@@ -0,0 +1,73 @@
+/*
+ *
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+#pragma once
```
FranklandJack, line 8: Although this is probably supported by almost any compiler, I think it is still non-standard, so it might be better to favour a traditional `#ifndef` guard.

source/common/ur_singleton.h
```diff
@@ -0,0 +1,73 @@
...
+#include <memory>
+#include <unordered_map>
+#include <mutex>
+
+//////////////////////////////////////////////////////////////////////////
+/// a abstract factory for creation of singleton objects
+template<typename _singleton_t, typename _key_t>
```
FranklandJack, line 15: I think prefixing types with `_` is UB since they are reserved by the compiler.

source/common/ur_util.h
```diff
@@ -0,0 +1,112 @@
+/*
+ *
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#if defined(__cplusplus)
+#pragma once
```
FranklandJack, line 10: Here too.

source/common/ur_util.h
```diff
@@ -0,0 +1,112 @@
...
+    char buffer[8];
+    auto rc = GetEnvironmentVariable(name, buffer, 8);
+    if (0 != rc && rc <= 8) {
+        env = buffer;
+    }
+#else
+    env = getenv(name);
+#endif
+
+    if( ( nullptr == env ) || ( 0 == strcmp( "0", env ) ) )
```
FranklandJack, line 109: I don't think `( 0 == strcmp( "0", env ) )` is neccesarry here since we do `return ( 0 == strcmp( "1", env ) );` on the line below which will return false if env is 0.

source/common/ur_singleton.h
```diff
@@ -0,0 +1,73 @@
...
+    /// if no instance exists, then creates a new instance
+    /// the params are forwarded to the ctor of the singleton
+    /// the first parameter must be the unique identifier of the instance
+    template<typename... Ts>
+    _singleton_t* getInstance( Ts&&... _params )
+    {
+        auto key = getKey( std::forward<Ts>( _params )... );
+
+        if(key == 0) // No zero keys allowed in map
+            return static_cast<_singleton_t*>(0);
```
FranklandJack, line 53: My preference would be to always have `{}` in any control flow statements.

source/drivers/null/ur_null.h
```diff
@@ -0,0 +1,36 @@
...
+    class __urdlllocal context_t
+    {
+    public:
+        ur_api_version_t version = UR_API_VERSION_0_9;
+
+        ur_dditable_t   urDdiTable = {};
+        context_t();
+        ~context_t() = default;
+
+        void* get( void )
```
FranklandJack, line 28: I don't think `void` is required here.

source/drivers/null/ur_null.h
```diff
@@ -0,0 +1,36 @@
+/*
+ *
+ * Copyright (C) 2019-2021 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * @file ur_null.h
+ *
+ */
+#pragma once
```
FranklandJack, line 10: I think we should use the `#ifndef` pattern here.

source/loader/linux/lib_init.cpp
```diff
@@ -0,0 +1,22 @@
+/*
+ *
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#include "../ur_lib.h"
```
FranklandJack, line 9: Is there a way to avoid including relative paths?

source/loader/linux/loader_init.cpp
```diff
@@ -0,0 +1,22 @@
+/*
+ *
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#include "../ur_loader.h"
```
FranklandJack, line 9: Here too, can we avoid using relative paths?

source/loader/platform_discovery.h
```diff
@@ -0,0 +1,20 @@
+/*
+ *
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#pragma once
```
FranklandJack, line 9: I think we should use `#ifndef` here.

source/loader/ur_ldrddi.h
```diff
@@ -0,0 +1,48 @@
+/*
+ *
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * @file ur_ldrddi.h
+ *
+ */
+#pragma once
```
FranklandJack, line 10: I think we should use `#ifndef` here.

source/loader/ur_lib.h
```diff
@@ -0,0 +1,41 @@
+/*
+ *
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * @file ur_lib.h
+ *
+ */
+#pragma once
```
FranklandJack, line 10: I think we should use `#ifndef` here.

source/loader/ur_loader.h
```diff
@@ -0,0 +1,49 @@
+/*
+ *
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+#pragma once
```
FranklandJack, line 8: I think we should use `#ifndef` here.

source/loader/ur_object.h
```diff
@@ -0,0 +1,39 @@
+/*
+ *
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * @file ur_object.h
+ *
+ */
+#pragma once
```
FranklandJack, line 10: I think we should use `#ifndef` here.

source/loader/windows/lib_init.cpp
```diff
@@ -0,0 +1,26 @@
+/*
+ *
+ * Copyright (C) 2021 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#include "../ur_lib.h"
```
FranklandJack, line 9: Is there a way to avoid relative file paths here?

source/loader/windows/loader_init.cpp
```diff
@@ -0,0 +1,14 @@
+/*
+ *
+ * Copyright (C) 2021 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#include "../ur_loader.h"
```
FranklandJack, line 9: And here.

source/common/ur_singleton.h
pbalcer: That's my preference as well. But I refrained from making any stylistic changes (the coding style is all over the place :-)). I suggest we pick on a single style and enforce it with clang-format (we were planning on doing that soon after this PR was merged).

I can do it with this PR, but a) it might conflict with the generated files, so it's not trivial addition, and b) it would probably affect the entire repo, not just the loader, so might clutter the PR.

kbenzie (changes requested):

CMakeLists.txt
```diff
@@ -69,7 +72,11 @@ install(
     TARGETS headers
     EXPORT ${PROJECT_NAME}-targets)
 
+# set the loader target name for linking with other components
+set(TARGET_LOADER_NAME ur_loader)
```
kbenzie, line 76: Is there really a need to set a variable for this target name?

I think it would be clearer to use the target name explicitly but provide an option to the user to change the name of the shared object via target property.

When adding support for `FetchContent`/`find_package` it became clear that its necessary to have internal/external target names. For example, the internal name for the header only library is `headers` while the external name is `unified-runtime::headers`.

Mu idea would be to follow that pattern for the loader library target as well, with the internal name `loader` and external name `unified-runtime::loader`.

.github/workflows/pull_request.yaml
```diff
@@ -1,29 +0,0 @@
...
-    steps:
-      - uses: actions/checkout@v3
-
-      - name: Install apt package
-        run: sudo apt-get install -y doxygen
-
-      - name: Install pip packages
-        run: pip install -r third_party/requirements.txt
-
-      - name: Generate specification, check for uncommitted diff
```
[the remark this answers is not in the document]
pbalcer: This step in the workflow I added performs that same check:
```
    - name: Generate source from spec
      run: cmake --build ${{github.workspace}}/build --target check-generated --config ${{env.BUILD_TYPE}}
```
I can improve the name to make it clear.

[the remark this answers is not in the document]
kbenzie: Does the `check-generated` target fail if there is a diff?

If so I missed that.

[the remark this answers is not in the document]
pbalcer: Yes, I added a comment now to clarify that.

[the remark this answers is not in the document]
kbenzie: Okay, thanks. I'll resolve this.

CMakeLists.txt
kbenzie: It occurs to me this is the kind of thing that should be written down somewhere for future reference. I don't think making this change is really a blocker, I'd be happy to address these changes and document why in a separate PR once this is merged.

kbenzie (approved):
Besides the minor things which could be addressed later, I think this looks good.

examples/hello_world/hello_world.cpp
```diff
@@ -0,0 +1,94 @@
...
+            ur_device_type_t device_type;
+            status = urDeviceGetInfo(d, UR_DEVICE_INFO_TYPE, sizeof(ur_device_type_t), static_cast<void *>(&device_type), nullptr);
+            if (status != UR_RESULT_SUCCESS)
+            {
+                std::cout << "urDeviceGet failed with return code: " << status << std::endl;
+                goto out;
+            }
+            static const size_t DEVICE_NAME_MAX_LEN = 1024;
+            char device_name[DEVICE_NAME_MAX_LEN] = {0};
+            status = urDeviceGetInfo(d, UR_DEVICE_INFO_NAME, DEVICE_NAME_MAX_LEN - 1, static_cast<void *>(&device_name), nullptr);
```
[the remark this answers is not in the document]
pbalcer: I'd rather keep the example short. Checking the length here doesn't add much - device_name will be a valid NULL-terminated C string, which `cout` should output correctly.

In a test, on the other hand, we should verify that if this function returns success, the name is populated with a non-zero length string.

scripts/json2src.py
```diff
@@ -0,0 +1,58 @@
...
+
+    for idx, specs in enumerate(input['specs']):
+        config = input['configs'][idx]
+        if args.sections == None or config['name'] in args.sections:
+            if args.lib:
+                generate_code.generate_lib(srcpath, config['name'], config['namespace'], config['tags'], args.ver, specs, input['meta'])
+            if args.loader:
+                generate_code.generate_loader(srcpath, config['name'], config['namespace'], config['tags'], args.ver, specs, input['meta'])
+            if args.layers:
+                generate_code.generate_layers(srcpath, config['name'], config['namespace'], config['tags'], args.ver, specs, input['meta'])
```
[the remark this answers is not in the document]
pbalcer: I don't want to remove it entirely since we will be adding the layer functionality soon. I commented out these lines, so it's impossible to call this a non-existent method.

CMakeLists.txt
pbalcer: I removed this variable and added the loader to the unified-runtime namespace.

source/loader/windows/loader_init.cpp
pbalcer: Done

source/loader/windows/lib_init.cpp
pbalcer: Done

source/loader/ur_object.h
pbalcer: Done

source/loader/ur_loader.h
pbalcer: Done

source/loader/ur_lib.h
pbalcer: Done

source/loader/ur_ldrddi.h
pbalcer: Done

source/loader/platform_discovery.h
pbalcer: Done

source/loader/linux/lib_init.cpp
pbalcer: Done

source/drivers/null/ur_null.h
pbalcer: Done

source/drivers/null/ur_null.h
pbalcer: Done

source/common/ur_singleton.h
pbalcer: Done. I'll add an issue about adding style checking to the build system.

source/common/ur_util.h
pbalcer: Done

source/common/ur_singleton.h
pbalcer: Done

.github/workflows/cmake.yml
pbalcer: Done

.gitignore
pbalcer: Done

pbalcer:
In addition to the review commits, I needed to address the conflicting workflow change to verify the python module. So I added a test/python/basic.py test that runs in the new cmake workflow. Once #94 is merged, we can modify that test to instantiate the UR_DDI class, and it should just work.

source/common/ur_singleton.h
pbalcer: #101

CMakeLists.txt
pbalcer: Done. I wanted to remove the `ur_` prefix, but that also changed the library name to `libloader.so`. It turns out there's a `LIBRARY_OUTPUT_NAME` property that can be used to change it back to what we need it to be.

source/common/ur_singleton.h
FranklandJack: Addressing them in a future issue works for me :+1:

FranklandJack (approved):
Thanks for addressing the comments I've made - LGTM.

CMakeLists.txt
```diff
@@ -90,9 +98,26 @@ configure_package_config_file(
...
+
 # Add the package files to the install.
 install(
     FILES
         ${PROJECT_BINARY_DIR}/cmake/${PROJECT_NAME}-config.cmake
         ${PROJECT_BINARY_DIR}/cmake/${PROJECT_NAME}-config-version.cmake
     DESTINATION lib/cmake/${PROJECT_NAME})
+
+# generate source from the specification
+add_custom_target(generate
```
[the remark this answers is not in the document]
kbenzie: Nice

PatKamin (approved):

.github/workflows/pull_request.yaml
```diff
@@ -1,37 +0,0 @@
...
-
-      - name: Generate specification, check for uncommitted diff, check python module.
-        run: |
-          cd scripts
-          python run.py --core
-          if [ $(git diff | wc -l) != 0 ]; then
-            git diff
-            exit 1
-          fi
-          cd ../include
```
FranklandJack, line 30: Did this check get lost in the move to doing this in CMake?

PR #89, opened 2022-12-06
Fix python module generation

FranklandJack:
* Fix bug in code that generates `ur.py` to allow for arbitrary levels
  of pointer derivedness.
* Add regexs for missing c_types in python generation.
Fixes #71.

PR #94, opened 2022-12-08
fix UR_DDI python class init

pbalcer:
This patch fixes all issues that were preventing the python bindings from initializing the loader.

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -1951,14 +1951,17 @@ class UR_DDI:
         if "Windows" == platform.uname()[0]:
             self.__dll = WinDLL("ur_loader.dll")
         else:
-            self.__dll = CDLL("ur_loader.so")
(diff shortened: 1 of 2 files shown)


pbalcer:
The loader PR has a conflict with the recent change to the `pull_request.yml` workflow (which I'm removing in favor of one that uses cmake targets) that added a check to verify if the python module even imports (#89). That inspired me to add simple test infrastructure using ctest and create a python that actually tried to run the python bindings. This was so that I could add running ctest to the cmake workflow.

Predictably, the python module didn't work ;) Hence this patch.

Once this is merged, I'll either update the loader PR to include the test or create a new PR.

FranklandJack (approved):
LGTM.

kbenzie (approved):
LGTM

PR #100, opened 2022-12-08
ur: Remove user event creation

### LINKED ISSUE #15: Reconsider urEventCreate
`urEventCreate` appears to be inspired by the [`clCreateUserEvent`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateUserEvent) API.

OpenCL user events can be painful to implement since they essentially require bidirectional host-device synchronization, which may not be something a platform can efficiently support. We should consider whether user events are really required in the Unified Runtime API and if not remove this entry point.

kbenzie:
OpenCL user events, which `urEventCreate` was an analogue of, allow the application to force a previously enqueued command to wait for the user event to be signalled before execution can begin. User events are also the only type of event that can synchronize across contexts, requiring a separate special case code path to implement. Lastly, there is no analogue to `clSetUserEventStatus` which actually enables the application to change a user events status, making this feature incomplete.

Until such a time as they are deemed absolutely necessary, with no viable alternative proposed, this patch removes support for user events.

Fixes #15.

### DIFF
--- a/scripts/core/event.yml
+++ b/scripts/core/event.yml
@@ -41,25 +41,6 @@ etors:
       desc: "A 64-bit value of current device counter in nanoseconds when the event has finished execution"
 --- #--------------------------------------------------------------------------
 type: function
-desc: "Create an event object. Events allow applications to enqueue commands that wait on an event to finish or signal a command completion."
-class: $xEvent
-name: Create
-ordinal: "0"
-analogue:
-    - "**clCreateUserEvent**"
-params:
-    - type: $x_context_handle_t
-      name: hContext
-      desc: "[in] handle of the context object"
-    - type: $x_event_handle_t*
-      name: phEvent
-      desc: "[out] pointer to handle of the event object created"
-returns:      
-    - $X_RESULT_ERROR_INVALID_CONTEXT
-    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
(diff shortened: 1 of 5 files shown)


FranklandJack (approved):
LGTM.

scripts/core/event.yml
```diff
@@ -41,25 +41,6 @@ etors:
...
-      desc: "[out] pointer to handle of the event object created"
-returns:      
-    - $X_RESULT_ERROR_INVALID_CONTEXT
-    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
-    - $X_RESULT_ERROR_OUT_OF_RESOURCES
---- #--------------------------------------------------------------------------
-type: function
 desc: "Get event object information"
 class: $xEvent
 name: GetInfo
```
jatinx, line 46: should we edit x_execution_info_t as well? it has some status info about user created events.

Cant seem to get this comment on that line, its on: 209 and 211

kbenzie: How does [this](https://github.com/oneapi-src/unified-runtime/compare/be9ab12ddb1d207211cbfcaf644e5d2c0857df8b..1013327ab435a178da3930d7c74c6670ae8ccf4e) look?

jatinx: LGTM

veselypeta (approved):
LGTM

Issue #75, opened 2022-11-30
Missing context info enum values 


jatinx:
- PI_CONTEXT_INFO_PLATFORM - Its a useful information, used by some backend to make certain decision (device->platform->name). But this can also be done by urDeviceGetInfo(UR_DEVICE_INFO_PLATFORM)->urGetPlatformInfo(PLATFORM_NAME). So we can skip this.
- PI_CONTEXT_INFO_PROPERTIES - Cant find its usage
- PI_CONTEXT_INFO_REFERENCE_COUNT  Cant find its usage too. UR has the concept of context acquire/release, but no way to query its reference count. Its inconsistent, for device which also has acq/rel, we have a UR_DEVICE_INFO_REFERENCE_COUNT. can add it to make it consistent.
- PI_CONTEXT_INFO_ATOMIC_MEMORY_[ORDER|SCOPE]_CAPABILITIES - used to populate sycl device info for order/scope capabilities, Should they be moved to device info instead of context?

callumfare:
We still seem to be missing `PI_CONTEXT_INFO_ATOMIC_MEMORY_[ORDER|SCOPE]_CAPABILITIES` which are needed for compatibility with PI

PR #108, opened 2022-12-09
Adding missing context enums

jatinx:
- PI_CONTEXT_INFO_PLATFORM - Its a useful information, used by some backend to make certain decision (device->platform->name). But this can also be done by context->getdevices->urDeviceGetInfo(UR_DEVICE_INFO_PLATFORM)->urGetPlatformInfo(PLATFORM_NAME). Not adding this.

- PI_CONTEXT_INFO_REFERENCE_COUNT - Cant find its usage. UR has the concept of context acquire/release, but no way to query its reference count. Its inconsistent, for device which also has acq/rel, we have a UR_DEVICE_INFO_REFERENCE_COUNT. Adding this to make it consistent.

- PI_CONTEXT_INFO_ATOMIC_MEMORY_[ORDER|SCOPE]_CAPABILITIES - used to populate sycl device info for order/scope capabilities.

Fixes: https://github.com/oneapi-src/unified-runtime/issues/75

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -240,11 +240,42 @@ class ur_rect_region_t(Structure):
     ]
 
 ###############################################################################
+## @brief Supported memory order capabilities
+class ur_memory_order_flags_v(IntEnum):
+    RELAXED = UR_BIT(0)                             ## Memory order relaxed
+    ACQUIRE = UR_BIT(1)                             ## Memory order acquire
+    RELEASE = UR_BIT(2)                             ## Memory order release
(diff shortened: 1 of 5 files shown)


FranklandJack:
Searching the DPC++ codebase for [`PI_CONTEXT_INFO_PLATFORM`](https://github.com/intel/llvm/blob/b3cbda570089eeed1564fcb32148563a773c2973/sycl/include/sycl/info/context_traits.def#L2) does show one use. It would be good to follow that and know it's definitely not a requirement if we aren't going to add it.

FranklandJack (changes requested):

PR #117, opened 2022-12-09
Fix include directories for loader

igchor:
When unified-runtime is built as part of another CMakeProject (through add_subdirectory) the paths will not be correct. It's better to use relative paths.

CMAKE_SOURCE_DIR would point to the top level of the other project directory, not unified-runtime top level dir.

### DIFF
--- a/source/loader/CMakeLists.txt
+++ b/source/loader/CMakeLists.txt
@@ -18,18 +18,15 @@ set_target_properties(loader PROPERTIES
 
 add_library(${PROJECT_NAME}::loader ALIAS loader)
 
-target_include_directories(loader PRIVATE
-    ${CMAKE_SOURCE_DIR}
-    ${CMAKE_SOURCE_DIR}/source/common
-    ${CMAKE_SOURCE_DIR}/source/loader
-)
+target_include_directories(loader PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
 
 set_target_properties(loader PROPERTIES
     VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}"
     SOVERSION "${PROJECT_VERSION_MAJOR}"
 )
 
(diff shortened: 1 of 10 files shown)


pbalcer:
Yes, that's what I'm thinking as well. We already have a loader and headers target, we just need to add a `common` interface target.

pbalcer:
I made a quick patch for that here: https://github.com/pbalcer/unified-runtime/commit/179de4fcd40e4418314ce010d6013fe33b293ece
@igchor take a look and see if this fixes your problem.

source/loader/CMakeLists.txt
```diff
@@ -19,9 +19,9 @@ set_target_properties(loader PROPERTIES
 add_library(${PROJECT_NAME}::loader ALIAS loader)
 
 target_include_directories(loader PRIVATE
-    ${CMAKE_SOURCE_DIR}
-    ${CMAKE_SOURCE_DIR}/source/common
-    ${CMAKE_SOURCE_DIR}/source/loader
+    ${CMAKE_CURRENT_SOURCE_DIR}/../common
+    ${CMAKE_CURRENT_SOURCE_DIR}/../..
+    ${CMAKE_CURRENT_SOURCE_DIR}
```
kbenzie, line 24: I actually think it would be better to use `${PROJECT_SOURCE_DIR}` which points to the root `CMakeLists.txt` in a similar way as `${CMAKE_SOURCE_DIR}` does when not part of an `add_subdirectoty`.

kbenzie (changes requested):

source/loader/CMakeLists.txt
FranklandJack: This seems like the simplest solution to fixing this.

igchor:
Ok, I've used Piotr's commit since we would probably need to create a 'common' library anyway once we start implementing the adapters.

I've tested this, and it seems to work with SYCL (I've done an initial integration which I'll publish on a PR soon).

PR #125, opened 2022-12-12
More CMake fixes

igchor:
Needed for integration with SYCL.

We might also consider adding some unique prefixes to 'loader' and other components' names (like 'ur_loader'/'ur_common') to avoid potential name collisions with other projects. For now, SYCL adds dependencies (level-zero, UR) through add_subdirectory.


pbalcer (approved):

kbenzie (approved):
LGTM

PR #126, opened 2022-12-13
Hello world - print API version

PatKamin:


examples/hello_world/hello_world.cpp
```diff
@@ -49,6 +49,10 @@ int main(int argc, char *argv[])
 
     for (auto p : platforms)
     {
+        ur_api_version_t api_version = {};
+        urPlatformGetApiVersion(platform, &api_version);
```
pbalcer, line 53: please add status check.

pbalcer (approved):
And I think this is where we should leave the null adapter and the hello world example for now. Adding anything else risks introducing unnecessary conflicts/breakages when making spec changes.

FranklandJack (approved):
LGTM.

PR #127, opened 2022-12-13
Add Windows build CI workflow

PatKamin:


pbalcer (approved):

source/common/ur_util.h
```diff
@@ -49,41 +49,6 @@ inline std::string create_library_path(const char *name, const char *path){
     return library_path;
 }
 
-#ifdef _WIN32
-inline std::string readUnifiedRuntimeLoaderLibraryPath() {
```
kbenzie, line 53: I'm not familiar with why `readUnifiedRuntimeLoaderLibraryPath` was introduced, or why its being removed now. Could you elaborate?

PatKamin: It was, I think mistakenly, incorporated alongside the `ur_util.h` file, based on the level zero loader's `util` file (with many other changes). However, this function is not used in the unified runtime and it's causing a Windows build failure. Focusing now on having a stable Windows build, I removed this unused function.

Perhaps there are more functions/structures that were introduced recently, but are not used anywhere which would require some more cleanup work.

kbenzie: Okay, makes sense. This is the kind of context and reasoning I'd really appreciate seeing in commit messages.

kbenzie (approved):

PR #129, opened 2022-12-13
[ur] support queue indices

### LINKED ISSUE #120: Support sycl_ext_intel_queue_index extension
https://github.com/intel/llvm/pull/7599 introduces support for the `sycl_ext_intel_queue_index` by adding an alternative signature of `piQueueCreate` which allows specifying a null terminated array of queue properties to enable the application to specify the desired queue index for devices which support it. Refer to the extension specification for more detail, and the PR for implementation details.

To resolve this ticket the UR spec should be updated to enable this extension.

veselypeta:
Closes #120

### DIFF
--- a/scripts/core/queue.yml
+++ b/scripts/core/queue.yml
@@ -29,7 +29,7 @@ etors:
       desc: "Queue size info"
 --- #--------------------------------------------------------------------------
 type: enum
-desc: "Queue properties"
+desc: "Queue property flags"
 class: $xQueue
 name: $x_queue_flags_t
 etors:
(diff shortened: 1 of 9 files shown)


scripts/core/queue.yml
```diff
@@ -46,6 +46,18 @@ etors:
       value: "$X_BIT(3)"
       desc: "Is the default queue for a device"
 --- #--------------------------------------------------------------------------
+type: enum
+desc: "Queue Properties"
+class: $xQueue
+name: $x_queue_props_t
```
veselypeta, line 52: Ideally this would be called `$x_queue_properties_x` however the code generation will match this as a properties struct and not an enum and fail to generate the spec [here](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/parse_specs.py#L691) which matches anything ending in `properties_t`.

kbenzie (changes requested):
Blocking until https://github.com/intel/llvm/pull/7599 is accepted.

kbenzie:
https://github.com/intel/llvm/pull/7599 has merged, I'll review this properly soon. Could you double check things between this PR and the other still line up @veselypeta?

veselypeta:
I've checked over the PR to DPC++ and this patch will match their interface.

kbenzie (approved):
LGTM

PR #131, opened 2022-12-13
Check ur.py module is valid

FranklandJack:
As part of the `check-generated` target check that the generated `ur.py`
module is valid python and can be included.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,9 +121,10 @@ add_custom_target(generate ALL
     DEPENDS ${API_JSON_FILE}
 )
 
-# generate source and check for uncommitted diffs
+# check for uncommitted diffs and valid python module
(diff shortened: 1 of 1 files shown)


pbalcer:
The correctness of the python bindings is verified as part of the `make test` target by running the [basic.py](https://github.com/oneapi-src/unified-runtime/blob/main/test/python/basic.py) test. This test import UR and instantiates the `UR_DDI` class.

FranklandJack:
> The correctness of the python bindings is verified as part of the `make test` target by running the [basic.py](https://github.com/oneapi-src/unified-runtime/blob/main/test/python/basic.py) test. This test import UR and instantiates the `UR_DDI` class.

Sorry I literally just saw this :facepalm:, will close now!

Issue #115, opened 2022-12-09
Add Missing `ur_device_affinity_domain_flag_t` Enumerations


FranklandJack:
# Issue

PI defines the [`pi_device_affinity_domain`](https://github.com/intel/llvm/blob/c4932957929f94afe7f87bee009eea5e2b5cd6bf/sycl/include/sycl/detail/pi.h#L635) bitfield which can take the following values:
* `PI_DEVICE_AFFINITY_DOMAIN_L4_CACHE`
* `PI_DEVICE_AFFINITY_DOMAIN_L3_CACHE`
* `PI_DEVICE_AFFINITY_DOMAIN_L2_CACHE`
* `PI_DEVICE_AFFINITY_DOMAIN_L1_CACHE`

which are used [here](https://github.com/intel/llvm/blob/7b47ebb68d87f760785dd4a409fa4ea24b8235c4/sycl/source/detail/device_info.hpp#L63), [here](https://github.com/intel/llvm/blob/7b47ebb68d87f760785dd4a409fa4ea24b8235c4/sycl/source/detail/device_info.hpp#L435), [here](https://github.com/intel/llvm/blob/7b47ebb68d87f760785dd4a409fa4ea24b8235c4/sycl/source/detail/device_info.hpp#L434) and [here](https://github.com/intel/llvm/blob/7b47ebb68d87f760785dd4a409fa4ea24b8235c4/sycl/source/detail/device_info.hpp#L433) respectively in the DPC++ runtime.

Unified Runtime defines the [`ur_device_affinity_domain_flag_t`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3109) enumeration but it is missing the above values.

# Task

Add the missing enumerations from above to Unified Runtime so that it meets the requirements of the DPC++ runtime.

FranklandJack:
There is a bit of a question of whether the DPC++ runtime actually needs to use these enumerations, see the comments on https://github.com/oneapi-src/unified-runtime/pull/132.

veselypeta:
These enums are likely not required for UR.

kbenzie:
> These enums are likely not required for UR.

I don't think I agree, they are required for the [SYCL 2020 spec](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-class), see table 23.

PR #132, opened 2022-12-13
Add missing affinity domains

FranklandJack:
Add missing cache affinity domains to device affinity domains
enumeration. Fixes https://github.com/oneapi-src/unified-runtime/issues/115.

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -791,7 +791,11 @@ class ur_device_exec_capability_flags_t(c_int):
 ## @brief Device affinity domain
 class ur_device_affinity_domain_flags_v(IntEnum):
     NUMA = UR_BIT(0)                                ## By NUMA
-    NEXT_PARTITIONABLE = UR_BIT(1)                  ## BY next partitionable
+    NEXT_PARTITIONABLE = UR_BIT(1)                  ## By next partitionable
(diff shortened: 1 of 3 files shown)


FranklandJack:
This might be worth reconsidering. 

Looking at the usage of `PI_DEVICE_AFFINITY_DOMAIN_L(1|2|3|4)_CACHE` none of the adaptors actually handle these enumeration values, so although they would be required replace the usage in the DPC++ codebase as is, perhaps they aren't really needed by the DPC++ codebase to begin with unless there are out of tree adapters using them?

kbenzie:
> ... perhaps they aren't really needed by the DPC++ codebase to begin with unless there are out of tree adapters using them?

Does make me wonder if we should hold off making this change then.

FranklandJack:
> > ... perhaps they aren't really needed by the DPC++ codebase to begin with unless there are out of tree adapters using them?
> 
> Does make me wonder if we should hold off making this change then.

Yeah I think so, this may be true of other missing PI features. But it would be useful to get confirmation from engineers who are familiar with that codebase.

FranklandJack:
I'm closing this since there is a question of if it is actually required.

Issue #128, opened 2022-12-13
loader windows platform discovery implementation


pbalcer:
# issue
The loader needs to be able to discover available adapter implementations, given a hardcoded list of names. On Linux, this is simple since we can rely on dlopen's builtin support for dynamic symbol discovery.
For windows, this gets complicated. The level-zero loader (https://github.com/oneapi-src/level-zero/blob/master/source/loader/windows/driver_discovery_win.cpp) looks for registry keys from the display adapter installation to discover the drivers. This approach might not necessarily work for UR.

# tasks

1. For the time being, we need a simple way of allowing the loader to find adapters on windows. This should include loading hardcoded adapters from some known directory (cwd ?) and support for `UR_ENABLE_ALT_DRIVERS` env variable.
2. We need to figure out and implement the best long-term approach for adapters to register themselves for the loader to discover. Maybe we need a registry folder where adapters info is placed? or maybe have a known single directory for all adapters?

kbenzie:
As a data point, the OpenCL-ICD-Loader defines a similar approach [here](https://github.com/KhronosGroup/OpenCL-ICD-Loader#registering-icds) of using a config file on Linux and registry keys on Windows.

pbalcer:
We've discussed this at the workgroup, and the decision is to keep things simple for now. The hard coded list of adapters will stay and be used on both platforms. In addition, we will:
 - create a new env variable (`UR_ADAPTERS_SEARCH_PATH`) that specifies a comma-separated list of directories to include in the adapter search. This can be implemented in a platform-agnostic way with the use of `std::filesystem` (because I don't think there's a clean way of adding a library search path on Linux. `LD_LIBRARY_PATH` needs to be defined on program start).
 - make the loader look for adapters in the loader's own location. This can be accomplished on Linux with `dlinfo(..., RTLD_DI_LINKMAP, ...);` and on Windows with `GetModuleFileName`.

PR #137, opened 2022-12-14
Search for known adapter library in the current working directory on Windows

PatKamin:
This is related to the task no. 1 from the issue #128.

This PR allows us to run the hello_world example with the null adapter library on Windows. Additionally, this PR makes it possible that some more adapters could be loaded in the future (with an env var) on Windows.

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -10,6 +10,7 @@
 import platform
 from ctypes import *
 from enum import *
+import os
 
 ###############################################################################
 __version__ = "1.0"
@@ -1945,7 +1946,9 @@ class UR_DDI:
     def __init__(self, version : ur_api_version_t):
         # load the ur_loader library
         if "Windows" == platform.uname()[0]:
-            self.__dll = WinDLL("ur_loader.dll")
+            os.add_dll_directory(os.getenv("UR_ADAPTERS_SEARCH_DIR"))
+            LOAD_LIBRARY_SEARCH_DEFAULT_DIRS = 0x00001000
+            self.__dll = WinDLL("ur_loader.dll", winmode=LOAD_LIBRARY_SEARCH_DEFAULT_DIRS)
         else:
             self.__dll = CDLL("libur_loader.so")
 
--- a/source/loader/windows/platform_discovery_win.cpp
+++ b/source/loader/windows/platform_discovery_win.cpp
@@ -7,6 +7,7 @@
  */
 
 #include "platform_discovery.h"
+#include "ur_util.h"
 
 #include <Windows.h>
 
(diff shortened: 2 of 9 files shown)


source/loader/ur_loader.cpp
```diff
@@ -25,6 +25,8 @@ namespace loader
             {
                 platforms.emplace_back();
                 platforms.rbegin()->handle = handle;
+            } else {
+                return UR_RESULT_ERROR_MODULE_LINK_FAILURE;
```
pbalcer, line 29: I don't think we want to completely fail init if one of the libraries doesn't load for some reason. Someone might just happen to have a completely different file with the path we expect.

Once we have logging infrastructure, we can print a warning/info about failing to load a library.

pbalcer:
Can you please add running tests to the windows build job?

source/loader/ur_loader.cpp
PatKamin: Done. Removed this failure.

PatKamin:
> Can you please add running tests to the windows build job?

Done. Had to call binaries directly as I couldn't make tests ran with ctest work properly with changed cwd. I think it's good enough as a temporary approach until the .dll loading mechanism will be changed.

test/python/basic.py
```diff
@@ -12,7 +12,9 @@
 
 # not ideal, but this solution is simple and portable. Alternative is to set
 # PYTHONPATH or test on installed sources.
-sys.path.insert(1, '../include')
+from pathlib import Path
+ur_module_path = str(Path(os.path.realpath(__file__)).parents[2].joinpath("include"))
```
pbalcer, line 16: This is too hacky :-) Maybe instead we can just remove this entirely and set PYTHONPATH like this in the CMake:
```
set_property(TEST python-basic APPEND PROPERTY
                ENVIRONMENT "PYTHONPATH=${PROJECT_SOURCE_DIR}")
```
?

this should allow us to just `import include.ur as ur` instead.

include/ur.py
```diff
@@ -1945,7 +1946,9 @@ class UR_DDI:
     def __init__(self, version : ur_api_version_t):
         # load the ur_loader library
         if "Windows" == platform.uname()[0]:
-            self.__dll = WinDLL("ur_loader.dll")
+            os.add_dll_directory(os.getcwd())
+            LOAD_LIBRARY_SEARCH_USER_DIRS = 0x00000400
+            self.__dll = WinDLL("ur_loader.dll", winmode=LOAD_LIBRARY_SEARCH_USER_DIRS)
```
pbalcer, line 1951: Please check if `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` solves this automatically.

source/loader/windows/platform_discovery_win.cpp
```diff
@@ -20,9 +21,40 @@
...
+    if (nullptr == result) {
+        return enabledPlatforms;
+    }
+    result = AddDllDirectory(libsPath);
+    if (nullptr == result) {
+        return enabledPlatforms;
+    }
+
+    // UR_ENABLE_ALT_DRIVERS is for development/debug only
+    altPlatforms = getenv("UR_ENABLE_ALT_DRIVERS");
```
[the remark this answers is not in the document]
PatKamin: Opened #145

pbalcer (approved):

pbalcer:
I still don't like having to create a windows-only env variable just so python tests work, but I can't think of a better idea for now.

PR #139, opened 2022-12-15
Remove `urKernelSetArg`

FranklandJack:
* Remove the `urKernelSetArg` entry point.
* Add thread safe `urKernelSetArgValue` and `urKernelSetArgLocal` entry
  points.


kbenzie (changes requested):
This also needs to update the old code example in PROG.rst that uses the removed entry point. That should also fix the CI.

scripts/core/kernel.yml
```diff
@@ -52,7 +49,24 @@ params:
...
+params:
+    - type: "$x_kernel_handle_t"
+      name: hKernel
+      desc: "[in] handle of the kernel object"
+    - type: "uint32_t"
+      name: argIndex
+      desc: "[in] argument index in range [0, num args - 1]"
+    - type: "size_t"
+      name: argSize
+      desc: "[in] size of the local buffer to be allocated by the runtime"
```
veselypeta, line 69: Maybe state the error codes that it returns:

```yml
returns:
        - $X_ERROR_INVALID_KERNEL_ARGUMENT_INDEX
        - $X_ERROR_INVALID_KERNEL_ARGUMENT_SIZE
```

veselypeta (approved):
Minor comment, but LGTM

scripts/core/kernel.yml
```diff
@@ -299,7 +311,6 @@ desc: "Set a Memory object as the argument value of a Kernel."
 class: $xKernel
 name: SetArgMemObj
 details:
-    - "The application must **not** call this function from simultaneous threads with the same kernel handle."
     - "The implementation of this function should be lock-free."
 params:
     - type: "$x_kernel_handle_t"
```
FranklandJack, line 316: Should we make the `hArgValue` optional here, this would be the equivalent of setting a buffer to null in OpenCL which I think is valid?

scripts/core/kernel.yml
FranklandJack: Yeah good point, these are missing from the other APIs too. Although this has opened questions about what should be allowed to be null.

scripts/core/kernel.yml
```diff
@@ -52,7 +49,24 @@ params:
       desc: "[in] size of argument type"
     - type: "const void*"
       name: pArgValue
-      desc: "[in][optional] argument value represented as matching arg type. If null then argument value is considered null."
+      desc: "[in] argument value represented as matching arg type."
+--- #--------------------------------------------------------------------------
+type: function
```
FranklandJack, line 54: Do we want to be able to set a local memory argument to NULL through this API? And if so how would that work?

veselypeta: Is this possible in OpenCL?

scripts/core/kernel.yml
veselypeta: Yeah I think that makes sense. Reading OpenCL it says `A NULL value can also be specified if the argument is a buffer object in which case a NULL value will be used as the value for the argument declared as a pointer to __global or __constant memory in the kernel.`. Although I'm not sure if UR has a concept of `__local`, `__global` etc..

scripts/core/kernel.yml
FranklandJack: I don't think so. In OpenCL you'd call `clSetKernelArg` with `NULL` as the argument and the required size of the local memory buffer in bytes in the size argument. So you could set the buffer size to zero (you could also do that here), but I don't think there is any way to actually set it to `NULL.`

To set a  global buffer to `NULL` in OpenCL you set the argument to `NULL` and the size to `sizeof(cl_mem)`.

scripts/core/kernel.yml
FranklandJack: I've made this optional.

scripts/core/kernel.yml
FranklandJack: Added these.

scripts/core/kernel.yml
FranklandJack: I've left this as it is since it is basically equivalent to what OpenCL has.

Issue #101, opened 2022-12-08
Add automatic code formatting and checking


pbalcer:
Now that we are about to have a much bigger codebase with the introduction of the loader, it's about time to add convenience utilities to the build system to automate tedious or repetitive tasks. One of which is enforcing a consistent coding style across the entire repository.

This task is to propose and integrate a code formatting tool with the selected coding style.
My proposal is clang-format with the LLVM style, augmented with whatever changes the team decides on.

Because many files in this repo are automatically generated, the template files will probably need to be manually modified to conform to the tool-enforced style.

PR #153, opened 2023-01-05
[ur] Adds platform testsuite

veselypeta:
This PR:
 * Implements a platform testsuite which tests the following API entry-points
   * urInit
   * urTearDown
   * urPlatformGet
   * urPlatformGetInfo
* Reduces cmake requirement to `3.14.0` - to match llvm/sycl CMake version
* Adds a bunch of fixtures which will be used in further testsuites


test/conformance/platform/urInit.cpp
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2022 Intel Corporation
+// SPDX-License-Identifier: MIT
+#include <uur/checks.h>
+
+TEST(urInitTest, Success) {
+  ur_platform_init_flags_t platform_flags = 0;
+  ur_device_init_flags_t device_flags = 0;
+  ASSERT_SUCCESS(urInit(platform_flags, device_flags));
+
+  // TODO - Question? Should we call urTearDown() in these tests. We aren't
```
igchor, line 10: IMO yes, we should always all resources allocated in a test.

test/conformance/platform/fixtures.h
```diff
@@ -0,0 +1,50 @@
...
+
+  std::vector<ur_platform_handle_t> platforms;
+};
+
+struct urPlatformTest : urPlatformsTest {
+
+  void SetUp() override {
+    UUR_RETURN_ON_FATAL_FAILURE(urPlatformsTest::SetUp());
+    ASSERT_GE(platforms.size(), 1);
+    platform = platforms[0]; // TODO - which to choose?
```
igchor, line 41: Isn't the platform already chosen by `platform_options.platform_name` in PlatformEnvironment? 

I see there is uur::urPlatformTest and uur::platform::urPlatformTest. What's the reason for having both of those?

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Copyright (C) 2022 Intel Corporation
 # SPDX-License-Identifier: MIT
 
-cmake_minimum_required(VERSION 3.22.0 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.16.0 FATAL_ERROR)
```
alycm, line 4: I see in the description that this was purposeful, but this was deliberately raised from 3.8.0 to 3.22.0 here: https://github.com/oneapi-src/unified-runtime/pull/147/files

It's not clear to me why that PR needed to bump the CMake version, but is 3.16.0 definitely high enough?

test/conformance/platform/fixtures.h
veselypeta: Not in this case. This test suite does not use any of the environments (Platform or Device). The reason is that the environment is a global setup run before any tests. This kind of assumes that certain parts of the API just work. These environments will be used in future test suites and are really useful in simplifying and parametrizing tests. This is fine since we have already tested those entry points in a previous test suite.

test/conformance/platform/urPlatformGet.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include "fixtures.h"
+
+using urPlatformGetTest = uur::platform::urTest;
+
+TEST_F(urPlatformGetTest, Success) {
+  uint32_t count;
+  ASSERT_SUCCESS(urPlatformGet(0, nullptr, &count));
+  std::vector<ur_platform_handle_t> platforms(count);
+  ASSERT_SUCCESS(urPlatformGet(count, platforms.data(), nullptr));
+  for (auto platform : platforms) {
```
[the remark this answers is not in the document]
veselypeta: 👍

test/conformance/platform/urInit.cpp
veselypeta: 👍

CMakeLists.txt
veselypeta: I'm not sure why the version was changed. From my perspective the version I have with ubuntu 20.04 is 3.16 which I think is default. I believe 3.22 is the new default version shipped with 22.04. I'm happy to update my version if it's worthwhile @pbalcer?

kbenzie: CMake 3.22.0 is very new, I don't think we should be this aggresive with requiring such a new CMake version with `FATAL_ERROR`. I think we should be matching the CMake version as https://github.com/intel/llvm requires since we are a direct dependency of that project.

test/conformance/source/DevicesEnvironmentMain.cpp
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2022 Intel Corporation
+// SPDX-License-Identifier: MIT
+
+#include <uur/environment.h>
+
+int main(int argc, char **argv) {
+  auto *environment = new uur::DevicesEnvironment(argc, argv);
```
kbenzie, line 7: I think we should just always have our own `main.cpp` and use a macro defined in CMake to switch between which environment class to instantiate. This way we don't have mostly identical code and two different conditinal logic in CMake to sometimes link `gtest_main`.

source/drivers/null/ur_null.cpp
```diff
@@ -9,89 +9,121 @@
  */
 #include "ur_null.h"
 
-namespace driver
-{
-    //////////////////////////////////////////////////////////////////////////
-    context_t d_context;
+namespace driver {
```
kbenzie, line 12: Why are there changes in this file?

kbenzie (changes requested):

source/drivers/null/ur_null.cpp
veselypeta: It wasn't spec conformant. I had to fix some error codes that were being returned and added some additional cases. 

Not sure about how to deal with the formatting changes - Since this is not an autogenerated file we don't have a formatting style I guess.

kbenzie: Oh, I assumed it was autogenerated.

As for formatting, yeah we don't have it yet but we have #101 to track it.

CMakeLists.txt
pbalcer: No, I agree we should roll back this change. @PatKamin was dealing with a very annoying problem on Windows that was solvable by using a very new cmake feature (which we realized was not really viable), but ultimately he managed to address the issue differently.

veselypeta: Sounds good with me. I'll rollback to version `3.14` to match intel's llvm/sycl version

test/conformance/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+endfunction()
+
+# add_subdirectory(context)
+# add_subdirectory(device)
+# add_subdirectory(enqueue)
+# add_subdirectory(event)
+# add_subdirectory(memory)
+add_subdirectory(platform)
+# add_subdirectory(queue)
+# add_subdirectory(usm)
```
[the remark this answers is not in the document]
veselypeta: 👍

test/conformance/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+
+function(add_conformance_test name)
+    set(TEST_TARGET_NAME test-${name})
+    add_executable(${TEST_TARGET_NAME}
+        ${ARGN}
+        ${UR_CONFORMANCE_TEST_DIR}/source/environment.cpp)
+    target_link_libraries(${TEST_TARGET_NAME}
+        PRIVATE
+        ${PROJECT_NAME}::loader
+        ${PROJECT_NAME}::common
```
pbalcer, line 25: You shouldn't need to link with common.

test/conformance/source/DevicesEnvironmentMain.cpp
veselypeta: Would you like individual an `main.cpp` for each test suite, or a single global main shared by every testsuite? Also by CMake macro do you mean define a compiler definition with CMake or a `macro()`?

My thought was something like the following:
```cpp
#include <uur/environment.h>

int main(int argc, char **argv) {
#ifdef DEVICES_ENVIRONMENT
  auto *environment = new uur::DevicesEnvironment(argc, argv);
#endif
#ifdef PLATFORM_ENVIRONMENT
  auto *environment = new uur::PlatformEnvironment(argc, argv);
#endif
  ::testing::InitGoogleTest(&argc, argv);
#if defined(DEVICES_ENVIRONMENT) || defined(PLATFORM_ENVIRONMENT)
  ::testing::AddGlobalTestEnvironment(environment);
#endif
  return RUN_ALL_TESTS();
}
```

test/conformance/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+        ${PROJECT_NAME}::headers
+        GTest::gtest_main)
+    target_include_directories(${TEST_TARGET_NAME} PRIVATE
```
pbalcer, line 28: You could create a `::testing` interface library for generic testing "framework" code, similar to what we have for common. This would let you avoid relative includes like this. And we will likely need some common testing utilities anyway (asserts and whatnot).

test/conformance/CMakeLists.txt
veselypeta: 👍

test/conformance/source/DevicesEnvironmentMain.cpp
kbenzie: A shared `main.cpp` makes senese for now. If we need separate ones later we can modify things. Yes, `#ifdef`s in the code is what I was aiming for.

kbenzie (approved):
LGTM

Issue #138, opened 2022-12-14
create logging infrastructure for use by loader and adapter implementations


pbalcer:
There's currently no simple way of outputting logs in the loader apart from using bare native functions.

Implement a basic multi-platform logging infrastructure that contains at least:
- log functions with predefined levels (info/warn/error/debug), with a way for the user to choose the verbosity level at runtime.
- option to control the output location of the logs (stdout, stderr, file, ?)
- rich compile-time and runtime asserts

Since this logging API is meant to be used in the loader and adapters, the log level and output location should be controllable per module. For example:
`UR_LOG="loader,null=warn+error;level-zero=info+warn+error"`

pbalcer:
One option is to use something that already exists, e.g., https://github.com/gabime/spdlog

pbalcer:
Done in #154

Issue #143, opened 2022-12-19
add tracing support in the loader


pbalcer:
The UR loader needs to support API call tracing to facilitate debugging and software introspection in general.
Level zero has a [custom tracing layer](https://github.com/oneapi-src/level-zero/tree/master/source/layers/tracing) that uses callbacks on entering and exiting a function. This approach is flexible and retains type information about the functions in callbacks. The downside is that the tracing layer requires a callback wrapper type for every UR function, doubling the overall API surface.

The tracing layer in L0 is primarily used to implement [XPTI](https://github.com/intel/llvm/blob/sycl/xptifw/doc/XPTI_Framework.md) instrumentation for the [L0 plugin](https://github.com/intel/llvm/blob/sycl/sycl/plugins/level_zero/tracing.cpp) in SYCL. XPTI is a generic tracing framework that is supported by a variety of sycl-related tooling.
Implementing XPTI support directly in UR will mean that all UR users get tracing support automatically without using the possibly cumbersome tracing callbacks. But this approach is less flexible, and software that wishes to trace UR will also require XPTI.

Depending on our needs, we have three options (ordered from least work to most work):
1. Natively integrate XPTI
2. Implement just the tracing layer, delegating XPTI integration to an upper layer
3. Implement the tracing layer and also integrate it with XPTI in the UR

This task is to decide on the approach and implement tracing support.

PR #154, opened 2023-01-05
Introduce logging mechanism

PatKamin:
This PR addresses #138


scripts/templates/libapi.cpp.mako
```diff
@@ -55,6 +57,13 @@ ${th.make_func_name(n, tags, obj)}(
     )
 {
 %if re.match("Init", obj['name']):
+    std::string logger_name = "loader";
+    auto logger = spdlog::stdout_color_st(logger_name);
+    assert(logger != nullptr);
```
igchor, line 62: Shouldn't we handle this by returning an error / printing something to stderr?

source/common/ur_util.h
```diff
@@ -76,4 +78,72 @@ inline bool getenv_tobool( const char* name )
     return 0 == strcmp( "1", env );
 }
 
+inline void parse_log_env_var(const char *env_var_name,
+                              std::pair<std::string, std::string> &kv,
+                              std::string logger_name) {
+    // example: UR_LOG_LEVELS="loader=info;adapters=warn"
+    const char* env_var_cstr = std::getenv(env_var_name);
```
igchor, line 85: It would be nice to implement this string parsing as some generic helper utility (not in this PR probably). It would be useful for parsing other env variables as well, for example: https://github.com/intel/llvm/blob/2c7ed980a43c4de633a6c32e7902116800a7876c/sycl/plugins/unified_runtime/ur/usm_allocator.cpp#L150

kbenzie:
Does this PR address #143?

PatKamin:
> Does this PR address #143?

This PR addresses #138

source/common/ur_util.h
```diff
@@ -76,4 +80,54 @@ inline bool getenv_tobool( const char* name )
...
+    while (std::getline(ss, key_value, main_delim)) {
+        std::string key;
+        std::string values;
+        std::stringstream kv_ss(key_value);
+        std::getline(kv_ss, key, key_value_delim);
+        std::getline(kv_ss, values);
+        if (values.empty()) {
+            std::cout << "Wrong format of the " << env_var_name << " environment variable! Value not found!" << std::endl;
+            std::cout << "Proper format is: ENV_VAR=\"param_1:value_1,value_2;param_2:value_1\"" << std::endl;
+            return -1;
```
igchor, line 118: You can consider just throwing an exception in this case. Then, you don't need to return any status and you could just return the map by value, instead of using a reference.

source/common/ur_util.h
```diff
@@ -76,4 +80,54 @@ inline bool getenv_tobool( const char* name )
     return 0 == strcmp( "1", env );
 }
 
+inline int parse_env_var(const char *env_var_name,
```
igchor, line 83: It would be good to have some simple unit test for this

igchor, line 83: I'm not sure about the name, it doesn't say to what the env is parsed, perhaps `getenv_to_map` would be better?

source/common/ur_util.h
```diff
@@ -76,4 +80,54 @@ inline bool getenv_tobool( const char* name )
...
+                         std::unordered_map<std::string, std::vector<std::string>> &map) {
+    // example: ENV_VAR="param_1:value_1,value_2;param_2:value_1"
+    char main_delim = ';';
+    char key_value_delim = ':';
+    char values_delim = ',';
+    int delim_pos = 0;
+
+    const char *env_var_c_str = std::getenv(env_var_name);
```
igchor, line 91: I think you might want to use the same logic as in `getenv_tobool` to support windows and linux (perhaps we can extract that as a standalone function).

examples/hello_world/CMakeLists.txt
```diff
@@ -19,7 +19,10 @@ if(MSVC)
     )
 endif()
 
+find_package(spdlog REQUIRED)
```
pbalcer, line 22: Let's implement our own simple logger. It's less of a hassle. I don't think it needs to be anything fancy - formatting can be done with `vsnprintf`, with a thread-local buffer, with printing to some configurable sink (but `stdout`/`stderr` should be sufficient for now).

source/common/ur_util.h
PatKamin: std::optional requires C++17

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,46 @@
...
+
+	    ret = std::fputs(buffer, stdout);
+        if (ret < 0) {
+            printf("Call to fputs failed!");
+        }
+        std::fputs("\n", stdout);
+    }
+
+private:
+    int print_to_buffer(char *buffer, const int buf_size, const char *format, ...) {
```
igchor, line 33: nit: this logic could be implemented as a standalone (global) function that returns std::string. This way you could use this formatting not only for logging/printing but also for creating messages (e.g. for exceptions).

Something like:
```
std::string format_to_str(const char *format, ...) {
    char buffer[8096];
    const int buf_size = 8096;

    va_list vlist;

    va_start(vlist, format);
    int ret = std::vsnprintf(buffer, buf_size, format, vlist);
    va_end(vlist);
    if (ret < 0) {
        throw std::runtime_error("vsnprintf failed");
    }

    return std::string(buffer);
}
```

source/common/ur_util.h
```diff
@@ -57,23 +60,79 @@ inline std::string create_library_path(const char *name, const char *path){
...
+    // example: ENV_VAR="param_1:value_1,value_2;param_2:value_1"
+    char main_delim = ';';
+    char key_value_delim = ':';
+    char values_delim = ',';
+    int delim_pos = 0;
+    std::unordered_map<std::string, std::vector<std::string>> map;
+
+    const char *env_var_c_str = ur_getenv(env_var_name);
+    if (env_var_c_str == nullptr) {
+        std::cout << "Environment variable " << env_var_name << " is not set!" << std::endl;
```
igchor, line 96: Use logger here

source/common/logger/ur_logger.h
```diff
@@ -0,0 +1,104 @@
...
+        this->sink = std::make_shared<sink::Sink>();
+    }
+
+    ~Logger() = default;
+
+    void setLevel(Level level) {
+        this->level = level;
+    }
+
+    void logDebug(const char *format, ...) {
```
pbalcer, line 39: I would just use `debug()` for the name so it's `log->debug("...");`

pbalcer, line 39: hm, this function is essentially repeated n times for each of the enum parameters. You could create a single template function (or a normal function that accepts the enum parameter...) and use it for every variant.

keryell (commented):
Good idea but do not use unsafe old C coding style in new C++ development.
Or is this an advertisement for using Rust in this project instead? :-)
Perhaps something to discuss during the next oneAPI TAB...

source/common/logger/ur_logger.h
```diff
@@ -0,0 +1,104 @@
...
+        sink->log("FATAL", format, vlist);
+        va_end(vlist);
+    }
+
+private:
+    Level level {Level::QUIET};
+    std::shared_ptr<sink::Sink> sink;
+};
+
+}
```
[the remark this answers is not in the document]
PatKamin: Done, thanks.

source/common/logger/ur_logger.h
PatKamin: I wanted the usage of logging to be as simple as can be, thus preferring debug("...") over log(DEBUG, "..."). Shortened functions a bit, though.

pbalcer:
> Good idea but do not use unsafe old C coding style in new C++ development. Or is this an advertisement for using Rust in this project instead? :-) Perhaps something to discuss during the next oneAPI TAB...

I wouldn't mind using Rust, to be perfectly honest :) We decided to use C-style formatting to keep this simple, but you are right - using idiomatic C++ for this isn't that much harder but will be much more ergonomic in the long term.

source/common/logger/ur_logger.h
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    void logInfo(const char *format, ...) {
+        va_list vlist;
+
+        if (Level::INFO < this->level) {
+            return;
+        }
+
+        va_start(vlist, format);
```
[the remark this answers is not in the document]
PatKamin: Replaced va_list with variadic templates.

source/common/logger/ur_level.h
```diff
@@ -0,0 +1,29 @@
...
+
+enum class Level {
+    DEBUG,
+    INFO,
+    WARNING,
+    ERROR,
+    QUIET
+};
+
+static const char *to_str(Level level) {
```
keryell, line 14: Make this `constexpr`.
More generally, put everything `constexpr` when you can.

test/unit/utils/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: MIT
+
+add_unit_test(getenv_to_map
+    getenv_to_map.cpp
+    )
+
```
keryell, line 7: Missing EOL.

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,66 @@
...
+            std::fputs("Call to fputs failed!\n", fd);
+        }
+    }
+
+private:
+    std::FILE *fd;
+
+    template<typename... Args>
+    int printToBuffer(char *buffer, const int buf_size, const char *format, Args&&... args) {
+        int ret = std::snprintf(buffer, buf_size, format, std::forward<Args>(args)...);
```
keryell, line 47: Well, I was thinking to get rid of unsafe C-like format at all.
The problem if you use them now in a new project, it will be here for the next decades in this project... :-(
Using something https://fmt.dev/latest/index.html#safety instead and removing it when you switch to C++20?

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,66 @@
...
+    {
+        if (os.rdbuf() == std::cerr.rdbuf()) {
+            this->fd = stderr;
+        }
+    }
+
+    ~Sink() = default;
+
+    template<typename... Args>
+    void log(const char *level, const char *format, Args&&... args) {
```
keryell, line 26: You could also provide a `log_concat` alternative without format, just concatenating the `std::to_string` of each argument to the log.
Useful instead of wasting time to write unneeded format strings for most use cases and avoid security holes.

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,66 @@
...
+        const int buf_size = 8192;
+        char buffer[buf_size];
+        int buf_size_left = buf_size;
+        unsigned position = 0;
+
+        appendToBuffer(position, buffer, buf_size_left, "[%s]:", level);
```
keryell, line 32: This could become for example
```suggestion
        concatToBuffer(position, buffer, buf_size_left, '[', level, "]:");
```

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,66 @@
...
+        appendToBuffer(position, buffer, buf_size_left, format, std::forward<Args>(args)...);
+        appendToBuffer(position, buffer, buf_size_left, "\n", level);
+
+	    int ret = std::fputs(buffer, fd);
+        if (ret < 0) {
+            std::fputs("Call to fputs failed!\n", fd);
```
keryell, line 38: Do you think retrying harder will work? :-)
Throw an exception?

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,66 @@
...
+        char buffer[buf_size];
+        int buf_size_left = buf_size;
+        unsigned position = 0;
+
+        appendToBuffer(position, buffer, buf_size_left, "[%s]:", level);
+        appendToBuffer(position, buffer, buf_size_left, format, std::forward<Args>(args)...);
+        appendToBuffer(position, buffer, buf_size_left, "\n", level);
+
+	    int ret = std::fputs(buffer, fd);
+        if (ret < 0) {
```
keryell, line 37: Weird indent. Could be just one line anyway.

source/common/logger/ur_level.h
```diff
@@ -0,0 +1,29 @@
...
+    switch(level) {
```
keryell, line 15: To be coherent with other parts of your code
```suggestion
    switch (level) {
```

test/unit/logger/logger.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <iostream>
+
+#include <gtest/gtest.h>
+
+#include "logger/ur_logger.h"
+
+TEST(logger, SetLevel) {
+    auto level = level::Level::WARNING;
+    auto sink = new sink::Sink();
```
keryell, line 12: Do not use unsafe code because you can. :-)
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-newdelete

source/common/logger/ur_level.h
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: MIT
+
+namespace level {
+
+enum class Level {
+    DEBUG,
```
keryell, line 7: Is such uppercase enforced by your coding style ?

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,66 @@
...
+            std::fputs("Call to fputs failed!\n", fd);
+        }
+    }
+
+private:
+    std::FILE *fd;
```
keryell, line 43: Looks scary.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#p8-dont-leak-any-resources

keryell (commented):
A few comments.
But at the end why reinventing yet-another-open-source-logging library?

PatKamin:
> A few comments. But at the end why reinventing yet-another-open-source-logging library?

The goal is not to add any new dependencies.

source/common/logger/ur_level.h
PatKamin: Yes, other enums in this project are uppercase. See Naming Convention here: https://github.com/oneapi-src/unified-runtime/blob/7bc754a218eee13d082bbcfcd8f221a541cbf211/scripts/core/INTRO.rst

source/common/logger/ur_sinks.h
PatKamin: Provided a simple formatting function to avoid adding a new dependency.

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,77 @@
...
+};
+
+class StderrSink : public Sink {
+  public:
+    StderrSink() { this->ostream = &std::cerr; }
+
+    ~StderrSink() = default;
+};
+
+class FileSink : public Sink {
```
igchor, line 59: Can we have a generic StreamSink as well? This would allow as to use it instead with stringstream and format messages for exceptions, etc.

source/common/logger/ur_logger.h
```diff
@@ -0,0 +1,52 @@
...
+// SPDX-License-Identifier: MIT
+
+#include "ur_level.h"
+#include "ur_sinks.h"
+
+namespace logger {
+
+class Logger {
+  public:
+    Logger(sink::Sink *sink) : sink(sink) {
```
igchor, line 11: I think it would be better to pass `sink` as `std::unique_ptr<Sink>` - ther logger takes ownership of the sink, right?

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,77 @@
...
+            throw std::runtime_error(
+                std::string("Failure while opening log file: ") + file_path);
+        }
+        this->ostream = &ofstream;
+    }
+
+    ~FileSink() { ofstream.close(); }
+
+  private:
+    const char *file_path;
```
igchor, line 73: If you want to store the `file_path` it should be `std::string`. Otherwise you're risking having a dangling pointer if the file path is created dynamically. But... do you really need to store this?

source/common/logger/ur_logger.h
PatKamin: Done

source/common/logger/ur_sinks.h
```diff
@@ -0,0 +1,77 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: MIT
+
+#include <iostream>
+
+namespace sink {
```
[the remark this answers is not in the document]
PatKamin: Right, it's better the way you suggested. Done.

source/common/logger/ur_sinks.h
PatKamin: Removed.

test/unit/utils/getenv_to_map.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include <gtest/gtest.h>
+
+#include "ur_util.h"
+
+#if defined(_WIN32)
+    #define setenv(name, value, overwrite) _putenv_s(name, value)
+    #define unsetenv(name) _putenv_s(name, "")
+#endif
+
+TEST(getenv_to_map, TwoParamsTwoValues) {
```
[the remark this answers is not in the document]
PatKamin: Extended tests with checks for map contents. Thanks, one bug found.

igchor (approved):

kbenzie (approved):
LGTM

source/common/ur_util.hpp
```diff
@@ -90,22 +93,81 @@ inline std::string create_library_path(const char *name, const char *path) {
...
 
+inline bool getenv_tobool(const char* name)
+{
+    const char* env = ur_getenv(name);
     if (nullptr == env)
         return false;
     return 0 == strcmp("1", env);
 }
 
+inline std::unordered_map<std::string, std::vector<std::string>> getenv_to_map(const char *env_var_name) {
```
igchor, line 118: this line looks pretty long, did you run clang-format on this file?

source/common/logger/ur_logger.hpp
```diff
@@ -0,0 +1,122 @@
...
+        if (level < this->level) {
+            return;
+        }
+
+        sink->log(logger::level_to_str(level), format,
+                  std::forward<Args>(args)...);
+    }
+};
+
+std::unique_ptr<Logger> logger;
```
igchor, line 70: Will this work correctly? I think you should use either static (if you want this to be per compilation unit) or extern.
But do we actually need to have this as a global variable? Perhaps we could store the logger in contexts? (e.g. `loader::context` and `ur_lib::context`).

igchor: Right, for common, we still need a global variable.

As for optional - it would work as well, but it requires C++17 (and we have C++14 I think).

source/common/ur_util.hpp
```diff
@@ -90,22 +93,81 @@ inline std::string create_library_path(const char *name, const char *path) {
...
+    const char *env_var_c_str = ur_getenv(env_var_name);
+    if (env_var_c_str == nullptr) {
+        std::cout << "Environment variable " << env_var_name << " is not set!" << std::endl;
+        return map;
+    }
+    std::string env_var(env_var_c_str);
+
+    // check if the env var is in the simple format
+    // example: ENV_VAR=1
+    if (env_var.find(':') == std::string::npos
```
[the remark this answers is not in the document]
igchor: perhaps we could just put the value under empty key here? `map[""] = values_vec` ?

source/common/logger/ur_logger.hpp
kbenzie: C++14 I think comes from being compiled as part of LLVM. [C++17 is now the default in LLVM](https://www.phoronix.com/news/LLVM-Clang-16-Default-GNU17) so, assuming https://github.com/intel/llvm has those patches, I'd be open to changing to C++17.

scripts/templates/libapi.cpp.mako
```diff
@@ -56,6 +57,17 @@ ${th.make_func_name(n, tags, obj)}(
 {
 %if re.match("Init", obj['name']):
     static ${x}_result_t result = ${X}_RESULT_SUCCESS;
+
+    // Init logger
```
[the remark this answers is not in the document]
PatKamin: Done.

source/common/logger/ur_logger.hpp
```diff
@@ -0,0 +1,79 @@
...
+#define UR_LOGGER_HPP 1
+
+#include "ur_level.hpp"
+#include "ur_sinks.hpp"
+
+namespace logger {
+
+class Logger {
+  public:
+    Logger(std::unique_ptr<logger::Sink> sink) : sink(std::move(sink)) {
```
pbalcer, line 14: does it have to be a pointer? Can't you accept the sink by value?

source/loader/ur_lib.cpp
```diff
@@ -8,10 +8,15 @@
  *
  */
 #include "ur_lib.hpp"
+#include "logger/ur_logger_api.hpp"
 #include "ur_loader.hpp"
 #include "ur_proxy_layer.hpp"
 #include "validation/ur_validation_layer.hpp"
 
+namespace logger {
+Logger logger = create_logger("loader");
```
pbalcer, line 17: The way we have this right now, this will cause an uncaught exception if something is misconfigured with the logger.
I suggest we catch all the exceptions in `create_logger` and print to stderr if something is wrong. Normally libraries should never print by default, but I guess printing to stderr if the user is trying to set up the logger is reasonable. There's no other simple way to return an error during global initialization.

pbalcer (commented):
You forgot `make cppformat` ;)

source/common/ur_util.hpp
```diff
@@ -90,22 +93,81 @@ inline std::string create_library_path(const char *name, const char *path) {
 #endif
 
 ///////////////////////////////////////////////////////////////////////////////
-inline bool getenv_tobool(const char *name) {
-    const char *env = nullptr;
-
+inline char *ur_getenv(const char* name) {
```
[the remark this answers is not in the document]
PatKamin: Changed to std::string

source/common/ur_util.hpp
```diff
@@ -90,22 +93,81 @@ inline std::string create_library_path(const char *name, const char *path) {
...
 #if defined(_WIN32)
-    char buffer[8];
-    auto rc = GetEnvironmentVariable(name, buffer, 8);
-    if (0 != rc && rc <= 8) {
-        env = buffer;
+    constexpr int buffer_size = 1024;
+    static char buffer[buffer_size];
```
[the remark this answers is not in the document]
PatKamin: Copying to std::string.

test/unit/utils/getenv_to_map.cpp
```diff
@@ -0,0 +1,127 @@
...
+                                         "param_1:value_1,value_2;:value_1",
+                                         ",;:", ",;"));
+
+TEST_P(getEnvToMapException, WrongEnvVarValues) {
+    ret = setenv("UR_TEST_ENV_VAR", env_var_value.c_str(), 1);
+    ASSERT_EQ(ret, 0);
+
+    EXPECT_THROW({ auto map = getenv_to_map("UR_TEST_ENV_VAR"); },
+                 std::invalid_argument);
+}
```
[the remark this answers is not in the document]
PatKamin: Added more tests.

source/common/logger/ur_logger.hpp
```diff
@@ -0,0 +1,93 @@
...
+#include <memory>
+
+#include "ur_logger_obj.hpp"
+#include "ur_util.hpp"
+
+namespace logger {
+extern Logger logger;
+
+template <typename... Args>
+void debug(const char *format, Args &&...args) {
```
pbalcer, line 17: inline

same in every function you define directly in the header.

source/common/logger/ur_sinks.hpp
```diff
@@ -0,0 +1,178 @@
...
+};
+
+auto sink_from_str(std::string name) {
+    struct sink_name {
+        std::string name;
+        std::unique_ptr<Sink> sink;
+    };
+
+    sink_name sink_names[] = {
+        {"stdout", std::make_unique<logger::StdoutSink>()},
```
pbalcer, line 153: You are unnecessarily creating instances of the sinks you have on the list. One alternative is to instead store functions that create the class you want.

source/common/logger/ur_sinks.hpp
```diff
@@ -0,0 +1,178 @@
...
+            return std::move(item.sink);
+        }
+    }
+    throw std::invalid_argument(
+        std::string("Parsing error: no valid sink for string '") + name +
+        std::string("'.") +
+        std::string("\nValid sink names are: stdout, stderr"));
+}
+
+auto sink_from_str(std::string name, std::string file_path) {
```
pbalcer, line 167: I think it would be simpler if you just had single function that always accepted two arguments, with the file path being optional.

source/common/logger/ur_sinks.hpp
```diff
@@ -0,0 +1,178 @@
...
+        ofstream = std::ofstream(file_path, std::ofstream::out);
+        if (ofstream.rdstate() != std::ofstream::goodbit) {
+            throw std::invalid_argument(
+                std::string("Failure while opening log file: ") + file_path +
+                std::string(" Check if given path exists."));
+        }
+        this->ostream = &ofstream;
+    }
+
+    ~FileSink() { ofstream.close(); }
```
keryell, line 140: https://en.cppreference.com/w/cpp/io/basic_ofstream
```suggestion
```

source/common/ur_util.hpp
```diff
@@ -87,23 +90,121 @@ inline std::string create_library_path(const char *name, const char *path) {
...
-        env = buffer;
+    constexpr int buffer_size = 1024;
+    char buffer[buffer_size];
+    auto rc = GetEnvironmentVariable(name, buffer, buffer_size);
+    if (0 != rc && rc < buffer_size) {
+        return std::string(buffer);
+    } else if (rc >= buffer_size) {
+        std::stringstream ex_ss;
+        ex_ss << "Environment variable " << name << " value too long!";
+        ex_ss << " Maximum length is 1023 characters.";
```
keryell, line 104: Value to compute

source/common/logger/ur_sinks.hpp
PatKamin: Replaced this code with an `if else` for simplicity.

source/common/ur_util.hpp
PatKamin: Done

source/common/logger/ur_sinks.hpp
PatKamin: Done

source/common/logger/ur_logger.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+    kv = map->find("output");
+    if (kv != map->end()) {
+        values = kv->second;
+    }
+
+    try {
+        sink = values.size() == 2 ? sink_from_str(values[0], values[1])
+                                  : sink_from_str(values[0]);
+    } catch (const std::invalid_argument &e) {
```
pbalcer, line 92: nit: instead of dotting the function around with try catch, you can do it once, for example by creating a new function. Especially since all the exceptions are exactly the same.

source/common/ur_util.hpp
```diff
@@ -87,23 +91,120 @@ inline std::string create_library_path(const char *name, const char *path) {
...
+    std::cerr << "Wrong format of the " << env_var_name << " environment variable!"
+              << " Proper format is: ENV_VAR=\"value_1,value_2,value_3\"" << std::endl;
+}
+
+static void wrong_format_map_msg(const char *env_var_name) {
+    std::cerr << "Wrong format of the " << env_var_name << " environment variable!"
+              << " Proper format is: ENV_VAR=\"param_1:value_1,value_2;param_2:value_1" << std::endl;
+}
+
+inline std::optional<std::vector<std::string>> getenv_to_vec(const char *env_var_name) {
```
pbalcer, line 131: please add briefs on what these two functions do. It's hard to get from just the names.

source/common/ur_util.hpp
```diff
@@ -87,23 +91,120 @@ inline std::string create_library_path(const char *name, const char *path) {
...
-    auto rc = GetEnvironmentVariable(name, buffer, 8);
-    if (0 != rc && rc <= 8) {
-        env = buffer;
+    constexpr int buffer_size = 1024;
+    char buffer[buffer_size];
+    auto rc = GetEnvironmentVariable(name, buffer, buffer_size);
+    if (0 != rc && rc < buffer_size) {
+        return std::string(buffer);
+    } else if (rc >= buffer_size) {
+        std::cerr << "Environment variable " << name << " value too long!"
```
pbalcer, line 102: please change it to an exception, so we don't print to screen. same below.

Ultimately, I think we shouldn't be as liberal with throwing exceptions as in this PR since ultimately the library interacts through a C API, but for now let's leave it as is. I wish we had `std::expected`... we might think about refactoring this with something simple based on `std::variant`.

pbalcer (commented):
mostly lgtm, just a few minor nits.

pbalcer (approved):

test/unit/helpers.h
```diff
@@ -0,0 +1,7 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: MIT
+
```
igchor, line 3: Missing include guard

igchor (approved):
LGTM, just one missing include guard

scripts/core/INTRO.rst
```diff
@@ -174,6 +174,26 @@ Environment Variables
 
 Specific environment variables can be set to control the behavior of unified runtime or enable certain features.
 
+.. envvar:: UR_LOG_LOADER/UR_LOG_NULL
+
+   Hold parameters for setting Unified Runtime logging. UR_LOG_LOADER sets up logging in the loader library and UR_LOG_NULL in the null adapter.
+   Valid format of values stored in these environment variables:
+      UR_LOG_*="[level:debug|info|warning|error][;flush:<debug|info|warning|error>][;output:file,<path>]"
+
+   Example of an environment variable for setting up the loader library logger with logging level set to `info`, flush level set to `warning` and output set to
```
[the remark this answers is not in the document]
PatKamin: Done

scripts/core/INTRO.rst
```diff
@@ -174,6 +174,26 @@ Environment Variables
...
+   the `out.log` file:
+          UR_LOG_LOADER="level:info;flush:warning;output:file,out.log"
```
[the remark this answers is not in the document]
PatKamin: Added an info that all are optional and an example where only some are set.

pbalcer (approved):

PR #155, opened 2023-01-06
[ur] Clearly specify that SetArg is thread safe

### LINKED ISSUE #14: Recondisder urKernelSetArg
`urKernelSetArg` appears to have been heavily influenced by the `clSetKernelArgs` OpenCL API.

The `clSetKernelArgs` entry point is problematic for two reasons:
1. It is [one of the only non-thread safe APIs in OpenCL](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_multiple_host_threads), this behavior seems to have been inherited by `urKernelSetArg`:
```
   * The application must not call this function from simultaneous threads with the same kernel handle.

   * The implementation of this function should be lock-free.
```
2. It only sets one argument at a time.

We should be careful about repeating history here. Firstly, the API should be thread safe unless there is a valid reason it can't be, second we should support an API that sets multiple kernel arguments at once since this may be more optimal on some platforms.

veselypeta:
We have decided that unlike OpenCL the SetArg APIs in UR should be thread-safe. This was mostly completed before, but I've added an additional statement to make this clear.

Closes #14

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -3339,6 +3339,8 @@ urKernelCreate(
 /// @brief Set kernel argument to a value.
 /// 
 /// @details
+///     - The application may call this function from simultaneous threads with
+///       the same kernel handle.
 ///     - The implementation of this function should be lock-free.
 /// 
 /// @returns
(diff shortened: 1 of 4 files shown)


kbenzie (approved):
LGTM

PR #157, opened 2023-01-06
[ur] Mark non-required arguments as optional

### LINKED ISSUE #33: Allow optional null arguments in API generation
Currently any pointer arguments to entry points in the API automatically have a error condition that the API should return `UR_RESULT_ERROR_INVALID_NULL_POINTER ` see [here](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/parse_specs.py#L667).

Whilst this makes sense for most pointer parameters, some are optional e.g. return events. We should generalize this logic to allow for optional pointer arguments.

veselypeta:
* `urEnqueue*` entry-points have an optional return event.
*  `urEventGetInfo` is fixed to allow optional return size and value

Closes #33


kbenzie (approved):
LGTM

Issue #88, opened 2022-12-06
Introduce a clGetDeviceAndHostTimer analogue


kbenzie:
https://github.com/intel/llvm/pull/7526 introduces the ability to query the device timestamp, this is an analogue of [clGetDeviceAndHostTimer](https://man.opencl.org/clGetDeviceAndHostTimer.html). We should also introduce this into the UR spec.

veselypeta:
We should make `pHostTimer` & `pDeviceTimer` optional parameters so that `urDeviceGetGlobalTimestamps` can be used as an anologue of both `clGetDeviceAndHostTimer` and `clGetHostTimer`

kbenzie:
Is this to match the behaviour in https://github.com/intel/llvm/pull/7526 ? Does the OpenCL analogue also make these optional?

veselypeta:
They are required arguments in OpenCL, but the OpenCL plugin in PI does the following:

```cpp
  if (DeviceTime) {
    uint64_t dummy;
    clGetDeviceAndHostTimer(deviceID, DeviceTime,
                            HostTime == nullptr ? &dummy : HostTime);

  } else if (HostTime) {
    clGetHostTimer(deviceID, HostTime);
  }
```

kbenzie:
Okay, so its to match the PI behaviour 👍

PR #160, opened 2023-01-06
[ur] Make host/device time parameters optional

veselypeta:
We should make device and host timer parameters optional in `urDeviceGetGlobalTimestamps` to allow it to be used as an analogue of both `clGetDeviceAndHostTimer` and `clGetHostTimer`.

See more detailed comment in #88


kbenzie (approved):
LGTM

PR #162, opened 2023-01-09
Add automatic code formatting

DamianDuy:
This PR adds code formatting using clang-format.
The .clang-format file was automatically generated and is based on the LLVM style so reviews about what to change/add are very welcome.
Clang-tidy will be added in the next PR for braces around single line statements since there's no such option in clang-format-10.
The rest of the cpp/header files will be formatted in another PR as well since they'd add too much clutter to this one. The hello_world.cpp file is left as a reference to how it looks after formatting.

Addresses #101

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,6 @@
+---
+Language:      Cpp
+BasedOnStyle:  LLVM
+IndentWidth:   4
+...
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2022 Intel Corporation
+# Copyright (C) 2022-2023 Intel Corporation
 # SPDX-License-Identifier: MIT
 
 cmake_minimum_required(VERSION 3.14.0 FATAL_ERROR)
@@ -7,6 +7,9 @@ project(unified-runtime VERSION 0.5.0)
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
 include(CTest)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+include(helpers)
 
 find_package(Python3 COMPONENTS Interpreter)
 
@@ -17,17 +20,17 @@ if(MSVC)
     set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin/$<CONFIG>)
 endif()
 
-# define rpath for libraries so that adapters can be found automatically
+# Define rpath for libraries so that adapters can be found automatically
 set(CMAKE_BUILD_RPATH "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
 
-#Define a path for custom commands to work around MSVC
+# Define a path for custom commands to work around MSVC
 set(CUSTOM_COMMAND_BINARY_DIR ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
 if(MSVC)
-    #MSVC implicitly adds $<CONFIG> to the output path
+    # MSVC implicitly adds $<CONFIG> to the output path
     set(CUSTOM_COMMAND_BINARY_DIR ${CUSTOM_COMMAND_BINARY_DIR}/$<CONFIG>)
 endif()
 
-#CXX compiler support
+# CXX compiler support
 if(NOT MSVC)
     include(CheckCXXCompilerFlag)
     CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX14)
@@ -42,15 +45,15 @@ if(NOT MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fpermissive -fPIC")
 endif()
 
-#MSVC compile flags
+# MSVC compile flags
 if(MSVC)
     string(REPLACE "/MDd" "/MTd" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}")
     string(REPLACE "/MD" "/MT" CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")
 
-    # treat warnings as errors
+    # Treat warnings as errors
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /WX /W3 /wd4996")
 
-    # enable multi-process compilation
+    # Enable multi-process compilation
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
 endif()
 
@@ -61,6 +64,44 @@ endif()
 
 # Option to disable tests
 option(${PROJECT_NAME}_BUILD_TESTS "Build unit tests." ON)
+
+# Option to run a code formatter
+option(${PROJECT_NAME}_FORMAT_CPP_STYLE "format code style of C++ sources" OFF)
+
+find_program(CLANG_FORMAT NAMES clang-format-10 clang-format-10.0 clang-format)
+
+set(CLANG_FORMAT_REQUIRED "10.0")
+
+if(CLANG_FORMAT)
+    get_program_version_major_minor(${CLANG_FORMAT} CLANG_FORMAT_VERSION)
+    message(STATUS "Found clang-format: ${CLANG_FORMAT} (version: ${CLANG_FORMAT_VERSION})")
+endif()
+
+# Check if the correct version of clang-format is available
+if(FORMAT_CPP_STYLE)
+    if(CLANG_FORMAT)
+        if(NOT (CLANG_FORMAT_VERSION VERSION_EQUAL CLANG_FORMAT_REQUIRED))
+            message(FATAL_ERROR "required clang-format version is ${CLANG_FORMAT_REQUIRED}")
+        endif()
+    else()
+        message(FATAL_ERROR "FORMAT_CPP_STYLE=ON, but clang-format not found (required version: ${CLANG_FORMAT_REQUIRED})")
+    endif()
+endif()
+
+# Obtain files for clang-format
+file(GLOB_RECURSE format_src
+    "*.h"
+    "*.cpp"
+)
+
+# Exclude files from build directory to ensure they're not passed to the formatter
+list(FILTER format_src EXCLUDE REGEX "${CMAKE_CURRENT_BINARY_DIR}.")
+
+# Add code formatter target
+add_custom_target(cppformat)
+
+# Add files to the formatter
+add_cppformat(src-formatter ${format_src})
 
 add_subdirectory(${THIRD_PARTY_DIR})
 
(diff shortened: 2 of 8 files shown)


CMakeLists.txt
```diff
@@ -114,16 +138,28 @@ install(
...
-# generate source and check for uncommitted diffs
+# Generate source and check for uncommitted diffs
 add_custom_target(check-generated
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
     COMMAND git diff --exit-code
     DEPENDS generate
 )
+
+# Add code formatter target
+add_custom_target(cppformat)
```
pbalcer, line 156: I think we should fail the build if format generates something.

CMakeLists.txt
```diff
@@ -85,6 +89,26 @@ add_subdirectory(source)
...
+
+if(CLANG_FORMAT)
+	get_program_version_major_minor(${CLANG_FORMAT} CLANG_FORMAT_VERSION)
+	message(STATUS "Found clang-format: ${CLANG_FORMAT} (version: ${CLANG_FORMAT_VERSION})")
+endif()
+
+# Check if the correct version of clang-format is available
+if(FORMAT_CPP_STYLE)
+	if(CLANG_FORMAT)
+		if(NOT (CLANG_FORMAT_VERSION VERSION_EQUAL CLANG_FORMAT_REQUIRED))
```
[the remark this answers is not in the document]
igchor: Unfortunately, the formatting might differ between versions, so we need to require a specific one. And this is a pain when upgrading a distribution on a dev machine - there's usually only one version of clang-format available in the package manager.

Some projects use a separate docker image with a specific clang-format version for that, e.g.: https://github.com/intel/CacheLib/blob/main/.github/workflows/clang-format-check.yml#L16. The drawback is that you cannot clang-format inside a different docker image, but it's not a big problem IMO.

CMakeLists.txt
igchor: We should probably have some DEVELOPER mode for that

pbalcer: I was thinking about adding `DEPENDS cppformat` to the `check-generated` target.

CMakeLists.txt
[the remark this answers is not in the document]
pbalcer: Well, that's a bummer. I just checked, and on Arch clang-format with the entire `clang` package, so there's no easy way to downgrade...
Let's do the docker image and a separate workflow like your cachelib example.

CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ project(unified-runtime VERSION 0.5.0)
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
 include(CTest)
+include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/helpers.cmake)
```
veselypeta, line 10: Might be a good idea to add this path to `CMAKE_MODULES_PATH`

README.md
```diff
@@ -49,6 +49,11 @@ Tools can be acquired via instructions in [third_party](/third_party/README.md).
 
 ## Building
 
+Requirements:
+- C++ compiler with C++14 support
+- cmake >= 3.22.0
```
veselypeta, line 54: We have decided that we'll drop the requirement to `3.8` https://github.com/oneapi-src/unified-runtime/pull/153#discussion_r1063601872

.clang-format
```diff
@@ -0,0 +1,137 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
```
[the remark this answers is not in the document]
kbenzie: Agreed, if this line is uncommented then everything else can likely be removed unless we diverge from llvm style.

CMakeLists.txt
kbenzie: So something like this @veselypeta?

```cmake
list(APPEND CMAKE_MODULES_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
include(helpers)
```

CMakeLists.txt
```diff
@@ -85,6 +89,26 @@ add_subdirectory(source)
 add_subdirectory(examples)
 add_subdirectory(test)
 
+find_program(CLANG_FORMAT NAMES clang-format-10 clang-format-10.0 clang-format)
+
+set(CLANG_FORMAT_REQUIRED "10.0")
+
+if(CLANG_FORMAT)
+	get_program_version_major_minor(${CLANG_FORMAT} CLANG_FORMAT_VERSION)
+	message(STATUS "Found clang-format: ${CLANG_FORMAT} (version: ${CLANG_FORMAT_VERSION})")
```
kbenzie, line 98: The new code in this file is using tabs but the rest of the project uses 4 spaces for indentation in CMake files.

If we could keep intendation consistent that would be best.

CMakeLists.txt
```diff
@@ -114,16 +138,28 @@ install(
...
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
     COMMAND git diff --exit-code
     DEPENDS generate
 )
+
+# Add code formatter target
+add_custom_target(cppformat)
+
+# Add files for clang-format
+add_cppformat(src ${CMAKE_CURRENT_SOURCE_DIR}/*/*.cpp
```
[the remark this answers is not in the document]
kbenzie: Alternatively we could wrap `add_executable`/`add_library`/`target_sources` to automatically add the files passed to them to the formatting target.

cmake/helpers.cmake
```diff
@@ -0,0 +1,39 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: MIT
+
+#
+# helpers.cmake -- helper functions for top-level CMakeLists.txt
+#
+
+# Sets ${ret} to version of program specified by ${name} in major.minor format
+function(get_program_version_major_minor name ret)
+	execute_process(COMMAND ${name} --version
```
kbenzie, line 10: This file is also using tabs instead of 4 spaces making it inconsistent with the rest of the project.

CMakeLists.txt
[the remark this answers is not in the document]
igchor: There is a question about whether we want to integrate this with CMake. Do we want a target that would run the formatting using the docker image? 

Also, With this approach, we would have a dev dependency on docker. Probably we would also need to host our docker image (on gh registry probably) since the image used by CacheLib is hosted here: https://github.com/jidicula/clang-format-action/pkgs/container/clang-format and I don't think we have any guarantees that those images will be always available.

CMakeLists.txt
veselypeta: Yes this was my idea. It might be useful if we add more modules in this directory.

cmake/helpers.cmake
```diff
@@ -0,0 +1,39 @@
...
+# as dependencies of global "cppformat" target.
+# Arguments are used as files to be checked.
+# ${name} must be unique.
+function(add_cppformat name)
+	if(NOT CLANG_FORMAT OR NOT (CLANG_FORMAT_VERSION VERSION_EQUAL CLANG_FORMAT_REQUIRED))
+		return()
+	endif()
+
+	if(${ARGC} EQUAL 0)
+        return()
```
PatKamin, line 27: One more indentation here.

CMakeLists.txt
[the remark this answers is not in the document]
pbalcer: hmm... integrating with CMake isn't a bad idea. But I think the target should first try to run a local `clang-format` if it exists in the correct version.
And yes, we should definitely have create own docker image.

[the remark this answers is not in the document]
igchor: So, probably the existing integration can stay, and we can just fall back to docker in case there is no clang format installed (in the required version). The docker integration could even be done in a separate PR.

CMakeLists.txt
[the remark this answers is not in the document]
DamianDuy: I tried adding Python script for obtaining all the cpp and header files. Would that be okay?

CMakeLists.txt
DamianDuy: Done. Or should I also add a developer mode?

CMakeLists.txt
DamianDuy: Done.

README.md
DamianDuy: Done.

CMakeLists.txt
DamianDuy: Okay. Done.

cmake/helpers.cmake
DamianDuy: Done.

cmake/helpers.cmake
DamianDuy: Done.

.clang-format
```diff
@@ -0,0 +1,137 @@
...
+    SortPriority:    0
+  - Regex:           '.*'
+    Priority:        1
+    SortPriority:    0
+IncludeIsMainRegex: '(Test)?$'
+IncludeIsMainSourceRegex: ''
+IndentCaseLabels: false
+IndentGotoLabels: true
+IndentPPDirectives: None
+IndentWidth:     2
```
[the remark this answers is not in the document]
DamianDuy: Done.

CMakeLists.txt
pbalcer: No, I think it's ok. But let's not enable this until we actually have the docker script integrated.

scripts/find_files.py
```diff
@@ -0,0 +1,35 @@
+"""
+ Copyright (C) 2022 Intel Corporation
```
PatKamin, line 2: 2023

scripts/find_files.py
```diff
@@ -0,0 +1,35 @@
...
+
+ SPDX-License-Identifier: MIT
+
+"""
+
+import os
+import sys
+import re
```
PatKamin, line 10: nit: Alphabetic order.

scripts/find_files.py
```diff
@@ -0,0 +1,35 @@
...
+        if build_dir in path:
+            continue
+        for file in files:
+            match = re.search("\.(h|cpp)$", file)
+            if match:
+                file_path = os.path.join(path, file)
+                print(file_path)
+
+def main():
+    if len(sys.argv) < 3:
```
PatKamin, line 28: `if not exactly two args` is the condition you should check. You might consider using `argparse` for having nice output when using this script with `-h/--help` flags or in case of a wrong number of arguments passed. Or even wrong type of arguments.

scripts/find_files.py
```diff
@@ -0,0 +1,35 @@
...
+import sys
+import re
+
+"""
+    helper for obtaining all of the h/cpp files recursively,
+    starting from root_dir, with the exclusion of the build_dir
+"""
+
+def find_cpp_h_files_rec(root_dir, build_dir):
+    for path, dirs, files in os.walk(root_dir):
```
PatKamin, line 18: `dirs` is unused. Use underscore `_` instead of this name for better readability:

` for path, _, files in os.walk(root_dir):`

CMakeLists.txt
DamianDuy: Done. I've commented it.

CMakeLists.txt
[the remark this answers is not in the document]
DamianDuy: After consulting this with @pbalcer, I've changed the approach to wrapping ```add_executable/add_library/target_sources```, mentioned above.

scripts/find_files.py
DamianDuy: I've changed the approach and wrapped the ```add_executable/add_library/target_sources``` functions to automatically add ```cppformat``` target in cmake files so I've deleted this script for now.

pbalcer (approved):
:+1: 

I've been looking for simpler solutions, and I found a "clang-format Python distribution", which "packages the clang-format utility as a Python package. It allows you to install clang-format directly from PyPI". Together with https://github.com/pre-commit/pre-commit, it's possible to avoid this whole clang-format brokenness problem.

But I think that's something for later when you are thinking about how to enable clang-format (and other checks, e.g., `clang-tidy`) across the entire repo. I also think the ur use-case is a bit unique since we want to have the format/tidy tools run immediately after generating the code (as part of the build system).

CMakeLists.txt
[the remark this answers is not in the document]
DamianDuy: I changed it to use ```file GLOB``` so it's much simpler right now.

PR #164, opened 2023-01-12
[UR] Document Environment Variables

### LINKED ISSUE #140: Introduce section to document environment variables
As the project grows it will inevitably introduce the necessity of controlling behavior via environment variables. As such, there should be a dedicated section in the specification where all environment variables should be enumerated with documentation about when and why they may be of user to a user.

veselypeta:
I couldn't find any reference to environment variables that we use currently so I've put in a placeholder currently to demonstrate the feature, although I'm not sure about placing in fake information in our docs. Perhaps we need a discussion of what environment variables we want or will these just be introduced based on requirements?

Closes #140

### DIFF
--- a/scripts/core/INTRO.rst
+++ b/scripts/core/INTRO.rst
@@ -168,3 +168,17 @@ and followed by a digit or uppercase letter are reserved for use by the implemen
 ## --validate=on
 Applications which use Unified Runtime must not provide definitions of these symbols. 
 This allows the Unified Runtime shared library to be updated with additional symbols for new API versions or extensions without causing symbol conflicts with existing applications.
+
+Environment Variables
+---------------------
+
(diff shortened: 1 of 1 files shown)


pbalcer:
I think the only env variable we have right now is `UR_ADAPTERS_FORCE_LOAD` which appends the comma-separated lists of names to the collection of libraries used by the loader for adapter discovery. For example, setting `UR_ADAPTERS_FORCE_LOAD=libur_foo.so,libur_bar.so` will force the loader to try to use adapter implementations from the `libur_foo.so` and `libur_bar.so` dynamic libraries, in addition to the predefined adapter libraries (`libur_null.so` at the moment).

veselypeta:
I'm removing the GitHub Pages changes from this PR. I've raised an issue to track this in #172 and will raise a follow up PR.

pbalcer (approved):

PR #182, opened 2023-01-23
Move Unified Memory Allocation sources to UR-common

igchor:
Initially, I wanted to move this into llvm repo, but having this here will simplify development. Implementation of the L0 memory provider will be a part of L0 adapter (so inside llvm for now). The same is true for a heap manager implementation. Right now, the idea is to use usm_allocator (this requires following 2 PRs to be merged: https://github.com/intel/llvm/pull/7853 https://github.com/intel/llvm/pull/7785)

List of things to do after this initial PR is merged:
- extend error codes
- add GetLastResult, as in UR
- consider creating a generic config structure for pools/providers
- allow to build UMA as shared library for testing (e.g. with openMP)


jandres742:
thanks @igchor .  could you clarify why are we adding the USM memory allocator to this repo (where we have the loader and spec), instead of directly to LLVM? From comment above, it's not clear to me whether that is just temporary while the other PRs are merged.

Or this is just a shim layer, that will call each adapter's memory allocator, and that's where we will have all optimizations?

source/common/unified_memory_allocation/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# Copyright (C) 2022 Intel Corporation
```
jandres742, line 1: nit: 2023

source/common/unified_memory_allocation/include/uma/memory_pool.h
```diff
@@ -0,0 +1,95 @@
...
+/// \brief Allocates size bytes of uninitialized storage of the specified hPool
+/// \param hPool specified memory hPool
+/// \param size number of bytes to allocate
+/// \return Pointer to the allocated memory
+///
+void *umaPoolMalloc(uma_memory_pool_handle_t hPool, size_t size);
+
+///
+/// \brief Allocates size bytes of uninitialized storage of the specified hPool
+/// with
```
jandres742, line 46: is this an extra line break that we can remove?

source/common/unified_memory_allocation/include/uma/memory_pool.h
```diff
@@ -0,0 +1,95 @@
...
+/// \brief Reallocates memory of the specified hPool
+/// \param hPool specified memory hPool
+/// \param ptr pointer to the memory block to be reallocated
+/// \param size new size for the memory block in bytes
+/// \return Pointer to the allocated memory
+///
+void *umaPoolRealloc(uma_memory_pool_handle_t hPool, void *ptr, size_t size);
+
+///
+/// \brief obtain size of block of memory allocated from the pool
```
jandres742, line 77: "Obtains" to follow format from previous comments above

source/common/unified_memory_allocation/include/uma/memory_pool.h
```diff
@@ -0,0 +1,95 @@
...
+/// \param hPool specified memory hPool
+/// \param ptr pointer to the allocated memory
+/// \return Number of bytes
+///
+size_t umaPoolMallocUsabeSize(uma_memory_pool_handle_t hPool, void *ptr);
+
+///
+/// \brief Free the memory space of the specified hPool pointed by ptr
```
jandres742, line 85: Same: "Frees"

pbalcer:
This repo is meant to contain what we call `common`, which are the shared bits between the loader and all the adapters (but still statically linked with each). This is to allow for adapter implementations to be independent of LLVM, located either in this repo or somewhere else entirely, with the only mandatory dependency being the UR itself.

For the time being, some of the common utilities live in the LLVM repo. Still, as time goes on, we want to successively move them here - so that they are genuinely 'common', shared among not just the adapters that have to live in the LLVM repo for now but also others that might be somewhere else.

source/common/unified_memory_allocation/include/uma/memory_pool_ops.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <uma/base.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// TODO: create macro for intializing this structure (including version fields)
+struct uma_memory_pool_ops_t {
+  uint64_t version_major;
```
pbalcer, line 20: you could create a base struct with versions and use it both here and in the provider.

pbalcer (commented):
Please add a simple test that at least touches every entry point.

source/common/unified_memory_allocation/include/uma/memory_pool.h
igchor: Sure, done, this was just some clang-format quirk

source/common/unified_memory_allocation/include/uma/memory_pool_ops.h
igchor: I've replaced it with a single variable and a macro to encode major/minor.

igchor:
> Please add a simple test that at least touches every entry point.

Done. I've thought about creating some C++ wrappers for memory pool/provider creation, like this: https://github.com/igchor/unified-runtime/commit/da0c5cbff6e8fecb626d6bac32f275ed5760c88e but decided against that for now. Since the library is in C, it would be good to have some pure C tests and the null/trace provider/pool can even be moved from tests to the actual library in the future.

igchor:
> thanks @igchor . could you clarify why are we adding the USM memory allocator to this repo (where we have the loader and spec), instead of directly to LLVM? From comment above, it's not clear to me whether that is just temporary while the other PRs are merged.
> 
> Or this is just a shim layer, that will call each adapter's memory allocator, and that's where we will have all optimizations?

At this moment, the PR only contains header files and the shim layer. The actual memory provider implementation will be provided by each adapter + we can have some generic ones (like OS provider) implemented here.

For L0 memory provider the plan would be to have it alongside the L0 adapter (currently in llvm). As for memory pool implementations, initially, I'd like to use the usm_allocator that is implemented inside llvm. I think it's OK to keep the usm_allocator there until we move adapters here.

jandres742:
> > thanks @igchor . could you clarify why are we adding the USM memory allocator to this repo (where we have the loader and spec), instead of directly to LLVM? From comment above, it's not clear to me whether that is just temporary while the other PRs are merged.
> > Or this is just a shim layer, that will call each adapter's memory allocator, and that's where we will have all optimizations?
> 
> @jandres742 At this moment, the PR only contains header files and the shim layer. The actual memory provider implementation will be provided by each adapter + we can have some generic ones (like OS provider) implemented here.
> 
> For L0 memory provider the plan would be to have it alongside the L0 adapter (currently in llvm). As for memory pool implementations, initially, I'd like to use the usm_allocator that is implemented inside llvm. I think it's OK to keep the usm_allocator there until we move adapters here.

Thanks @igchor . Yes, that is a good plan.

igchor:
In my previous comment, I wrote that we could keep the usm_allocator (foundation for our generic memory pool implementation) inside llvm since the L0 adapter is in llvm. However, what about CUDA adapter?

@kbenzie @alycm are you already working on the CUDA adapter? Will the implementation be done here or in llvm repo? If it's here, then perhaps having the memory pool implementation in UR repo would actually be better,

test/unified_memory_allocation/memoryPool.cpp
```diff
@@ -0,0 +1,53 @@
...
+            mallocs++;
+        } else if (std::string(name) == "calloc") {
+            callocs++;
+        } else if (std::string(name) == "realloc") {
+            reallocs++;
+        } else if (std::string(name) == "aligned_malloc") {
+            aligned_mallocs++;
+        } else if (std::string(name) == "malloc_usable_size") {
+            malloc_usable_sizes++;
+        } else {
```
pbalcer, line 28: else if for free and assert in else please ;)

source/common/unified_memory_allocation/src/memory_pool.c
```diff
@@ -0,0 +1,69 @@
...
+
+enum uma_result_t umaPoolCreate(struct uma_memory_pool_ops_t *ops, void *params,
+                                uma_memory_pool_handle_t *hPool) {
+  uma_memory_pool_handle_t pool = malloc(sizeof(struct uma_memory_pool_t));
+  if (!pool) {
+    return UMA_RESULT_RUNTIME_ERROR;
+  }
+
+  // TODO: handle any version mismatches
+  pool->ops = *ops;
```
vinser52, line 26: I would suggest adding assert for now. Since it is the first version we do not expect a version mismatch so far. And in future, the assert will help to not forget to add required support for backward compatibility.

The same is applicable for provider ops.

vinser52 (changes requested):

source/common/unified_memory_allocation/include/uma/memory_pool_ops.h
```diff
@@ -0,0 +1,47 @@
...
+#define UMA_MEMORY_POOL_OPS_H 1
+
+#include <uma/base.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// TODO: create macro for intializing this structure (including version fields)
+struct uma_memory_pool_ops_t {
```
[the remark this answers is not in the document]
igchor: I've added the comment but I'm not sure about the naming. Right now, we don't use the word "heap" anywhere. If we'd mention it here, we would need to explain what it is exactly and what is the relation between pool and heap. And I think it's not exactly clear because those operations we define here are actually responsible for "pooling" (managing memory from the provider, etc.)

 I think I'd rather rename `pool` to `heap` if you think that makes more sense.

source/common/unified_memory_allocation/include/uma/memory_pool_ops.h
```diff
@@ -0,0 +1,47 @@
...
+  uint64_t version;
+
+  ///
+  /// \brief Intializes memory pool
+  /// \param params pool-specific params
```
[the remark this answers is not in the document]
igchor: Please see my previous answer.

source/common/unified_memory_allocation/include/uma/memory_pool_ops.h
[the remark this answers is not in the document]
vinser52: Ok, I agree with you regarding naming. let's go with pool.

source/common/unified_memory_allocation/include/uma/memory_pool_ops.h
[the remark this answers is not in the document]
vinser52: Resolved

vinser52 (approved):

source/common/unified_memory_allocation/include/uma/base.h
```diff
@@ -0,0 +1,59 @@
...
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define UMA_VERSION_MAJOR ((uint64_t)0)
+#define UMA_VERSION_MINOR ((uint64_t)9)
+
+#define UMA_VERSION_ENCODE(major, minor) (((uint64_t) major << 32) | (uint64_t) minor)
```
vinser52, line 22: I think the user should not be responsible for passing version numbers to this macro. This macro should return the version based on the values specified in this header.

igchor: Yeah, that was my initial implementation but I'm not sure. In some cases, it might be useful to encode/decode particular (not necessarily newest) version. It's similar to how it's done for UR: https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L4006

Perhaps we should just have those generic macros to decode/encode + something like 'UMA_VERSION_CURRENT'?

igchor: Done. I've also make the naming of VERSION macros consistent with the ones in UR.

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -0,0 +1,63 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct uma_memory_provider_t *uma_memory_provider_handle_t;
+
+///
+/// \brief Creates new memory provider
+/// \param ops instance of uma_memory_provider_ops_t
+/// \param params pointer to provider-specific parameters
```
PatKamin, line 24: Missing `*hProvider` param description.

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -0,0 +1,63 @@
...
+/// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure
+///
+enum uma_result_t
+umaMemoryProviderCreate(struct uma_memory_provider_ops_t *ops, void *params,
+                        uma_memory_provider_handle_t *hProvider);
+
+///
+/// \brief Destroys memory provider
+/// \param hPool handle to the memory provider
```
PatKamin, line 33: hPool -> hProvider

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -0,0 +1,63 @@
...
+                                         void **ptr);
+
+///
+/// \brief Frees the memory space pointed by ptr from the memory provider
+/// \param hProvider handle to the memory provider
+/// \param ptr pointer to the allocated memory
+/// \param size size of the allocation
+///
+enum uma_result_t umaMemoryProviderFree(uma_memory_provider_handle_t hProvider,
+                                        void *ptr, size_t size);
```
PatKamin, line 57: Why including `size` as a param in a `free()` function? Is it for tracking of the used memory inside a memory provider?

source/common/unified_memory_allocation/src/memory_pool.c
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  assert(ops->version == UMA_VERSION_CURRENT);
+
+  pool->ops = *ops;
+
+  void *pool_priv;
+  enum uma_result_t ret = ops->initialize(params, &pool_priv);
+  if (ret != UMA_RESULT_SUCCESS) {
+    return ret;
```
PatKamin, line 33: Free the memory allocated for `pool`.

source/common/unified_memory_allocation/src/memory_provider.c
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  assert(ops->version == UMA_VERSION_CURRENT);
+
+  provider->ops = *ops;
+
+  void *provider_priv;
+  enum uma_result_t ret = ops->initialize(params, &provider_priv);
+  if (ret != UMA_RESULT_SUCCESS) {
+    return ret;
```
PatKamin, line 35: Free memory allocated for `provider` before returning.

test/unified_memory_allocation/common/main.cpp
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: MIT
+
+#include <gtest/gtest.h>
+
+int main(int argc, char **argv) {
```
PatKamin, line 6: Adding a custom `main()` function that doesn't do anything special is not necessary.

source/common/unified_memory_allocation/include/uma/memory_pool_ops.h
```diff
@@ -0,0 +1,51 @@
...
+extern "C" {
+#endif
+
+
+/// This structure comprises function pointers used by corresponding  umaPool*
+/// calls. Each memory pool implementation should initialize all function pointers.
+struct uma_memory_pool_ops_t {
+  /// Version of the ops structure.
+  /// Should be initialized using UMA_VERSION_CURRENT
+  uint32_t version;
```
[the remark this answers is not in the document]
vinser52: A client that uses UMA might be compiled with a different version of UMA. This field allows UMA to understand which version of ops structure is used by the client to support ABI compatibility.

source/common/unified_memory_allocation/include/uma/memory_provider.h
vinser52: Not all memory providers can determine the size of the memory pointed by `ptr`. E.g. `munmap` requires the size.
It is the heap manager's responsibility to track allocation sizes.

source/common/unified_memory_allocation/include/uma/memory_pool_ops.h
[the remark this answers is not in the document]
PatKamin: And this is specifically to allow for future changes in the `ops` structure, like adding new fields or removing existing ones? In such a case, `ops` created with an older UMA version will result in a failure when creating a new memory pool, right?

source/common/unified_memory_allocation/include/uma/memory_provider.h
PatKamin: Got it, thanks for the explanation.

source/common/unified_memory_allocation/include/uma/memory_pool_ops.h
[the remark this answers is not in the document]
vinser52: yeah, the version field is introduced to support compatibility with previous versions of the UMA in case ops structure is modified. Regarding what should be done in case of version mismatch, if ops structure is just extended with some field and we can use some default behavior than we do not need to return the error. But in general you are right, based on version field we need to implement some logic in case of version mismatch.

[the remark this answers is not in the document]
PatKamin: There is also another implication of this. We can either bump UMA version **only** when providing significant changes in ops structure, ie. ones that makes the old ops unusable or we can check for version range (this would be cumbersome).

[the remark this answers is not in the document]
vinser52: Yeah, it depends on the meaning of the `version` field in the ops structure.

- If it contains the version of the library, then the library version is increased on each release and we should check for the because ops structure might stay unchanged between versions).
- If the `version` field contains the version of the ops structure itself then the version is increased only when the ops structure is changed.

test/unified_memory_allocation/common/main.cpp
igchor: What do you mean? This is necessary for gtest I belive.

PatKamin: Gtest will provide it's main function and run all tests by default. See new tests added here, ie: https://github.com/oneapi-src/unified-runtime/pull/154/commits/c65c4417dfc1845bd9796f463d9589d0bf2c77ff

igchor: I see, thanks. Removed.

kbenzie (approved):
LGTM

PR #184, opened 2023-01-24
[ur] Disable testing against null adapter

veselypeta:
We have decided that there is no purpose in running the conformance tests against the null adapter since it will never be a conformant implementation.
This MR:
 * Removes the job to run tests against the null adapter
 * Re-enables disabled tests

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -34,10 +34,6 @@ jobs:
     - name: Build
       run: cmake --build ${{github.workspace}}/build
 
-    - name: Test
-      working-directory: ${{github.workspace}}/build
-      run: ctest -C ${{env.BUILD_TYPE}} --output-on-failure
-
   windows-build:
     name: Build - Windows
     runs-on: windows-latest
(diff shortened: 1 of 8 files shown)


pbalcer:
I think that it would be better to remove the null adapter from being loaded by default: https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/linux/platform_discovery_lin.cpp#L19 (same on Windows), and then keep the tests in CI. That way, you'd have no platforms and tests should (?) pass.
Alternatively, we could have an additional env variable that forces the loader to use only one single adapter implementation (or change the behavior of the current env variable to clear the list of adapters...). This would let us have separate CI jobs that run tests with specified adapters, regardless of what was built.

veselypeta:
> I think that it would be better to remove the null adapter from being loaded by default: https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/linux/platform_discovery_lin.cpp#L19 (same on Windows), and then keep the tests in CI. That way, you'd have no platforms and tests should (?) pass. Alternatively, we could have an additional env variable that forces the loader to use only one single adapter implementation (or change the behavior of the current env variable to clear the list of adapters...). This would let us have separate CI jobs that run tests with specified adapters, regardless of what was built.

It will just fail [here](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/source/environment.cpp#L100) instead

pbalcer (approved):
Makes sense, LGTM.

I also see there's a way to filter platforms in the test framework itself, so my earlier point about using env variable for per-adapter CI jobs is moot.

Issue #144, opened 2022-12-19
add automatic api validation support


pbalcer:
Implement an optional validation API layer that verifies the correctness of UR calls at runtime. In addition to performing basic sanity checks (e.g., checking if provided handles are valid), it should eventually be extended to perform more complex semantic checks, like checking for missing arguments in kernels or tracking object ownership.
This implementation can be modeled after a similar functionality in [level-zero](https://github.com/oneapi-src/level-zero/tree/master/source/layers/validation).

Overall, we need to:
1. Implement validation layer infrastructure (controlled with an env variable, `UR_VALIDATION=none/basic/complex`?)
2. Analyze the UR API and create a documentation of useful checks that the validation layer could perform.
3. Implement those checks, starting from high-impact ones.

pbalcer:
Done with #318 #221

PR #188, opened 2023-01-26
Add initial validation layer

kswiecicki:
This PR adds an initial validation layer to build infrastructure and parameter validation functionality.
The validation layer can be controlled through the following environmental variables:
- UR_ENABLE_VALIDATION_LAYER
- UR_ENABLE_PARAMETER_VALIDATION

Templates and the solution comes from: https://github.com/oneapi-src/level-zero.

### DIFF
--- a/examples/hello_world/hello_world.cpp
+++ b/examples/hello_world/hello_world.cpp
@@ -5,16 +5,41 @@
  * SPDX-License-Identifier: MIT
  *
  */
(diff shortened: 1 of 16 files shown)


igchor (commented):
Would you add some tests to this PR? We can probably work on full/better coverage later, but it would be good to have some basic ones here.

Issue #119, opened 2022-12-12
Add missing `ur_device_partition_property_flag_t` entries


FranklandJack:
# Task

PI defined a [`pi_device_partition_property`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L629) enumeration (as an `intptr_t`) that can take the following values:
* `PI_DEVICE_PARTITION_EQUALLY`
* `PI_DEVICE_PARTITION_BY_COUNTS`
* `PI_DEVICE_PARTITION_BY_COUNTS_LIST_END`
* `PI_DEVICE_PARTITION_BY_AFFINITY_DOMAIN`
* `PI_EXT_INTEL_DEVICE_PARTITION_BY_CSLICE`

UR defines a [`ur_device_partition_property_flag_t`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L2954) enumeration which takes the following values:
* `UR_DEVICE_PARTITION_PROPERTY_FLAG_EQUALLY`
* `UR_DEVICE_PARTITION_PROPERTY_FLAG_BY_COUNTS`
* `UR_DEVICE_PARTITION_PROPERTY_FLAG_BY_AFFINITY_DOMAIN`
* `UR_DEVICE_PARTITION_PROPERTY_FLAG_FORCE_UINT32`

Of the PI values `PI_DEVICE_PARTITION_BY_COUNTS_LIST_END` (which is used [here](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/source/detail/device_impl.cpp#L192)) and `PI_EXT_INTEL_DEVICE_PARTITION_BY_CSLICE` (which is used [here](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/source/detail/device_impl.cpp#L265)) are missing from UR.

# Task

Add `UR_DEVICE_PARTITION_BY_COUNTS_LIST_END` and `PI_EXT_INTEL_DEVICE_PARTITION_BY_CSLICE` to Unified Runtime so that it meets the requirements of implementing the DPC++ runtime.

FranklandJack:
[...]

PR #189, opened 2023-01-26
[ur] Clarify device partition properties

### LINKED ISSUE #183: Clarify how device partitioning properties are terminated
In https://github.com/oneapi-src/unified-runtime/blob/00b8fd6aef508d8bd83089322d9ae0b84f647e6c/include/ur_api.h#L3138
///< [in] null-terminated array of <property, value> pair of the requested partitioning.

But the input is an array of struct {uint32_t, uint32_t}, so there is no pointer to indicate end of the array.

veselypeta:
Open to suggestions on this one. 

On the WG call we decided that we should just revert how we specify device properties to be just an `intptr_t` - so i've created a typedef for `ur_device_partition_property_t` to `intptr_t`. However I have added an additional enum to specify the partitioning options `ur_device_partition_t` - which maybe isn't ideal. We could specify the values with macros but I'm not sure what I prefer.

Closes #183 & #119

### DIFF
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -230,7 +230,7 @@ etors:
     - name: PARENT_DEVICE
       desc: "$x_device_handle_t: return parent device handle"
     - name: PARTITION_PROPERTIES
-      desc: "uint32_t: return a bit-field of partition properties $x_device_partition_property_flags_t"
+      desc: "$x_device_partition_property_t[]: Returns the list of partition types supported by the device"
     - name: PARTITION_MAX_SUB_DEVICES
       desc: "uint32_t: maximum number of sub-devices when the device is partitioned"
     - name: PARTITION_AFFINITY_DOMAIN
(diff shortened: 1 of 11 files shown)


scripts/core/device.yml
```diff
@@ -395,10 +393,10 @@ params:
       name: hDevice
       desc: |
             [in] handle of the device to partition.
-    - type: "$x_device_partition_property_value_t*"
+    - type: const $x_device_partition_property_t*
       name: Properties
       desc: |
-            [in] null-terminated array of <property, value> pair of the requested partitioning.
+            [in] null-terminated array of <property, value> pairs.
```
kbenzie, line 399: I think this description needs to mention what enumeration is expected to be used, otherwise there would be no relation between `ur_device_partition_t` and this list of properties.

```suggestion
            [in] null-terminated array of <$_device_partition_t enum, value> pairs.
```

scripts/core/device.yml
```diff
@@ -395,10 +393,10 @@ params:
       name: hDevice
       desc: |
             [in] handle of the device to partition.
-    - type: "$x_device_partition_property_value_t*"
+    - type: const $x_device_partition_property_t*
       name: Properties
```
kbenzie, line 397: this is a pointer argument so should have the `p` prefix

```suggestion
      name: pProperties
```

kbenzie (changes requested):
A couple of tweaks, otherwise looks good.

kbenzie (approved):
LGTM

PR #190, opened 2023-01-26
[ur] Remove prop-value structs

veselypeta:
PR #189 fixed how we specify properties for entry-points and we moved away from using the prop-value struct so it makes sense to make this consistent with the other 2 remaining entry-points that used this style `urQueueCreate` and `urSamplerCreate`

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -577,6 +577,11 @@ class ur_queue_flags_t(c_int):
 
 
 ###############################################################################
+## @brief Queue property type
+class ur_queue_property_t(c_intptr_t):
+    pass
+
+###############################################################################
 ## @brief Queue Properties
 class ur_queue_properties_v(IntEnum):
     FLAGS = -1                                      ## [::ur_queue_flags_t]: the bitfield of queue flags
@@ -586,14 +591,6 @@ class ur_queue_properties_t(c_int):
     def __str__(self):
         return str(ur_queue_properties_v(self.value))
 
-
-###############################################################################
-## @brief Queue property value
-class ur_queue_property_value_t(Structure):
-    _fields_ = [
-        ("propertyType", ur_queue_properties_t),                        ## [in] queue property
-        ("propertyValue", c_ulong)                                      ## [in] queue property value
-    ]
 
 ###############################################################################
 ## @brief Get sample object information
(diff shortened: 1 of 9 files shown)


include/ur.py
```diff
@@ -1509,9 +1503,9 @@ class ur_kernel_dditable_t(Structure):
 ###############################################################################
 ## @brief Function-pointer for urSamplerCreate
 if __use_win_types:
-    _urSamplerCreate_t = WINFUNCTYPE( ur_result_t, ur_context_handle_t, POINTER(ur_sampler_property_value_t), POINTER(ur_sampler_handle_t) )
+    _urSamplerCreate_t = WINFUNCTYPE( ur_result_t, ur_context_handle_t, POINTER(ur_samples_property_t), POINTER(ur_sampler_handle_t) )
```
jandres742, line 1506: is "samples" instead of "sampler" correct?

include/ur.py
```diff
@@ -1509,9 +1503,9 @@ class ur_kernel_dditable_t(Structure):
...
 else:
-    _urSamplerCreate_t = CFUNCTYPE( ur_result_t, ur_context_handle_t, POINTER(ur_sampler_property_value_t), POINTER(ur_sampler_handle_t) )
+    _urSamplerCreate_t = CFUNCTYPE( ur_result_t, ur_context_handle_t, POINTER(ur_samples_property_t), POINTER(ur_sampler_handle_t) )
```
jandres742, line 1508: same

include/ur_api.h
```diff
@@ -2522,6 +2517,10 @@ typedef enum ur_sampler_properties_t
 
 } ur_sampler_properties_t;
 
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Sampler Properties type
+typedef intptr_t ur_samples_property_t;
```
jandres742, line 2522: same

include/ur.py
veselypeta: done.

include/ur.py
veselypeta: done.

include/ur_api.h
veselypeta: done.

jandres742 (commented):
+1

kbenzie (approved):
LGTM

PR #191, opened 2023-01-26
[ur] Correct name for urEnqueueUSMMemAdvise

### LINKED ISSUE #187: MemAdvice being used instead of MemAdvise as in other APIs
Advise (verb) instead of advice (noun) is used in other APIs ("MemAdvise sets an Advice"). See

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ge37112fc1ac88d0f6bab7a945e48760a

https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=advise#_CPPv428zeCommandListAppendMemAdvise24ze_command_list_handle_t18ze_device_handle_tPKv6size_t18ze_memory_advice_t

But here we have only the noun. Minor thing but still :)

https://github.com/oneapi-src/unified-runtime/blob/746786d6c223005335de7939127c413a8411fb9c/include/ur_api.h#L1296 

Dont know if we could change it w/o breaking compatibility.

veselypeta:
Renames `urEnqueueUSMMemAdvice` to `urEnqeueuUSMMemAdvise`

Closes #187

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -1783,11 +1783,11 @@ else:
     _urEnqueueUSMPrefetch_t = CFUNCTYPE( ur_result_t, ur_queue_handle_t, c_void_p, c_size_t, ur_usm_migration_flags_t, c_ulong, POINTER(ur_event_handle_t), POINTER(ur_event_handle_t) )
 
 ###############################################################################
-## @brief Function-pointer for urEnqueueUSMMemAdvice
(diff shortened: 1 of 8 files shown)


jandres742 (commented):
+1

kbenzie (approved):
LGTM

PR #192, opened 2023-01-30
[ur] HostPtr should be optional in urMemBufferCreate

veselypeta:
`hostPtr` argument in urMemBufferCreate is not required for all uses of this entry-point and is only required when specific flags are set such as `UR_MEM_FLAGS_USE_HOST_POINTER`.
This also clarifies the spec that `UR_RESULT_ERROR_INVALID_HOST_PTR` should be returned instead of `UR_RESULT_ERROR_INVALID_NULL_POINTER`


jandres742 (commented):
+1

kbenzie (approved):
LGTM

PR #195, opened 2023-01-30
Change C++ header files extension to hpp

igchor:
This makes it easy to distinguish them from from C headers. Until now, they both had '.h' extension.

IDEs/editors usually just look at the extension so this helps with detecting the language automaticaly.


PatKamin (commented):
Remove new empty file source/common/ur_uma_helpers.h

pbalcer:
You need to change the templates, not the generated files ;)

veselypeta:
You might want to add "*.hpp" to the formatter here https://github.com/oneapi-src/unified-runtime/blob/main/CMakeLists.txt#L94

igchor:
> Remove new empty file source/common/ur_uma_helpers.h

Done.

igchor:
> You need to change the templates, not the generated files ;)

Done.

igchor:
> You might want to add "*.hpp" to the formatter here https://github.com/oneapi-src/unified-runtime/blob/main/CMakeLists.txt#L94

Done.

PR #196, opened 2023-01-31
[ur] Adds usm testsuite

veselypeta:
Adds testsuite for USM testsuite to ur. Tests the following entrypoints

* `urMemFree`
* `urMemGetMemAllocInfo`
* `urUSMHostAlloc`
* `urUSMDeviceAlloc`
* `urUSMSharedAlloc`

There are some additional fixes to the spec as part of implementing this testsuite
- in `urMemGetMemAllocInfo` - `pPropValue` & `pPropValueSizeRet` arguments should be optional
- defined the `ur_usm_type_t` enumeration which was missing as the return type for the query of `UR_MEM_ALLOC_INFO_ALLOC_TYPE`


test/conformance/usm/urUSMSharedAlloc.cpp
```diff
@@ -0,0 +1,2 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: MIT
```
kbenzie, line 2: This is missing files for testing `urMemFree` and `urMemGetMemAllocInfo` which are both int he [USM section](https://oneapi-src.github.io/unified-runtime/core/api.html#usm). There's perhaps some inconsistency with the naming of these entry points.

kbenzie (changes requested):

test/conformance/usm/urUSMSharedAlloc.cpp
veselypeta: Yes I was confused by this slightly - I suppose they should be called `urUSMMemFree` & `urUSMGetMemAllocInfo`?

kbenzie: Something along those lines, although the `M` in `USM` stands for memory so not including `Mem` would remove some reduncandy form the names. An entry point name change is deffo out of scope for this PR though.

veselypeta: https://github.com/oneapi-src/unified-runtime/issues/201

kbenzie (approved):
LGTM

PR #197, opened 2023-01-31
[ur] Add clang-format check in CI

veselypeta:


CMakeLists.txt
```diff
@@ -96,12 +96,20 @@ file(GLOB_RECURSE format_src
 
 # Exclude files from build directory to ensure they're not passed to the formatter
 list(FILTER format_src EXCLUDE REGEX "${CMAKE_CURRENT_BINARY_DIR}.")
+# Exclude generated files from formatting
+list(FILTER format_src EXCLUDE REGEX "${CMAKE_PROJECT_DIR}/include/.")
```
pbalcer, line 100: I don't think we should be excluding the generated files. Instead, these files should be formatted immediately after being generating, so that they don't fail the `check-generated` check. Or, alternatively, we could manually fix the templates so that are correct upfront, but that might be a pita...

.github/workflows/cmake.yml
```diff
@@ -16,11 +16,19 @@ jobs:
...
-    - name: Install apt package
+    - name: Install apt packages
       run: |
         sudo apt-get update
-        sudo apt-get install -y doxygen
+        sudo apt-get install -y doxygen libncurses5
+  
+    - name: Install clang-format-10
+      run: |
+        cd ${{github.workspace}}
```
pbalcer, line 26: I was thinking about using this https://github.com/ssciwr/clang-format-wheel, which should be easier to setup.

veselypeta: Nice - I hadn't seen this. The default `apt install clang-format` installs the latest version (`14.0` I believe) and does not have a target for `clang-format-10`. Hence this solution

CMakeLists.txt
veselypeta: Sounds good - I've made the generate target run `cppformat`

include/ur_api.h
```diff
@@ -133,137 +133,171 @@ typedef struct ur_mem_handle_t_ *ur_mem_handle_t;
...
+    UR_RESULT_ERROR_OVERLAPPING_REGIONS =
+        62, ///< [Validation] copy operations do not support overlapping regions
+            ///< of memory
+    UR_RESULT_ERROR_INVALID_HOST_PTR = 63, ///< Invalid host pointer
+    UR_RESULT_ERROR_INVALID_USM_SIZE = 64, ///< Invalid USM size
+    UR_RESULT_ERROR_OBJECT_ALLOCATION_FAILURE =
+        65, ///< Objection allocation failure
+    UR_RESULT_ERROR_ADAPTER_SPECIFIC =
+        66, ///< An adapter specific warning/error has been reported and can be
+            ///< retrieved via the urGetLastResult entry point.
```
veselypeta, line 255: Maybe it would be best to exclude this file from formatting - I think it was better formatted before

pbalcer: yea, I agree.

cmake/helpers.cmake
```diff
@@ -36,3 +36,23 @@ function(add_cppformat name)
 
     add_dependencies(cppformat cppformat-${name})
 endfunction()
+
+function(check_cppformat name)
```
pbalcer, line 40: I don't think this is necessary, we can verify if the code is formatted correctly by just formatting it in CI and looking if that caused any changes. So just like we do it for code generation. See [this line](https://github.com/oneapi-src/unified-runtime/blob/main/CMakeLists.txt#L173) (the docker comment is no longer relevant since we found the pip clang-format package).

PR #200, opened 2023-02-01
[ur] remove platform init flags for urInit

### LINKED ISSUE #21: Clarify UR_PLATFORM_INIT_FLAG_LEVEL_ZERO platform init flag
[`urInit`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L4093) accepts a `ur_platform_init_flags_t platform_flags` argument. Currently the only valid value for this flag is [UR_PLATFORM_INIT_FLAG_LEVEL_ZERO](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L4053).

The documentation for this flag isn't particularly clear:
```
initialize Unified Runtime platform drivers
```
But its name makes it sound like it is specific to a level zero adapter (although perhaps this is just an artifact of Unified Runtime previously being called Level Zero Runtime?)

Either way, we should either rename this value, or clarify its meaning.

veselypeta:
`ur_platform_init_flags` are now removed. Originally (I believe - we have little documentation of the ur spec before we got it) this was a flag to select the desired backend. However, this is now handled with the loader and the backend is selected through an environment variable. Moreover, the loader actually does nothing with these parameters [here](https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_lib.cpp#L25). This then also raises the question if we need `ur_device_init_flags_t`....

Closes #21

### DIFF
--- a/examples/hello_world/hello_world.cpp
+++ b/examples/hello_world/hello_world.cpp
@@ -20,7 +20,7 @@ int main(int argc, char *argv[]) {
     ur_device_handle_t pDevice = nullptr;
 
     // Initialize the platform
-    status = urInit(0, 0);
+    status = urInit(0);
     if (status != UR_RESULT_SUCCESS) {
         std::cout << "urInit failed with return code: " << status << std::endl;
         return 1;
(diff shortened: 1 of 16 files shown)


pbalcer (approved):

kbenzie (approved):
LGTM

PR #203, opened 2023-02-01
[RFC] Add C++ helpers for UMA

igchor:
To allow defining memory providers and pools as C++ classes. This avoids having to deal with packing input parameters and casting void* pool/provider to the appropriate type on each call. I assume most (or all) providers in UR to be implemented entirely in C++.

@vinser52 is my assumption correct? do you think it's reasonable to add those helpers?

Example usage would be:
```c++
class provider {
public:
    provider(ze_context_handle_t context):
        context(context), protection(protection) {}

    enum uma_result_t alloc(size_t size, size_t alignment, void **ptr) {
        return zeStatusToUMA(zeMemAllocHost(context, NULL, size, alignment, ptr));
    }

    enum uma_result_t free(void *ptr, size_t size) {
        ...
    }
private:
    ze_context_handle_t context;
};

auto providerHandle = memoryProviderMakeUnique<provider>(context);
umaMemoryProviderAlloc(providerHandle.get(), size, alignment, ptr);
```

I can add some tests if we agree on having those helpers here.

### DIFF
--- a/source/common/uma_helpers.hpp
+++ b/source/common/uma_helpers.hpp
@@ -1,0 +1,130 @@
+/*
+ *
+ * Copyright (C) 2023 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#ifndef UR_UMA_HELPERS_H
+#define UR_UMA_HELPERS_H 1
+
+#include <uma/memory_pool.h>
+#include <uma/memory_pool_ops.h>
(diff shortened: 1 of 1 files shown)


igchor:
> If those helpers will be just implementation details of UR I am fine. In the case of public API, we have to be careful and go through discussion and approval on oneMemory WG

Yes, that's the idea. This should not be part of the public API (that's why I haven't put it in uma/include).

PR #209, opened 2023-02-03
[uma] allow building UMA as shared library

igchor:
This might be useful for prototype integration with OpenMP

### DIFF
--- a/source/common/unified_memory_allocation/CMakeLists.txt
+++ b/source/common/unified_memory_allocation/CMakeLists.txt
@@ -1,10 +1,21 @@
 # Copyright (C) 2023 Intel Corporation
 # SPDX-License-Identifier: MIT
 
-add_library(unified_memory_allocation STATIC
+set(UMA_SOURCES
     src/memory_pool.c
     src/memory_provider.c
 )
+
+if(UMA_BUILD_SHARED_LIBRARY)
+    message(WARNING "Unified Memory Allocation is still an early work in progress."
+                     "There are no API/ABI backward compatibility guarantees. There will be breakages."
+                     "Do not use the shared library in production software.")
+    add_library(unified_memory_allocation SHARED
+        ${UMA_SOURCES})
+else()
+    add_library(unified_memory_allocation STATIC
+        ${UMA_SOURCES})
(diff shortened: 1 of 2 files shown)


source/common/unified_memory_allocation/CMakeLists.txt
```diff
@@ -1,11 +1,19 @@
...
 # SPDX-License-Identifier: MIT
 
-add_library(unified_memory_allocation STATIC
+set(UMA_SOURCES
     src/memory_pool.c
     src/memory_provider.c
 )
 
+if(UMA_BUILD_SHARED_LIBRARY)
+    add_library(unified_memory_allocation SHARED
```
pbalcer, line 10: `message(WARNING, "Unified Memory Allocation is still an early work in progress. There are no API/ABI backward compatibility guarantees. There will be breakages. Do not use the shared library in production software.")`

Or something to that effect.

source/common/unified_memory_allocation/CMakeLists.txt
```diff
@@ -1,11 +1,19 @@
 # Copyright (C) 2023 Intel Corporation
 # SPDX-License-Identifier: MIT
 
-add_library(unified_memory_allocation STATIC
+set(UMA_SOURCES
     src/memory_pool.c
     src/memory_provider.c
 )
 
+if(UMA_BUILD_SHARED_LIBRARY)
```
kbenzie, line 9: Is `USM_BUILD_SHARED_LIBRARY` only defined when the user sets it during CMake configuration?

Making it an [option](https://cmake.org/cmake/help/latest/command/option.html) would be nice.

source/common/unified_memory_allocation/CMakeLists.txt
igchor: Done.

source/common/unified_memory_allocation/CMakeLists.txt
igchor: Right, done.

pbalcer (approved):

vinser52 (approved):

PR #211, opened 2023-02-06
enable all warnings and add developer mode

pbalcer:
For now, the developer mode only turns warnings into errors, but over time it will also encompass other various checks that we don't want to happen in
normal builds.

I also did a small cleanup of the main cmake.
The UR_ENABLE_TESTS option is now BUILD_TESTS.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,10 +1,6 @@
 name: CMake
 
-on:
-  push:
-    branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
+on: [push, pull_request]
 
 env:
   BUILD_TYPE: Release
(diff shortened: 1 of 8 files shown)


pbalcer:
I also wanted to enable asan and ubsan, but that's going to be its own PR since it's more work than I expected.

veselypeta (approved):
LGTM

DamianDuy:
From what I see on Windows check on github, there's a lot of linker warnings in logs. Shouldn't this fail since the -WX option is on?

.github/workflows/cmake.yml
```diff
@@ -26,7 +22,7 @@ jobs:
       run: pip install -r third_party/requirements.txt
 
     - name: Configure CMake
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DUR_ENABLE_TESTS=ON
+      run: cmake -B ${{github.workspace}}/build -DDEVELOPER_MODE=ON -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DBUILD_TESTS=ON
```
PatKamin, line 25: Add DEVELOPER_MODE to the list of cmake options in the README (perhaps with info that it enables Werror compiler flag?)

pbalcer: done

PR #212, opened 2023-02-06
add support for ubsan/asan/msan in cmakes

pbalcer:
depends on #211

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,9 +17,12 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # Build Options
-option(DEVELOPER_MODE "enable developer checks, treats warnings as errors" OFF)
-option(BUILD_TESTS "Build unit tests." ON)
-option(FORMAT_CPP_STYLE "format code style of C++ sources" OFF)
+option(UR_DEVELOPER_MODE "enable developer checks, treats warnings as errors" OFF)
+option(UR_BUILD_TESTS "Build unit tests." ON)
+option(UR_FORMAT_CPP_STYLE "format code style of C++ sources" OFF)
+option(UR_USE_ASAN "enable AddressSanitizer" OFF)
+option(UR_USE_UBSAN "enable UndefinedBehaviorSanitizer" OFF)
+option(UR_USE_MSAN "enable MemorySanitizer" OFF)
 
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
@@ -41,7 +44,8 @@ endif()
 # CXX flags setup
 if(NOT MSVC)
     add_compile_options(-fPIC -Wall)
-    if(DEVELOPER_MODE)
+    if(UR_DEVELOPER_MODE)
+        add_compile_options(-fno-omit-frame-pointer)
         add_compile_options(-Werror)
     endif()
 elseif(MSVC)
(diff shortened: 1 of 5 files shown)


source/common/ur_util.hpp
```diff
@@ -13,6 +13,28 @@
...
+#define SANITIZER_ADDRESS
+#endif
+
+/* define for running with memory sanitizer */
+#if CLANG_HAS_FEATURE(memory_sanitizer)
+#define SANITIZER_MEMORY
+#endif
+
+/* define for running with any sanitizer runtime */
+#if defined(SANITIZER_MEMORY) || defined(SANITIZER_ADDRESS)
```
veselypeta, line 34: what about UBSAN?

pbalcer: It does not require that we avoid `RTLD_DEEPBIND`. GCC also doesn't have `__SANITIZE_ADDRESS__`, and clang only recently added `__has_feature(undefined_behavior_sanitizer)`, so there's no reliable way of checking this anyway.

veselypeta: 👍

veselypeta (approved):

CMakeLists.txt
```diff
@@ -56,6 +60,22 @@ elseif(MSVC)
     endif()
 endif()
 
+if(USE_ASAN)
+    add_sanitizer_flag(adress)
```
PatKamin, line 64: adress -> address

PatKamin (approved):

CMakeLists.txt
```diff
@@ -13,6 +13,17 @@ include(helpers)
...
 
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+# Build Options
+option(DEVELOPER_MODE "enable developer checks, treats warnings as errors" OFF)
+option(BUILD_TESTS "Build unit tests." ON)
+option(FORMAT_CPP_STYLE "format code style of C++ sources" OFF)
+option(USE_ASAN "enable AddressSanitizer" OFF)
+option(USE_UBSAN "enable UndefinedBehaviorSanitizer" OFF)
```
kbenzie, line 24: ASan and UBSan can be enabled at the same time, e.g. `-fsanitize=address,undefined`, what happens if `USE_ASAN=ON` and `USE_UBSAN=ON`?

CMakeLists.txt
```diff
@@ -13,6 +13,17 @@ include(helpers)
...
+option(USE_MSAN "enable MemorySanitizer" OFF)
```
kbenzie, line 25: I feel these CMake options should have the `UR_` prefix since we integrate into other projects we should avoid using option names which might conflict with those projects. Taking LLVM as an example, all options in that project have the `LLVL_` prefix unless they are options builtin to CMake.

cmake/helpers.cmake
```diff
@@ -36,3 +36,20 @@ function(add_cppformat name)
...
+
+	check_cxx_compiler_flag("-fsanitize=${flag}" CXX_HAS_SANITIZER)
+	if(CXX_HAS_SANITIZER)
+		add_compile_options(-fsanitize=${flag})
+		add_link_options(-fsanitize=${flag})
+	else()
+		message("${flag} sanitizer not supported")
+	endif()
+
+	set(CMAKE_REQUIRED_LIBRARIES ${SAVED_CMAKE_REQUIRED_LIBRARIES})
```
kbenzie, line 54: Tabs instead of spaces.

source/common/ur_util.hpp
```diff
@@ -13,6 +13,28 @@
...
 
+/* for compatibility with non-clang compilers */
+#if defined(__has_feature)
+#define CLANG_HAS_FEATURE(x) __has_feature(x)
+#else
+#define CLANG_HAS_FEATURE(x) 0
+#endif
+
+/* define for running with address sanitizer */
+#if CLANG_HAS_FEATURE(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
```
kbenzie, line 24: GCC also supports sanitizers, its not a clang only feature which this seems to imply.

pbalcer: hmm, `HAS_FEATURE_IF_CLANG`? What I wanted to convey is that this macro checks for features if running under cmake - which I think is better than just having something generic like `HAS_FEATURE` or simply defining clang's `__has_feature__` if it's not present.

CMakeLists.txt
pbalcer: it works fine.

cmake/helpers.cmake
pbalcer: done

CMakeLists.txt
pbalcer: Done - I added `UR_`. I removed the `${PROJECT_NAME}` prefix because it was cumbersome and wasn't being used correctly.

kbenzie: Yeah, `${PROJECT_NAME}` prefix is cumbersom.

kbenzie (approved):
LGTM

PR #217, opened 2023-02-07
[ur] Implement UMA memory pool: usmPool

igchor:
This PR moves usm_allocator implementation from intel/llvm repo.
"[ur] Remove OwnZeMemHandle from USMAllocator]" is cherry-picked from: https://github.com/intel/llvm/pull/7853

Remaining commits implement the UMA memory pool.

I wasn't sure what to do with the licence. I left the original one for usm_allocator for now. LLVM files are under Apache license and the UR code is under MIT. @kbenzie @pbalcer should we keep the original license and just remove the mention that those file are part of llvm project?

### DIFF
--- a/source/common/CMakeLists.txt
+++ b/source/common/CMakeLists.txt
@@ -12,9 +12,8 @@ target_include_directories(ur_common INTERFACE
 )
 
 add_subdirectory(unified_memory_allocation)
-target_link_libraries(ur_common INTERFACE unified_memory_allocation ${CMAKE_DL_LIBS} ${PROJECT_NAME}::headers)
-
-target_sources(ur_common INTERFACE uma_helpers.hpp)
+add_subdirectory(uma_pools)
+target_link_libraries(ur_common INTERFACE unified_memory_allocation disjoint_pool ${CMAKE_DL_LIBS} ${PROJECT_NAME}::headers)
 
 if(WIN32)
     target_sources(ur_common
--- a/source/common/uma_pools/disjoint_pool.hpp
+++ b/source/common/uma_pools/disjoint_pool.hpp
@@ -12,66 +12,72 @@
 #include <atomic>
 #include <memory>
 
-// USM system memory allocation/deallocation interface.
-class SystemMemory {
-public:
-  virtual void *allocate(size_t size) = 0;
-  virtual void *allocate(size_t size, size_t aligned) = 0;
-  virtual void deallocate(void *ptr) = 0;
-  virtual ~SystemMemory() = default;
-};
+#include "../uma_helpers.hpp"
 
-class USMLimits {
-public:
-  // Maximum memory left unfreed
-  size_t MaxSize = 16 * 1024 * 1024;
-
-  // Total size of pooled memory
-  std::atomic<size_t> TotalSize = 0;
-};
+namespace usm {
 
 // Configuration for specific USM allocator instance
-class USMAllocatorParameters {
-public:
-  const char *memoryTypeName = "";
+class DisjointPoolConfig {
+  public:
+    DisjointPoolConfig();
 
-  // Minimum allocation size that will be requested from the system.
-  // By default this is the minimum allocation size of each memory type.
-  size_t SlabMinSize = 0;
+    std::string name = "";
 
-  // Allocations up to this limit will be subject to chunking/pooling
-  size_t MaxPoolableSize = 0;
+    struct SharedLimits {
+        SharedLimits() : TotalSize(0) {}
 
-  // When pooling, each bucket will hold a max of 4 unfreed slabs
-  size_t Capacity = 0;
+        // Maximum memory left unfreed
+        size_t MaxSize = 16 * 1024 * 1024;
 
-  // Holds the minimum bucket size valid for allocation of a memory type.
-  size_t MinBucketSize = 0;
+        // Total size of pooled memory
+        std::atomic<size_t> TotalSize;
+    };
 
-  // Holds size of the pool managed by the allocator.
-  size_t CurPoolSize = 0;
+    // Minimum allocation size that will be requested from the system.
+    // By default this is the minimum allocation size of each memory type.
+    size_t SlabMinSize = 0;
 
-  // Whether to print pool usage statistics
-  int PoolTrace = 0;
+    // Allocations up to this limit will be subject to chunking/pooling
+    size_t MaxPoolableSize = 0;
 
-  std::shared_ptr<USMLimits> limits;
+    // When pooling, each bucket will hold a max of 4 unfreed slabs
+    size_t Capacity = 0;
+
+    // Holds the minimum bucket size valid for allocation of a memory type.
+    size_t MinBucketSize = 0;
+
+    // Holds size of the pool managed by the allocator.
+    size_t CurPoolSize = 0;
+
+    // Whether to print pool usage statistics
+    int PoolTrace = 0;
+
+    std::shared_ptr<SharedLimits> limits;
 };
 
-class USMAllocContext {
-public:
-  // Keep it public since it needs to be accessed by the lower layer(Buckets)
-  class USMAllocImpl;
+class DisjointPool {
+  public:
+    class AllocImpl;
+    using Config = DisjointPoolConfig;
 
-  USMAllocContext(std::unique_ptr<SystemMemory> memHandle,
-                  USMAllocatorParameters params);
-  ~USMAllocContext();
+    uma_result_t initialize(uma_memory_provider_handle_t *providers,
+                            size_t numProviders,
+                            DisjointPoolConfig parameters) noexcept;
+    void *malloc(size_t size) noexcept;
+    void *calloc(size_t, size_t) noexcept;
+    void *realloc(void *, size_t) noexcept;
+    void *aligned_malloc(size_t size, size_t alignment) noexcept;
+    size_t malloc_usable_size(void *) noexcept;
+    void free(void *ptr) noexcept;
+    enum uma_result_t get_last_result(const char **ppMessage) noexcept;
 
-  void *allocate(size_t size);
-  void *allocate(size_t size, size_t alignment);
-  void deallocate(void *ptr);
+    DisjointPool();
+    ~DisjointPool();
 
-private:
-  std::unique_ptr<USMAllocImpl> pImpl;
+  private:
+    std::unique_ptr<AllocImpl> impl;
 };
+
+} // namespace usm
 
 #endif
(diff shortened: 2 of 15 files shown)


pbalcer:
If USM is composed of contributions that come exclusively from Intel folks (or if we can easily get approval), then we are free to relicense it to MIT - which I think would be the easiest. Thankfully MIT and Apache licenses are compatible, so I also don't think there's any issue with us just using the Apache-licensed usmPool component - but we'd need to be careful to maintain proper attributions.

source/common/uma_pools/usm_allocator_config.hpp
```diff
@@ -0,0 +1,34 @@
...
+// Parses and stores configuration for all instances of USM allocator
+class USMAllocatorConfig {
+  public:
+    USMAllocatorConfig(const std::string &config = "", bool trace = true);
+
+    size_t EnableBuffers = 1;
+    USMAllocatorParameters Configs[MemType::All];
+
+    // String names of memory types for printing in limits traces.
+    static constexpr const char *MemTypeNames[MemType::All] = {
```
pbalcer, line 29: does this need to be in the header? I think you can get the same information from `Configs[MemType::whatever].memoryTypeName`.

kbenzie:
> If USM is composed of contributions that come exclusively from Intel folks (or if we can easily get approval), then we are free to relicense it to MIT - which I think would be the easiest. Thankfully MIT and Apache licenses are compatible, so I also don't think there's any issue with us just using the Apache-licensed usmPool component - but we'd need to be careful to maintain proper attributions.

If we do end up with both MIT and Apache in the repo the [LICENSE](https://github.com/oneapi-src/unified-runtime/blob/main/LICENSE) file would need to be updated. I personally would prefer everything under a single license, its easier for end users that way, however I don't have a strong opinion on this topic.

I think the question about licensing is probably one we should bring to the WG call. This will also affect bringing in the plugins/adapters once we're in a position to do so.

source/common/uma_pools/usm_allocator_config.hpp
igchor: Well, this is just mapping from enum to string used to initialize the `Configs[MemType::whatever].memoryTypeName` - I can put it into usm_allocator_config.cpp. I did not want to do too many changes in this PR but eventually, I'd like to get rid of the `enum MemType` at all and use some UR type.

source/common/uma_pools/usm_allocator.cpp
```diff
@@ -0,0 +1,866 @@
...
+  assert(Alignment > 0);
+  return (Val + Alignment - 1) & (~(Alignment - 1));
+}
+
+class Bucket;
+
+// Represents the allocated memory block of size 'SlabMinSize'
+// Internally, it splits the memory block into chunks. The number of
+// chunks depends of the size of a Bucket which created the Slab.
+// The chunks
```
PatKamin: Looks like some missing info?

source/common/uma_pools/usm_allocator.cpp
```diff
@@ -0,0 +1,866 @@
...
+// Note: Bucket's method are responsible for thread safety of Slab access,
```
PatKamin: nit: method -> methods

source/common/uma_pools/usm_allocator.cpp
```diff
@@ -0,0 +1,866 @@
...
+  MemPtr = Bkt.getMemHandle().allocate(SlabSize);
+  regSlab(*this);
+}
+
+Slab::~Slab() {
+  unregSlab(*this);
+  bucket.getMemHandle().deallocate(MemPtr, true /* OwnZeMemHandle */);
+}
+
+// Return the index of the first available chunk, -1 otherwize
```
PatKamin: nit: otherwize -> otherwise

source/common/uma_pools/usm_allocator.cpp
```diff
@@ -0,0 +1,866 @@
...
+  using ListIter = std::list<std::unique_ptr<Slab>>::iterator;
+
+  // Store iterator to the corresponding node in avail/unavail list
+  // to achieve O(1) removal
+  ListIter SlabListIter;
+
+  // Hints where to start search for free chunk in a slab
+  size_t FirstFreeChunkIdx = 0;
+
+  // Return the index of the first available chunk, -1 otherwize
```
PatKamin: `-1` will never be returned as the type returned is `size_t`. Perhaps ssize_t should be used instead. Otherwise this return value should be documented as SIZE_MAX and it would be better to return SIZE_MAX explicitly in this method's implementation rather than implicitly returning it from `-1` value.

source/common/uma_pools/usm_allocator.cpp
```diff
@@ -0,0 +1,866 @@
...
+// Return the index of the first available chunk, -1 otherwize
+size_t Slab::FindFirstAvailableChunkIdx() const {
+  // Use the first free chunk index as a hint for the search.
+  auto It = std::find_if(Chunks.begin() + FirstFreeChunkIdx, Chunks.end(),
+                         [](auto x) { return !x; });
+  if (It != Chunks.end()) {
+    return It - Chunks.begin();
+  }
+
+  return static_cast<size_t>(-1);
```
PatKamin: Casting a negative int to the unsigned type.

source/common/CMakeLists.txt
```diff
@@ -10,3 +10,7 @@ target_include_directories(common INTERFACE
 
 add_subdirectory(unified_memory_allocation)
 target_link_libraries(common INTERFACE unified_memory_allocation)
+
+add_subdirectory(uma_pools)
+target_include_directories(uma_usm_pool PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
```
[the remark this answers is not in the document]
igchor: Yes, I actually already renamed the class and file names - I just forgot about the cmake. I will change that

source/common/uma_pools/usm_allocator.cpp
igchor: Removed.

source/common/uma_pools/usm_allocator.cpp
igchor: Done.

source/common/uma_pools/usm_allocator.cpp
igchor: Done.

source/common/uma_pools/usm_allocator_config.hpp
igchor: Removed.

pbalcer (approved):
lgtm, but let's resolve the licensing issue first.

source/common/uma_pools/disjoint_pool.hpp
```diff
@@ -0,0 +1,78 @@
+//===---------- disjoint_pool.hpp - Allocator for USM memory --------------===//
```
vinser52, line 1: Why disjoint_pool implementation is not part of the `unified_memory_allocation` directory? Is it just an intermediate step?

source/common/usm_pool_config.hpp
```diff
@@ -0,0 +1,80 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef USM_POOL_CONFIG
+#define USM_POOL_CONFIG
+
+#include "disjoint_pool.hpp"
+#include "ur_api.h"
```
DamianDuy, line 13: When I tried to include the `usm_pool_config.hpp` file in the test it failed because It could not find the `ur_api.h` so I think this include needs a path perhaps?

source/common/uma_pools/disjoint_pool.hpp
igchor: As I mentioned offline, it makes more sense to keep it as part of UR rather than UMA for now, so that we can use UR utils (like logger) + it makes it clear that we can safely change the API (e.g. for config).

source/common/usm_pool_config.hpp
igchor: I forgot to add usm_pool_config to 'common' lib. It should work now in tests, as they already link with common.

source/common/uma_helpers.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+#include <iostream>
+
+namespace uma {
+using ur_memory_provider_handle_unique =
+    std::unique_ptr<uma_memory_provider_t, decltype(&umaMemoryProviderDestroy)>;
+using ur_memory_pool_handle_unique =
+    std::unique_ptr<uma_memory_pool_t, decltype(&umaPoolDestroy)>;
+
+// C++14 version of std::make_from_tuple (but uses operator new)
```
PatKamin, line 29: You can use `std::make_from_tuple` now as the current standard in UR is C++17.

igchor: It's still problematic since I need to allocate the `T` dynamically. I could probably use something like this: `new T(std::make_from_tuple<T>(args...);` but that would only work for moveable types

source/common/usm_pool_config.cpp
```diff
@@ -0,0 +1,267 @@
...
+            }
+        }
+        if (More) {
+            More = ParamParser(Params, Configs[LM].Capacity, ParamWasSet);
+            if (ParamWasSet && poolDesc.type == UR_USM_TYPE_UNKOWN) {
+                for (auto &Config : Configs) {
+                    Config.second.Capacity = Configs[LM].Capacity;
+                }
+            }
+        }
```
DamianDuy, line 135: I'd like to ask, what was the intended behavior for when invalid type is passed in the string config (like ` foo:0,3,2m`)? Since, as it is now,  the invalid type's parameters are overwriting other config types in the `ParamParser` (found in the test), except for the first parameter- `MaxPoolableSize` because in the `MemTypeParser` the invalid type name is never erased from the string so this parameter isn't returned from the `GetValue`. Perhaps It'd be a good idea to add  some checking to the `ParamParser` if the type isn't valid and erase it and it's parameters from the config string?

igchor: Yes, I think it's a good idea. We should also log each invalid type and say we're going to use the defaults. There is some code for printing the configuration in this file but I'd like to replace it with our logging mechanism in the future.

source/common/uma_helpers.hpp
```diff
@@ -0,0 +1,185 @@
+/*
+ *
+ * Copyright (C) 2023 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
```
omarahmed1111, line 7: Just wanted to give an update that the license of the project have been updated in this [PR](https://github.com/oneapi-src/unified-runtime/pull/557). So, we should be able to update all the files license to the llvm one now.

PR #221, opened 2023-02-08
Add initial validation layer

kswiecicki:
This PR adds an initial validation layer to build infrastructure and parameter validation functionality.
The validation layer can be controlled through the following environmental variables:
- UR_ENABLE_VALIDATION_LAYER
- UR_ENABLE_PARAMETER_VALIDATION

Templates and the solution comes from: https://github.com/oneapi-src/level-zero.
Replaces: https://github.com/oneapi-src/unified-runtime/pull/188.


kswiecicki:
@pbalcer I've seen some tests involving parameter validation inside conformance tests, so I've enabled the validation layer for those tests for now.
In a separate PR, I plan to test most functions inside ur_valddi.cpp and remove duplicates from conformance tests. What do you think?

source/loader/layers/ur_proxy_layer.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include "ur_util.hpp"
+
+///////////////////////////////////////////////////////////////////////////////
+class __urdlllocal proxy_layer_context_t
+{
+    public:
+        ur_api_version_t version = UR_API_VERSION_0_9;
+
+        virtual bool isEnabled() = 0;
+        virtual ur_result_t init(ur_dditable_t *dditable) = 0;
```
pbalcer, line 23: For now, I think this is fine, but it would be best if we created a generic `init` function to avoid duplication (code for the tracing layer will be nearly identical) and, in the future, allow the layers to be dynamically loaded just as implementations of UR spec, and not some custom API.

pbalcer:
> @pbalcer I've seen some tests involving parameter validation inside conformance tests, so I've enabled the validation layer for those tests for now. In a separate PR, I plan to test most functions inside ur_valddi.cpp and remove duplicates from conformance tests. What do you think?

I'm not sure what's the best approach. But I don't think we should be messing with conformance tests - they are supposed to test if the adapter implementation is compliant and provide exactly the behavior specified in the spec. If you enable the validation layer you can actually make the implementation more restrictive than what the specification requires, potentially messing up the conformance tests - although that shouldn't happen unless it includes some negative tests. Later, it might make sense to optionally enable validation layers in the CI as another job.

I think we should have tests for the proxy layer implementation. I would hold off on implementing tests for every single function in valddi. The UR is evolving, and having manually written tests for everything would be annoying. Given the relative simplicity of the parameter validation functionality, I think it would be sufficient to test it on one single function.

source/loader/layers/ur_proxy_layer.hpp
pbalcer: I think this could be done as an implementation of a "dynamic" proxy layer that loads the appropriate symbols.

kswiecicki:
> 
> I'm not sure what's the best approach. But I don't think we should be messing with conformance tests - they are supposed to test if the adapter implementation is compliant and provide exactly the behavior specified in the spec. If you enable the validation layer you can actually make the implementation more restrictive than what the specification requires, potentially messing up the conformance tests - although that shouldn't happen unless it includes some negative tests. Later, it might make sense to optionally enable validation layers in the CI as another job.
> 
> I think we should have tests for the proxy layer implementation. I would hold off on implementing tests for every single function in valddi. The UR is evolving, and having manually written tests for everything would be annoying. Given the relative simplicity of the parameter validation functionality, I think it would be sufficient to test it on one single function.

Alright. I've added two simple validation tests and reverted the changes in conformance tests.

test/layers/validation/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: MIT
+
+set(UR_VALIDATION_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+
+function(add_validation_test name)
```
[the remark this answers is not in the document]
kswiecicki: > LGTM, but please rebase and `make cppformat`.

I've rebased and applied cppformat formatting.

> these tests will not run - for now we are filtering which tests are ran by label, see here: https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/cmake.yml#L35
> 
> Please add a label and update the workflow file to include it.

I've added a 'validation' label to the workflow.

pbalcer:
Looks like the checking if code is formatted doesn't work properly...  clang-format behaves differently depending on whether code generation is done as part of the same target as formatting. No idea why. My best guess is that `generate` creating new files somehow breaks the list of things to be formatted...

For now, format using `make generate` and that will make the build pass. I'll try to think of a solution.

Issue #215, opened 2023-02-06
Add tests for `urEventRetain` & `urEventRelease`


veselypeta:
We should add tests for the entry-points: `urEventRetain` & `urEventRelease`. They should effectively be the same as previous release/retain tests such as `urDeviceRelease`

PR #247, opened 2023-02-08
Add urEventRetain and urEventRelease tests

fabiomestre:
Add urEventRetain and urEventRelease tests 
Update urEventWait tests to avoid allocating buffer when running negative tests

Addresses issue https://github.com/oneapi-src/unified-runtime/issues/215

### DIFF
--- a/test/conformance/event/fixtures.h
+++ b/test/conformance/event/fixtures.h
@@ -1,0 +1,50 @@
+// Copyright (C) 2022-2023 Intel Corporation
+// SPDX-License-Identifier: MIT
+
+#ifndef UR_CONFORMANCE_EVENT_FIXTURES_H_INCLUDED
+#define UR_CONFORMANCE_EVENT_FIXTURES_H_INCLUDED
+
+#include <uur/fixtures.h>
+
+namespace uur {
+namespace event {
+
+/**
+ * Test fixture that is intended to be used when testing reference count APIs
+ * (i.e. urEventRelease and urEventRetain). Does not handle destruction of the
+ * event.
+ */
+struct urEventReferenceTest : uur::urQueueTest {
+
+    void CreateEvent() {
+        ASSERT_SUCCESS(urMemBufferCreate(context, UR_MEM_FLAG_WRITE_ONLY, size,
+                                         nullptr, &buffer));
+
+        input.assign(count, 42);
+        ASSERT_SUCCESS(urEnqueueMemBufferWrite(
+            queue, buffer, false, 0, size, input.data(), 0, nullptr, &event));
+    }
+
+    void checkEventReferenceCount(uint32_t expected_value) {
+        uint32_t reference_count{};
+        urEventGetInfo(event, ur_event_info_t::UR_EVENT_INFO_REFERENCE_COUNT,
+                       sizeof(uint32_t), &reference_count, nullptr);
+        ASSERT_EQ(reference_count, expected_value);
+    }
+
+    void Cleanup() {
+        if (buffer) {
+            EXPECT_SUCCESS(urMemRelease(buffer));
+        }
+    }
+
+    const size_t count = 1024;
+    const size_t size = sizeof(uint32_t) * count;
+    ur_mem_handle_t buffer = nullptr;
+    ur_event_handle_t event = nullptr;
+    std::vector<uint32_t> input;
+};
+} // namespace event
+} // namespace uur
+
+#endif // UR_CONFORMANCE_EVENT_FIXTURES_H_INCLUDED
--- a/test/conformance/event/fixtures.h
+++ b/test/conformance/event/fixtures.h
@@ -16,7 +16,8 @@ namespace event {
  */
 struct urEventReferenceTest : uur::urQueueTest {
 
-    void CreateEvent() {
+    void SetUp() override {
+        UUR_RETURN_ON_FATAL_FAILURE(urQueueTest::SetUp());
         ASSERT_SUCCESS(urMemBufferCreate(context, UR_MEM_FLAG_WRITE_ONLY, size,
                                          nullptr, &buffer));
 
@@ -25,17 +26,18 @@ struct urEventReferenceTest : uur::urQueueTest {
             queue, buffer, false, 0, size, input.data(), 0, nullptr, &event));
     }
 
-    void checkEventReferenceCount(uint32_t expected_value) {
-        uint32_t reference_count{};
-        urEventGetInfo(event, ur_event_info_t::UR_EVENT_INFO_REFERENCE_COUNT,
-                       sizeof(uint32_t), &reference_count, nullptr);
-        ASSERT_EQ(reference_count, expected_value);
-    }
-
-    void Cleanup() {
+    void TearDown() override {
         if (buffer) {
             EXPECT_SUCCESS(urMemRelease(buffer));
         }
+        urQueueTest::TearDown();
+    }
+
+    bool checkEventReferenceCount(uint32_t expected_value) {
+        uint32_t reference_count{};
+        urEventGetInfo(event, ur_event_info_t::UR_EVENT_INFO_REFERENCE_COUNT,
+                       sizeof(uint32_t), &reference_count, nullptr);
+        return reference_count == expected_value;
     }
 
     const size_t count = 1024;
--- a/test/conformance/event/urEventRetain.cpp
+++ b/test/conformance/event/urEventRetain.cpp
@@ -7,32 +7,25 @@ using urEventRetainTest = uur::event::urEventReferenceTest;
 
 /* Check that urEventRetain returns Success */
 TEST_P(urEventRetainTest, Success) {
-    ASSERT_NO_FATAL_FAILURE(CreateEvent());
     ASSERT_SUCCESS(urEventRetain(event));
     ASSERT_SUCCESS(urEventRelease(event));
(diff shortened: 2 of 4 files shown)


test/conformance/event/urEventWait.cpp
```diff
@@ -4,7 +4,7 @@
 #include <uur/fixtures.h>
 
 struct urEventWaitTest : uur::urQueueTest {
-    void SetUp() override {
+    void WriteSourceBuffer() {
```
fabiomestre, line 7: I don't think the negative tests should be creating buffers and copying memory when running. Negative tests should be as simple as possible. I couldn't think of a better way to handle this issue other than renaming SetUp and TearDown so that they have to be explicitly called. But I'm open to suggestions on how to handle this.

test/conformance/event/urEventRetain.cpp
```diff
@@ -1,2 +1,38 @@
...
+    EXPECT_NO_FATAL_FAILURE(Cleanup());
+}
+
+TEST_P(urEventRetainTest, InvalidNullHandle) {
+    ASSERT_EQ(urEventRetain(nullptr), UR_RESULT_ERROR_INVALID_NULL_HANDLE);
+}
+
+TEST_P(urEventRetainTest, InvalidEvent) {
+    ur_event_handle_t event{};
+    ASSERT_EQ(urEventRetain(event), UR_RESULT_ERROR_INVALID_EVENT);
```
fabiomestre, line 35: Would this be the correct way to test UR_RESULT_ERROR_INVALID_EVENT?

veselypeta: A handle is just a pointer so just give it some random invalid address that isn't null.

```cpp
ur_event_handle_t event = 0xdeadbeef;
```

test/conformance/event/fixtures.h
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    void checkEventReferenceCount(uint32_t expected_value) {
+        uint32_t reference_count{};
+        urEventGetInfo(event, ur_event_info_t::UR_EVENT_INFO_REFERENCE_COUNT,
+                       sizeof(uint32_t), &reference_count, nullptr);
+        ASSERT_EQ(reference_count, expected_value);
+    }
+
+    void Cleanup() {
```
veselypeta, line 35: This can be `TearDown()` - mean you don't have to call it manually

test/conformance/event/urEventWait.cpp
```diff
@@ -17,7 +17,7 @@ struct urEventWaitTest : uur::urQueueTest {
         ASSERT_SUCCESS(urEventWait(1, &event));
     }
 
-    void TearDown() override {
+    void Cleanup() {
```
veselypeta, line 20: I still think this can be `TearDown()` - same as above

veselypeta (commented):
I left some comments.

I understand that the way the tests are currently setup we shouldn't have to do some setup code. I still wonder if it would be such a problem to use a different fixture for the negative tests. Obviously this would mean that they would be part of a different test suite, but maybe that wouldn't be so bad?

```cpp
using urEventReleaseTest = uur::urEventTest;
TEST_P(urEventReleaseTest, Success){}

using urEventReleaseNegativeTest = uur::urQueueTest:
TEST_P(urEventReleaseNegativeTest, Invalid...){}

```

test/conformance/event/fixtures.h
```diff
@@ -0,0 +1,50 @@
...
+namespace event {
+
+/**
+ * Test fixture that is intended to be used when testing reference count APIs
+ * (i.e. urEventRelease and urEventRetain). Does not handle destruction of the
+ * event.
+ */
+struct urEventReferenceTest : uur::urQueueTest {
+
+    void CreateEvent() {
```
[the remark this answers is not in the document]
fabiomestre: No, they are being implicily called because I'm not overriding the functions

fabiomestre:
> I left some comments.
> 
> I understand that the way the tests are currently setup we shouldn't have to do some setup code. I still wonder if it would be such a problem to use a different fixture for the negative tests. Obviously this would mean that they would be part of a different test suite, but maybe that wouldn't be so bad?
> 
> ```c++
> using urEventReleaseTest = uur::urEventTest;
> TEST_P(urEventReleaseTest, Success){}
> 
> using urEventReleaseNegativeTest = uur::urQueueTest:
> TEST_P(urEventReleaseNegativeTest, Invalid...){}
> ```

Yeah, so that's the issue that I was not sure how to approach. Having 2 separate test suites seemed to go against the current design of the CTS (i.e. one test suite per entry point). And that's why I chose this solution. But, if we think 2 test suits is ok,  I don't mind changing. It certainly makes the code cleaner.

test/conformance/event/urEventRetain.cpp
fabiomestre: How would an implementation verify that a pointer is pointing to a random place in memory? I think that would lead to segmentation faults. Unless we keep a list of valid pointers internally. Not sure if that's something we would like to do.

test/conformance/event/fixtures.h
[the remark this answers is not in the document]
veselypeta: 👍

test/conformance/event/urEventRetain.cpp
veselypeta: In UnitUR there was a test for memory that used this https://github.com/oneapi-src/unified-runtime/pull/233/files#diff-8bcf0034cb139b390002ba870b39e7dd2fb245f47d618c02ad74914725b937fdR73. However for allocations we do track valid pointers so can check if a pointer is valid. I don't know if we mandate the same for events

kbenzie: Those tests are testing something different @veselypeta, `UR_RESULT_ERROR_INVALID_NULL_HANDLE` rather than `UR_RESULT_ERROR_INVALID_<OBJECT>`, or a fill patten for a setting memory.

I don't think we can reliable trigger `UR_RESULT_ERROR_INVALID_EVENT` without making assumptions about the implementation of an adapter.

test/conformance/event/urEventWait.cpp
kbenzie: Events are a bit different because they are implicitly tied to enqueuing commands to let the user determine when the start and finish and to create depedencies. OpenCL does have the concept of user events but we deliberately removed them from Unified Runtime because they are a pain to implement if its even possible when layering on top of e.g. CUDA.

test/conformance/event/fixtures.h
kbenzie: I agree, having `CreateEvent()` and `Cleanup()` instead of just doing work in overridden `SetUp()` and `TearDown()` actually adds unnecessary repetition to the test cases.

It also diverges from the patterns in all the other test fixtures which I found jarring when reviewing this code, being different requires extra thought for the reader.

test/conformance/event/fixtures.h
```diff
@@ -0,0 +1,50 @@
...
+        input.assign(count, 42);
+        ASSERT_SUCCESS(urEnqueueMemBufferWrite(
+            queue, buffer, false, 0, size, input.data(), 0, nullptr, &event));
+    }
+
+    void checkEventReferenceCount(uint32_t expected_value) {
+        uint32_t reference_count{};
+        urEventGetInfo(event, ur_event_info_t::UR_EVENT_INFO_REFERENCE_COUNT,
+                       sizeof(uint32_t), &reference_count, nullptr);
+        ASSERT_EQ(reference_count, expected_value);
```
kbenzie, line 32: This could return `bool` instead and use `ASSERT_TRUE(checkEventReferenceCount(event))` in the test cases instead of using `ASSERT_NO_FATAL_FAILURE` every time.

kbenzie:
> Yeah, so that's the issue that I was not sure how to approach. Having 2 separate test suites seemed to go against the current design of the CTS (i.e. one test suite per entry point). And that's why I chose this solution. But, if we think 2 test suits is ok, I don't mind changing. It certainly makes the code cleaner.

There's no problem adding test fixtures that are only available in a single `.cpp` test file if that makes the most sense. The reason there is not more of that yet is mainly because we've not go to the more complicated things to test (compilation, enqueuing multiple commands, etc).

test/conformance/event/urEventWait.cpp
kbenzie: > Or do you think it's okay to setup buffers and copy memory on negative tests (even though those buffers are not actually used)?

I would have done this myself. Its a small amount of extra work but it could easiler just be a separate test fixture, no issue with that.

I'd much prefer keeping `SetUp()` and `TearDown()` rather than having these renamed member functions which are called manually.

test/conformance/event/fixtures.h
fabiomestre: Done

test/conformance/event/fixtures.h
fabiomestre: Done

test/conformance/event/urEventWait.cpp
fabiomestre: I ended up creating new fixtures. Because reusing the same fixtures would lead to memory leaks on negative tests because urEventRelease is not being called on TearDown.

kbenzie: Sounds good 👍

veselypeta (approved):
LGTM

PR #249, opened 2023-02-10
[ur] Add return code for nullptr function pointers

veselypeta:
For entry points `urContextSetExtendedDeleter` and `urEventSetCallback` they accept a function pointer. We should clarify that the spec expects these values to not be null and if so return the appropriate error code.

Additionally the userdata parameter should be optional in `urContextSetExtendedDeleter` - this was already the case is `urEventSetCallback`.


kbenzie (approved):
LGTM

PR #250, opened 2023-02-10
Refactor the compiler API.

### LINKED ISSUE #9: urModuleCreate and the void **pfnNotify parameter
[`urModuleCreate`](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/module.yml#L13)is declared as:
```
UR_APIEXPORT ur_result_t UR_APICALL urModuleCreate(ur_context_handle_t hContext, const void *pIL, uint32_t length, const char *pOptions, void **pfnNotify, void *pUserData, ur_module_handle_t *phModule)
```

It accepts a [`pfnNotify`](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/module.yml#L36) argument that is supposed to be called when program compilation is complete, this is confusing for two reasons:

1. `urModuleCreate` doesn't have to do any compilation (it seems to just be an in memory representation of a SPIR-V string program, although maybe this is a misunderstanding), this may be handled further down the pipeline when you create the program, so why pass a callback here? In OpenCL [`clBuildProgram`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clBuildProgram) takes a callback to be called when the program is built. `clBuildProgram` takes a `cl_program` object that has already been created via [`clCreateProgramWith(Source|IL|Binary)`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_creating_program_objects). The `clCreateProgramWith.*` APIs do not take a callback, if these are the analogous entry points then should `urModuleCreate` take a callback?
2.  The type of `pfnNotify` is `void **`, which is not a function pointer. This is probably a typo and should be `void (*pfnNotify)` or `void * (*pfnNotify)` although the latter is unlikely since callbacks don't often return values, so the second `*` may also be a typo.

### LINKED ISSUE #20: Consider relaxing ur_module_handle_t SPIR-V requirement
There seems to be an implicit requirement that `ur_module_handle_t` objects are SPIR-V modules. From the [`urProgramCreate`](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3712) entry point:

```
Create Program from input SPIR-V modules. 
```

Do we want to restrict the `ur_module_handle_t` objects to contain SPIR-V only. Or is it possible that higher level language runtimes sitting on top of Unified Runtime could feed it non SPIR-V based intermediate representations for compilation to target specific executables in the adapter via the `urModuleCreate` -> `urProgramCreate` path?

aarongreig:
The module object has been removed, and the program object now more closely
mirrors its OpenCL and PI equivalents.

Closes #9
Closes #20


scripts/core/program.yml
```diff
@@ -10,29 +10,47 @@ type: header
...
+--- #--------------------------------------------------------------------------
 type: function
 desc: "Create Program from input SPIR-V modules."
 class: $xProgram
 name: Create
 decl: static
 ordinal: "0"
 analogue:
-    - "**clCreateProgram**"
+    - "**clCreateProgramWithIL** and **clBuildProgram**"
```
kbenzie, line 32: I feel like this is closer to `clCompileProgram` since `clBuildProgram` does an implicit link.

scripts/core/program.yml
```diff
@@ -264,6 +310,8 @@ desc: "Set a Program object specialization constant to a specific value"
 class: $xProgram
 name: SetSpecializationConstant
 ordinal: "0"
+analogue: "**clSetProgramSpecializationConstant**"
+details: "`hProgram` must not be an executable program, or a program created from a device binary."
```
kbenzie, line 314: probably works but best to be consistent and make this an array of strings.

```suggestion
details:
    - "`hProgram` must not be an executable program, or a program created from a device binary."
```

kbenzie (commented):
This is a good start. Once we finalize the design it would be good to enumerate the different use cases in the programming guide.

scripts/core/program.yml
kbenzie: This comment also applies to the `ERROR_PROGRAM_BUILD_FAILURE` error code.

scripts/core/program.yml
```diff
@@ -264,6 +310,10 @@ desc: "Set a Program object specialization constant to a specific value"
 class: $xProgram
 name: SetSpecializationConstant
 ordinal: "0"
+analogue:
+    - "**clSetProgramSpecializationConstant**"
+details:
+    - "`hProgram` must not be an executable program, or a program created from a device binary."
 params:
```
kbenzie, line 317: #269 is looking to add a kernel argument to enable better support for specialization constants on CUDA.

scripts/core/program.yml
```diff
@@ -10,29 +10,47 @@ type: header
...
+    - type: size_t
+      name: length
+      desc: "[in] length of IL in bytes."
     - type: const char*
       name: pOptions
-      desc: "[in][optional] pointer to linker options null-terminated string."
+      desc: "[in] pointer to compiler options null-terminated string."
+    - type: $x_programcreate_callback_t
+      name: pfnNotify
+      desc: "[in][optional] A function pointer to a notification routine that is called when program compilation is complete."
```
kbenzie, line 50: Based on the fact that no plugins currently support the callback in #9 it probably makes sense to drop this callback. Also with the addition of `ur_program_properties_t` in #275 if we wanted to add the callback back in later we could do it with a `pNext` chain.

aarongreig: agreed, always nice to delete stuff if you can

kbenzie (approved):
LGTM

veselypeta (approved):
LGTM

PR #251, opened 2023-02-10
Add urEventGetInfo tests

fabiomestre:


test/conformance/event/urEventGetInfo.cpp
```diff
@@ -1,2 +1,137 @@
...
+        break;
+    }
+    case UR_EVENT_INFO_REFERENCE_COUNT: {
+        auto returned_reference_count =
+            reinterpret_cast<uint32_t *>(data.data());
+        EXPECT_EQ(1, *returned_reference_count);
+        break;
+    }
+    default:
+        ASSERT_FALSE(false);
```
veselypeta, line 85:
```suggestion
        FAIL() << "Invalid event info enumeration";
```

test/conformance/event/urEventGetInfo.cpp
```diff
@@ -1,2 +1,137 @@
...
+
+    /* Invalid hEvent */
+    EXPECT_EQ(
+        urEventGetInfo(nullptr, UR_EVENT_INFO_COMMAND_QUEUE, 0, nullptr, &size),
+        UR_RESULT_ERROR_INVALID_NULL_HANDLE);
+
+    /* Valid hEvent but invalid pPropValue and pPropValueSizeRet */
+    EXPECT_EQ(
+        urEventGetInfo(event, UR_EVENT_INFO_COMMAND_QUEUE, 0, nullptr, nullptr),
+        UR_RESULT_ERROR_INVALID_NULL_HANDLE);
```
veselypeta, line 114: I think this is a valid invocation - should return `UR_RESULT_SUCCESS`. Essentially does nothing though.

test/conformance/event/urEventGetInfo.cpp
```diff
@@ -1,2 +1,137 @@
...
+
+TEST_P(urEventGetInfoNegativeTest, InvalidNullHandle) {
+    ur_event_info_t info_type = UR_EVENT_INFO_COMMAND_QUEUE;
+    size_t size;
+    ASSERT_SUCCESS(urEventGetInfo(event, info_type, 0, nullptr, &size));
+    ASSERT_NE(size, 0);
+    std::vector<uint8_t> data(size);
+
+    /* Invalid hEvent */
+    EXPECT_EQ(
```
veselypeta, line 107:
```suggestion
    ASSERT_EQ(
```

test/conformance/event/urEventGetInfo.cpp
veselypeta: I think you can just remove it.

fabiomestre: On second thought, both parameters are marked as optional. So I guess this behaviour is implied. I don't mind removing it then

veselypeta (approved):

PR #253, opened 2023-02-10
[ur] Clarify some device return codes

veselypeta:
This PR clarifies some device entry-points and what error codes they should return.

* I added `UR_RESULT_ERROR_INVALID_DEVICE_PARTITION_COUNT` to be returned from `urDeviceParition` when by-counts paritioning is used and invalid counts are provided. (to match OpenCL)
* I added `UR_RESULT_ERROR_DEVICE_PARTITION_FAILED` to be returned from `urDeviceParition` when the paritioning scheme is supported but the device could not be futher paritioned. (to march OpenCL)
* Adds `UR_RESULT_ERROR_INVALID_VALUE` error code to `urDeviceGetInfo` & `urDeviceSelectBinary`

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -164,56 +164,58 @@ typedef enum ur_result_t
     UR_RESULT_ERROR_DEVICE_LOST = 21,               ///< Device hung, reset, was removed, or driver update occurred
     UR_RESULT_ERROR_DEVICE_REQUIRES_RESET = 22,     ///< Device requires a reset
     UR_RESULT_ERROR_DEVICE_IN_LOW_POWER_STATE = 23, ///< Device currently in low power state
-    UR_RESULT_ERROR_INVALID_WORK_ITEM_SIZE = 24,    ///< Invalid work item size
-    UR_RESULT_ERROR_INVALID_WORK_DIMENSION = 25,    ///< Invalid work dimension
-    UR_RESULT_ERROR_INVALID_KERNEL_ARGS = 26,       ///< Invalid kernel args
-    UR_RESULT_ERROR_INVALID_KERNEL = 27,            ///< Invalid kernel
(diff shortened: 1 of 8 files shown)


martygrant (approved):
Looks good!

fabiomestre (approved):
Looks good to me. I'm not familiar with all the places that need to be updated though. So maybe someone else should double check.

include/ur_api.h
```diff
@@ -3212,12 +3217,14 @@ urDevicePartition(
...
+///     - ::UR_RESULT_ERROR_INVALID_SIZE
 UR_APIEXPORT ur_result_t UR_APICALL
 urDeviceSelectBinary(
     ur_device_handle_t hDevice,                     ///< [in] handle of the device to select binary for.
     const uint8_t** ppBinaries,                     ///< [in] the array of binaries to select from.
-    uint32_t NumBinaries,                           ///< [in] the number of binaries passed in ppBinaries. Must greater than or
-                                                    ///< equal to zero.
+    uint32_t NumBinaries,                           ///< [in] the number of binaries passed in ppBinaries. 
+                                                    ///< Must greater than or equal to zero otherwise
+                                                    ///< ::UR_RESULT_ERROR_INVALID_SIZE is returned.
```
omarahmed1111, line 3227: shouldn't `UR_RESULT_ERROR_INVALID_SIZE` be the interpretation of 0 value? as from my understanding to this from [ur current specification](https://oneapi-src.github.io/unified-runtime/core/api.html#_CPPv416ur_device_info_t). it says:
> enumerator UR_RESULT_ERROR_INVALID_SIZE
 [Validation] size argument is invalid (e.g., must not be zero)

fabiomestre: That's a good point. UR_RESULT_ERROR_INVALID_VALUE probably makes more sense?

veselypeta: Done.

omarahmed1111 (approved):

PR #254, opened 2023-02-12
[ur] Remove extra semicolons

jandres742:
Signed-off-by: Jaime Arteaga <jaime.a.arteaga.molina@intel.com>

### DIFF
--- a/source/loader/ur_loader.cpp
+++ b/source/loader/ur_loader.cpp
@@ -33,13 +33,13 @@ ur_result_t context_t::init() {
         intercept_enabled = true;
 
     return UR_RESULT_SUCCESS;
-};
+}
 
 ///////////////////////////////////////////////////////////////////////////////
 context_t::~context_t() {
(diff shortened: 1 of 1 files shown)


pbalcer:
Makes me wonder if we shouldn't enable `-Wpedantic` just like SYCL to avoid these types of issues in the future.

jandres742:
> Makes me wonder if we shouldn't enable `-Wpedantic` just like SYCL to avoid these types of issues in the future.

maybe? I would say that we should try to have the compilation as close as possible between the two. However, in this case, I think it is because we are building the UR loader in SYCL. Once we dont build it, but just link to the ur_loader.so/dll, it shouldn't be needed.

jandres742:
@pbalcer , @kbenzie :
actually, I found more issues, so yes, it makes sense as you mention. Please see here https://github.com/oneapi-src/unified-runtime/pull/265

PR #259, opened 2023-02-14
[ur] Fix clang-format

DamianDuy:
Refers to: #101

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,5 +2,5 @@
 Language:      Cpp
 BasedOnStyle:  LLVM
 IndentWidth:   4
+InsertBraces: true
 ...
-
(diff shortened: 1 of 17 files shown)


pbalcer (approved):

pbalcer:
Once we merge this, PRs will require that they are formatted with clang-format 15.0 - the target `make cppformat` will do that automatically, but everyone will need to have the appropriate version of `clang-format`. This can be accomplished, for example, using `pip install clang-format==15.0.7`. Some distros already have it in their repositories (arch, latest ubuntu, fedora rawhide), so be mindful about conflicts with your distro's binaries (e.g., use `venv` if you are using the pip package).

kbenzie:
> Some distros already have it in their repositories (arch, latest ubuntu, fedora rawhide), so be mindful about conflicts with your distro's binaries (e.g., use `venv` if you are using the pip package).

It's also worth noting that https://apt.llvm.org/ exists for older versions of Debian/Ubuntu which doesn't package new LLVM versions.

kbenzie (approved):
LGTM

PR #261, opened 2023-02-14
[ur] Remove null adapter default load

veselypeta:


test/conformance/source/environment.cpp
```diff
@@ -100,11 +106,18 @@ uur::PlatformEnvironment::PlatformEnvironment(int argc, char **argv)
 
 void uur::PlatformEnvironment::SetUp() {
     if (!error.empty()) {
-        FAIL() << error;
+        if (error == "Could not load adapter") {
```
pbalcer, line 109: nit - Using strings to differentiate between different error scenarios isn't ideal. For tests it's fine, but defining something like might make it cleaner:
```
constexpr std::string_view ERROR_NO_ADAPTER = "Could not load adapter";
...
if (error == ERROR_NO_ADAPTER) {
    ...
}
```

.github/workflows/cmake.yml
```diff
@@ -32,7 +32,7 @@ jobs:
 
     - name: Test
       working-directory: ${{github.workspace}}/build
-      run: ctest -C ${{env.BUILD_TYPE}} --output-on-failure -L "python|uma|loader|validation"
+      run: UR_ADAPTERS_FORCE_LOAD={{github.workspace}}/build/lib/libur_null.so ctest UR_ENABLE_VALIDATION_LAYER=1 UR_ENABLE_PARAMETER_VALIDATION=1 -C ${{env.BUILD_TYPE}} --output-on-failure -L "python|uma|loader|validation"
```
pbalcer, line 35: hm, maybe in test cmake scripts we could have a logic like this:
```
if (UR_ADAPTERS_FORCE_LOAD is not set) {
  set UR_ADAPTERS_FORCE_LOAD = null adapter;
}
```

so that uma or other basic tests work locally when just running `make test`.

You shouldn't need to set the validation layer parameters (they are set as part of the test script)

kbenzie: Agreed, I've been trying a couple things out and think I have something that could work for that.

kbenzie: Doesn't check if `UR_ADAPTERS_FORCE_LOAD` is set but does allow python and loader tests to work with `make test`.

```diff
diff --git a/test/loader/CMakeLists.txt b/test/loader/CMakeLists.txt
index a6aa2a7..b0f3e27 100644
--- a/test/loader/CMakeLists.txt
+++ b/test/loader/CMakeLists.txt
@@ -1,5 +1,13 @@
 # Copyright (C) 2022 Intel Corporation
 # SPDX-License-Identifier: MIT

-add_test(NAME example-hello-world COMMAND hello_world)
+if(UNIX)
+    set(null_adapter libur_null.so)
+else()
+    set(null_adapter ur_null.dll)
+endif()
+
+add_test(NAME example-hello-world COMMAND
+    ${CMAKE_COMMAND} -E env "UR_ADAPTERS_FORCE_LOAD=${null_adapter}"
+    $<TARGET_FILE:hello_world>)
 set_tests_properties(${example-hello-world} PROPERTIES LABELS "loader")
diff --git a/test/python/CMakeLists.txt b/test/python/CMakeLists.txt
index 0dd4db3..18daa6b 100644
--- a/test/python/CMakeLists.txt
+++ b/test/python/CMakeLists.txt
@@ -7,18 +7,23 @@ function(add_python_test name)
         COMMAND ${Python3_EXECUTABLE} -B -m pytest ${CMAKE_CURRENT_SOURCE_DIR}/${name}.py
         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
     set_tests_properties(${TEST_NAME} PROPERTIES LABELS "python")
+    # variable to collect all the environment variables we'd like to set.
+    set(ENVVARS)
     # python uses LD_LIBRARY_PATH (PATH on Windows) to search for dynamic libraries,
     # so set it to the location where it can find the loader.
     if(UNIX)
-        set_property(TEST ${TEST_NAME} PROPERTY
-            ENVIRONMENT "LD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+        list(APPEND ENVVARS
+            "LD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}"
+            "UR_ADAPTERS_FORCE_LOAD=libur_null.so")
     else()
+        list(APPEND ENVVARS "UR_ADAPTERS_FORCE_LOAD=ur_null.dll")
         set_property(TEST ${TEST_NAME} PROPERTY
             ENVIRONMENT_MODIFICATION "PATH=cmake_list_prepend:${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")
     endif()
     # this is for importing the include/ur.py module in other python files
-    set_property(TEST ${TEST_NAME} APPEND PROPERTY
-       ENVIRONMENT "PYTHONPATH=${PROJECT_SOURCE_DIR}")
+    list(APPEND ENVVARS "PYTHONPATH=${PROJECT_SOURCE_DIR}")
+    # only set the ENVIRONMENT property once as a list to avoid overwriting it.
+    set_property(TEST ${TEST_NAME} PROPERTY ENVIRONMENT "${ENVVARS}")
 endfunction()

 add_python_test(basic)
```

test/conformance/source/environment.cpp
veselypeta: No `std::string_view` on C++14 unfortunately, but will fix.

.github/workflows/cmake.yml
veselypeta: This was mostly me just testing. I have added the required environment variables as test properties for it to work now.

source/loader/windows/platform_discovery_win.cpp
```diff
@@ -21,22 +21,14 @@
...
 
     // UR_ADAPTERS_FORCE_LOAD  is for development/debug only
     char *altPlatforms = nullptr;
     _dupenv_s(&altPlatforms, NULL, "UR_ADAPTERS_FORCE_LOAD");
 
-    if (altPlatforms == nullptr) {
-        for (auto libName : knownAdaptersNames) {
-            enabledPlatforms.emplace_back(libName);
-        }
-    } else {
```
veselypeta, line 39: Unfortunately I had to remove `knownAdaptersNames` from the windows build as the MSVC compiler is quite picky about zero-length arrays. Perhaps I should remove from linux equivalent too?

igchor: You could use std::array. zero-length std:array is well defined so Windows should not complain. But if we don't have any adapters yet removing it will work as well I guess.

pbalcer: Yea, I wouldn't worry about removing this - this code needs to be refactored anyway.

pbalcer (approved):

PR #263, opened 2023-02-14
Run clang-format twice on generated sources

igchor:
Clang-format is sometimes not idempotent. This is probably a bug: https://stackoverflow.com/questions/61754044/clang-format-makes-changes-to-an-already-formatted-file .


pbalcer:
That's an hour wasted spent debugging that I'm never getting back...

I *think* setting `ReflowComments: false` addresses the problem.

PR #265, opened 2023-02-14
[ur] Add pedantic option

jandres742:
This to align with SYCL compiler chain.

Signed-off-by: Jaime Arteaga <jaime.a.arteaga.molina@intel.com>


source/drivers/null/ur_nullddi.cpp
```diff
@@ -389,9 +389,9 @@ __urdlllocal ur_result_t UR_APICALL urEnqueueMemBufferReadRect(
     size_t bufferSlicePitch, ///< [in] length of each 2D slice in bytes in the
                              ///< buffer object being read
     size_t hostRowPitch,     ///< [in] length of each row in bytes in the host
-                         ///< memory region pointed by dst
-    size_t hostSlicePitch, ///< [in] length of each 2D slice in bytes in the
-                           ///< host memory region pointed by dst
+                             ///< memory region pointed by dst
```
jandres742, line 392: all these changes here are because of clang-format.

pbalcer (approved):

PR #267, opened 2023-02-15
disable ReflowComments in .clang-format

pbalcer:
This option is not idempotent, potentially causing issues when formatting multiple times.

Alternative to #263

### DIFF
--- a/source/drivers/null/ur_nullddi.cpp
+++ b/source/drivers/null/ur_nullddi.cpp
@@ -12,12 +12,11 @@
 namespace driver {
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Intercept function for urContextCreate
-__urdlllocal ur_result_t UR_APICALL urContextCreate(
-    uint32_t DeviceCount, ///< [in] the number of devices given in phDevices
-    ur_device_handle_t
-        *phDevices, ///< [in][range(0, DeviceCount)] array of handle of devices.
-    ur_context_handle_t
-        *phContext ///< [out] pointer to handle of context object created
+__urdlllocal ur_result_t UR_APICALL
+urContextCreate(
+    uint32_t DeviceCount,          ///< [in] the number of devices given in phDevices
+    ur_device_handle_t *phDevices, ///< [in][range(0, DeviceCount)] array of handle of devices.
+    ur_context_handle_t *phContext ///< [out] pointer to handle of context object created
 ) {
     ur_result_t result = UR_RESULT_SUCCESS;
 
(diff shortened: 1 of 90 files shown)


kbenzie:
Interesting, I recall a discussion about not formatting the `ur_api.h` header because of comments being reflowed. With this option maybe that doesn't look as bad as it did before?

source/drivers/null/ur_nullddi.cpp
```diff
@@ -78,20 +75,19 @@ __urdlllocal ur_result_t UR_APICALL urContextRelease(
...
-                        ///< if propSize is not equal to or greater than the
-                        ///< real number of bytes needed to return the info then
-                        ///< the ::UR_RESULT_ERROR_INVALID_SIZE error is
-                        ///< returned and pContextInfo is not used.
-    size_t *pPropSizeRet ///< [out][optional] pointer to the actual size in
-                         ///< bytes of data queried by ContextInfoType.
+    ///< if propSize is not equal to or greater than the real number of bytes
+    ///< needed to return
+    ///< the info then the ::UR_RESULT_ERROR_INVALID_SIZE error is returned and
+    ///< pContextInfo is not used.
```
kbenzie, line 84: this comment seems a bit mangled

source/drivers/null/ur_nullddi.cpp
```diff
@@ -179,37 +172,35 @@ __urdlllocal ur_result_t UR_APICALL urContextSetExtendedDeleter(
...
+    const size_t *
+        pGlobalWorkSize, ///< [in] pointer to an array of workDim unsigned values that specify the
+    ///< number of global work-items in workDim that will execute the kernel
+    ///< function
+    const size_t *
+        pLocalWorkSize, ///< [in][optional] pointer to an array of workDim unsigned values that
+    ///< specify the number of local work-items forming a work-group that will
+    ///< execute the kernel function.
+    ///< If nullptr, the runtime implementation will choose the work-group
+    ///< size.
```
kbenzie, line 190: seems like the old breaks are still present but are no longer aligned, was this originally one long line before the previous formatting?

pbalcer:
I was wrong. The `ur_api.h` header looks decent when `ColumnLimit` is set to higher. I added a new commit where I changed it to 140, so you can check it out. This change also helps with comments around function arguments.

kbenzie:
That looks not bad, there's some functions that are changed from one argument per line. Maybe `BinPackArguments` set to `false` would be useful. That's actually come up [here](https://github.com/oneapi-src/unified-runtime/pull/255#discussion_r1106984285) today.

pbalcer:
Added a commit with `BinPackArguments`. I'm not sure if this is much of an improvement, but I'm fine with either setting.

kbenzie:
Hmm, that did not have the effect I expected and you're right its not an improvement. I'd be fine with dropping the `BinPackArguments` option.

pbalcer:
I rebased this and now changed the `ColumnLimit` to 0. With 140, as I had previously, clang-format was generating super long lines. With no column limit, "clang-format will respect the input’s line breaking decisions within statements unless they contradict other rules.". So we can have long documentation comments in the headers  but also can manually format code to some reasonable column limit.
We could also write a simple script to enforce a soft column limit by yelling in CI when a non-comment line exceeds the limit.

Any opposition to merging this?

kbenzie (approved):

PR #274, opened 2023-02-16
Make array of devices passed to urContextCreate const

### LINKED ISSUE #273: `urContextCreate` should take a const devices parameter
This entry point currently takes a pointer to a device list 'phDevices':
```cpp
urContextCreate(uint32_t DeviceCount, ur_device_handle_t *phDevices, ur_context_handle_t *phContext)
```
However in PI the equivalent has a `const` device list parameter:
```cpp
pi_result piContextCreate(
    const pi_context_properties *properties, pi_uint32 num_devices,
    const pi_device *devices,
    void (*pfn_notify)(const char *errinfo, const void *private_info, size_t cb,
                       void *user_data),
    void *user_data, pi_context *ret_context)
```

This mismatch between const/non-const is problematic for translating between PI and UR, so the `phDevices` parameter in urContextCreate should instead be `const ur_device_handle_t *`. This also brings UR in line with clCreateContext in OpenCL (which takes `const cl_device_id* devices`).

kbenzie:
Add `const` to the `phDevices` argument of `urContextCreate` to match up
with PI/OpenCL and avoid unnecessary potential for `const_cast` in
adapter implementations.

Fixes #273

### DIFF
--- a/cmake/helpers.cmake
+++ b/cmake/helpers.cmake
@@ -31,6 +31,7 @@ function(add_cppformat name)
                 --style=file
                 --i
                 ${ARGN}
+            COMMENT "Format CXX source files"
             )
     endif()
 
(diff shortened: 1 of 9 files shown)


callumfare:
LGTM

jandres742 (commented):
+1

PR #275, opened 2023-02-16
Add ur_program_properties_t for program creation

### LINKED ISSUE #95: Reconsider urProgramCreateWithBinary signature
# Issue

The `piProgramCreateWithBinary` and `urProgramCreateWithBinary` entry points have different semantics.

`piProgramCreateWithBinary` has the following signature:
```
__SYCL_EXPORT pi_result piProgramCreateWithBinary(
    pi_context context, pi_uint32 num_devices, const pi_device *device_list,
    const size_t *lengths, const unsigned char **binaries,
    size_t num_metadata_entries, const pi_device_binary_property *metadata,
    pi_int32 *binary_status, pi_program *ret_program);
```

and `urProgramCreateWithBinary` has the following signature:
```
urProgramCreateWithBinary(
    ur_context_handle_t hContext,                   ///< [in] handle of the context instance
    ur_device_handle_t hDevice,                     ///< [in] handle to device associated with binary.
    size_t size,                                    ///< [in] size in bytes.
    const uint8_t* pBinary,                         ///< [in] pointer to binary.
    ur_program_handle_t* phProgram                  ///< [out] pointer to handle of Program object created.
    );

```

From the above we can see the following differences:
1. `piProgramCreateWithBinary` takes a list of devices to create the program for (via the `num_devices` & `device_lists` params) where `urProgramCreateWithBinary` just takes a single device (via `hDevice`).
2. `piProgramCreateWithBinary` takes a list of binaries to create  a program with (via `lengths` and `binaries` params) whereas `urProgramCreateWithBinary` takes single binary (via `size` and `pBinary params`).
3. `piProgramCreateWithBinary` takes a list of metadata entries (via the `num_metadata_entries` and `metadata` params)
4. `piProgramCreateWithBinary` takes a `binary_status` param.

Examining the [usage of `piProgramCreateWithBinary`](https://github.com/intel/llvm/blob/b4aaca7bc3801e23b4df49ebe35a1057881a7414/sycl/source/detail/program_manager/program_manager.cpp#L86) we see that the DPC++ only passes one device to the entry point and takes one binary so differences 1. and 2. are valid. We do see however that the DPC++ runtime passes a metadata argument to the entry point, so difference 3. is invalid. Finally, the `binary_status` out parameter is an array of statuses, one for each device in the input, if there is only ever one device then this argument is redundant and can be folded into the return value.

As an aside, it looks like some adapters may only be able to support a single device e.g. [L0](https://github.com/intel/llvm/blob/c4932957929f94afe7f87bee009eea5e2b5cd6bf/sycl/plugins/level_zero/pi_level_zero.cpp#L4519).

# Task

Add the metadata arguments to the `urProgramCreateWithBinary` so that the UR entry point matches the requirements of the DPC++ runtime.

kbenzie:
The new `ur_program_properties_t` struct allows the provision of
additional metadata to be passed into the
`urProgramCreate`/`urProgramCreateWithBinary` entry points.
Additionally, it supports `stype` and `pNext` which enable extending
program creation without breaking ABI.

Fixes #95

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,14 +176,14 @@ install(
 set(API_JSON_FILE ${PROJECT_BINARY_DIR}/unified_runtime.json)
 
 # Generate source from the specification
-add_custom_target(generate-code
+add_custom_target(generate-code USES_TERMINAL
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/scripts
     COMMAND ${Python3_EXECUTABLE} run.py --api-json ${API_JSON_FILE}
     COMMAND ${Python3_EXECUTABLE} json2src.py --api-json ${API_JSON_FILE} ${PROJECT_SOURCE_DIR}
 )
 
 # Generate and format source from the specification
-add_custom_target(generate
+add_custom_target(generate USES_TERMINAL
     COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --target generate-code
     COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --target cppformat
 )
--- a/include/ur.py
+++ b/include/ur.py
@@ -205,6 +205,7 @@ class ur_result_t(c_int):
 ## @brief Defines structure types
 class ur_structure_type_v(IntEnum):
     IMAGE_DESC = 0                                  ## ::ur_image_desc_t
+    PROGRAM_PROPERTIES = 1                          ## ::ur_program_properties_t
 
 class ur_structure_type_t(c_int):
     def __str__(self):
@@ -1035,6 +1036,55 @@ class ur_api_version_t(c_int):
 
 
 ###############################################################################
+## @brief Program metadata property type.
+class ur_program_metadata_type_v(IntEnum):
+    UINT32 = 0                                      ## type is a 32-bit integer.
+    UINT64 = 1                                      ## type is a 64-bit integer.
+    BYTE_ARRAY = 2                                  ## type is a byte array.
+    STRING = 3                                      ## type is a null-terminated string.
+
+class ur_program_metadata_type_t(c_int):
+    def __str__(self):
+        return str(ur_program_metadata_type_v(self.value))
+
+
+###############################################################################
+## @brief Program metadata value union.
+class ur_program_metadata_value_t(Structure):
+    _fields_ = [
+        ("data32", c_ulong),                                            ## [in] inline storage for the 32-bit data, type
+                                                                        ## ::UR_PROGRAM_METADATA_TYPE_UINT32.
+        ("data64", c_ulonglong),                                        ## [in] inline storage for the 64-bit data, type
+                                                                        ## ::UR_PROGRAM_METADATA_TYPE_UINT64.
+        ("pString", c_char_p),                                          ## [in] pointer to null-terminated string data, type
+                                                                        ## ::UR_PROGRAM_METADATA_TYPE_STRING.
+        ("pData", c_void_p)                                             ## [in] pointer to binary data, type
+                                                                        ## ::UR_PROGRAM_METADATA_TYPE_BYTE_ARRAY.
+    ]
+
+###############################################################################
+## @brief Program metadata property.
+class ur_program_metadata_t(Structure):
+    _fields_ = [
+        ("pName", c_char_p),                                            ## [in] null-terminated metadata name.
+        ("type", ur_program_metadata_type_t),                           ## [in] the type of metadata value.
+        ("size", c_size_t),                                             ## [in] size in bytes of the data pointed to by value.pData, or 0 when
+                                                                        ## value size is less than 64-bits and is stored directly in value.data.
+        ("value", ur_program_metadata_value_t)                          ## [in] the metadata value storage.
+    ]
+
+###############################################################################
+## @brief Program creation properties.
+class ur_program_properties_t(Structure):
+    _fields_ = [
+        ("stype", ur_structure_type_t),                                 ## [in] type of this structure
+        ("pNext", c_void_p),                                            ## [in,out][optional] pointer to extension-specific structure
+        ("count", c_ulong),                                             ## [in] the number of entries in pMetadatas, if count is greater than
+                                                                        ## zero then pMetadatas must not be null.
+        ("pMetadatas", POINTER(ur_program_metadata_t))                  ## [in][optional][range(0,count)] pointer to array of metadata entries.
+    ]
+
+###############################################################################
 ## @brief Get Program object information
 class ur_program_info_v(IntEnum):
     REFERENCE_COUNT = 0                             ## Program reference count info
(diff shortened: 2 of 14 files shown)


CMakeLists.txt
```diff
@@ -176,14 +176,14 @@ install(
 set(API_JSON_FILE ${PROJECT_BINARY_DIR}/unified_runtime.json)
 
 # Generate source from the specification
-add_custom_target(generate-code
+add_custom_target(generate-code USES_TERMINAL
```
jandres742, line 179: why do we need USES_TERMINAL?

scripts/core/program.yml
```diff
@@ -10,6 +10,67 @@ type: header
...
+name: $x_program_metadata_type_t
+etors:
+    - name: UNKNOWN  # TODO: Is this necessary?
+      desc: "type is unknown."
+    - name: UINT32
+      desc: "type is a 32-bit integer."
+    - name: UINT64
+      desc: "type is a 32-bit integer."
+    - name: BYTE_ARRAY
+      desc: "type is a byte array."  # TODO: null-terminated?
```
jandres742, line 24: if the question is if we should ask users to provide the null-terminated character, I'd say yes.

scripts/core/program.yml
```diff
@@ -10,6 +10,67 @@ type: header
 desc: "Intel $OneApi Level-Zero Runtime APIs for Program"
 ordinal: "2"
 --- #--------------------------------------------------------------------------
+type: enum
+desc: "Program metadata property type."
+name: $x_program_metadata_type_t
+etors:
+    - name: UNKNOWN  # TODO: Is this necessary?
```
jandres742, line 17: this enum is to define the type, so UR knows what to do with it, if it not known, kind of contradicts the purpose, no?

include/ur_api.h
```diff
@@ -4111,6 +4112,48 @@ urGetLastResult(
...
+    ur_program_metadata_value_t value; ///< [in] the metadata value storage.
+
+} ur_program_metadata_t;
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Program creation properties.
+typedef struct ur_program_properties_t {
+    ur_structure_type_t stype;               ///< [in] type of this structure
+    void *pNext;                             ///< [in,out][optional] pointer to extension-specific structure
+    uint32_t count;                          ///< [in] the number of entries in pMetadatas.
```
jandres742, line 4152: should we clarify that if count is non-zero, then pMetaDatas should be non-null? kind of obvious but maybe need to add it?

scripts/core/program.yml
kbenzie: I should deleted this comment

scripts/core/program.yml
kbenzie: Unknown is actually in [PI already](https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/pi.h#L709-L715) which I was unsure about.

include/ur_api.h
kbenzie: Sounds like a good idea, Ideally there would be error codes for this also. I've not figured out how to do that with the generator yet.

CMakeLists.txt
kbenzie: This is not essential but I find it makes iterating on spec changes faster.

scripts/core/program.yml
kbenzie: We actually have a `size` paremeter [here](https://github.com/oneapi-src/unified-runtime/pull/275/files#diff-49df2b979e9b956e061467dd2f8e1a6a07d5a2ac09f33acf90dac1603cb3ceefR54-R56) so null termination is not strictly necessary.

scripts/core/program.yml
kbenzie: I have removed `UNKNOWN` for now. If its actually being used in SYCL RT we can add it back in later.

include/ur_api.h
kbenzie: Updated the documentation and also added error codes to `urProgramCreate[WithBinary]`.

include/ur.py
```diff
@@ -1033,6 +1034,50 @@ def __str__(self):
         return str(ur_api_version_v(self.value))
 
 
+###############################################################################
+## @brief Program metadata property type.
+class ur_program_metadata_type_v(IntEnum):
+    UINT32 = 0                                      ## type is a 32-bit integer.
+    UINT64 = 1                                      ## type is a 32-bit integer.
```
martygrant, line 1041: Typo in the comment for UINT64 saying it's 32-bit?

include/ur_api.h
```diff
@@ -4111,6 +4112,48 @@ urGetLastResult(
 #if !defined(__GNUC__)
 #pragma region program
 #endif
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Program metadata property type.
+typedef enum ur_program_metadata_type_t {
+    UR_PROGRAM_METADATA_TYPE_UINT32 = 0,     ///< type is a 32-bit integer.
+    UR_PROGRAM_METADATA_TYPE_UINT64 = 1,     ///< type is a 32-bit integer.
```
martygrant, line 4119: Typo in the comment for UINT64 saying it's 32-bit?

scripts/core/program.yml
```diff
@@ -10,6 +10,65 @@ type: header
...
 ordinal: "2"
 --- #--------------------------------------------------------------------------
+type: enum
+desc: "Program metadata property type."
+name: $x_program_metadata_type_t
+etors:
+    - name: UINT32
+      desc: "type is a 32-bit integer."
+    - name: UINT64
+      desc: "type is a 32-bit integer."
```
martygrant, line 20: Typo for UINT64 saying it's 32-bit?

source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -3246,20 +3247,29 @@ urProgramCreate(
         if (NULL == phProgram) {
```
martygrant, line 3247: Curious why `NULL` is used over `nullptr`?

martygrant (approved):

scripts/core/program.yml
```diff
@@ -10,6 +10,65 @@ type: header
...
+class: $xProgram
+name: $x_program_metadata_value_t
+members:
+    - type: char*
+      name: pString
+      desc: "[in] pointer to null-terminated string data."
+    - type: void*
+      name: pData
+      desc: "[in] pointer to binary data."
+    - type: uint64_t
```
veselypeta, line 37: Will a `uint32_t` be store here?

i.e. 
```cpp
uint32_t my_data = 42;
ur_program_metadata_value_t md_value{.data = static_cast<uint64_t>(my_data)};
ur_program_metadata_t md{ "my_metadata", UR_PROGRAM_METADATA_TYPE_UINT32, 0, md_value};

```

veselypeta (approved):
LGTM

source/loader/layers/validation/ur_valddi.cpp
kbenzie: For this `NULL == phProgram` is automatically generated, I'm not sure where the source of it is. For the new checks I decided to match the existing usage of `NULL` to make the spec consisten.

I beleive the reason to use `NULL` is that the API defined in `ur_api.h` is a C API where `nullptr` is not a valid keyword.

scripts/core/program.yml
kbenzie: Correct.

source/loader/layers/validation/ur_valddi.cpp
martygrant: Good to know, thanks!

scripts/core/program.yml
pbalcer: Given that this is a union, are we gaining anything by not having a dedicated `uint32_t` field? Fewer fields?

kbenzie: That's a good point, the original PI interface was not a union. I'll make this a separate member.

aarongreig (approved):
LGTM

PR #278, opened 2023-02-17
[ur] Remove semicolon from extern C

veselypeta:
The `-Wpedantic` option was added to the UR build - when building I get the following error.

```
[build] ../source/drivers/null/ur_nullddi.cpp:3404:2: error: extra ‘;’ [-Werror=pedantic]
[build]  3404 | };
```

This additional semicolon is unnecessary for `extern "C" {}` construct so I have removed it.


kbenzie (approved):
LGTM

martygrant (approved):
Looks good

PR #279, opened 2023-02-17
Add tests for urEnqueueUSMMemcpy

fabiomestre:


martygrant (approved):
LGTM

test/conformance/enqueue/urEnqueueUSMMemcpy.cpp
```diff
@@ -1,2 +1,140 @@
...
+
+        bool device_usm{false};
+        ASSERT_SUCCESS(
+            urDeviceGetInfo(device, UR_DEVICE_INFO_USM_DEVICE_SUPPORT, sizeof(bool), &device_usm, nullptr));
+
+        if (!device_usm) {
+            GTEST_SKIP_("Device USM is not supported");
+        }
+
+        ur_usm_mem_flags_t flags;
```
veselypeta, line 19:
```suggestion
        ur_usm_mem_flags_t flags{};
```
I think you should zero the flags - there is no guarantee that it will be zero and may result in some of the flags being set.

test/conformance/enqueue/urEnqueueUSMMemcpy.cpp
```diff
@@ -1,2 +1,140 @@
...
+ */
+TEST_P(urEnqueueUSMMemcpyTest, WaitForDependencies) {
+
+    ASSERT_SUCCESS(
+        urEnqueueUSMMemcpy(queue, true, device_dst, device_src, sizeof(int), 1, &memset_event, nullptr));
+    ASSERT_TRUE(verifyData());
+}
+
+TEST_P(urEnqueueUSMMemcpyTest, InvalidNullQueueHandle) {
+
```
veselypeta, line 115: What happens to the `memset_event` in these tests?

fabiomestre: It gets released on TearDown()

test/conformance/enqueue/urEnqueueUSMMemcpy.cpp
```diff
@@ -1,2 +1,140 @@
...
+};
+
+/**
+ * Test that urEnqueueUSMMemcpy blocks when the blocking parameter is set to true.
+ */
+TEST_P(urEnqueueUSMMemcpyTest, Blocking) {
+
+    ASSERT_SUCCESS(urEventWait(1, &memset_event));
+    ASSERT_SUCCESS(urEnqueueUSMMemcpy(queue, true, device_dst, device_src, allocation_size, 0, nullptr, nullptr));
+    ASSERT_TRUE(verifyData());
```
kbenzie, line 67: I think it make sense to assert that the event has actually completed before verifying the data:

```cpp
ur_event_status_t memset_event_status;
ASSERT_SUCCESS(urEventGetInfo(memset_event, UR_EVENT_INFO_COMMAND_EXECUTION_STATUS,
    sizeof(ur_event_status_t), &memset_event_status, nullptr));
ASSERT_EQ(UR_EVENT_STATUS_COMPLETE, memset_event_status);
```

This also applies to the other tests which verify the data.

test/conformance/enqueue/urEnqueueUSMMemcpy.cpp
veselypeta: I mean does it mater that you don't wait for the event to complete with `urEventWait` like you've done in the previous tests?

fabiomestre: I don't think it matters. I'm not expecting a successful memory copy in negative tests. So the fact that the memset might not be complete is irrelevant to the outcome of the test.

veselypeta: Sounds good. I wasn't sure if you had to wait for the event to complete before releasing it 👍

veselypeta (approved):

test/conformance/enqueue/urEnqueueUSMMemcpy.cpp
fabiomestre: Done :)

PR #282, opened 2023-02-20
[ur] Complete Native Handle Tests

### LINKED ISSUE #222: Consider removing `CreateWithNativeHandle` Tests
In the CTS it's probably going to be difficult if not impossible to test these entry points since it relies on the higher level runtime having a handle to the specific adapter.

We could still however test all the error codes of these entry-points, but lack a 'happy path' test.

veselypeta:
Since we don't have a `native_handle` the only valid test we can write is for when we have a `nullptr` native handle. This PR adds the same test to `event`. `device`, `context`, `memory`, `queue` were completed previously. 

This Closes #222 & #224


fabiomestre (approved):

martygrant (approved):
LGTM

PR #286, opened 2023-02-20
Add UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST to enqueue entrypoints

fabiomestre:


scripts/core/enqueue.yml
```diff
@@ -54,6 +54,7 @@ returns:
     - $X_RESULT_ERROR_INVALID_QUEUE
     - $X_RESULT_ERROR_INVALID_KERNEL
     - $X_RESULT_ERROR_INVALID_EVENT
+    - $X_RESULT_ERROR_INVALID_EVENT_WAIT_LIST
```
veselypeta, line 57:
```suggestion
    - $X_RESULT_ERROR_INVALID_EVENT_WAIT_LIST
        - "phEventWaitList == NULL && numEventsInWaitList > 0"
        - "phEventWaitList != NULL && numEventsInWaitList == 0"
```

I saw @kbenzie do something like this [here](https://github.com/oneapi-src/unified-runtime/pull/275/files#diff-49df2b979e9b956e061467dd2f8e1a6a07d5a2ac09f33acf90dac1603cb3ceefR141). I think it would be nice adding this so that the conditions for this error code are documented per entry-point. Not 100% sure though.

test/conformance/enqueue/urEnqueueMemImageCopy.cpp
```diff
@@ -1,2 +1,4 @@
 // Copyright (C) 2023 Intel Corporation
 // SPDX-License-Identifier: MIT
+
+//TODO Add UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST negative test when this is implemented.
```
veselypeta, line 4: I think this comment is unnecessary. The error code will still be covered in the description of the issue i.e. https://github.com/oneapi-src/unified-runtime/issues/235

veselypeta (commented):
Looks good. Just a couple of questions

scripts/core/enqueue.yml
fabiomestre: I think that makes sense. I added the details to every entrypoint

test/conformance/enqueue/urEnqueueMemImageCopy.cpp
fabiomestre: Fair enough!

scripts/core/enqueue.yml
fabiomestre: Seems like this is used as input for validation layers :O
Need to update it again and keep only the code conditions

fabiomestre: Seems to be passing CI now :+1:

jandres742 (commented):
+1

scripts/core/enqueue.yml
veselypeta: nice!

veselypeta: nice!

veselypeta (approved):
LGTM

kbenzie (approved):
LGTM

PR #287, opened 2023-02-20
[ur] CTS release/retain tests increment reference counts

### LINKED ISSUE #248: We should check Release/Retain entry-points actually increment reference count.
Currently the ported tests from UnitUR do not include tests that check that release/retain entry-points actually increment the reference count. @fabiomestre added tests that check for this in `urEventRetain` & `urEventRelease` - we should mimic these tests for the remaining release/retain entry-points.

veselypeta:
`urRetain*` and `urRelease*` entry-points should increase and decrease the reference count of an object respectivelly. We have getInfo queries that can query the reference count from each object. These queries have the caveat that their value cannot be used reliably, especially in a multithreaded context. However, they are useful in a testing scenario where we disallow multithreading and assert that the reference count has changed after a release/retain call. 

This PR:
* Extends object lifetime testing and asserts that the reference count has changed after each release/retain call
* Clarifies the spec that `REFERENCE_COUNT` queries cannot be used reliably in an application context, especially a multithreaded one.

Closes #248

### DIFF
--- a/scripts/core/context.yml
+++ b/scripts/core/context.yml
@@ -69,7 +69,10 @@ etors:
     - name: DEVICES
       desc: "[$x_context_handle_t...] The array of the device handles in the context"
     - name: REFERENCE_COUNT
-      desc: "[uint32_t] Reference count of the context object."
+      desc: |
+            [uint32_t] Reference count of the context object.
+            The reference count returned should be considered immediately stale. 
+            It is unsuitable for general use in applications. This feature is provided for identifying memory leaks.
     - name: USM_MEMCPY2D_SUPPORT
       desc: "[bool] to indicate if the $xEnqueueUSMMemcpy2D entrypoint is supported."
     - name: USM_FILL2D_SUPPORT
--- a/scripts/core/context.yml
+++ b/scripts/core/context.yml
@@ -65,19 +65,16 @@ class: $xContext
 name: $x_context_info_t
 etors:
     - name: NUM_DEVICES
-      value: "1"
       desc: "[uint32_t] The number of the devices in the context"
     - name: DEVICES
-      value: "2"
       desc: "[$x_context_handle_t...] The array of the device handles in the context"
+    - name: REFERENCE_COUNT
+      desc: "[uint32_t] Reference count of the context object."
     - name: USM_MEMCPY2D_SUPPORT
-      value: "3"
       desc: "[bool] to indicate if the $xEnqueueUSMMemcpy2D entrypoint is supported."
     - name: USM_FILL2D_SUPPORT
-      value: "4"
       desc: "[bool] to indicate if the $xEnqueueUSMFill2D entrypoint is supported."
     - name: USM_MEMSET2D_SUPPORT
-      value: "5"
       desc: "[bool] to indicate if the $xEnqueueUSMMemset2D entrypoint is supported."
 --- #--------------------------------------------------------------------------
 type: function
--- a/test/conformance/testing/uur/utils.h
+++ b/test/conformance/testing/uur/utils.h
@@ -14,19 +14,24 @@ namespace uur {
 inline std::string GTestSanitizeString(const std::string &str) {
     auto str_cpy = str;
     std::replace_if(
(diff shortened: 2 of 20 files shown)


scripts/core/context.yml
```diff
@@ -65,19 +65,16 @@ class: $xContext
 name: $x_context_info_t
 etors:
     - name: NUM_DEVICES
-      value: "1"
       desc: "[uint32_t] The number of the devices in the context"
     - name: DEVICES
-      value: "2"
       desc: "[$x_context_handle_t...] The array of the device handles in the context"
+    - name: REFERENCE_COUNT
```
pbalcer, line 71: Is this useful in scenarios besides simple tests? This sort of functionality can suffer from time of check/time of use (TOCTOU) races.

For testing, instead of just checking if `fetch_and_add` worked, it would be more useful to test the outcome - whether a proper sequence of calls appropriately retains or deallocates memory. Typically I would do this by integrating valgrind, but in UR case we'd probably need an API to query heap statistics.

veselypeta: I'm inclined to agree, these queries exist in OpenCL with the caveat that `The reference count returned should be considered immediately stale. It is unsuitable for general use in applications. This feature is provided for identifying memory leaks. `.

pbalcer: My experience is that people rarely read docs and are happy to misuse an API if it appears to help with their problem at first glance. In this case, I'm not even sure if you could use this for leak checking. I can't imagine it would be safe to call this on an object that's already released. You'd be reading from freed/unitialized memory, or, even worse, from another allocation.

@kswiecicki has been working on a validation layer to track object lifetimes, #289, which I think is going to be a better solution for this particular problem.

veselypeta: Through some discussions we think that it's useful to assert that the reference count has changed after a release/retain entry point. During testing we are eliminating the concern for a multi-threaded context and thus race conditions like you suggested. I have however updated the documentation to clarify the danger with these queries.

pbalcer: Well, I'm going to choose my battles, but I still don't entirely agree that this functionality is useful for detecting leaks :)

test/conformance/testing/uur/utils.h
```diff
@@ -42,6 +47,46 @@ inline std::string GetPlatformAndDeviceName(ur_device_handle_t device) {
     return GetPlatformName(GetPlatform()) + "__" + GetDeviceName(device);
 }
 
+inline std::pair<uint32_t, bool>
+urContextGetReferenceCount(ur_context_handle_t context) {
```
kbenzie, line 51: Since these are in the `uur` namespace the `ur` prefix seems unnecessary and potentially misleading as these are not part of the spec.

kbenzie (approved):
Can potentially make some of the code cleaner once #293 is merged but looks good anyway.

fabiomestre (approved):

test/conformance/testing/uur/utils.h
```diff
@@ -42,6 +48,39 @@ inline std::string GetPlatformAndDeviceName(ur_device_handle_t device) {
...
+    }
+    if constexpr (std::is_same_v<T, ur_device_handle_t>) {
+        return GetInfo<uint32_t>(object, UR_DEVICE_INFO_REFERENCE_COUNT, urDeviceGetInfo);
+    }
+    if constexpr (std::is_same_v<T, ur_event_handle_t>) {
+        return GetInfo<uint32_t>(object, UR_EVENT_INFO_REFERENCE_COUNT, urEventGetInfo);
+    }
+    if constexpr (std::is_same_v<T, ur_queue_handle_t>) {
+        return GetInfo<uint32_t>(object, UR_QUEUE_INFO_REFERENCE_COUNT, urQueueGetInfo);
+    }
```
kbenzie, line 80: Should we also handle the other handle types so it doesn't need modified later?

kbenzie (approved):
LGTM

test/conformance/testing/uur/utils.h
veselypeta: Done

PR #290, opened 2023-02-21
Exclude `<ENUM>_FORCE_UINT32` from Doxygen

kbenzie:
Enums generated in C or C++ files include an `<ENUM>_FORCE_UINT32`
enumeration to force the compiler to use the equivalent of `uint32_t`
for the underlying type. Unfortunately, this also made its way into the
Doxygen and therefore the generated HTML specification. This patch
unconditionally excludes the `<ENUM>_FORCE_UINT32` enumeration from the
Doxygen and thus also from the HTML specification.


fabiomestre (approved):

veselypeta (approved):
+1

PR #294, opened 2023-02-22
[ur] Add tests for queues with properties

### LINKED ISSUE #271: Add tests urQueueCreate properties
We should add conformance tests for urQueueCreate that test supported flag values (defined in ur_queue_flag_t).

Required Coverage:

At least one successful (i.e. called with correct arguments and assert on expected result, not applicable to CreateWithNativeHandle entry points)
If this a query entry-point or one which accepts flags etc : we should tests every possible enumeration (where possible).
A test case for each returned error code (exclude disaster situations i.e. OUT_OF_MEMORY etc)

veselypeta:
Closes #271

### DIFF
--- a/test/conformance/queue/urQueueCreate.cpp
+++ b/test/conformance/queue/urQueueCreate.cpp
@@ -12,6 +12,41 @@ TEST_P(urQueueCreateTest, Success) {
     ASSERT_SUCCESS(urQueueRelease(queue));
 }
 
+using urQueueCreateWithParamTest = uur::urContextTestWithParam<ur_queue_flag_t>;
+UUR_TEST_SUITE_P(
+    urQueueCreateWithParamTest,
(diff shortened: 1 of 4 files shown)


aarongreig:
LGTM, but would it be worth testing more/all of the property flags if we can query support for them?

test/conformance/queue/urQueueCreate.cpp
```diff
@@ -12,6 +12,22 @@ TEST_P(urQueueCreateTest, Success) {
     ASSERT_SUCCESS(urQueueRelease(queue));
 }
 
+TEST_P(urQueueCreateTest, SuccessWithProperties) {
```
fabiomestre, line 15: Would it make sense to check the other valid flags as well using a parameterized test?

fabiomestre (approved):

PR #295, opened 2023-02-22
[ur] use explicit function pointers

pbalcer:
UR spec defined 'fptr' as function types instead of actual pointers.
This is fine for function parameters but made it difficult to populate
params structs (e.g., `params.pfnNotify = &notify` didn't work).

> A function designator is an expression that has function type. Except when it is the
> operand of the *sizeof operator* or the unary *&* operator, a function designator with
> type "function returning type" is converted to an expression that has type "pointer to
> function returning type".
> C11 6.3.2.1-4

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -460,7 +460,7 @@ urContextCreateWithNativeHandle(
 
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Context's extended deleter callback function with user data.
-typedef void(ur_context_extended_deleter_t)(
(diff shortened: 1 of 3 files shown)


kbenzie (approved):
I recall there being some issues with the generated Python module when support for function pointers was added to the specification, not sure if we testing coverage of that though.

Otherwise LGTM.

pbalcer:
Yes, there's a test to validate the python module. My earlier attempt to fix this (by adding `*` to the parameters, not the typedef) actually broke CI because the python test stopped working.

Issue #166, opened 2023-01-12
Check urDeviceGetGlobalTimestamps returns a synchronized time.


veselypeta:
When testing the entry-point `urDeviceGetGlobalTimestamps` we should have an assertion to check that the timestamps returned from host and device are synchronized.

* There isn't really a definition of what 'synchronized' means but maybe we can enforce an upper bound on the latency between host/device.

PR #296, opened 2023-02-22
[CTS] Assert that device/host timers are synchronized

veselypeta:
This test is adapted from the OpenCL-CTS test [test_device_timer.cpp](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/device_timer/test_device_timer.cpp). I've assumed the same accuracy of %0.5 between the timers and the same delay of 5s. We might have to alter this accuracy once we have an implementation to pass the tests.

### DIFF
--- a/test/conformance/device/urDeviceGetGlobalTimestamps.cpp
+++ b/test/conformance/device/urDeviceGetGlobalTimestamps.cpp
@@ -1,7 +1,27 @@
 // Copyright (C) 2022-2023 Intel Corporation
 // SPDX-License-Identifier: MIT
 
+#include <chrono>
+#include <thread>
+#include <type_traits>
 #include <uur/fixtures.h>
+
+// WARNING  - This is the precision that is used in the OpenCL-CTS.
+//          - We might need to modify this value per-adapter.
+//          - Currently we are saying the error between host/device
+//          - timers is 0.5%
+constexpr double allowedTimerError = 0.005;
+
+/// @brief Return the absolute difference between two numeric values.
+/// @tparam T An numeric type.
+/// @param[in] a First value.
+/// @param[in] b Second value.
+/// @return The absolute difference between `a` and `b`.
+template <class T,
+          typename std::enable_if_t<std::is_arithmetic<T>::value, bool> = true>
+T absolute_difference(T a, T b) {
+    return std::max(a, b) - std::min(a, b);
+}
 
 using urDeviceGetGlobalTimestampTest = uur::urAllDevicesTest;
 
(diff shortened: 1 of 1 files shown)


kbenzie:
Is this addressing #166?

If so, remember to put the ticket being closed by the PR in the commit/description.

veselypeta:
> Is this addressing #166?
> 
> If so, remember to put the ticket being closed by the PR in the commit/description.

My bad - updated now.

kbenzie (approved):
LGTM

test/conformance/device/urDeviceGetGlobalTimestamps.cpp
```diff
@@ -30,6 +51,50 @@ TEST_F(urDeviceGetGlobalTimestampTest, SuccessNoTimers) {
...
+                                                   &hostStartTime));
+        ASSERT_SUCCESS(
+            urDeviceGetGlobalTimestamps(device, nullptr, &hostOnlyStartTime));
+        ASSERT_NE(deviceStartTime, 0);
+        ASSERT_NE(hostStartTime, 0);
+        ASSERT_NE(hostOnlyStartTime, 0);
+        ASSERT_GE(hostOnlyStartTime, hostStartTime);
+
+        // wait for timers to increment
+        std::this_thread::sleep_for(std::chrono::seconds(5));
```
aarongreig, line 70: could we make this less? 1 second is already a while for any modern processor and 5 seconds is a perceptible wait for the human running the test

aarongreig (commented):
LGTM pending comments

test/conformance/device/urDeviceGetGlobalTimestamps.cpp
veselypeta: I'm happy to reduce this, this was just the time used by the OpenCL-CTS.  What would be a sensible amount - maybe 500ms?

test/conformance/device/urDeviceGetGlobalTimestamps.cpp
```diff
@@ -30,6 +51,50 @@ TEST_F(urDeviceGetGlobalTimestampTest, SuccessNoTimers) {
     }
 }
 
+TEST_F(urDeviceGetGlobalTimestampTest, SuccessSynchronizedTime) {
+    for (auto device : devices) {
+        uint64_t deviceStartTime, deviceEndTime = 0;
```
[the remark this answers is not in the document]
veselypeta: TIL - I had assumed it would initialize them all.

test/conformance/device/urDeviceGetGlobalTimestamps.cpp
kbenzie: > this was just the time used by the OpenCL-CTS

ah yes, the test suite that's notoriously long running 😀 

>  What would be a sensible amount - maybe 500ms?

what's the resolution of the timers? perhaps 10x that amount?

test/conformance/device/urDeviceGetGlobalTimestamps.cpp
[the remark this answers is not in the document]
aarongreig: yeah I was surprised when I learned this, kinda disappointing but :shrug:

test/conformance/device/urDeviceGetGlobalTimestamps.cpp
aarongreig: I was going to suggest keeping it the same order of magnitude and going with 1s, still human perceptible but not by much

veselypeta: > what's the resolution of the timers? perhaps 10x that amount?
You can query the resolution with `UR_DEVICE_INFO_PROFILING_TIMER_RESOLUTION` - this would make it scalable. I quite like that idea.

PR #300, opened 2023-02-23
add initial tracing layer integrated with xpti

### LINKED ISSUE #143: add tracing support in the loader

pbalcer:
... and an example demonstrating its use.

This patch adds a tracing layer to the loader that transparently enables users to trace and hook into any UR call. This is done directly by leveraging SYCL's XPTI instrumentation framework.

For now, this is disabled by default and only runs in CI.

Closes #143

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ option(UR_USE_ASAN "enable AddressSanitizer" OFF)
 option(UR_USE_UBSAN "enable UndefinedBehaviorSanitizer" OFF)
 option(UR_USE_MSAN "enable MemorySanitizer" OFF)
 option(UMA_BUILD_SHARED_LIBRARY "Build UMA as shared library" OFF)
+option(UR_ENABLE_TRACING "enable api tracing through xpti" OFF)
 
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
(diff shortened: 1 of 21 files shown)


pbalcer:
This is a jumping-off point to a more fully-fledged overall tracing support, including tools and more complete use of xpti. I'll create github issues for those. For now, this brings UR on par in terms of basic tracing functionality with the level-zero loader.

If we all agree that tracing through xpti is a sufficient replacement for the callback-based approach, I'll also remove all the callback typedefs from the ur header.

cmake/helpers.cmake
```diff
@@ -54,3 +54,23 @@ macro(add_sanitizer_flag flag)
 
     set(CMAKE_REQUIRED_LIBRARIES ${SAVED_CMAKE_REQUIRED_LIBRARIES})
 endmacro()
+
+include(FetchContent)
+
+function(FetchContentSparse_Declare name GIT_REPOSITORY GIT_TAG GIT_DIR)
```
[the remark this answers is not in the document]
pbalcer: Sure.
Yes, it's to avoid downloading the entire llvm repo. Even the sparse checkout is fairly lengthy...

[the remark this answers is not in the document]
pbalcer: Done.

igchor (approved):

CMakeLists.txt
```diff
@@ -65,6 +62,34 @@ elseif(MSVC)
     endif()
 endif()
 
+if(UR_ENABLE_TRACING)
+    add_compile_definitions(UR_ENABLE_TRACING)
+
+    # fetch xpti proxy library for the tracing layer
+    FetchContentSparse_Declare(xpti https://github.com/intel/llvm.git "f25cc4006f4da3acd9f4050cd150065ab0857f4d" "xpti")
```
PatKamin, line 69: Can you use the name of the tag here, like "sycl-nightly/20230223" instead? This would tell us more how old is the current version of XPTI for possible future updates. However, if we don't plan to update XPTI from time to time, then it's irrelevant.

CMakeLists.txt
```diff
@@ -65,6 +62,34 @@ elseif(MSVC)
...
+    set(XPTI_ENABLE_TESTS OFF CACHE INTERNAL "Turn off xptifw tests")
+    FetchContentSparse_Declare(xptifw https://github.com/intel/llvm.git "f25cc4006f4da3acd9f4050cd150065ab0857f4d" "xptifw")
+    FetchContent_MakeAvailable(xptifw)
+    set_target_properties(xptifw PROPERTIES
+        LIBRARY_OUTPUT_DIRECTORY ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
+    )
+    if (MSVC)
+        set_property(TARGET xpti PROPERTY MSVC_RUNTIME_LIBRARY "${CMAKE_MSVC_RUNTIME_LIBRARY}")
+        set_property(TARGET xpti PROPERTY MSVC_RUNTIME_LIBRARY "${CMAKE_MSVC_RUNTIME_LIBRARY}")
+        # disable warning warning C4267
```
PatKamin, line 88: nit: Double "warning"

CMakeLists.txt
```diff
@@ -65,6 +62,34 @@ elseif(MSVC)
...
+    set(XPTI_DIR ${xpti_SOURCE_DIR})
+    set(XPTI_ENABLE_TESTS OFF CACHE INTERNAL "Turn off xptifw tests")
+    FetchContentSparse_Declare(xptifw https://github.com/intel/llvm.git "f25cc4006f4da3acd9f4050cd150065ab0857f4d" "xptifw")
+    FetchContent_MakeAvailable(xptifw)
+    set_target_properties(xptifw PROPERTIES
+        LIBRARY_OUTPUT_DIRECTORY ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
+    )
+    if (MSVC)
+        set_property(TARGET xpti PROPERTY MSVC_RUNTIME_LIBRARY "${CMAKE_MSVC_RUNTIME_LIBRARY}")
+        set_property(TARGET xpti PROPERTY MSVC_RUNTIME_LIBRARY "${CMAKE_MSVC_RUNTIME_LIBRARY}")
```
PatKamin, line 87: Did you mean `xptifw` here?

CMakeLists.txt
PatKamin, line 88: nit: Be more elaborate on what is this warning about, ie.:
```
(...): The compiler detected a conversion from size_t to a smaller type.
```

examples/collector/README.md
```diff
@@ -0,0 +1,25 @@
...
+You will also need to setup your chosen adapter (`UR_ADAPTERS_FORCE_LOAD`).
+
+For example, to run the `hello_world` example with the collector, use:
+
+```
+$ mkdir build
+$ cd build
+$ cmake .. -DUR_ENABLE_TRACING=ON
+$ make
+$ UR_ADAPTERS_FORCE_LOAD=./lib/libur_null.so XPTI_TRACE_ENABLE=1 XPTI_FRAMEWORK_DISPATCHER=./lib/libxptifw.so XPTI_SUBSCRIBERS=./lib/libcollector.so ./bin/hello_world
```
PatKamin, line 22: Can you set the XPTI-related env vars in CMakeLists during configuration or building? So the user won't have to type it all himself? After all, if I configure UR with UR_ENABLE_TRACING=ON then I most probably want to use this feature.

CMakeLists.txt
pbalcer: When I was writing this patch, I had to use this specific tag because it was the first commit with my fixes for the xptifw build system. Ideally, there'd be a tag for xpti version with a stable ABI, but no such luck.
I'll use a daily tag.

examples/collector/README.md
pbalcer: I'm not sure what you are asking. These are runtime variables that the xpti proxy library and the dispatcher look for during initialization. I cannot bake them into a binary, nor should they be.
Having all the environment variables spelled out explicitly helps the user see how to use the tracing functionality without anything being hidden by scripts.

examples/collector/collector.cpp
```diff
@@ -0,0 +1,135 @@
...
+                                     unsigned int minor_version,
+                                     const char *version_str,
+                                     const char *stream_name) {
+    if (!stream_name || std::string_view(stream_name) != UR_STREAM_NAME) {
+        std::cout << "Invalid stream name: " << stream_name << ". Expected " << UR_STREAM_NAME << ". Aborting." << std::endl;
+        return;
+    }
+
+    if (UR_MAKE_VERSION(major_version, minor_version) != UR_API_VERSION_CURRENT) {
+        std::cout << "Invalid stream version: " << major_version << "." << minor_version << ". Expected " << UR_MAJOR_VERSION(UR_API_VERSION_CURRENT) << "." << UR_MINOR_VERSION(UR_API_VERSION_CURRENT) << ". Aborting." << std::endl;
```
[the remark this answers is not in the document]
pbalcer: Enforced by `clang-format`. We've decided to uncap column width in the `.clang-format` as a compromise between developer experience (i.e., having to deal with an occasional long column) and nice-looking header files.
See #267 for more context.

PR #302, opened 2023-02-24
Remove a requirement for type: struct

igchor:
That disallows member to be 'handle_t'.

I'd like to create an opaque handle: `ur_usm_pool_handle_t` and have it as a member in `ur_usm_desc_t` in: https://github.com/oneapi-src/unified-runtime/pull/264.

@jandres742 said that this requirement: `It is just to avoid having in a struct handles, which by the moment we use this struct, might not be defined` so, in my opinion it should be safe to lift this requirement as long as we make sure that the handle is defined.

If you see some other problems with this change, I can just leave the `void*` as the type of pool handle or pass `ur_usm_pool_handle_t` as a separate argument to allocation function, but I think having everything in the desc struct is a better option.

### DIFF
--- a/scripts/YaML.md
+++ b/scripts/YaML.md
@@ -464,7 +464,7 @@ class ur_name_flags_v(IntEnum):
       - `in,out` is used for members that are both read and write; typically this is used for pointers to other data structures that contain both read and write members
     + `desc` may include one the following annotations: {`"[optional]"`}
       - `optional` is used for members that are pointers where it is legal for the value to be `nullptr`
-    + `type` must be an ISO-C standard identifier; except it may **not** be a `handle_t`
+    + `type` must be an ISO-C standard identifier
     + `name` must be a unique ISO-C standard identifier
   - A member may take the following optional scalar field: {`init`, `version`}
     + `init` will be used to initialize the C++ struct|union member's value
--- a/scripts/parse_specs.py
+++ b/scripts/parse_specs.py
@@ -259,9 +259,6 @@ def _validate_doc(f, d, tags, line_num):
             if not annotation:
                 raise Exception(prefix+"'desc' must start with {'[in]', '[out]', '[in,out]'}")
 
-            if type_traits.is_handle(item['type']):
-                raise Exception(prefix+"'type' must not be '*_handle_t': %s"%item['type'])
-
             if item['type'].endswith("flag_t"):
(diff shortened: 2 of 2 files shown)


jandres742:
> That disallows member to be 'handle_t'.
> 
> I'd like to create an opaque handle: `ur_usm_pool_handle_t` and have it as a member in `ur_usm_desc_t` in: #264.
> 
> @jandres742 said that this requirement: `It is just to avoid having in a struct handles, which by the moment we use this struct, might not be defined` so, in my opinion it should be safe to lift this requirement as long as we make sure that the handle is defined.
> 
> If you see some other problems with this change, I can just leave the `void*` as the type of pool handle or pass `ur_usm_pool_handle_t` as a separate argument to allocation function, but I think having everything in the desc struct is a better option.

thanks @igchor .

I think we should be typesafe and not use the void* for a handle. If we want the handle in a struct, then we should. I dont remember why L0 scripts (from which we derived the UR scripts) didnt allow that, so adding others in case they remember.

@bashbaug
@bmyates

pbalcer (approved):

bmyates:
@igchor   The loader does handle translation for all handles passed as function arguments in https://github.com/oneapi-src/unified-runtime/blob/6fb8e2620f1474428a539ef46d6dd47043c7d59b/source/loader/ur_ldrddi.cpp#L43

Adding handles inside structs will likely require some additional code in the loader to ensure they are translated correctly.

bashbaug:
One more thought: putting handles inside structs also means they cannot hash or serialize/de-serialize deterministically.  Even though we currently have pointers in structs, the data they point to will be invariant from run-to-run, and this would not be true for handles.

jandres742:
> One more thought: putting handles inside structs also means they cannot hash or serialize/de-serialize deterministically. Even though we currently have pointers in structs, the data they point to will be invariant from run-to-run, and this would not be true for handles.

umm, that's a very good point @bashbaug.

igchor:
Thanks @bmyates @bashbaug for pointing those things out. 

My only reason for putting `handle_t` inside the `desc_t` structure was to minimize number of arguments for alloc* functions. However, if having one extra argument means we can hash and serialize/de-serialize the desc_t deterministically it's probably worth it.

I'll get feedback on https://github.com/oneapi-src/unified-runtime/pull/264 if we are OK with having one more param to those functions and close this PR for now.

PR #303, opened 2023-02-27
[spec] Clarify USM allocation entry-points

veselypeta:
This PR clarifies USM allocation entry-points:
* Add additional error code checks for support and alignment
* Fix tests to check for USM support
* Test `urUSMFree` with each type of USM allocation


pbalcer (approved):
Why are flags passed by pointer in the first place? This is changing in #264, so I'm curious if that PR needs to change.

@igchor ping

veselypeta:
> > Why are flags passed by pointer in the first place?
> 
> Good question, its odd, I'm not sure. I think the pointer to struct approach makes much more sense.
> 
> > This is changing in #264, so I'm curious if that PR needs to change.
> 
> Yeah, I think this will need to reflect the changes in #264.

I agree, I'll rebase once that change is merged

kbenzie (approved):

igchor (approved):

PR #304, opened 2023-02-27
Add tests for urEnqueueUSMFill2D

### LINKED ISSUE #240: Adds CTS tests for `urEnqueueUSMFill2D` entry-point
We should add conformance tests for `urEnqueueUSMFill2D`.

Required Coverage:

fabiomestre:
Closes #240

### DIFF
--- a/test/conformance/enqueue/urEnqueueUSMFill2D.cpp
+++ b/test/conformance/enqueue/urEnqueueUSMFill2D.cpp
@@ -1,2 +1,288 @@
 // Copyright (C) 2023 Intel Corporation
 // SPDX-License-Identifier: MIT
+
+#include <random>
+#include <uur/fixtures.h>
+
+struct testParametersFill2D {
+    size_t pitch;
+    size_t width;
+    size_t height;
+    size_t pattern_size;
+};
+
+template <typename T>
+inline std::string printFill2DTestString(const testing::TestParamInfo<typename T::ParamType> &info) {
+    const auto device_handle = std::get<0>(info.param);
+    const auto
+        platform_device_name = uur::GetPlatformAndDeviceName(device_handle);
+    std::stringstream test_name;
+    test_name << platform_device_name << "__pitch__"
+              << std::get<1>(info.param).pitch
+              << "__width__" << std::get<1>(info.param).width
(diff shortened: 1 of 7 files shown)


test/conformance/enqueue/urEnqueueUSMFill2D.cpp
```diff
@@ -1,2 +1,288 @@
...
+
+    void TearDown() override {
+        if (ptr) {
+            EXPECT_SUCCESS(urUSMFree(context, ptr));
+        }
+
+        UUR_RETURN_ON_FATAL_FAILURE(uur::urQueueTest::TearDown());
+    }
+
+    const size_t pitch = 16;
```
veselypeta, line 185: I would avoid making member variables `const` : [see here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-constref)

veselypeta: make it `static constexpr` instead

veselypeta (commented):
Looks good. Just some suggestions.

test/conformance/enqueue/urEnqueueUSMFill2D.cpp
```diff
@@ -1,2 +1,288 @@
...
+
+struct urEnqueueUSMFill2DTestWithParam
+    : uur::urQueueTestWithParam<testParametersFill2D> {
+    void SetUp() override {
+        UUR_RETURN_ON_FATAL_FAILURE(urQueueTestWithParam::SetUp());
+
+        pitch = std::get<1>(GetParam()).pitch;
+        width = std::get<1>(GetParam()).width;
+        height = std::get<1>(GetParam()).height;
+        pattern_size = std::get<1>(GetParam()).pattern_size;
```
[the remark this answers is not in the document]
fabiomestre: Thanks for the tip. I think there is a bit too much going on with the structured bindings approach (create 4 new variables to then make a tuple and finally assign that tuple to the variable).  So I prefer the original approach.

veselypeta (approved):
LGTM

martygrant (approved):
LGTM

kbenzie (approved):
LGTM

PR #305, opened 2023-02-28
[CTS] Add tests for urEnqueueUSMMemcpy2D

### LINKED ISSUE #243: Adds CTS tests for `urEnqueueUSMMemcpy2D` entry-point
We should add conformance tests for `urEnqueueUSMMemcpy2D`.

Required Coverage:

veselypeta:
Closes #243


test/conformance/enqueue/urEnqueueUSMMemcpy2D.cpp
```diff
@@ -1,2 +1,167 @@
...
+    : uur::urQueueTestWithParam<uur::TestParameters2D> {
+    void SetUp() override {
+        UUR_RETURN_ON_FATAL_FAILURE(
+            uur::urQueueTestWithParam<uur::TestParameters2D>::SetUp());
+        const auto device_usm =
+            uur::GetDeviceInfo<bool>(device, UR_DEVICE_INFO_USM_DEVICE_SUPPORT);
+        ASSERT_TRUE(device_usm.has_value());
+        if (device_usm.value()) {
+            GTEST_SKIP() << "device USM not supported.";
+        }
```
omarahmed1111, line 17: Shouldn't this be `if(!device_usm.value())` ?

test/conformance/enqueue/urEnqueueUSMMemcpy2D.cpp
```diff
@@ -1,2 +1,167 @@
...
+        memcpy_event, UR_EVENT_INFO_COMMAND_EXECUTION_STATUS);
+    ASSERT_TRUE(eventStatus.has_value());
+    ASSERT_EQ(eventStatus.value(), UR_EVENT_STATUS_COMPLETE);
+
+    ASSERT_NO_FATAL_FAILURE(verifyMemcpySucceeded());
+}
+
+using urEnqueueUSMMemcpy2DNegativeTest = urEnqueueUSMMemcpy2DTestWithParam;
+UUR_TEST_SUITE_P(urEnqueueUSMMemcpy2DNegativeTest,
+                 ::testing::Values(uur::TestParameters2D{16, 16, 16}),
```
aarongreig, line 95: It's going to make naming this suite tricky (`urEnqueueUSMMemcpy2D16x16Test`?) but would it be worth putting in a test that checks for success with these dimensions?

aarongreig (approved):
LGTM pending comment

test/conformance/enqueue/urEnqueueUSMMemcpy2D.cpp
veselypeta: Done.

test/conformance/enqueue/urEnqueueUSMMemcpy2D.cpp
veselypeta: Good spot. I actually forgot to add the parametrized versions for the success case. I've changed the negative tests to use {1,1,1,} which is a tested success case.

omarahmed1111 (approved):
LGTM

PR #306, opened 2023-02-28
ur: Introduce urKernelSetSpecializationConstants

### LINKED ISSUE #269: Update urProgramSetSpecializationConstant to follow PI changes for CUDA
https://github.com/intel/llvm/pull/7946 introduces a change/additional version of the entry point to set specialization constants to enable better support for the feature in the CUDA plugin.

Track the changes made to the interface and update the UR specification to enable the same use case.

kbenzie:
The issue #269, and the PR intel/llvm#7946, outline a desire to enable adapters to opt-in to setting specialization constants on a kernel object rather than a program object. On these adapters it is not required to perform compilation or linking when setting specialization constants and as such they can be set immediately prior to `urEnqueueKernelLaunch` with very low overhead.

This patch:

* Adds the `UR_DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS` device query to determine if `urKernelSetSpecializationConstants` is supported by the device.
* Adds `ur_specialization_constant_info_t` to represent the data required to set an individual specialization constant
* Adds `urKernelSetSpecializationConstants` entry point, which adapters can opt-in to supporting, which enables setting multiple specialization constants on a kernel object.
* Replaces the singular `urProgramSetSpecializationConstant` with the plural `urProgramSetSpecializationConstants` to enable setting multiple specialization constants at once on a program object and have a consistent API between the program and kernel variations.

Fixes #269

### DIFF
--- a/scripts/core/kernel.yml
+++ b/scripts/core/kernel.yml
@@ -345,6 +345,32 @@ returns:
     - $X_RESULT_ERROR_INVALID_KERNEL_ARGUMENT_INDEX
 --- #--------------------------------------------------------------------------
 type: function
+desc: "Set an array of specialization constants on a Kernel."
+class: $xKernel
+name: SetSpecializationConstants
+details:
+    - "This entry point is optional, the application should query for support with device query $X_DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS passed to $xDeviceGetInfo."
+    - "Adapters which are capable of setting specialization constants immediately prior to $xEnqueueKernelLaunch with low overhead should implement this entry point."
+    - "Otherwise, if setting specialization constants late requires recompiling or linking a program, adapters should not implement this entry point."
+    - "The application may call this function from simultaneous threads for the same device."
+    - "The implementation of this function should be thread-safe."
+params:
+    - type: $x_kernel_handle_t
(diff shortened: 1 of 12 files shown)


jandres742:
thanks @kbenzie . Proposal is looking good in my opinion.

scripts/core/kernel.yml
```diff
@@ -345,6 +345,30 @@ returns:
     - $X_RESULT_ERROR_INVALID_KERNEL_ARGUMENT_INDEX
 --- #--------------------------------------------------------------------------
 type: function
+desc: "Set an array of specialization constants on a Kernel."
+class: $xKernel
+name: SetSpecializationConstants
+details:
+    - "This entry point is optional, the application should query for support with device query $X_DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS passed to $xDeviceGetInfo."
```
veselypeta, line 352: should we have the usual spiel here about the entry-points being 'thread-safe'

```yml
details:
    - "The application may call this function from simultaneous threads for the same device."
    - "The implementation of this function should be thread-safe."
```

aarongreig (approved):
LGTM

veselypeta (approved):
LGTM

Issue #50, opened 2022-11-24
ur: add enqueue fill image


veselypeta:
PI currently has an API function `piEnqueueMemImageFill` which is analogous to the openCL version `clEnqueueFillImage`

We should add this to Unified Runtime

veselypeta:
Turns out this is already added to the spec, but does not appear in the generated header since it's not supported by level-zero 
https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/enqueue.yml#L675

kbenzie:
We should discuss why there are disabled entry points with the WG.

kbenzie:
Along with `urEnqueueMemImageFill`, `urEnqueueMemImageMap` is also currently disabled.

veselypeta:
* `piEnqueueMemImageFill` is only implemented in the `OpenCL` plugin
* `piEnqueueMemImageMap` does not exist it PI.

I suppose we do not need to add these entry-points

kbenzie:
If these are necessary to implement SYCL 2020, they should be added to the UR spec.

aarongreig:
These entry points are still in the spec files. I would have already opened the PR to get them deleted *but* strangely MemImageFill does appear to get used in the sycl runtime: https://github.com/intel/llvm/blob/sycl/sycl/source/detail/memory_manager.cpp#L851 and not just in a sort of stubbed out helper function or whatever.

I think what's happening here is that SYCL has this `fill` function that takes an `accessor`, which is a potentially multi-dimensional array view type object that sits on top of an allocation. You can see there's a little bit of extra code where I linked in the "not image" branch that deals with it possibly being a > 1D operation for buffers. It looks like in the "yes image" branch they're taking advantage of the fact that there's this magic API for images that deals with that nonsense for them, it seems more like a use out of convenience than out of necessity.

Technically I think we could implement ImageFill for all the adapters that support images, but for everything that isn't opencl it would basically be a regular fill with the additional faff of keeping track of which mem_obj belongs to which image and figuring out what our pixel size and strides etc are. All that seems kind of pointless when that one use that I linked has that info on hand and leaves it out when it calls the ImageFill entry point instead of the BufferFill one.

Anyway we should probably bring this up to sycl maintainer people because if we delete ImageFill now we should also figure out how we can delete it from PI and memory_manager so it doesn't come back to bite us later when it's time to port.

PR #310, opened 2023-03-01
add function ids to the spec

pbalcer:
XPTI function tracing requires "a stable ABI function id" that
is used to programmatically identify a traced entry point.
This patch adds a requirement for functions to have a unique id,
populates that field for existing functions, and creates an enum
with each function id. It also adds a script that automatically
adds those ids if they are missing.


pbalcer:
This is a bit janky, so I'm open to other ideas. In SYCL they have an autogenerated enum. Right now in #300 I'm just using a loop counter inside of the template. Both of those implicitly depend on the order of functions in the spec.
I also considered hashing, but making the ids explicit seemed like a better idea.

scripts/generate_ids.py
```diff
@@ -0,0 +1,55 @@
...
+        if match:
+            valid_ids.discard(int(match.group(1)))
+
+# go over every function in the spec and add a unique id if it's missing
+for filename in yml_files:
+    with open(filename, 'r+') as f:
+        infunc = False
+        hasid = False
+
+        lines = f.readlines()
```
PatKamin, line 33: Storing the whole file is unnecessary, try having the following loop this way:

`for line in f`

And later when checking for a newline at EOF:
```
f.seek(-1,2) 
if f.read() != '\n'
```

scripts/generate_ids.py
```diff
@@ -0,0 +1,55 @@
...
+        f.seek(0)
+        for line in lines:
+            if line.startswith('---'):
+                if infunc and not hasid:
+                    f.write(generate_id_str())
+                infunc = False
+                hasid = False
+
+            if line.startswith('type: function'):
```
PatKamin, line 42: It's an `elif` actually, as the same line can't start with all of these strings.

pbalcer: done

scripts/generate_ids.py
pbalcer: That won't work because the file is being overwritten. I'd have to buffer the output lines, so it would ultimately come down to the same thing perf- or memory-wise (not that it matters whether this executes in 0.05s or 0.1s... :P).

include/ur_api.h
```diff
@@ -7848,6 +7848,128 @@ typedef struct ur_callbacks_t {
...
+    UR_FUNCTION_ENQUEUE_MEM_BUFFER_READ_RECT = 22,
+    UR_FUNCTION_ENQUEUE_MEM_BUFFER_WRITE_RECT = 23,
+    UR_FUNCTION_ENQUEUE_MEM_BUFFER_COPY = 24,
+    UR_FUNCTION_ENQUEUE_MEM_BUFFER_COPY_RECT = 25,
+    UR_FUNCTION_ENQUEUE_MEM_BUFFER_FILL = 26,
+    UR_FUNCTION_ENQUEUE_MEM_IMAGE_READ = 27,
+    UR_FUNCTION_ENQUEUE_MEM_IMAGE_WRITE = 28,
+    UR_FUNCTION_ENQUEUE_MEM_IMAGE_COPY = 29,
+    UR_FUNCTION_ENQUEUE_MEM_BUFFER_MAP = 31,
+    UR_FUNCTION_ENQUEUE_MEM_UNMAP = 33,
```
PatKamin, line 7884: Missing 30 and 32, isn't this enum autogenerated?

pbalcer: I had to rebase this many times now, which is very fiddly... In general, ids don't have to be monotonically increasing.

pbalcer: Turns out it works fine, entry points with ids 30 and 32 are disabled, see #50.

PatKamin (approved):

kbenzie:
> This is a bit janky, so I'm open to other ideas. In SYCL they have an autogenerated enum. Right now in #300 I'm just using a loop counter inside of the template. Both of those implicitly depend on the order of functions in the spec.
> I also considered hashing, but making the ids explicit seemed like a better idea.

In Khronos specifications they have the concept of a registry, this is used for hardware vendors to reserve ranges of enumeration values for vendor extensions among other things. I wonder if it would make sense to have a `registry.yml`, that would updated by the script and committed to the repo, to keep these unique IDs consistent over time?

pbalcer:
> > This is a bit janky, so I'm open to other ideas. In SYCL they have an autogenerated enum. Right now in #300 I'm just using a loop counter inside of the template. Both of those implicitly depend on the order of functions in the spec.
> > I also considered hashing, but making the ids explicit seemed like a better idea.
> 
> In Khronos specifications they have the concept of a registry, this is used for hardware vendors to reserve ranges of enumeration values for vendor extensions among other things. I wonder if it would make sense to have a `registry.yml`, that would be updated by the script and committed to the repo, to keep these unique IDs consistent over time?

I like that idea. The script would be simpler, and such a registry would let keep the autogenerated id separate from the actual spec. On the other hand, accessing the id variable in templates will be a little bit more involved, but nothing a helper function can't solve. I'll sleep on it.
Thanks!

kbenzie (approved):
Assuming the IDs only need to be consistent in a particular version of the headers and not across versions this looks good.

pbalcer:
It needs to be stable across versions as if it's part of the ABI. XPTI-based tools will target this and shouldn't break when, e.g., a new function is added in the middle of some yml file. That's why this version adds the ids to the spec, and the generate script respects existing ids.

I started reworking this in line with your suggestion for ids to be in a separate file. I'll probably close this PR and open a new one.

pbalcer:
Closing in favor of a different approach in #343

PR #312, opened 2023-03-01
Add test for urEnqueueUSMPrefetch

### LINKED ISSUE #245: Adds CTS tests for `urEnqueueUSMPrefetch` entry-point
We should add conformance tests for `urEnqueueUSMPrefetch`.

Required Coverage:

fabiomestre:
Close #245

### DIFF
--- a/test/conformance/enqueue/urEnqueueUSMPrefetch.cpp
+++ b/test/conformance/enqueue/urEnqueueUSMPrefetch.cpp
@@ -33,14 +33,14 @@ UUR_INSTANTIATE_DEVICE_TEST_SUITE_P(urEnqueueUSMPrefetchTest);
 TEST_P(urEnqueueUSMPrefetchTest, InvalidNullHandleQueue) {
     ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_NULL_HANDLE,
                      urEnqueueUSMPrefetch(nullptr, ptr, sizeof(int),
-                                          UR_MEM_ADVICE_DEFAULT, 0, nullptr,
+                                          UR_USM_MIGRATION_FLAG_DEFAULT, 0, nullptr,
                                           nullptr));
 }
 
 TEST_P(urEnqueueUSMPrefetchTest, InvalidNullPointerMem) {
     ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_NULL_POINTER,
                      urEnqueueUSMPrefetch(queue, nullptr, sizeof(int),
-                                          UR_MEM_ADVICE_DEFAULT, 0, nullptr,
+                                          UR_USM_MIGRATION_FLAG_DEFAULT, 0, nullptr,
                                           nullptr));
 }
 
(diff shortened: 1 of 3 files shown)


test/conformance/enqueue/urEnqueueUSMPrefetch.cpp
```diff
@@ -1,2 +1,82 @@
...
+using urEnqueueUSMPrefetchWithParamTest =
+    uur::urUSMDeviceAllocTestWithParam<ur_usm_migration_flag_t>;
+
+UUR_TEST_SUITE_P(urEnqueueUSMPrefetchWithParamTest,
+                 ::testing::Values(UR_USM_MIGRATION_FLAG_DEFAULT),
+                 uur::deviceTestWithParamPrinter<ur_usm_migration_flag_t>);
+
+TEST_P(urEnqueueUSMPrefetchWithParamTest, Success) {
+    ur_event_handle_t prefetch_event = nullptr;
+    ASSERT_SUCCESS(urEnqueueUSMPrefetch(queue, ptr, sizeof(int),
```
aarongreig, line 15: could we make it so the `urUSMDeviceAllocTestWithParam` fixture has a variable denoting the size of the `ptr` allocation? instead of doing `sizeof(int)` everywhere

veselypeta: Fair point. Maybe we can have an `allocation_size` or `ptr_size` member.

test/conformance/enqueue/urEnqueueUSMPrefetch.cpp
```diff
@@ -1,2 +1,82 @@
...
+                                          nullptr));
+
+    ur_event_handle_t validEvent;
+    ASSERT_SUCCESS(urEnqueueEventsWait(queue, 0, nullptr, &validEvent));
+
+    ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST,
+                     urEnqueueUSMPrefetch(queue, ptr, sizeof(int),
+                                          UR_USM_MIGRATION_FLAG_DEFAULT, 0, &validEvent,
+                                          nullptr));
+}
```
veselypeta, line 82: Probably not required for this PR but maybe we can have a test which uses the `pEventWaitlist`?

veselypeta (approved):
LGTM

test/conformance/enqueue/urEnqueueUSMPrefetch.cpp
fabiomestre: I added an allocation_size variable to the fixture.

test/conformance/enqueue/urEnqueueUSMPrefetch.cpp
fabiomestre: I added a new test for this. I'm not sure if this would be the best way to test this behaviour. Please let me know if you know of a better way.

test/conformance/enqueue/urEnqueueUSMPrefetch.cpp
```diff
@@ -1,2 +1,135 @@
...
+
+TEST_P(urEnqueueUSMPrefetchTest, InvalidEventWaitList) {
+    ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST,
+                     urEnqueueUSMPrefetch(queue, ptr, allocation_size,
+                                          UR_USM_MIGRATION_FLAG_DEFAULT, 1,
+                                          nullptr,
+                                          nullptr));
+
+    ur_event_handle_t validEvent;
+    ASSERT_SUCCESS(urEnqueueEventsWait(queue, 0, nullptr, &validEvent));
```
veselypeta, line 128: Would this be successful? I don't think `validEvent` is a valid handle to an event object.

fabiomestre: validEvent is an output parameter. At least that's my understanding of the enqueue entrypoints. urEnqueueEventsWait will return an event which can then be waited on. 

"phEvent – [in,out][optional] return an event object that identifies this particular command instance."

I think it should be just [out] instead of [in,out]. So perhaps, I can make another PR to update this in the spec.

veselypeta: Ah yes that sounds right.

PR #315, opened 2023-03-03
[uma] memory tracking provider

igchor:
This PR adds a `umaPoolByPtr` function. It's implemented on top of (internal) memoryTracker.

This PR also requires users to pass all memory providers to memory pool through `providers` array parameter in initialize method. This allows UMA to wrap each provider and register all allocations/deallocations.

### DIFF
--- a/source/common/uma_helpers.hpp
+++ b/source/common/uma_helpers.hpp
@@ -14,6 +14,7 @@
 #include <uma/memory_provider.h>
 #include <uma/memory_provider_ops.h>
 
+#include <functional>
 #include <memory>
 #include <stdexcept>
 #include <tuple>
@@ -22,9 +23,11 @@
 namespace uma {
 
 using pool_unique_handle_t =
-    std::unique_ptr<uma_memory_pool_t, decltype(&umaPoolDestroy)>;
+    std::unique_ptr<uma_memory_pool_t,
+                    std::function<void(uma_memory_pool_handle_t)>>;
 using provider_unique_handle_t =
-    std::unique_ptr<uma_memory_provider_t, decltype(&umaMemoryProviderDestroy)>;
+    std::unique_ptr<uma_memory_provider_t,
+                    std::function<void(uma_memory_provider_handle_t)>>;
 
 /// @brief creates UMA memory provider based on given T type.
 /// T should implement all functions defined by
@@ -45,7 +48,7 @@ auto memoryProviderMakeUnique(Args &&...args) {
         auto provider = new T;
         *obj = provider;
         return std::apply(&T::initialize,
-                          std::tuple_cat(std::make_tuple(*provider), *tuple));
+                          std::tuple_cat(std::make_tuple(provider), *tuple));
     };
     ops.finalize = [](void *obj) { delete reinterpret_cast<T *>(obj); };
     ops.alloc = [](void *obj, auto... args) {
@@ -74,24 +77,33 @@ auto memoryProviderMakeUnique(Args &&...args) {
 /// replaced by dtor). All arguments passed to this function are
 /// forwarded to T::initialize(). All functions of T
 /// should be noexcept.
-template <typename T, typename... Args> auto poolMakeUnique(Args &&...args) {
+template <typename T, typename... Args>
+auto poolMakeUnique(uma_memory_provider_handle_t *providers,
+                    size_t numProviders, Args &&...args) {
     uma_memory_pool_ops_t ops;
     auto argsTuple = std::make_tuple(std::forward<Args>(args)...);
-    static_assert(
-        noexcept(std::declval<T>().initialize(std::forward<Args>(args)...)));
+    static_assert(noexcept(std::declval<T>().initialize(
+        providers, numProviders, std::forward<Args>(args)...)));
 
     ops.version = UMA_VERSION_CURRENT;
-    ops.initialize = [](void *params, void **obj) {
+    ops.initialize = [](uma_memory_provider_handle_t *providers,
+                        size_t numProviders, void *params, void **obj) {
         auto *tuple = reinterpret_cast<decltype(argsTuple) *>(params);
         auto pool = new T;
         *obj = pool;
-        return std::apply(&T::initialize,
-                          std::tuple_cat(std::make_tuple(*pool), *tuple));
+        return std::apply(
+            &T::initialize,
+            std::tuple_cat(std::make_tuple(pool, providers, numProviders),
+                           *tuple));
     };
     ops.finalize = [](void *obj) { delete reinterpret_cast<T *>(obj); };
     ops.malloc = [](void *obj, auto... args) {
         static_assert(noexcept(reinterpret_cast<T *>(obj)->malloc(args...)));
         return reinterpret_cast<T *>(obj)->malloc(args...);
+    };
+    ops.calloc = [](void *obj, auto... args) {
+        static_assert(noexcept(reinterpret_cast<T *>(obj)->calloc(args...)));
+        return reinterpret_cast<T *>(obj)->calloc(args...);
     };
     ops.aligned_malloc = [](void *obj, auto... args) {
         static_assert(
@@ -118,7 +130,7 @@ template <typename T, typename... Args> auto poolMakeUnique(Args &&...args) {
     };
 
     uma_memory_pool_handle_t hPool = nullptr;
-    auto ret = umaPoolCreate(&ops, &argsTuple, &hPool);
+    auto ret = umaPoolCreate(&ops, providers, numProviders, &argsTuple, &hPool);
     return std::pair<uma_result_t, pool_unique_handle_t>{
         ret, pool_unique_handle_t(hPool, &umaPoolDestroy)};
 }
--- a/source/common/unified_memory_allocation/include/uma/memory_pool.h
+++ b/source/common/unified_memory_allocation/include/uma/memory_pool.h
@@ -10,7 +10,7 @@
 #define UMA_MEMORY_POOL_H 1
 
 #include <uma/base.h>
-#include <uma/memory_pool_ops.h>
+#include <uma/memory_provider.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -18,13 +18,19 @@ extern "C" {
 
 typedef struct uma_memory_pool_t *uma_memory_pool_handle_t;
 
+struct uma_memory_pool_ops_t;
+
 ///
 /// \brief Creates new memory pool
 /// \param ops instance of uma_memory_pool_ops_t
+/// \param providers array of memory providers that will be used for coarse-grain allocations. Should contain at least one memory provider.
+/// \param numProvider number of elements in the providers array
 /// \param params pointer to pool-specific parameters
 /// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure
 ///
-enum uma_result_t umaPoolCreate(struct uma_memory_pool_ops_t *ops, void *params,
+enum uma_result_t umaPoolCreate(struct uma_memory_pool_ops_t *ops,
+                                uma_memory_provider_handle_t *providers,
+                                size_t numProviders, void *params,
                                 uma_memory_pool_handle_t *hPool);
 
 ///
@@ -88,6 +94,12 @@ size_t umaPoolMallocUsableSize(uma_memory_pool_handle_t hPool, void *ptr);
 void umaPoolFree(uma_memory_pool_handle_t hPool, void *ptr);
 
 ///
+/// \brief Frees the memory space pointed by ptr if it belongs to UMA pool, does nothing otherwise
+/// \param ptr pointer to the allocated memory
+///
+void umaFree(void *ptr);
+
+///
 /// \brief Retrieve string representation of the underlying pool specific
 ///        result reported by the last API that returned
 ///        UMA_RESULT_ERROR_POOL_SPECIFIC or NULL ptr (in case of allocation
@@ -110,6 +122,12 @@ void umaPoolFree(uma_memory_pool_handle_t hPool, void *ptr);
 ///         returned indicates that the adapter specific result is an error.
 enum uma_result_t umaPoolGetLastResult(uma_memory_pool_handle_t hPool,
                                        const char **ppMessage);
+
+///
+/// \brief Retrieve memory pool associated with a given ptr.
+/// \param ptr pointer to memory belonging to a memory pool
+/// \return handle to a memory pool that contains ptr or NULL if pointer does not belong to any UMA pool
+uma_memory_pool_handle_t umaPoolByPtr(const void *ptr);
 
 #ifdef __cplusplus
 }
--- a/source/common/unified_memory_allocation/include/uma/memory_provider.h
+++ b/source/common/unified_memory_allocation/include/uma/memory_provider.h
@@ -41,7 +41,7 @@ void umaMemoryProviderDestroy(uma_memory_provider_handle_t hProvider);
 /// \param hProvider handle to the memory provider
 /// \param size number of bytes to allocate
 /// \param alignment alignment of the allocation
-/// \param ptr returns pointer to the allocated memory
+/// \param ptr will be updated with pointer to the allocated memory
 /// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure
 ///
 enum uma_result_t umaMemoryProviderAlloc(uma_memory_provider_handle_t hProvider,
--- a/source/common/unified_memory_allocation/src/memory_pool.c
+++ b/source/common/unified_memory_allocation/src/memory_pool.c
@@ -6,18 +6,43 @@
  *
  */
 
+#include "memory_tracker.h"
 #include <uma/memory_pool.h>
+#include <uma/memory_pool_ops.h>
 
 #include <assert.h>
 #include <stdlib.h>
 
 struct uma_memory_pool_t {
-    struct uma_memory_pool_ops_t ops;
     void *pool_priv;
+    struct uma_memory_pool_ops_t ops;
+
+    // Holds array of memory providers. All providers are wrapped
+    // by memory tracking providers (owned and released by UMA).
+    uma_memory_provider_handle_t *providers;
+
+    size_t numProviders;
 };
 
-enum uma_result_t umaPoolCreate(struct uma_memory_pool_ops_t *ops, void *params,
+static void
+destroyMemoryProviderWrappers(uma_memory_provider_handle_t *providers,
+                              size_t numProviders) {
+    for (size_t i = 0; i < numProviders; i++) {
+        umaMemoryProviderDestroy(providers[i]);
+    }
+
+    free(providers);
+}
+
+enum uma_result_t umaPoolCreate(struct uma_memory_pool_ops_t *ops,
+                                uma_memory_provider_handle_t *providers,
+                                size_t numProviders, void *params,
                                 uma_memory_pool_handle_t *hPool) {
+    if (!numProviders || !providers) {
+        return UMA_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    enum uma_result_t ret = UMA_RESULT_SUCCESS;
     uma_memory_pool_handle_t pool = malloc(sizeof(struct uma_memory_pool_t));
     if (!pool) {
         return UMA_RESULT_ERROR_OUT_OF_HOST_MEMORY;
@@ -25,24 +50,47 @@ enum uma_result_t umaPoolCreate(struct uma_memory_pool_ops_t *ops, void *params,
 
     assert(ops->version == UMA_VERSION_CURRENT);
 
-    pool->ops = *ops;
-
-    void *pool_priv;
-    enum uma_result_t ret = ops->initialize(params, &pool_priv);
-    if (ret != UMA_RESULT_SUCCESS) {
-        free(pool);
-        return ret;
+    pool->providers =
+        calloc(numProviders, sizeof(uma_memory_provider_handle_t));
+    if (!pool->providers) {
+        ret = UMA_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+        goto err_providers_alloc;
     }
 
-    pool->pool_priv = pool_priv;
+    size_t providerInd = 0;
+    pool->numProviders = numProviders;
+
+    // Wrap each provider with memory tracking provider.
+    for (providerInd = 0; providerInd < numProviders; providerInd++) {
+        ret = umaTrackingMemoryProviderCreate(providers[providerInd], pool,
+                                              &pool->providers[providerInd]);
+        if (ret != UMA_RESULT_SUCCESS) {
+            goto err_providers_init;
+        }
+    }
+
+    pool->ops = *ops;
+    ret = ops->initialize(pool->providers, pool->numProviders, params,
+                          &pool->pool_priv);
+    if (ret != UMA_RESULT_SUCCESS) {
+        goto err_pool_init;
+    }
 
     *hPool = pool;
-
     return UMA_RESULT_SUCCESS;
+
+err_pool_init:
+err_providers_init:
+    destroyMemoryProviderWrappers(pool->providers, providerInd);
+err_providers_alloc:
+    free(pool);
+
+    return ret;
 }
 
 void umaPoolDestroy(uma_memory_pool_handle_t hPool) {
     hPool->ops.finalize(hPool->pool_priv);
+    destroyMemoryProviderWrappers(hPool->providers, hPool->numProviders);
     free(hPool);
 }
 
@@ -71,7 +119,18 @@ void umaPoolFree(uma_memory_pool_handle_t hPool, void *ptr) {
     hPool->ops.free(hPool->pool_priv, ptr);
 }
 
+void umaFree(void *ptr) {
+    uma_memory_pool_handle_t hPool = umaPoolByPtr(ptr);
+    if (hPool) {
+        umaPoolFree(hPool, ptr);
+    }
+}
+
 enum uma_result_t umaPoolGetLastResult(uma_memory_pool_handle_t hPool,
                                        const char **ppMessage) {
     return hPool->ops.get_last_result(hPool->pool_priv, ppMessage);
+}
+
+uma_memory_pool_handle_t umaPoolByPtr(const void *ptr) {
+    return umaMemoryTrackerGetPool(umaMemoryTrackerGet(), ptr);
 }
--- a/source/common/unified_memory_allocation/src/memory_tracker.cpp
+++ b/source/common/unified_memory_allocation/src/memory_tracker.cpp
@@ -1,0 +1,188 @@
+/*
+ *
+ * Copyright (C) 2023 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#include "memory_tracker.h"
+#include <uma/memory_provider.h>
+#include <uma/memory_provider_ops.h>
+
+#include <map>
+#include <mutex>
+#include <shared_mutex>
+
+// TODO: reimplement in C and optimize...
+struct uma_memory_tracker_t {
+    enum uma_result_t add(void *pool, const void *ptr, size_t size) {
+        std::unique_lock<std::shared_mutex> lock(mtx);
+
+        if (size == 0) {
+            return UMA_RESULT_SUCCESS;
+        }
+
+        auto ret =
+            map.try_emplace(reinterpret_cast<uintptr_t>(ptr), size, pool);
+        return ret.second ? UMA_RESULT_SUCCESS : UMA_RESULT_ERROR_UNKNOWN;
+    }
+
+    enum uma_result_t remove(const void *ptr, size_t size) {
+        std::unique_lock<std::shared_mutex> lock(mtx);
+
+        map.erase(reinterpret_cast<uintptr_t>(ptr));
+
+        // TODO: handle removing part of the range
+        (void)size;
+
+        return UMA_RESULT_SUCCESS;
+    }
+
+    void *find(const void *ptr) {
+        std::shared_lock<std::shared_mutex> lock(mtx);
+
+        auto intptr = reinterpret_cast<uintptr_t>(ptr);
+        auto it = map.upper_bound(intptr);
+        if (it == map.begin()) {
+            return nullptr;
+        }
+
+        --it;
+
+        auto address = it->first;
+        auto size = it->second.first;
+        auto pool = it->second.second;
+
+        if (intptr >= address && intptr < address + size) {
+            return pool;
+        }
+
+        return nullptr;
+    }
+
+  private:
+    std::shared_mutex mtx;
+    std::map<uintptr_t, std::pair<size_t, void *>> map;
+};
+
+static enum uma_result_t
+umaMemoryTrackerAdd(uma_memory_tracker_handle_t hTracker, void *pool,
+                    const void *ptr, size_t size) {
+    return hTracker->add(pool, ptr, size);
+}
+
+static enum uma_result_t
+umaMemoryTrackerRemove(uma_memory_tracker_handle_t hTracker, const void *ptr,
+                       size_t size) {
+    return hTracker->remove(ptr, size);
+}
+
+extern "C" {
+
+uma_memory_tracker_handle_t umaMemoryTrackerGet() {
+    static uma_memory_tracker_t tracker;
+    return &tracker;
+}
+
+void *umaMemoryTrackerGetPool(uma_memory_tracker_handle_t hTracker,
+                              const void *ptr) {
+    return hTracker->find(ptr);
+}
+
+struct uma_tracking_memory_provider_t {
+    uma_memory_provider_handle_t hUpstream;
+    uma_memory_tracker_handle_t hTracker;
+    uma_memory_pool_handle_t pool;
+};
+
+typedef struct uma_tracking_memory_provider_t uma_tracking_memory_provider_t;
+
+static enum uma_result_t trackingAlloc(void *hProvider, size_t size,
+                                       size_t alignment, void **ptr) {
+    uma_tracking_memory_provider_t *p =
+        (uma_tracking_memory_provider_t *)hProvider;
+    enum uma_result_t ret = UMA_RESULT_SUCCESS;
+
+    ret = umaMemoryProviderAlloc(p->hUpstream, size, alignment, ptr);
+    if (ret != UMA_RESULT_SUCCESS) {
+        return ret;
+    }
+
+    ret = umaMemoryTrackerAdd(p->hTracker, p->pool, *ptr, size);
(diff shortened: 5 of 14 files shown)


source/common/unified_memory_allocation/src/memory_tracker.cpp
```diff
@@ -0,0 +1,85 @@
...
+        map.erase(reinterpret_cast<uintptr_t>(ptr));
+
+        // TODO: handle removing part of the range
+        (void)size;
+
+        return UMA_RESULT_SUCCESS;
+    }
+
+    void *find(const void *ptr) {
+        std::shared_lock<std::shared_mutex> lock(mtx);
```
igchor, line 36: I actually just realized std::map might not be threadsafe (even for upper_bound) so I'll need to reimplement this probably.

igchor:
Open question: do we want to define a more rigid interface between pools and providers? For example, require memory pool to accept a (list of) provider(s)? This way we could make sure that each allocation is tracked by automatically wrapping each provider by `memoryTracking` provider - we wouldn't even need to expose the `memoryTracking` provider to the user (although then user loses ability to exchange the memory tracking implementation)

source/common/uma_helpers.hpp
```diff
@@ -0,0 +1,142 @@
...
+    return createUmaObject<MemProvider>(ops, &umaMemoryProviderCreate, &umaMemoryProviderDestroy, args...);
+}
+
+/// @brief creates UMA memory pool based on given MemPool type.
+/// MemPool should implement all functions defined by
+/// uma_memory_pool_ops_t, except for intialize and finalize (those are
+/// replaced by ctor and dtor). All arguments passed to this function are
+/// forwarded to MemProvider ctor.
+template <typename MemPool, typename... Args>
+auto poolMakeUnique(Args &&...args) {
```
pbalcer, line 128: Instead of doing all this, wouldn't it be simpler and more type safe to have some abstract class that implements a conversion operator to uma types?

source/common/unified_memory_allocation/src/memory_provider.c
```diff
@@ -28,37 +29,112 @@ umaMemoryProviderCreate(struct uma_memory_provider_ops_t *ops, void *params,
...
+    *ret = provider;
+    return UMA_RESULT_SUCCESS;
+}
+
+static void trackingFinalize(void *provider) {
+    free(provider);
+}
+
+static struct uma_memory_provider_ops_t umaTrackingMemoryProviderOps() {
+    struct uma_memory_provider_ops_t ops = {
```
pbalcer, line 118: I don't think you need to wrap this in a function. Just create a static variable.

source/common/unified_memory_allocation/src/memory_provider.c
```diff
@@ -28,37 +29,112 @@ umaMemoryProviderCreate(struct uma_memory_provider_ops_t *ops, void *params,
...
+    ret = umaMemoryTrackerRemove(p->hTracker, ptr, size);
+    if (ret != UMA_RESULT_SUCCESS) {
+        // TODO
+    }
+
+    return ret;
+}
+
+static enum uma_result_t trackingInitialize(void *params, void **ret) {
+    uma_tracking_memory_provider_t *provider = (uma_tracking_memory_provider_t *)malloc(sizeof(uma_tracking_memory_provider_t));
```
pbalcer, line 107: `if (provider == NULL) ...`

source/common/unified_memory_allocation/src/memory_tracker.cpp
pbalcer: What do you mean? None of this is thread-safe since you are not using a const map. That's why you needed a lock.
And, even if you had used a lock-free map, you'd need a single safe find 'less equals' method.

source/common/unified_memory_allocation/src/memory_tracker.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  private:
+    std::shared_mutex mtx;
+    std::map<uintptr_t, std::pair<size_t, void *>> map;
+};
+
+extern "C" {
+
+uma_memory_tracker_handle_t umaMemoryTrackerGet() {
+    static uma_memory_tracker_t tracker;
```
pbalcer, line 60: I'm not sure if I like this being a global.

pbalcer:
> Open question: do we want to define a more rigid interface between pools and providers? For example, require memory pool to accept a (list of) provider(s)? This way we could make sure that each allocation is tracked by automatically wrapping each provider by `memoryTracking` provider - we wouldn't even need to expose the `memoryTracking` provider to the user (although then user loses ability to exchange the memory tracking implementation)

I don't know if the ability to change the tracking provider is that useful. There are not that many possible implementations, and there's no reason why we can't provide a good one (i.e., a lock-free one for reads). I'm not sure if there's any trade-off - maybe if an allocator allocates from the provider very frequently it might make sense to optimize the write path, making reads slower. But that's a stretch.

As for accepting a list of providers in the pools - I'm not sure. We'd be once again removing type safety in favor of dubious "simplicity". Maybe it could be better if we had some way of pools to select providers for particular roles like metadata/data split. Something like:
`providers = { {.base = provider_dram(), .type = PROVIDER_METADATA}, {.base = provider_hbm(), .type = PROVIDER_DATA} };`
This way, users wouldn't have to know that `providers[0]` has to be this and `provider[1]` has to be that.

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -75,6 +75,16 @@ enum uma_result_t umaMemoryProviderFree(uma_memory_provider_handle_t hProvider,
...
 
+///
+/// \brief Create a proxy memory provider that associates each allocation with a pool
+///        and forwards calls to the upstream provider. By calling umaPoolByPtr()
+///        it's possible to obtain pool associated with the pointer.
+/// \param hUpstream all calls are forwarded to this provider
+/// \param pool pool to be associated with all allocations
+/// \param [out] returns instance of tracking provider
+enum uma_result_t umaTrackingMemoryProviderCreate(uma_memory_provider_handle_t hUpstream,
+                                                  void *pool, uma_memory_provider_handle_t *hTrackingProvider);
```
vinser52, line 86: Why `void*` instead of `uma_memory_pool_handle_t`?

source/common/unified_memory_allocation/src/memory_pool.c
```diff
@@ -6,14 +6,15 @@
...
  */
 
+#include "memory_tracker_internal.h"
 #include <uma/memory_pool.h>
 
 #include <assert.h>
 #include <stdlib.h>
 
 struct uma_memory_pool_t {
+    void *priv; // so that &pool->priv == pool
```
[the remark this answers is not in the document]
vinser52: You just moved `void*` field to the beginning of the `uma_memory_pool_t` structure. I did not get what is confusing you.

source/common/unified_memory_allocation/src/memory_provider.c
```diff
@@ -6,14 +6,15 @@
...
  */
 
+#include "memory_tracker_internal.h"
 #include <uma/memory_provider.h>
 
 #include <assert.h>
 #include <stdlib.h>
 
 struct uma_memory_provider_t {
+    void *priv;
```
vinser52, line 16: It looks like a simple refactoring, isn't it? If so, I think it is better to do that in a separate PR.

source/common/unified_memory_allocation/src/memory_tracker.cpp
igchor: I wasn't sure if `const` version of `upper_bound` is thread safe but it turns out it is - so this code should be safe as long as we make sure we use the const version.

Eventually, we'll need to reimplement this anyway with something faster (ideally with lock-free reads), for example with https://github.com/kilobyte/critnib

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -75,6 +75,16 @@ enum uma_result_t umaMemoryProviderFree(uma_memory_provider_handle_t hProvider,
...
 enum uma_result_t umaMemoryProviderGetLastResult(uma_memory_provider_handle_t hProvider, const char **ppMessage);
 
+///
+/// \brief Create a proxy memory provider that associates each allocation with a pool
+///        and forwards calls to the upstream provider. By calling umaPoolByPtr()
+///        it's possible to obtain pool associated with the pointer.
+/// \param hUpstream all calls are forwarded to this provider
+/// \param pool pool to be associated with all allocations
+/// \param [out] returns instance of tracking provider
+enum uma_result_t umaTrackingMemoryProviderCreate(uma_memory_provider_handle_t hUpstream,
```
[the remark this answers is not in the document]
igchor: But how will the memory provider know which pool the memory belongs to? This function is needed to pass that `pool` handle/pointer to the provider. Another approach is to add 'providers' param to pool::initialize and force users to only pass memory providers through this param - then we can automatically wrap those providers in umaPoolCreate.

source/common/uma_helpers.hpp
igchor: But you still need a way to create a uma handle from that object (call umaCreatePool/Provider).

igchor:
> > Open question: do we want to define a more rigid interface between pools and providers? For example, require memory pool to accept a (list of) provider(s)? This way we could make sure that each allocation is tracked by automatically wrapping each provider by `memoryTracking` provider - we wouldn't even need to expose the `memoryTracking` provider to the user (although then user loses ability to exchange the memory tracking implementation)
> 
> I don't know if the ability to change the tracking provider is that useful. There are not that many possible implementations, and there's no reason why we can't provide a good one (i.e., a lock-free one for reads). I'm not sure if there's any trade-off - maybe if an allocator allocates from the provider very frequently it might make sense to optimize the write path, making reads slower. But that's a stretch.
> 
> As for accepting a list of providers in the pools - I'm not sure. We'd be once again removing type safety in favor of dubious "simplicity". Maybe it could be better if we had some way of pools to select providers for particular roles like metadata/data split. Something like: `providers = { {.base = provider_dram(), .type = PROVIDER_METADATA}, {.base = provider_hbm(), .type = PROVIDER_DATA} };` This way, users wouldn't have to know that `providers[0]` has to be this and `provider[1]` has to be that.

Yes, regarding the roles of providers - I think it makes sense. However, I was thinking about just exposing those types/properties through some kind of function (umaMemoryProviderGetProperites()) or something like this. Then, the pool can just specify list of requirements.

Also, there is a case where pool might not need any provider (e.g. malloc-based pool). For this case, I think we can still pass `NULL` provider (that does nothing but we still track calls to that provider).

source/common/uma_helpers.hpp
igchor: I've removed this file (and tests) from this PR for now. I'll add them once we agree on API and https://github.com/oneapi-src/unified-runtime/pull/322 is merged

source/common/unified_memory_allocation/src/memory_tracker.cpp
igchor: Another option is to expose trackerCreate method to the user, but then the handle needs to be passed trackingMemoryProviderCreate or to poolCreate (if we decide to wrap providers ourselves) and it will mean that different applications using UMA will not share the tracker - this is problematic for cases like MPI.

source/common/unified_memory_allocation/include/uma/memory_pool.h
```diff
@@ -108,6 +108,13 @@ void umaPoolFree(uma_memory_pool_handle_t hPool, void *ptr);
 ///         returned indicates that the adapter specific result is an error.
 enum uma_result_t umaPoolGetLastResult(uma_memory_pool_handle_t hPool, const char **ppMessage);
 
+///
+/// \brief returns memory pool associated with a given ptr. Memory pointed to by ptr
+///        must have been tracked using provider obtained by calling umaTrackingMemoryProviderCreate
+/// \param ptr pointer to memory belonging to memory pool
+/// \return handle to memory pool that contains ptr
+uma_memory_pool_handle_t umaPoolByPtr(const void *ptr);
```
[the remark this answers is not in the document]
igchor: Yes, I think just returning NULL is fine in that case, I just forgot to write that in the comment. Done.

source/common/unified_memory_allocation/include/uma/memory_provider.h
igchor: I've changed it to `uma_memory_pool_handle_t` but I'm not sure about this one. In current design this pointer is what user gets within `pool::initialize` function as `void** pool`. If we make it `uma_memory_pool_handle_t` the user needs to do casting.

source/common/unified_memory_allocation/src/memory_pool.c
```diff
@@ -26,51 +27,51 @@ enum uma_result_t umaPoolCreate(struct uma_memory_pool_ops_t *ops, void *params,
     assert(ops->version == UMA_VERSION_CURRENT);
 
     pool->ops = *ops;
-
-    void *pool_priv;
-    enum uma_result_t ret = ops->initialize(params, &pool_priv);
+    enum uma_result_t ret = ops->initialize(params, &pool->priv);
```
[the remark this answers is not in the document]
igchor: I've reverted the unrelated changes - I've included them by mistake. I've created separate PR for the leak.

source/common/unified_memory_allocation/src/memory_pool.c
```diff
@@ -6,42 +6,61 @@
...
-    pool->ops = *ops;
+    pool->providers = calloc(numProviders, sizeof(uma_memory_provider_handle_t));
+    if (!pool->providers) {
+      free(pool);
+      return UMA_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+    }
+  
+    pool->numProviders = numProviders;
+    for (size_t i = 0; i < numProviders; i++) {
+      enum uma_result_t ret = umaTrackingMemoryProviderCreate(providers[i], (uma_memory_pool_handle_t)pool, &pool->providers[i]);
```
vinser52, line 39: I have another concern: technically the same provider instance might be used by multiple pools. At least our API does not prohibit this.
Strictly speaking, to determine the pool by ptr we need to map ptr to heap manager that requested allocation. But for the MPI use case we need to determine the properties of the memory - it means that we need to map ptr to the memory provider. Because the memory provider defines the properties of the memory.

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -39,7 +39,8 @@ void umaMemoryProviderDestroy(uma_memory_provider_handle_t hProvider);
 /// \param hProvider handle to the memory provider
 /// \param size number of bytes to allocate
 /// \param alignment alignment of the allocation
-/// \param ptr returns pointer to the allocated memory
+/// \param ptr[in][out] If memory provider is of type UMA_MEMORY_PROVIDER_TYPE_TRACKER, value of *ptr will be used for tracking purposes.
```
igchor, line 42: This is to support provider-less pools (like malloc-based ones).

My idea is to only create provider of type UMA_MEMORY_PROVIDER_TYPE_TRACKER inside umaPoolCreate if the list of providers is empty (and forbid user from passing tracker with this type) so we can be sure that UMA_MEMORY_PROVIDER_TYPE_TRACKER  provider will handle this properly but still, this feels like a hack. Maybe you have some better ideas?

source/common/unified_memory_allocation/include/uma/memory_pool.h
```diff
@@ -10,21 +10,41 @@
...
+/// \brief Descriptor of a memory provider. Consists of handle to the provider and it's type.
+struct uma_memory_provider_desc_t {
+    uma_memory_provider_handle_t hProvider;
+    enum uma_memory_provider_type providerType;
+};
+
 ///
 /// \brief Creates new memory pool
 /// \param ops instance of uma_memory_pool_ops_t
+/// \param providers array of memory provider descriptors, if numProviders is 0, UMA will create a single memory provider
```
igchor, line 41: I though about requiring numProviders to be > 0 so that number of providers here is the same as  the number passed to pool::initialize but I think this approach is less verbose, you don't need to create a dummy provider if it's not needed.

igchor:
Tests coming after https://github.com/oneapi-src/unified-runtime/pull/322 is merged

source/common/unified_memory_allocation/include/uma/memory_provider.h
igchor: We could just expose memoryTrackerAdd/Remove functions - it's more explicit but it might be harder to 'enforce' tracking the memory. With `UMA_MEMORY_PROVIDER_TYPE_TRACKER ` the pool at least gets an instance that is supposed to be used for tracking.

vinser52: From the user perspective, I think UMA itself should be responsible for memory tracking (not the user). And we should try to hide tracking functionality inside UMA and do not expose anything related to memory tracking.

source/common/unified_memory_allocation/src/memory_pool.c
```diff
@@ -6,43 +6,101 @@
...
-    void *pool_priv;
-    enum uma_result_t ret = ops->initialize(params, &pool_priv);
-    if (ret != UMA_RESULT_SUCCESS) {
-        free(pool);
-        return ret;
+    size_t providerInd = 0;
+    if (!numProviders) {
+        pool->numProviders = 1;
+        pool->providers[0].providerType = UMA_MEMORY_PROVIDER_TYPE_TRACKER;
+        ret = umaTrackingMemoryProviderCreate(NULL, pool, &pool->providers[0].hProvider);
```
vinser52, line 60: Maybe for consistency, we should always require at least one memory provider for data?
What is the use case when the user does not provide a memory provider?

source/common/unified_memory_allocation/src/memory_pool.c
```diff
@@ -6,43 +6,101 @@
...
+    if (!providers && numProviders) {
+        return UMA_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    for (size_t i = 0; i < numProviders; i++) {
+        // UMA_MEMORY_PROVIDER_TYPE_TRACKER can only be created by the UMA library
+        if (providers[i].providerType == UMA_MEMORY_PROVIDER_TYPE_TRACKER) {
+            return UMA_RESULT_ERROR_INVALID_ARGUMENT;
+        }
+    }
```
vinser52, line 40: if the `UMA_MEMORY_PROVIDER_TYPE_TRACKER` type were not exposed to the user we would not need this check.

source/common/unified_memory_allocation/src/memory_pool.c
```diff
@@ -6,43 +6,101 @@
...
 
     *hPool = pool;
-
     return UMA_RESULT_SUCCESS;
+
+err_pool_init:
+err_providers_init:
+    for (size_t i = 0; i < providerInd; i++) {
+        umaMemoryProviderDestroy(pool->providers[i].hProvider);
```
vinser52, line 87: I think we should not destroy memory providers that were not created by us.
If it is created and passed by the user as a parameters to `umaPoolCreate` function than user is responsible to destroy them.

source/common/unified_memory_allocation/src/memory_pool.c
```diff
@@ -6,43 +6,101 @@
...
+    free(pool);
+
+    return ret;
 }
 
 void umaPoolDestroy(uma_memory_pool_handle_t hPool) {
     hPool->ops.finalize(hPool->pool_priv);
+
+    for (size_t i = 0; i < hPool->numProviders; i++) {
+        umaMemoryProviderDestroy(hPool->providers[i].hProvider);
```
vinser52, line 100: the same is here

vinser52 (changes requested):

vinser52:
A general comment from my side.
I am worried that we have not included examples (with TBB & jemalloc heap managers and NUMA bindings) from our initial prototype. As a result, we are not checking if the changes we are making do not break these important examples.

source/common/unified_memory_allocation/include/uma/memory_provider.h
igchor: Right, this is just a special case for when pool does not use any memory provider for allocations. I agree it would be much easier if we just require having at least one memory provider for data.

source/common/unified_memory_allocation/include/uma/memory_pool.h
```diff
@@ -10,21 +10,42 @@
...
 typedef struct uma_memory_pool_t *uma_memory_pool_handle_t;
 
+struct uma_memory_pool_ops_t;
+
+/// \brief Type of a memory provider, defines purpose of the provider.
+enum uma_memory_provider_type {
+    UMA_MEMORY_PROVIDER_TYPE_METADATA,          ///< Provider should be used to allocate metadata only
+    UMA_MEMORY_PROVIDER_TYPE_DATA,              ///< Provider should be used to allocate data only
+    UMA_MEMORY_PROVIDER_TYPE_DATA_AND_METADATA, ///< Provider should be used to allocate both data and metadata
+    UMA_MEMORY_PROVIDER_TYPE_TRACKER            ///< Provider should not be used to request memory, it's only purpose is to track memory allocations.
```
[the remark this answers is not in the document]
igchor: Again, it's mostly to support the case when pool does not use any memory provider for allocations.

> More generic question: do we need a tracking memory provider at all? Perhaps we can extend the umaMemoryProviderAlloc() function to track allocations?

We would need to extend `umaMemoryProviderAlloc` with a `pool` parameter so that the allocation can be associated with the correct pool. The approach with wrapping the memory provider is pretty generic and the only `ugly` interface is to support 'no memory provider' use case.

source/common/unified_memory_allocation/src/memory_pool.c
igchor: Yes, that's true. So we probably need to decide whether 'no memory provider' use case is something we want to support.

source/common/unified_memory_allocation/src/memory_pool.c
igchor: That would certainly simplify things.

The use case I had in mind is a malloc-based provider (you cannot really implement things like realloc with if you on memory provider level. More generically, the usecase is "any memory pool that forwards calls to some exiting allocator interface that support malloc/realloc/calloc-like API".

source/common/unified_memory_allocation/src/memory_pool.c
igchor: Yes, and we are not destroying the user-provider providers - just the wrappers (trackingMemoryProviders) that were created by UMA.

source/common/unified_memory_allocation/src/memory_pool.c
igchor: same as above.

igchor:
> A general comment from my side. I am worried that we have not included examples (with TBB & jemalloc heap managers and NUMA bindings) from our initial prototype. As a result, we are not checking if the changes we are making do not break these important examples.

There are no real fundamental changes here but we can definitely add jemalloc/tbb examples (or test) in a next PR.

source/common/unified_memory_allocation/include/uma/memory_pool.h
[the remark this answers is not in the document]
vinser52: Thinking more about returning NULL... The return type is `uma_memory_pool_handle_t` not `void*`. So that we should define a "null" or "empty" handle.

source/common/unified_memory_allocation/src/memory_pool.c
vinser52: Ahh, got it. Could you please add a comment?

source/common/unified_memory_allocation/src/memory_pool.c
vinser52: Probably, we can wrap this into a separate function.

source/common/unified_memory_allocation/include/uma/memory_pool.h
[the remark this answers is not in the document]
igchor: We can do that but we define `uma_memory_pool_handle_t` to be a pointer to structure in the public header anyway so I don't see anything wrong with NULL

source/common/unified_memory_allocation/src/memory_pool.c
igchor: Sure, done.

source/common/unified_memory_allocation/src/memory_pool.c
igchor: Done. I've also added description to umaTrackingMemoryProviderCreate

source/common/unified_memory_allocation/include/uma/memory_pool.h
```diff
@@ -10,21 +10,41 @@
...
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 typedef struct uma_memory_pool_t *uma_memory_pool_handle_t;
 
+struct uma_memory_pool_ops_t;
+
+/// \brief Type of a memory provider, defines purpose of the provider.
+enum uma_memory_provider_type {
```
pbalcer, line 24: bit flags?

pbalcer (commented):
Do we want to expose this functionality through UR? If so, what would be the API?

`ur_result_t urPlatformGetPoolByPtr(ur_platform_handle_t hPlatform, void *ptr, ur_usm_pool_handle_t **ppPool);`
or
`ur_result_t urGetPoolByPtr(void *ptr, ur_usm_pool_handle_t **ppPool);`

I think we can all agree that the latter would be best, but it would also be difficult to implement in the loader (unless we want to do a for loop over platforms).

We need to consider in this API how to make the loader the registrar (or a "provider") for pool tracking information. A callback for tracking pages? That would make this whole thing external to UMA.

source/common/unified_memory_allocation/include/uma/memory_pool.h
igchor: Or perhaps we should just have METADATA and DATA types and let user pass the same provider two times (with both DATA and METADATA) if he wants. 

Right now, nothing prevents user from doing that anyway, and we basically have two ways of marking a provider as data and metadata. Also, if in such case, we can just wrap the data provider for memory tracking and ignore the metadata one.

igchor:
> Do we want to expose this functionality through UR? If so, what would be the API?
> 
> `ur_result_t urPlatformGetPoolByPtr(ur_platform_handle_t hPlatform, void *ptr, ur_usm_pool_handle_t *ppPool);` or `ur_result_t urGetPoolByPtr(void *ptr, ur_usm_pool_handle_t *ppPool);`
> 
> I think we can all agree that the latter would be best, but it would also be difficult to implement in the loader (unless we want to do a for loop over platforms).
> 
> We need to consider in this API how to make the loader the registrar (or a "provider") for pool tracking information. A callback for tracking pages? That would make this whole thing external to UMA.

Right now we have the possibility to query a pointer within a given context through urUSMGetMemAllocInfo to get 'UR pool handle' (that is different from uma pool handle): https://github.com/oneapi-src/unified-runtime/pull/348

I assume this functionality would leverage the UMA memory tracking mechanism. We can store a reference to UR USM pool within the UMA pool (we could even implement a generic proxy pool for storing this info). Then we can just do something like this:
``` urUSMGetMemAllocInfo (ptr, ...) { return (ur_usm_pool_handle_t*) umaGetParams(umaPoolByPtr(ptr)); } ```

I'm not yet sure what `umaGetParams` should look like but at the very least it can just return parameters passed during pool creation.

As for the second option (where we don't pass adapter/context handle to UR function): can we even implement this with the current UR architecture? We don't have any shared state between adapters.

Ideally, we could just use a single shared instance of UMA (then allocations are tracked across all adapters). But that would require having a shared state and also some interface between the loader and the adapters (loader would need to know either about what `umaGetParams` returns and how to interpret it or it would need to talk with some UR pool manager that is used by adapters).

File: source/common/unified_memory_allocation/include/uma/memory_pool.h (hunk 20 above)
Comment (igchor) [reply source/common/unified_memory_allocation/include/uma/memory_pool.h:24]: I've removed the DATA_AND_METADATA type.

**Comment by pbalcer** [comment]
Yes, it shouldn't be a problem. We'd just need some callbacks in UR:
```
ur_platform_usm_register_on_page_alloc_cb(..., void (*cb)(void *ptr, size_t size, void *arg), void *arg);
(... ditto for free)
```

File: source/common/unified_memory_allocation/include/uma/memory_pool.h (hunk 20 above)
Comment (igchor) [reply source/common/unified_memory_allocation/include/uma/memory_pool.h]: The immediate reason for this API was to avoid tracking the metadata allocations yes. I don't think it will have a huge overhead so we can definitely wait with adding this.

Another reason was to allow user to pass some specialized metadata providers. However, this is an unrelated feature so I think it would be best to add it separately (in another PR) to keep this one simple.

I've removed this for now and will create another PR so we can discuss this API futher.

File: source/common/unified_memory_allocation/include/uma/memory_pool.h (hunk 10 above)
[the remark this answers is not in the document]
Comment (igchor) [reply source/common/unified_memory_allocation/include/uma/memory_pool.h:130]: UR uses plain 'NULL'. Here's an example: https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L425

File: source/common/unified_memory_allocation/include/uma/memory_pool.h (hunk 20 above)
Comment (vinser52) [reply source/common/unified_memory_allocation/include/uma/memory_pool.h]: > Another reason was to allow user to pass some specialized metadata providers. However, this is an unrelated feature so I think it would be best to add it separately (in another PR) to keep this one simple.

It is a good point. Introducing types for providers will help a heap manager (that requires more than 1 provider) to understand the purpose of each provider. But I agree with you: we can do it as a separate PR.

File: source/common/unified_memory_allocation/include/uma/memory_pool.h (hunk 10 above)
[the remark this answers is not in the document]
Comment (pbalcer) [reply source/common/unified_memory_allocation/include/uma/memory_pool.h:130]: Yea, I agree - let's just use NULL. A handle is just a pointer to an opaque type...

File: test/unified_memory_allocation/memoryPool.hpp [hunk 21]
```diff
@@ -52,4 +74,42 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
...
 }
 
+// TODO: add similar tests for realloc/aligned_alloc, etc.
+// TODO: add multithreaded tests
+TEST_P(umaMultiPoolTest, memoryTracking) {
+    static constexpr int allocSizes[] = {1,  2,  4,  8,   16,   20,
+                                         32, 40, 64, 128, 1024, 4096};
+    static constexpr auto nAllocs = 256;
+
+    std::mt19937_64 g(0); // TODO: test with random seeds?
```
pbalcer, line 84: no, tests should always use a fixed seed.

source/common/unified_memory_allocation/src/memory_tracker.cpp
```diff
@@ -0,0 +1,170 @@
...
+    enum uma_result_t ret;
+    uma_tracking_memory_provider_t *p =
+        (uma_tracking_memory_provider_t *)hProvider;
+
+    ret = umaMemoryTrackerRemove(p->hTracker, ptr, size);
+    if (ret != UMA_RESULT_SUCCESS) {
+        return ret;
+    }
+
+    // TODO: is it fine that deallocation and removing memory from the tracker is not a single atomic action
```
pbalcer, line 124: yes, because nothing else should have any pointer in this specific memory range. that would be race at a higher level.

source/common/unified_memory_allocation/src/memory_tracker.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    return ret;
+}
+
+static enum uma_result_t trackingFree(void *hProvider, void *ptr, size_t size) {
+    enum uma_result_t ret;
+    uma_tracking_memory_provider_t *p =
+        (uma_tracking_memory_provider_t *)hProvider;
+
+    ret = umaMemoryTrackerRemove(p->hTracker, ptr, size);
```
pbalcer, line 119: might be useful to add a comment that the order of removing from tracker and deallocating the pages is important for thread safety.

source/common/unified_memory_allocation/src/memory_tracker.cpp
```diff
@@ -0,0 +1,170 @@
...
+        std::shared_lock<std::shared_mutex> lock(mtx);
+
+        auto intptr = reinterpret_cast<uintptr_t>(ptr);
+        auto it = map.upper_bound(intptr);
+        if (it == map.begin()) {
+            return nullptr;
+        }
+
+        --it;
+        if (intptr >= it->first && intptr < it->first + it->second.first) {
```
pbalcer, line 48: this looks like magic if you don't know what you are looking at. Maybe create more useful variable names or add comments?
`size = it->first +it->second.first`

pbalcer (approved):
mostly lgtm, but you have a lot of TODOs in the code. Please create issues for at least the big things (e.g., optimizing the lookup data structure).

source/common/unified_memory_allocation/src/memory_tracker.cpp
igchor: Right, I've added a comment.

source/common/unified_memory_allocation/src/memory_tracker.cpp
igchor: Done.

igchor:
> mostly lgtm, but you have a lot of TODOs in the code. Please create issues for at least the big things (e.g., optimizing the lookup data structure).

Done: https://github.com/oneapi-src/unified-runtime/issues/442.

test/unified_memory_allocation/memoryPool.hpp
pbalcer: I strongly dislike when CI tests randomly fail because of some non-determinism in a test. If you need something with a lot of cases, I'd rather we do proper fuzzing in a separate job.

source/common/unified_memory_allocation/src/memory_tracker.cpp
```diff
@@ -0,0 +1,177 @@
...
+    params.hUpstream = hUpstream, params.hTracker = umaMemoryTrackerGet(),
+    params.pool = hPool;
+
+    struct uma_memory_provider_ops_t trackingMemoryProviderOps;
+    trackingMemoryProviderOps.version = UMA_VERSION_CURRENT;
+    trackingMemoryProviderOps.initialize = trackingInitialize;
+    trackingMemoryProviderOps.finalize = trackingFinalize;
+    trackingMemoryProviderOps.alloc = trackingAlloc;
+    trackingMemoryProviderOps.free = trackingFree;
+    trackingMemoryProviderOps.get_last_result = NULL;
```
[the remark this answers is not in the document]
igchor: Right, looks like a rebase issue. I fIxed and added a test for that

source/common/unified_memory_allocation/src/memory_tracker_internal.h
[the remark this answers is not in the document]
igchor: Done.

vinser52 (approved):

Issue #289, opened 2023-02-21
Validation layer for object lifetime tracking


pbalcer:
Lifetime of UR objects is managed manually by calls to release/retain. This can be error-prone and lead to memory leaks, use-after-free or double-free (e.g., releasing twice).

Create a validation layer that tracks object lifetimes and verifies that they are all appropriately freed on exit and that there are no redundant release calls. If any objects weren't released, print a stacktrace for when they were first created. 

In the next step, consider designing and implementing an additional feature that protects against using already released or incorrect handles in functions. This should be able to leverage the same lifetime information.

PR #318, opened 2023-03-03
Introduce basic leak checking into the validation layer

kswiecicki:
Leak checking can be turned on/off using `UR_ENABLE_LEAK_CHECKING` environmental variable.
Depends on: https://github.com/oneapi-src/unified-runtime/pull/154.


source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -9,7 +9,51 @@
...
+#include <atomic>
+#include <functional>
+#include <map>
+#include <mutex>
+#include <shared_mutex>
+#include <utility>
+
 namespace validation_layer {
+struct {
+    std::map<void *, std::atomic<int>> counts;
```
igchor, line 21: Why std::map and not std::unordered_map?

igchor, line 21: The reference count in UR is `uint32_t` so I think you should use a type that can hold at least UINT32_MAX references (`int64_t`)

source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -9,7 +9,51 @@
...
+    std::pair<void *, int> incrementRefCount(void *ptr) {
+        std::shared_lock<std::shared_timed_mutex> lock(mutex);
+        auto it = counts.find(ptr);
+        if (it != counts.end()) {
+            lock.unlock();
+            std::unique_lock<std::shared_timed_mutex> ulock(mutex);
+            it->second.fetch_add(1, std::memory_order_relaxed);
+            return std::make_pair(it->first, it->second.load());
+        } else {
+            counts.emplace(ptr, 1);
```
igchor, line 33: I think it should be the other way around - you should hold unique_lock here since you modify the entire data structure and shared_lock when incrementing a single element - also see my comment above.

source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -9,7 +9,51 @@
...
+struct {
+    std::map<void *, std::atomic<int>> counts;
+    std::shared_timed_mutex mutex;
+
+    std::pair<void *, int> incrementRefCount(void *ptr) {
+        std::shared_lock<std::shared_timed_mutex> lock(mutex);
+        auto it = counts.find(ptr);
+        if (it != counts.end()) {
+            lock.unlock();
+            std::unique_lock<std::shared_timed_mutex> ulock(mutex);
```
igchor, line 29: There's a potential race here. After you call unlock, some other thread might call decrementRefCount() and remove the element pointed to by `it` before you take the `unique_lock`.

The simplest solution would be to just always use `std::mutex` and `std::unique_lock`. Or you can use something like https://en.wikipedia.org/wiki/Double-checked_locking if you want to optimize for existing elements. However, I'm not entirely sure if find() on std::map is actually thread-safe... (standard does not specify it I think and there might be some internal state updates even on find).

source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -9,7 +9,51 @@
...
+    }
+
+    std::pair<void *, int> decrementRefCount(void *ptr) {
+        std::shared_lock<std::shared_timed_mutex> lock(mutex);
+        auto it = counts.find(ptr);
+        if (it != counts.end()) {
+            if (it->second.fetch_sub(1, std::memory_order_release) == 1) {
+                lock.unlock();
+                std::unique_lock<std::shared_timed_mutex> ulock(mutex);
+                counts.erase(it);
```
igchor, line 45: I'm wondering if we should also keep track of how many times decrement was called on an already released objects. E.g. when the ref count is already zero, and someone calls `decrementRefCount` you'll just return -1 but what if there's another `decrementRefCount` call to that object. Do we want to know that?

source/loader/layers/validation/ur_valddi.cpp
igchor: It would be good if you added some multithreaded tests.

source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -9,7 +9,51 @@
...
+        }
+    }
+
+    std::pair<void *, int> decrementRefCount(void *ptr) {
+        std::shared_lock<std::shared_timed_mutex> lock(mutex);
+        auto it = counts.find(ptr);
+        if (it != counts.end()) {
+            if (it->second.fetch_sub(1, std::memory_order_release) == 1) {
+                lock.unlock();
+                std::unique_lock<std::shared_timed_mutex> ulock(mutex);
```
igchor, line 44: Race here, similarly as in `incrementRefCount`

pbalcer:
#289 

Two thoughts. Just printing pointers to leaked objects is not super useful, consider printing backtraces (see `man 3 backtrace`). I also would prefer for most of the logic to live in the non-generated files.

source/loader/layers/validation/ur_valddi.cpp
kswiecicki: I've changed it to std::unordered_map.

kswiecicki: Done.

source/loader/layers/validation/ur_valddi.cpp
kswiecicki: Done.
I've omitted a case where I call urContextCreate by multiple threads because sometimes urContextCreate provides the same handle in different calls and it interferes with my .out.match tests. I'm not sure if urContextCreate is threadsafe.

source/loader/layers/validation/ur_valddi.cpp
kswiecicki: Since such information might be useful I'm now tracking the handles with negative refcount values.

source/loader/layers/validation/ur_valddi.cpp
kswiecicki: Should be resolved now.

kswiecicki:
> #289
> 
> Two thoughts. Just printing pointers to leaked objects is not super useful, consider printing backtraces (see `man 3 backtrace`). I also would prefer for most of the logic to live in the non-generated files.

I've added backtrace logging for the references retained after `urTearDown()`.
Also, I've moved most of the leak-checking logic to a separate file.

source/loader/layers/validation/ur_leak_check.hpp
```diff
@@ -0,0 +1,113 @@
...
+#include <utility>
+
+#define MAX_BACKTRACE_FRAMES 64
+
+namespace validation_layer {
+class RefCountContext {
+  private:
+    struct RefRuntimeInfo {
+        int64_t refCount;
+        char **originBacktrace;
```
igchor, line 19: You might want to wrap originBacktrace in std::unique_ptr here, you wouldn't need to call `clearRefInfo` manually and it would also prevent leaks in case of unhandled exceptions.

Or alternatively, you could just convert `backtraceStrings` you get from `backtrace_symbols` to std::string inside `createRefInfo` and just store a string here.

source/loader/layers/validation/ur_leak_check.hpp
```diff
@@ -0,0 +1,113 @@
...
+        int backtraceLines;
+    };
+
+    std::shared_mutex mutex;
```
igchor, line 23: you can use std::mutex since you're always using unique_lock with it

test/layers/validation/leaks_mt.cpp
```diff
@@ -0,0 +1,68 @@
...
+            ASSERT_EQ(urContextRetain(context), UR_RESULT_SUCCESS);
+            ASSERT_NE(nullptr, context);
+        });
+    }
+
+    for (auto &thread : threads) {
+        thread.join();
+    }
+
+    threads.clear();
```
igchor, line 29: nit: no need to call this

test/layers/validation/leaks_mt.cpp
```diff
@@ -0,0 +1,68 @@
...
+// SPDX-License-Identifier: MIT
+
+#include "fixtures.hpp"
+
+#include <iostream>
+#include <thread>
+#include <vector>
+
+INSTANTIATE_TEST_SUITE_P(threadCountForValDeviceTest, valDeviceTestMultithreaded,
+                         ::testing::Values(8));
```
PatKamin, line 11: Add to parameter values: 2 threads, and max threads, ie. [hardware_concurrency](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency)

source/loader/layers/validation/ur_leak_check.hpp
kswiecicki: I've changed `char **` into a `std::vector` of `std::string`.

source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -703,7 +786,9 @@ urMemImageCreate(
         }
     }
 
-    return pfnImageCreate(hContext, flags, pImageFormat, pImageDesc, pHost, phMem);
+    ur_result_t result = pfnImageCreate(hContext, flags, pImageFormat, pImageDesc, pHost, phMem);
```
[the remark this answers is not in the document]
kswiecicki: This function doesn't have a *Release counterpart.

test/layers/validation/leaks_mt.cpp
```diff
@@ -0,0 +1,68 @@
...
+            ASSERT_EQ(urContextRelease(context), UR_RESULT_SUCCESS);
+        });
+    }
+
+    for (auto &thread : threads) {
+        thread.join();
+    }
+
+    threads.clear();
+}
```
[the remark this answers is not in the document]
kswiecicki: Done.

kswiecicki:
@pbalcer what do you think about linking the loader with dbghelp Windows lib? `target_link_libraries(loader dbghelp)`
Should I introduce a compile time variable like `UR_ENABLE_VALIDATION_BACKTRACE` and link the necessary sources and dbghelp lib only when it is set?

source/loader/layers/validation/ur_leak_check.hpp
```diff
@@ -0,0 +1,102 @@
...
+        return RefRuntimeInfo{refCount, backtrace};
+    }
+
+    std::pair<void *, int64_t> incrementRefCount(void *ptr) {
+        auto it = counts.find(ptr);
+        if (it != counts.end()) {
+            it->second.refCount++;
+            return std::make_pair(it->first, it->second.refCount);
+        } else {
+            counts.emplace(ptr, createRefInfo(1));
```
pbalcer, line 38: I think we should differentiate between create/retain/release. It will allow you to find issues when retain is called on an object that wasn't created properly.

source/loader/layers/validation/ur_leak_check.hpp
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    std::pair<void *, int64_t> decrementRefCount(void *ptr) {
+        auto it = counts.find(ptr);
+        if (it != counts.end()) {
+            it->second.refCount--;
+            return std::make_pair(it->first, it->second.refCount);
+        } else {
+            counts.emplace(ptr, createRefInfo(-1));
```
pbalcer, line 49: isn't this an error anyway? you can't decrement something that doesn't exist. We should log this as an error.

source/loader/layers/validation/ur_leak_check.hpp
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+  public:
+    enum RefCountUpdateType {
+        REFCOUNT_INCREASE,
+        REFCOUNT_DECREASE,
+    };
+
+    void updateRefCount(void *ptr, enum RefCountUpdateType type) {
```
pbalcer, line 60: nit: these functions should be the other way around:

```
private:
  updateRefcount(ptr, type) {
    lock lock();
    if counts doesn't have ptr {
       counts.insert(ptr);
    }
    int value = 0;
    match(type) {
      case INC:
        value = 1;
       ...
    }
    counts[ptr] += value;
  }
public:
 inc(ptr) {
    update(ptr, INC);
 }
 dec(ptr) {
    update(ptr, DEC);
 }
```

At least, I think that would be cleaner and would reduce the repetitions and abstraction leaks you have here. Also, notice that you created an implicit requirement to acquire the mutex before calling the inc/dec functions. This isn't ideal - you should always try to keep the scope of the locks narrow and as close to the data they are protecting as possible.

pbalcer:
> @pbalcer what do you think about linking the loader with dbghelp Windows lib? Should I introduce a compile time variable like `UR_ENABLE_VALIDATION_BACKTRACE` and link the necessary sources and dbghelp lib only when it is set?

According to [docs](https://learn.microsoft.com/en-us/windows/win32/debug/calling-the-dbghelp-library), "DbgHelp.dll ships with all versions of Windows", so I don't see a problem with always linking it in.

test/layers/validation/leaks_mt.cpp
kswiecicki: Done.

source/loader/layers/validation/ur_leak_check.hpp
kswiecicki: Done.

source/loader/layers/validation/ur_leak_check.hpp
```diff
@@ -0,0 +1,113 @@
...
+        auto it = counts.find(ptr);
+        if (it != counts.end()) {
+            it->second.refCount--;
+            return std::make_pair(it->first, it->second.refCount);
+        } else {
+            counts.emplace(ptr, createRefInfo(-1));
+            return std::make_pair(ptr, -1);
+        }
+
+        return std::make_pair(nullptr, 0);
```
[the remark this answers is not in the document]
kswiecicki: Done.

source/loader/layers/validation/ur_leak_check.hpp
kswiecicki: The origin of those calls should be visible in backtrace.

source/loader/layers/validation/ur_leak_check.hpp
kswiecicki: There were suggestions that tracking *Release calls that have gone into the negative could be useful when the UR API wouldn't throw an error for some reason.

source/loader/layers/validation/ur_leak_check.hpp
kswiecicki: Understood, I've reversed the dependency of those functions.

source/loader/layers/validation/ur_leak_check.hpp
```diff
@@ -0,0 +1,98 @@
...
+            counts[ptr].refCount--;
+            break;
+        }
+
+        if (counts[ptr].refCount < 0 && type == REFCOUNT_DECREASE) {
+            context.logger.debug("Attempting to release nonexistent handle {}",
+                                 ptr);
+        }
+
+        context.logger.debug("Reference count for handle {} changed to {}", ptr,
```
pbalcer, line 59: Please use different log levels for events depending on their expected occurrence.

Someone might want only to see the errors in their app, not every single refcount change.

source/loader/layers/validation/ur_leak_check.hpp
pbalcer: Right, but the way you have it, doing:
```
  retain(0xCAFE);
  retain(0xCAFE);
  release(0xCAFE);
  release(0xCAFE);
```
Would be a perfectly valid sequence of calls. That's why I think special-casing `create` here makes sense.

source/loader/layers/validation/ur_leak_check.hpp
pbalcer: My suggestion was to differentiate between the scenario where retain is called on a completely invalid object (e.g., `0xCAFE`) and where retain is called on a previously valid one (going into negative).

scripts/core/INTRO.rst
```diff
@@ -271,3 +271,13 @@ Specific environment variables can be set to control the behavior of unified run
...
     This environment variable should be used together with "UR_ENABLE_VALIDATION_LAYER".
+
+.. envvar:: UR_ENABLE_LEAK_CHECKING
+
+   Holds the value '0' or '1'. By setting it to '1' you enable leak checking for Unified Runtime API calls involving
+   object creation/destruction. Leak checking depends on the logging mechanism.
+
+   .. note::
+
+    This environment variable should be used together with "UR_ENABLE_VALIDATION_LAYER" and "UR_LOG_VALIDATION".
```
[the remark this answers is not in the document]
pbalcer: nvm, I think there should be different verbosity levels. But please document the `UR_ENABLE_VALIDATION_LAYER` variable separately.

source/loader/layers/validation/backtrace_lin.cpp
```diff
@@ -0,0 +1,34 @@
...
+    if (backtraceStr == nullptr) {
+        return std::vector<BacktraceLine>(1, "Failed to acquire a backtrace");
+    }
+
+    std::vector<BacktraceLine> backtrace;
+    for (int i = 0; i < frameCount; i++) {
+        backtrace.push_back(BacktraceLine(backtraceStr[i]));
+    }
+
+    free(backtraceStr);
```
igchor, line 29: nit: there's still a (small) chance for a leak here: if push_back or std::string ctor throws, free() is never called.

scripts/core/INTRO.rst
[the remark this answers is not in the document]
kswiecicki: Done.

source/loader/layers/validation/ur_leak_check.hpp
kswiecicki: As we discussed, to differentiate between invalid references and valid references that have gone into negative we need to keep a record of all involved references. The memory cost is probably not worth the effort.

source/loader/layers/validation/backtrace_lin.cpp
kswiecicki: I've changed `push_back` to `emplace_back`, supposedly it's more efficient. The `std::bad_alloc` seems to be the only possible exception in this case. I've put the `emplace_back` inside a try-catch block to handle this exception.

source/loader/layers/validation/backtrace_lin.cpp
```diff
@@ -0,0 +1,34 @@
+/*
+ *
+ * Copyright (C) 2023 Intel Corporation
```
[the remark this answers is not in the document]
kswiecicki: Done.

source/loader/layers/validation/backtrace_lin.cpp
```diff
@@ -0,0 +1,39 @@
...
+    char **backtraceStr = backtrace_symbols(backtraceFrames, frameCount);
+
+    if (backtraceStr == nullptr) {
+        return std::vector<BacktraceLine>(1, "Failed to acquire a backtrace");
+    }
+
+    std::vector<BacktraceLine> backtrace;
+    try {
+        for (int i = 0; i < frameCount; i++) {
+            backtrace.emplace_back(backtraceStr[i]);
```
PatKamin, line 27: nit: You can already `reserve()` memory for 64 entries in the vector beforehand, avoiding vector copying when increasing its size.

source/loader/layers/validation/backtrace_win.cpp
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+    } catch (std::bad_alloc &) {
+        SymCleanup(process);
+        return std::vector<BacktraceLine>(1, "Failed to acquire a backtrace");
+    }
+
+    SymCleanup(process);
+
+    return backtrace;
```
PatKamin, line 45: If sth goes wrong and the `CaptureStackBackTrace` returns zero, then the returned backtrace is empty. I think it's better to return a backtrace with a single entry and a failure string.

source/loader/layers/validation/ur_leak_check.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include "ur_validation_layer.hpp"
+
+#include <mutex>
+#include <unordered_map>
+#include <utility>
+
+#define MAX_BACKTRACE_FRAMES 64
+
+namespace validation_layer {
+struct RefCountContext {
```
PatKamin, line 16: nit: An empty line at the beginning (and before the end) of the namespace scope for better readability, just as in other files.

source/loader/layers/validation/backtrace_lin.cpp
pbalcer: I wouldn't worry about things like that in non-performance-sensitive code. It just muddies things.

source/loader/layers/validation/backtrace_win.cpp
kswiecicki: Done.

source/loader/layers/validation/ur_leak_check.hpp
kswiecicki: Done.

scripts/templates/valddi.cpp.mako
```diff
@@ -16,18 +17,24 @@ from templates import helper as th
  * @file ${name}.cpp
  *
  */
-#include "${x}_validation_layer.hpp"
+#include "ur_leak_check.hpp"
```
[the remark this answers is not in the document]
kswiecicki: I'd forgotten about it. Should be fine now.

source/loader/layers/validation/ur_leak_check.hpp
kswiecicki: I've modified the leak-checking mechanism to differentiate between *Create/*Retain/*Release calls.

pbalcer (approved):

PR #319, opened 2023-03-03
Remove urEnqueueUSMMemset entrypoints and add urEnqueueUSMFill

### LINKED ISSUE #313: Remove urUSMMemset and urUSMMemset2D entrypoints. Add urUSMFill
urUSMMemset entrypoints can be removed from UR because their behaviour can always be achieved by using urUSMFill with a pattern of size 1.

In addition, an entrypoint for the 1 dimension version of urUSMFill entrypoint should be added to the specification.

This issue is a follow up to the discussion that happened during the work group weekly meeting of 01/03/2023.

fabiomestre:
Closes #313


kbenzie (commented):
These changes look good but I have one concern and that's how to implement patters larger 1 byte in CUDA. I suspect we'll need two paths dependant on the pattern size where `cudaMemset2D` is used when its 1 byte and another solution, potentially with `cudaMemcpy` in a loop, for larger patter sizes.

igchor (approved):

fabiomestre:
> These changes look good but I have one concern and that's how to implement patters larger 1 byte in CUDA. I suspect we'll need two paths dependant on the pattern size where `cudaMemset2D` is used when its 1 byte and another solution, potentially with `cudaMemcpy` in a loop, for larger patter sizes.

Yes, for efficiency, it will probably be good to have an if condition that uses cudaMemset when the pattern has 1 byte and memcpy for a larger pattern. I think an if condition inside fill() is still preferable than having a separate memset entrypoint that is only useful for cuda (opencl and level zero only have fill)

There is also other functions such as cuMemsetD16 and cuMemsetD32 which can be used to further optimize patterns of 2 bytes / 4 bytes.

PR #321, opened 2023-03-06
[CTS] Add device query helpers

veselypeta:
- Refactor the `testing` target from a simple header only `ALIAS` target into a traditional static library.
- Add helper functions for device queries to use within tests using out params

exceptions seems to me like a good solution but is bad for a few reasons:

```cpp
ASSERT_NO_THROW(const auto info = uur::GetDeviceInfo<T>(...));
// info out of scope :(

bool info;
ASSERT_NO_THROW(info = uur:GetDeviceInfo<T>(...));
// no advantage here to out params
// Also must remember to use `ASSERT_NO_THROW()`
// ASSERT_SUCCESS() is ubiquitous


// if this throws it will stop and fail the test
// but gtest will lose the point and which the test failed - pain to debug.
const auto info = uur::GetDeviceInfo<T>(...);

```


test/conformance/testing/include/uur/assert.h
```diff
@@ -0,0 +1,23 @@
...
+#ifndef UUR_CONFORMANCE_TESTING_ASSERT_H_INCLUDED
+#define UUR_CONFORMANCE_TESTING_ASSERT_H_INCLUDED
+
+#include <cstdio>
+#include <cstdlib>
+
+#define UUR_ASSERT(CONDITION, MESSAGE)                                     \
+    if (!(CONDITION)) {                                                    \
+        std::fprintf(stderr, "%s: %d: %s\n", __FILE__, __LINE__, MESSAGE); \
+        std::abort();                                                      \
```
omarahmed1111, line 13: I am just curious, does gtest catch the `SIGABRT` by default and declare the test as failed? (just concerned if that may terminate the whole tests if it fails or may not terminate the test without calling some necessary destructors like static variables if any)

omarahmed1111 (approved):
LGTM, just a small question on aborting the tests

test/conformance/testing/include/uur/assert.h
kbenzie: Neither of these macros seem to be used in this PR, maybe it shouldn't be added?

omarahmed1111: I guess they were used inside `test/conformance/testing/source/utils.cpp`

kbenzie: Oh, I was searching on a partial diff because GitHub is super helpful and doesn't load the whole thing... Disregard my comment then.

veselypeta: > I am just curious, does gtest catch the `SIGABRT` by default and declare the test as failed? (just concerned if that may terminate the whole tests if it fails or may not terminate the test without calling some necessary destructors like static variables if any)

I tested this locally and it just aborts the entire process. No destructors are called. Maybe this is an issue, but this is how it was handled in UnitCL

kbenzie: Yeah, this could be a real pain for a partially implemented adapter hitting aborts because a query failed. I'd be much better to propogate the error code and fail a fixture/test than abort.

kbenzie: Its not as nice as `std::expected` but we should probably always return a `ur_result_t` and get the value as an out param.

kbenzie (changes requested):
We need to avoid aborting on unimplemented queries and instead fail or skip the test fixture/case. Having the whole test suite abort in this situation will hide real test failures which occur after the failed query making iteration times longer and more of a pain.

veselypeta:
> We need to avoid aborting on unimplemented queries and instead fail or skip the test fixture/case. Having the whole test suite abort in this situation will hide real test failures which occur after the failed query making iteration times longer and more of a pain.

Can we throw exceptions? Perhaps we could do something like:

```cpp
ASSERT_NO_THROW(const auto addr_bits = uur::GetDeviceAddressBits(device));
```

kbenzie (approved):
LGTM

PR #322, opened 2023-03-06
[uma] Add more test helpers and a few simple tests

igchor:


PatKamin (approved):

source/common/uma_helpers.hpp
```diff
@@ -0,0 +1,144 @@
...
+#include <uma/memory_pool_ops.h>
+#include <uma/memory_provider.h>
+#include <uma/memory_provider_ops.h>
+
+#include <memory>
+#include <stdexcept>
+#include <tuple>
+#include <utility>
+
+namespace uma {
```
pbalcer, line 22: I made a quick and dirty prototype of what I meant: https://github.com/pbalcer/unified-runtime/commit/c2a09310cdcf889b8687f453f38908d49fcf0d23#diff-bce8177ea074557c7f61ceebeb5396d305a41b2689101c92174b69108f6dbb34R24

It's much simpler than this monstrosity (or I don't understand what this is for) ;)

igchor: Well, I just wanted to avoid manually implementing wrappers around all the functions (malloc, free, malloc_usable_size, ...) and just have a macro for that (hence UMA_OPS_GENERATE_FUNCTION). In your case it's just 2 lines, but it would be good to add some exception handling, etc.

And I'm not really a fan of the conversion operator - especially if it returns a raw UMA handle, it's quite easy to get a leak in that case. I'd rather have a function to create an instance of a pool/provider explicitly.
 
I can get rid of most of this code if we don't care about type safety in a pool/provider ctor - most of it is just to pack args... into a tuple and then back to args...

pbalcer (approved):

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -0,0 +1,54 @@
...
+    static constexpr size_t allocSize = 64;
+    auto *ptr = umaPoolMalloc(pool.get(), allocSize);
+    ASSERT_NE(ptr, nullptr);
+    std::memset(ptr, 0, allocSize);
+    umaPoolFree(pool.get(), ptr);
+}
+
+TEST_P(umaPoolTest, pow2AlignedAlloc) {
+#ifdef _WIN32
+    // TODO: implement support for windows
```
pbalcer, line 37: please add an issue to fix aligned allocs on windows.

PR #324, opened 2023-03-06
[ur] extend ur_usm_pool_limits_desc_t and rename members

igchor:


scripts/core/usm.yml
```diff
@@ -132,14 +132,17 @@ members:
...
+      name:  maxPoolSizePerDevice
+      desc: "[in] Maximum size of a memory pool per device"
     - type: "size_t"
       name:  maxPoolableSize
       desc: "[in] Allocations up to this limit will be subject to pooling"
     - type: "size_t"
-      name:  capacity
-      desc: "[in] When pooling, each bucket will hold a max of 4 unfreed slabs"
+      name:  bucketCapacity
+      desc: "[in] Maximum number of unfreed allocations in each bucket"
```
pbalcer, line 143: what do you mean by unfreed allocations?
This should be in bytes or pages.
And even then, I'm not sure how practical this is. Enforcing a single limit for all types of buckets might not make much sense. 4 pages is plenty for 16 byte objects, but very little for 2kb ones.

scripts/core/usm.yml
```diff
@@ -132,14 +132,17 @@ members:
     - type: "size_t"
       name:  maxPoolSize
       desc: "[in] Maximum size of a memory pool"
+    - type: "size_t"
+      name:  maxPoolSizePerDevice
+      desc: "[in] Maximum size of a memory pool per device"
```
pbalcer, line 137: I would have expected `memPoolSize` to be per device. Otherwise, how would an implementation enforce these max pool size limits? Fail to allocate on one device because some global limit has been reached due to allocations on another device? Is this useful behavior for some applications/use cases?

igchor: Well, in UR, pools are per-context, so limiting the amount of memory for the entire context might make sense (it is a set of 'somehow' related devices, after all). Yes, the implementation would fail to allocate from the pool and request the memory from the driver. I'm not entirely sure about the usefulness of this, but for L0 PI I belive there is a limit for overall memory pool size (not even per context): https://github.com/intel/llvm/blob/sycl/sycl/plugins/unified_runtime/ur/usm_allocator_config.cpp#L76

scripts/core/usm.yml
igchor: That's true, perhaps this should be a percentage then?

pbalcer: I think we just shouldn't have this. I'm not sure how changing bucket sizes is meant to be observable by actual applications or how developers are intended to pick a correct value for this setting without any runtime statistics that can be used to deduce fragmentation per bucket.

scripts/core/usm.yml
pbalcer: Artificially limiting the use of the fine-granular allocator in favor of falling back to allocating [blocks](https://github.com/intel/llvm/blob/b706bbd4659e99fcd4d665607b39dd342fc0617d/sycl/plugins/unified_runtime/ur/usm_allocator.cpp#L593) is likely to have a negative impact on fragmentation, increasing the overall memory consumption of an application. I genuinely don't understand when this might be useful from the application's point of view. Maybe I'm missing something.

scripts/core/usm.yml
igchor: Right, I'm fine with removing this but I'm wondering if we need to keep backward compatibility. 

Perhaps we could just allow setting this as part of `UR_LEVEL_ZERO_USM_CONFIG` env variable (we could make this env variable have the same format as the env variable in SYCL) and remove it from the programmatic API?

scripts/core/usm.yml
igchor: Yes, you are right, I'll try to think of some better name. One other thing - the USM allocator implementation has only one instance of this variable (common for all devices/contexts). Do you think that makes sense or we should have this per device/context like I proposed. I'm not sure if we need to support the global setting for compatibility.

pbalcer: Yes, I think this should only ever be a per-physical pool memory setting. From the user's perspective, this is intended to prevent situations where a single process is hogging all the physical memory. Having a single global limit might hamper allocation performance for one device without necessarily helping to alleviate memory pressure in the memory pool that's actually consuming too much space.

scripts/core/usm.yml
pbalcer: If we absolutely need to have it, I would leave it in the API but specify that it's a hint, and implementations are free to ignore it.

igchor: I removed this: turns out this option was only part of a DEBUG/EXPERIMENTAL API, so we don't need to keep it.

scripts/core/usm.yml
igchor: I agree. However, I'm wondering whether we should specify what a 'physical pool' is. I'm not 100% sure but I belive in some cases we could have a single physical pool per context (with multiple devices, if they have appropriate P2P capabilities), and in some cases (most of the time) we would want per-device pool. So, perhaps just leaving this as is (without specifying if it's per-device or context) is the best option? (but then, does it make sense to have the same settings for those 2 cases...?)

pbalcer: Good point. The topology can get complex. Maybe we can somehow define this in the context of 'local'/'device' memory and 'elsewhere'/'global' memory. Meaning that a pool would hold on to only X 'local' memory, and Y 'elsewhere' memory.

Jemalloc defines the concept of 'decay' that, instead of specifying limits in terms of space, defines it in terms of time - for how long does the pool is allowed to keep memory pages around. I think that's better because it wouldn't hamper performance if the pages are actually actively used, and would be meaningful for any type of memory - since the 'timer' is associated with each page/extent, not device in the pool. So if the allocator actively prefers local memory, any 'elsewhere' memory would naturally decay and get freed once it's no longer needed.

igchor:
I've created an issue to discuss this in more detail and closed this PR for now: https://github.com/oneapi-src/unified-runtime/issues/369

PR #325, opened 2023-03-06
[ur] move align param to ur_usm_desc_t

igchor:
The alignment is an optional paramater (can be set to 0).

We already have ur_usm_desc_t that groups flags and pool ptr that are also optional so it makes sense to store all of those parameters together.


pbalcer (approved):

veselypeta (approved):

igchor:
> please rebase.

Done.

Issue #10, opened 2022-11-09
What SPIR-V dialect does Unified Runtime consume?


FranklandJack:
The [`urProgramCreate`](https://spec.oneapi.io/unified-runtime/latest/core/api.html#urprogramcreate) API seems to mandate that `ur_module_handle_t`s are SPIR-V modules:
```
Create Program from input SPIR-V modules. 
```
However we don't specify which SPIR-V dialect we can consume. It should probably either reference the OpenCL environment spec or a new Unified Runtime dialect.

kbenzie:
For adapters which consume SPIR-V, it is Kernel capability SPIR-V as defined for OpenCL. Not all adapters will consume SPIR-V though.

aarongreig:
so we do have the device [IL_VERSION](https://oneapi-src.github.io/unified-runtime/core/api.html#_CPPv4N16ur_device_info_t25UR_DEVICE_INFO_IL_VERSIONE) property, maybe it would be enough to spec that out a bit more and leave the specifics to individual plugins as an implementation detail?

alycm:
Really any SPIR-V consumer should specify an environment for how SPIR-V is interpreted, e.g. as with [OpenCL](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html), [Vulkan](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#spirvenv), or [Level Zero](https://spec.oneapi.io/level-zero/latest/core/SPIRV.html).  Often the SPIR-V specification will defer definitions of behaviour to a "client API", so without an environment specification to pair with the SPIR-V specification is incomplete.

For unified runtime it may be sufficient to just state that either the OpenCL or Level Zero one applies.  But is that true?  Level Zero and OpenCL SPIR-V requirements are certainly broadly similar at a glance, but are they exactly the same?

That is an argument for letting the adapter specify the SPIR-V environment, but could that mean that N SPIR-V using adapters could need to pack N SPIR-V variants in per binary?  N is currently 2, but going to at least 3 is plausible.  It could be that Unified Runtime has a single SPIR-V environment for the front-end, and if an adapter needs very different SPIR-V in the back-end then it needs to translate it.  Like how `clvk` can consume kernel SPIR-V from OpenCL but then generates shader SPIR-V for Vulkan.

aarongreig:
[...]

PR #327, opened 2023-03-07
[UR] `urDeviceSelectBinary` spec changes 

### LINKED ISSUE #114: Add ur_device_binary_type enum
# Issue

PI defines a [`pi_device_binary_type`](https://github.com/intel/llvm/blob/c4932957929f94afe7f87bee009eea5e2b5cd6bf/sycl/include/sycl/detail/pi.h#L708) enumeration that can take the following values:
* `PI_DEVICE_BINARY_TYPE_NONE`
* `PI_DEVICE_BINARY_TYPE_NATIVE`
* `PI_DEVICE_BINARY_TYPE_SPIRV`
* `PI_DEVICE_BINARY_TYPE_LLVMIR_BITCODE`
which are used [here](https://github.com/intel/llvm/blob/b083e0543c264a1fb210bf5990a912bac41ba04a/sycl/source/detail/device_binary_image.cpp#L165), [here](https://github.com/intel/llvm/blob/f4fc927095949ed9403b049662ec60923d56b067/sycl/source/detail/pi.cpp#L644), [here](https://github.com/intel/llvm/blob/b4aaca7bc3801e23b4df49ebe35a1057881a7414/sycl/source/detail/program_manager/program_manager.cpp#L251) and [here](https://github.com/intel/llvm/blob/b4aaca7bc3801e23b4df49ebe35a1057881a7414/sycl/source/detail/program_manager/program_manager.cpp#L1403) respectively in the DPC++ runtime.

There is no `ur_device_binary_type` enumeration in Unified Runtime.

# Task 

Add the `ur_device_binary_type` enumeration to Unified Runtime so that it meets the requirements for implementing the DPC++ runtime.

### LINKED ISSUE #118: Consider whether `ur_device_binary_struct` is required
# Issue

The PI interface defines a [`pi_device_binary_struct`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L796) struct that is (via a `pi_device_binary` alias) is passed through the [`piextDeviceSelectBinary`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L1046) and [`piProgramCreateWithBinary`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L1246). 

The `pi_device_binary_struct` type contains various bits of metadata about the binary. There is no equivalent `ur_device_binary_struct` in Unified Runtime, instead [urDeviceSelectBinary](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3032) and [urProgramCreateWithBinary](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L3946) take the binary arguments as `uint8_t*`s.

The DPC++ runtime [makes use of the `pi_deivce_binary`](https://github.com/intel/llvm/blob/1c77f9a78a59ecf60bafe5ed73b3878dea96fae1/sycl/source/detail/device_binary_image.hpp#L123) alias. If UR is to be a replacement for PI in the DPC++ runtime we need to decide whether this is a requirement to add to Unified Runtime or not.

If it is decided we need a `ur_device_binary_struct` we may also need to port the associated types and values to UR:
* [`_pi_device_binary_property_struct`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L696)
* [`_pi_device_binary_property_set_struct`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L706)
* [`PI_DEVICE_BINARY_VERSION`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L727)
* [`PI_DEVICE_BINARY_OFFLOAD_KIND_SYCL`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L730)
* [`pi_device_binaries_struct`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L893)
* [`PI_DEVICE_BINARIES_VERSION`](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L888)

and the [various target identification strings](https://github.com/intel/llvm/blob/5995c61840153b26817ea5b45e7bc0f49b6f4c3e/sycl/include/sycl/detail/pi.h#L732-L784).

This is also related to https://github.com/oneapi-src/unified-runtime/issues/95 which adds a metadata parameter to `urProgramCreateWithBinary`. If the parameter is required and added, its type would be `ur_device_binary_property` which would be a field in the `ui_device_binary_property_set` which is in turn would be a field in `ur_device_binary_struct`.

# Task

Investigate whether the above PI API objects need to be ported to the Unified Runtime spec. If they are, port them if not then no action is required.

### LINKED ISSUE #171: Clarify `urDeviceSelectBinaryTest`
Currently we have a placeholder test for `urDeviceSelectBinary` which only calls the entry-point with valid arguments, but does not actually assert for any behaviour. We cloud discuss what this entry point should do an it's requirements once we have a clearer understanding of the compiler requirements for UR.

### LINKED ISSUE #252: `urDeviceSelectBinary` should take an array of binary sizes
The `urDeviceSelectBinary` entry point currently has the following signature: 

```cpp
urDeviceSelectBinary(
    ur_device_handle_t hDevice,                     ///< [in] handle of the device to select binary for.
    const uint8_t** ppBinaries,                     ///< [in] the array of binaries to select from.
    uint32_t NumBinaries,                           ///< [in] the number of binaries passed in ppBinaries. Must greater than or
                                                    ///< equal to zero.
    uint32_t* pSelectedBinary                       ///< [out] the index of the selected binary in the input array of binaries.
                                                    ///< If a suitable binary was not found the function returns ${X}_INVALID_BINARY.
    );
```

We should pass an additional argument, which will be an array of sizes for the length of each binary.

veselypeta:
PI uses a `pi_device_binary` type which includes a bunch of additional metadata that is used by the program manager in SYCL. However the only place where this binary struct is used is with the `urDeviceSelectBinary` entry point. Every implemented plugin selects the correct device binary by comparing the correct `DeviceTargetSpec` within the plugin. The extra complexity inherited in PI from the program manager is unnecessary for UR and has been removed. I have left the option of extending this struct with the `ur_base_desc_t` class.

Closes #252 
Closes #118 
Closes #171 
Closes #114

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -233,6 +233,7 @@ typedef enum ur_structure_type_t {
     UR_STRUCTURE_TYPE_USM_DESC = 7,                         ///< ::ur_usm_desc_t
     UR_STRUCTURE_TYPE_USM_POOL_DESC = 8,                    ///< ::ur_usm_pool_desc_t
     UR_STRUCTURE_TYPE_USM_POOL_LIMITS_DESC = 9,             ///< ::ur_usm_pool_limits_desc_t
+    UR_STRUCTURE_TYPE_DEVICE_BINARY = 10,                   ///< ::ur_device_binary_t
     /// @cond
     UR_STRUCTURE_TYPE_FORCE_UINT32 = 0x7fffffff
     /// @endcond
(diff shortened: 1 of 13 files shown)


veselypeta:
The current `piextDeviceSelectBinary` signature is:
```cpp
pi_result piextDeviceSelectBinary(pi_device device,
                                  pi_device_binary *binaries,
                                  pi_uint32 num_binaries,
                                  pi_uint32 *selected_binary_ind);
```

However:
```cpp 
using pi_device_binary = pi_device_binary_struct *;
```

So in effect:
```cpp
pi_result piextDeviceSelectBinary(pi_device device,
                                  pi_device_binary_struct **binaries,
                                  pi_uint32 num_binaries,
                                  pi_uint32 *selected_binary_ind);
```

Since this is just an array of binaries is there any reason to have this additional layer of indirection?

scripts/core/program.yml
```diff
@@ -68,7 +19,7 @@ members:
     - type: uint32_t
       name: count
       desc: "[in] the number of entries in pMetadatas, if count is greater than zero then pMetadatas must not be null."
-    - type: const $x_program_metadata_t*
+    - type: const $x_device_binary_property_t*
```
kbenzie, line 22: This is not my favorite change. Seems much less clear reading this here, also its not just a binary property its also passed to `CreateWithIL`.

veselypeta: Yeah I think this is poorly names. I was copying the existing style that is in PI - is `piProgramCreateWithBinary` equivalent to `urProgramCreateWithIL`?

kbenzie: > Yeah I think this is poorly names. I was copying the existing style that is in PI - is `piProgramCreateWithBinary` equivalent to `urProgramCreateWithIL`?

No, `piProgramCreateWithBinary` == `urProgramCreateWithBinary` and `piProgramCreate` == `urProgramCreateWithIL` although when I ported the metadata structs I added them to both of the UR entry points. Currently the CUDA plugin passes PTX, which is an IL, to `piProgramCreateWithBinary` but logically it should be `urProgramCreateWithIL` which is why I added it to both entrypoints.

veselypeta: My mistake thanks. Do you think it can just be called something generic like `ur_metadata_property_t`?

kbenzie (approved):
LGTM

include/ur_api.h
```diff
@@ -233,6 +233,7 @@ typedef enum ur_structure_type_t {
     UR_STRUCTURE_TYPE_USM_DESC = 7,                         ///< ::ur_usm_desc_t
     UR_STRUCTURE_TYPE_USM_POOL_DESC = 8,                    ///< ::ur_usm_pool_desc_t
     UR_STRUCTURE_TYPE_USM_POOL_LIMITS_DESC = 9,             ///< ::ur_usm_pool_limits_desc_t
+    UR_STRUCTURE_TYPE_DEVICE_BINARY = 10,                   ///< TODO - ur_device_binary_desc_t not yet implemented.
```
kbenzie, line 236: Is this still a TODO?

scripts/core/device.yml
```diff
@@ -11,6 +11,73 @@ type: header
...
+desc: |
+      Target identification strings for $x_device_binary_t.pDeviceTargetSpec 
+      A device type represented by a particular target triple requires specific 
+      binary images. We need to map the image type onto the device target triple
+name: "$X_DEVICE_BINARY_TARGET_UNKNOWN"
+value: "\"<unknown>\""
+--- #--------------------------------------------------------------------------
+type: macro
+desc: |
+     SPIR-V 32-bit image <-> "spir", 32-bit OpenCL device
```
aarongreig, line 24: I would hesitate to name-drop OpenCL in here, still not sure what we're going to do about #10. Could say something like "32-bit address width"

kbenzie: I'm guessing this was lifted from PI?

veselypeta: This is essentially copy pasta from PI.

scripts/core/device.yml
```diff
@@ -453,9 +520,10 @@ params:
       name: pSelectedBinary
       desc: |
             [out] the index of the selected binary in the input array of binaries.
-            If a suitable binary was not found the function returns ${X}_INVALID_BINARY.
+            If a suitable binary was not found the function returns $X_RESULT_ERROR_INVALID_BINARY.
 returns:
-    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_INVALID_VALUE:
```
aarongreig, line 525: `$X_RESULT_ERROR_INVALID_SIZE`

include/ur_api.h
```diff
@@ -544,6 +545,77 @@ urGetLastResult(
...
+#define UR_DEVICE_BINARY_TARGET_AMDGCN "amdgcn"
+#endif // UR_DEVICE_BINARY_TARGET_AMDGCN
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Device Binary Type
+typedef struct ur_device_binary_t {
+    ur_structure_type_t stype;     ///< [in] type of this structure, must be ::UR_STRUCTURE_TYPE_DEVICE_BINARY
+    const void *pNext;             ///< [in][optional] pointer to extension-specific structure
+    const char *pDeviceTargetSpec; ///< [in] null-terminated string representation of the device's target architecture.
+                                   ///< Must be one of
```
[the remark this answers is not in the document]
veselypeta: 👍

include/ur_api.h
veselypeta: Fixed.

scripts/core/device.yml
aarongreig: alright I'll add a note to #10 anyway

scripts/core/device.yml
veselypeta: 👍

aarongreig (approved):

PR #329, opened 2023-03-08
[ur] Enable better formating for tests

veselypeta:
This PR:
* Removes `ColumnLimit: 0` from the root `.clang-format`.
* Adds override `.clang-format` file for autogenerated files in `/include`
* Applies all remaining formating changes

Important files [`.clang-format`](https://github.com/oneapi-src/unified-runtime/pull/329/files#diff-1026e0038b722990204a42bed8a6f7c0ec2302aa79e3fad1959d62ba968edfa2) [`include/.clang-format`](https://github.com/oneapi-src/unified-runtime/pull/329/files#diff-f7bc9dbee0ee89027baab27fda48147ed6a4f5abc7998a9042f9ad3c6e133bae)

### DIFF
--- a/examples/collector/collector.cpp
+++ b/examples/collector/collector.cpp
@@ -25,14 +25,17 @@
 #include "ur_api.h"
 #include "xpti/xpti_trace_framework.h"
 
-constexpr uint16_t TRACE_FN_BEGIN = static_cast<uint16_t>(xpti::trace_point_type_t::function_with_args_begin);
-constexpr uint16_t TRACE_FN_END = static_cast<uint16_t>(xpti::trace_point_type_t::function_with_args_end);
+constexpr uint16_t TRACE_FN_BEGIN =
+    static_cast<uint16_t>(xpti::trace_point_type_t::function_with_args_begin);
+constexpr uint16_t TRACE_FN_END =
+    static_cast<uint16_t>(xpti::trace_point_type_t::function_with_args_end);
 constexpr std::string_view UR_STREAM_NAME = "ur";
 
 /**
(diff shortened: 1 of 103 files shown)


test/.clang-format
```diff
@@ -0,0 +1,6 @@
+---
+Language:      Cpp
+BasedOnStyle:  LLVM
+IndentWidth:   4
+InsertBraces: true
+...
```
pbalcer, line 6: Please use `ReflowComments: false`. See #267 for context.
Otherwise lgtm.

pbalcer:
Hm, to be more consistent across the codebase, perhaps we could create a `.clang-format` in the include directory with `ColumnLimit: 0`, and then remove that setting from the `.clang-format` in the main directory?

kbenzie:
> Hm, to be more consistent across the codebase, perhaps we could create a `.clang-format` in the include directory with `ColumnLimit: 0`, and then remove that setting from the `.clang-format` in the main directory?

This thought has been hanging around in the back of my mind too. I've found it harder to review non-generated code recently with the unlimited line lengths.

test/.clang-format
veselypeta: I think the fact that it is non-idempotent won't affect code that is not autogenerated.

pbalcer: It might, because in CI we are essentially formatting the entire repo and failing if there's a diff. So it doesn't matter whether it's autogenerated or not.

veselypeta: 👍

pbalcer (approved):

pbalcer:
Everyone needs to rebase PR after this merge.

PR #331, opened 2023-03-08
ur: Update spec to v0.6

kbenzie:
Change version in preparation for oneAPI HW SIG to differentiate the current state of the specification from previous meetings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MIT
 
 cmake_minimum_required(VERSION 3.14.0 FATAL_ERROR)
-project(unified-runtime VERSION 0.5.0)
+project(unified-runtime VERSION 0.6.0)
 
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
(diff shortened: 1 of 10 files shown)


kbenzie:
I don't actually know how to do that, do you @jandres742 ?

alycm:
I don't think that it should redirect, it makes sense to have both a "released" version at deliberate points and the live version that updates with every accepted PR.

jandres742:
Ah ok,  we have two URLs for the spec:
https://spec.oneapi.io/unified-runtime/latest/index.html
https://oneapi-src.github.io/unified-runtime/

I guess the former is the official, updated every once in a while (like every release), and the latter is updated on every commit.

in that case, we shouldn't redirect, as one is the stable spec, and the other is the development spec.

I dont know how to update the one in  https://spec.oneapi.io/unified-runtime/latest/index.html, but I can find out. Let me ask around.

kbenzie:
> I dont know how to update the one in https://spec.oneapi.io/unified-runtime/latest/index.html, but I can find out. Let me ask around.

Thank you!

PR #339, opened 2023-03-09
[ur] Refactor adapters load

PatKamin:
Move code specific for adapters loading to a new header.


source/loader/lib_loader.hpp
```diff
@@ -0,0 +1,133 @@
...
+ *
+ */
+
+#include <array>
+
+#include "ur_api.h"
+#include "ur_ddi.h"
+#include "ur_object.hpp"
+
+#if defined(_WIN32)
```
pbalcer, line 15: as a general rule, we should be using platform ifdefs very sparingly. In this case, I suggest creating a simple abstraction for managing dynamic libraries and then providing platform-specific implementations.

source/loader/lib_loader.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <Windows.h>
+#else
+#include <dlfcn.h>
+#endif
+
+/* for compatibility with non-clang compilers */
+#if defined(__has_feature)
```
pbalcer, line 22: this should stay in utils.

source/loader/lib_loader.hpp
```diff
@@ -0,0 +1,133 @@
...
+#define CLANG_HAS_FEATURE(x) __has_feature(x)
+#else
+#define CLANG_HAS_FEATURE(x) 0
+#endif
+
+/* define for running with address sanitizer */
+#if CLANG_HAS_FEATURE(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
```
kswiecicki, line 29: Shouldn't sanitizer defines also stay in utils?

source/loader/lib_loader.hpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    void discoverEnabledPlatforms() {
+        std::optional<std::string> altPlatforms;
+
+        // UR_ADAPTERS_FORCE_LOAD  is for development/debug only
+        try {
+            altPlatforms = ur_getenv("UR_ADAPTERS_FORCE_LOAD");
+        } catch (const std::invalid_argument &e) {
+            std::cerr << e.what();
```
igchor, line 95: why not logger?

source/loader/lib_loader.hpp
PatKamin: Reverted

source/loader/lib_loader.hpp
PatKamin: Created files for linux- and windows-specific functions.

source/loader/linux/lib_loader.cpp
```diff
@@ -0,0 +1,30 @@
...
+#if defined(SANITIZER_ANY)
+#define LOAD_DRIVER_LIBRARY(NAME) dlopen(NAME, RTLD_LAZY | RTLD_LOCAL)
+#else
+#define LOAD_DRIVER_LIBRARY(NAME) dlopen(NAME, RTLD_LAZY | RTLD_LOCAL | RTLD_DEEPBIND)
+#endif
+
+namespace loader {
+
+HMODULE LibLoader::loadAdapterLibrary(const char *name) {
+    return LOAD_DRIVER_LIBRARY(name);
```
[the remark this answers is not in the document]
PatKamin: Done

source/loader/lib_loader.hpp
```diff
@@ -0,0 +1,84 @@
...
+                platforms.emplace_back(std::move(handle));
+            }
+        }
+    }
+
+  private:
+    static constexpr std::array<char *, 0> knownPlatformNames{};
+    std::vector<std::string> discoveredPlatforms;
+
+    static LibHandle loadAdapterLibrary(const char *name);
```
pbalcer, line 50: It would be cleaner to create a standalone abstraction (with its own class) for this in common. And then the loader can just instantiate that class to access the functions.

PatKamin: Split the code into three parts: loading libraries, finding adapters, and platform_t structure. Moved first two abstractions to common.

source/common/ur_adapters_detector.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include "logger/ur_logger.hpp"
+#include "ur_util.hpp"
+
+namespace loader {
+
+class AdaptersDetector {
+  public:
+    std::vector<std::string> discoveredPlatforms;
+
+    void discoverEnabledPlatforms() {
```
igchor, line 22: Why not run this code in the constructor? This function should only be called once anyway - otherwise you'll get duplicated entires.

igchor: Alternatively, you could make this function static (or freestanding, do you even need a class here?) and just return a vector or platforms - this way, even if you call it twice, you get good result.

source/common/ur_adapters_detector.hpp
```diff
@@ -0,0 +1,57 @@
...
+ */
+#ifndef UR_ADAPTERS_DETECTOR_HPP
+#define UR_ADAPTERS_DETECTOR_HPP 1
+
+#include <array>
+
+#include "logger/ur_logger.hpp"
+#include "ur_util.hpp"
+
+namespace loader {
```
pbalcer, line 16: This is part of the loader, so keep it in the loader directory.

source/loader/linux/ur_loader_lin.cpp
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+#include "ur_loader.hpp"
+
+namespace loader {
+
+FUNC_PTR platform_t::getFunctionPtr(const char *func_name) {
+    return dlsym(handle.get(), func_name);
```
pbalcer, line 13: You should keep all the dynamic shared library abstractions in one place - `ur_lib_loader`.

source/common/ur_adapters_detector.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+class AdaptersDetector {
```
pbalcer, line 18: I don't like this name :P `AdaptersRegistry` ?

scripts/templates/ldrddi.hpp.mako
```diff
@@ -19,6 +19,8 @@ from templates import helper as th
 #ifndef UR_LOADER_LDRDDI_H
 #define UR_LOADER_LDRDDI_H 1
 
+#include "ur_object.hpp"
```
kswiecicki, line 22: `#include "${x}_object.hpp"`. Change the copyright date here and the ldrddi.cpp.mako file.

source/common/ur_lib_loader.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <memory>
+
+#include "ur_util.hpp"
+
+namespace loader {
+
+class LibLoader {
+  public:
+    struct lib_dtor {
+        typedef HMODULE pointer;
```
kswiecicki, line 20: Do you use the `pointer` somewhere?

source/common/ur_adapters_detector.hpp
PatKamin: AdapterRegistry is an object with an iterator-like behavior now and it's placed in the context_t.

source/common/ur_util.hpp
```diff
@@ -38,35 +38,22 @@
 #if defined(SANITIZER_MEMORY) || defined(SANITIZER_ADDRESS)
 #define SANITIZER_ANY
 #endif
-
 ///////////////////////////////////////////////////////////////////////////////
 #if defined(_WIN32)
 #include <Windows.h>
```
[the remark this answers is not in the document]
PatKamin: Done

PR #341, opened 2023-03-10
[ur] Replace extern logger object with a singleton

PatKamin:
Replace extern logger object with a singleton one. Remove the need for initializing logger when logger is not used.

Add printing logger name at the beginning of messages.

### DIFF
--- a/source/common/logger/ur_logger.hpp
+++ b/source/common/logger/ur_logger.hpp
@@ -11,31 +11,41 @@
 #include "ur_util.hpp"
 
 namespace logger {
-extern Logger logger;
+
+Logger create_logger(std::string logger_name);
+
+inline Logger &get_logger(std::string name = "common") {
+    static Logger logger = create_logger(name);
+    return logger;
+}
+
+inline void init(std::string name) {
+    get_logger(name);
+}
 
 template <typename... Args>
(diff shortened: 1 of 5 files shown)


source/common/logger/ur_sinks.hpp
```diff
@@ -13,6 +13,7 @@ class Sink {
   public:
     template <typename... Args>
     void log(logger::Level level, const char *fmt, Args &&...args) {
+        *ostream << "<" << logger_name << ">";
         *ostream << "[" << level_to_str(level) << "]:";
```
pbalcer, line 17: I think there should be a space after `:`

PatKamin: Done

igchor (commented):
LGTM

source/common/logger/ur_logger.hpp
```diff
@@ -11,31 +11,41 @@
 #include "ur_util.hpp"
 
 namespace logger {
-extern Logger logger;
+
+Logger create_logger(std::string logger_name);
+
+inline Logger &get_logger(std::string name = "") {
```
pbalcer, line 17: This means that tests and other places you log without explicit init will print `<>[INFO]: message`. Maybe instead we can use 'common' ?

PR #343, opened 2023-03-10
[ur] create function id registry 

pbalcer:
XPTI function tracing requires "a stable ABI function id" that
is used to programmatically identify a traced entry point.
This patch creates a function id registry enum that stores such
stable unique ids for all functions. This registry is part of the
spec and must be kept stable across versions.

To make things easier, this patch also adds a script that
automatically generates ids for newly added spec functions
while preserving existing identifiers.

Also uses the newly added function id registry in the xpti
tracing layer


scripts/generate_ids.py
```diff
@@ -0,0 +1,69 @@
...
+    updated = False
+
+    new_registry = list()
+    for fname in functions:
+        etor_name = util.to_snake_case(fname)
+        id = etors.get(etor_name)
+        if id is None:
+            updated = True
+            id = valid_ids.pop()
+        new_registry.append({'name': util.to_snake_case(fname), 'desc': 'Enumerator for $x'+fname, 'id': id})
```
kbenzie, line 49: What happens if a function is removed from the spec?

I think it would no longer appear in the registry but the ID would not be used again but I'm not sure. Does that sound correct?

kbenzie (approved):
LGTM just one question

scripts/generate_ids.py
pbalcer: Yup. The id will become available the next time around. I considered this and concluded that addressing this is not worth complicating the script (I'd have to reverse map the etors to function names in the spec).

igchor (approved):
LGTM

PR #345, opened 2023-03-10
Add a bit of spec detail to the PROGRAM_BUILD_INFO_OPTIONS query.

aarongreig:
The new wording should make a bit more sense with the latest changes to the compiler API.

### DIFF
--- a/scripts/core/program.yml
+++ b/scripts/core/program.yml
@@ -310,21 +310,21 @@ etors:
             The reference count returned should be considered immediately stale. 
             It is unsuitable for general use in applications. This feature is provided for identifying memory leaks.
     - name: CONTEXT
-      desc: "Program context info"
+      desc: "Program context info."
     - name: NUM_DEVICES
-      desc: "Return number of devices associated with Program"
+      desc: "Return number of devices associated with Program."
     - name: DEVICES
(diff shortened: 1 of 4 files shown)


omarahmed1111 (approved):
LGTM

scripts/core/program.yml
```diff
@@ -391,7 +391,7 @@ etors:
     - name: STATUS
       desc: "Program build status, return type $x_program_build_status_t"
     - name: OPTIONS
-      desc: "Program build options, return type char[]"
+      desc: "Options string specified by last build, compile or link operation performed on the program. Return type char[]"
```
kbenzie, line 394: Should maybe be `char*` for the return type and mention it being null-terminated.

kbenzie (approved):
LGTM, one comment

scripts/core/program.yml
aarongreig: CL also does the [] thing but I don't think it makes a whole load of sense. I guess for char[] stuff it's like you're returning char[pPropSizeRet]? but that breaks down for stuff like int32_t[] so yeah changed all instances of that to use `*`

pbalcer (approved):

PR #347, opened 2023-03-10
ur: Add parameter valid usage for enqueue buffer commands

### LINKED ISSUE #19: Consider semantic parameter verification/valid usage for arguments
Unified Runtime currently supports parameter verification for some arguments based on their type. For example, pointer and handle (which is technically still a pointer) arguments are automatically checked against `NUL`. The generation of the error language in the spec appears to be automatic and based only on the type of the argument. 

(Aside: In some cases this is too strict, e.g. the out parameter for the event argument on enqueue APIs should probably be optional: https://github.com/oneapi-src/unified-runtime/blob/main/include/zer_api.h#L859.)

More generally, we should consider whether (like OpenCL) we want to have verification for the semantics of arguments where appropriate e.g. checking that a read/write command will not overflow a buffer. In OpenCL the [`clEnqueue(Read|Write)Buffer`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_reading_writing_and_copying_buffer_objects) APIs require that the entry point returns:

```
CL_INVALID_VALUE if the region being read or written specified by (offset, size) is out of bounds or if ptr is a NULL value.
```

We need to decide whether we want parameter verification of this kind since in total for all the APIs there will be a lot of it. We will also need to write negative tests for all these cases once we have a test suite.

kbenzie:
Define the conditions under which the `UR_RESULT_ERROR_INVALID_SIZE`
error code should be returned for the following entry points:

* `urEnqueueMemBufferRead`
* `urEnqueueMemBufferReadRect`
* `urEnqueueMemBufferWrite`
* `urEnqueueMemBufferWriteRect`
* `urEnqueueMemBufferCopy`
* `urEnqueueMemBufferCopyRect`
* `urEnqueueMemBufferFill`
* `urEnqueueMemBufferMap`

Fixes #19

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -4204,6 +4204,8 @@ urEnqueueEventsWaitWithBarrier(
 ///         + `phEventWaitList != NULL && numEventsInWaitList == 0`
 ///         + If event objects in phEventWaitList are not valid events.
 ///     - ::UR_RESULT_ERROR_INVALID_MEM_OBJECT
+///     - ::UR_RESULT_ERROR_INVALID_SIZE
+///         + If `offset + size` results in an out-of-bounds access.
 ///     - ::UR_RESULT_ERROR_OUT_OF_HOST_MEMORY
 ///     - ::UR_RESULT_ERROR_OUT_OF_RESOURCES
 UR_APIEXPORT ur_result_t UR_APICALL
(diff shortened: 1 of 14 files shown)


pbalcer (approved):

martygrant (approved):
LGTM

Issue #326, opened 2023-03-06
Extend urUSMGetMemAllocInfo to return memory pool handle?


igchor:
It might be useful if someone wants to allocate memory from a pool that contains a specific pointer.

kbenzie:
This makes sense to me, I think that would just mean adding `UR_USM_ALLOC_INFO_POOL` to [ur_usm_alloc_info_t](https://oneapi-src.github.io/unified-runtime/core/api.html#ur-usm-alloc-info-t) and adding it to the list of enums to test [here](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/usm/urUSMGetMemAllocInfo.cpp#L9).

PR #348, opened 2023-03-11
[ur] extend ur_usm_alloc_info_t with pool info

igchor:
Closes: https://github.com/oneapi-src/unified-runtime/issues/326

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -841,6 +841,7 @@ class ur_usm_alloc_info_v(IntEnum):
     BASE_PTR = 1                                    ## Memory allocation base pointer info
     SIZE = 2                                        ## Memory allocation size info
     DEVICE = 3                                      ## Memory allocation device info
+    POOL = 4                                        ## Memory allocation pool info
 
 class ur_usm_alloc_info_t(c_int):
     def __str__(self):
--- a/scripts/core/usm.yml
+++ b/scripts/core/usm.yml
@@ -70,6 +70,8 @@ etors:
       desc: "Memory allocation size info"
     - name: DEVICE
       desc: "Memory allocation device info"
+    - name: POOL
+      desc: "Memory allocation pool info"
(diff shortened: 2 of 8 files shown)


include/ur.py
```diff
@@ -843,6 +843,7 @@ class ur_usm_alloc_info_v(IntEnum):
     BASE_PTR = 1                                    ## Memory allocation base pointer info
     SIZE = 2                                        ## Memory allocation size info
     DEVICE = 3                                      ## Memory allocation device info
+    POOL = 4                                        ## Memory allocation pool handle
```
PatKamin, line 846: Did you mean "pool info" in the comment?

igchor: Not really. Well, 'info' would be consistent with the other ones, but this option is actually used to obtain a 'handle' to the pool

pbalcer (approved):

pbalcer:
Please rebase.

igchor:
> Please rebase.

Done.

scripts/core/usm.yml
```diff
@@ -70,6 +70,8 @@ etors:
       desc: "Memory allocation size info"
     - name: DEVICE
       desc: "Memory allocation device info"
+    - name: POOL_HANDLE
```
kbenzie, line 73: The `DEVICE` is also a handle, `POOL` without `_HANDLE` would be consistent with the other property names.

kbenzie (approved):
LGTM - one comment

include/ur.py
kbenzie: I didn't notice this discussion and [commented](https://github.com/oneapi-src/unified-runtime/pull/348#discussion_r1137327335) again on this. I think the *device info* is a mistake and should be *device handle* in the docs.

scripts/core/usm.yml
igchor: Good point, I changed it back.

PR #358, opened 2023-03-15
[cmake] add quotes to cmd output, make it robust

yingcong-wu:
I encounter the problem when having local clang-format but it fails to run because of some missing library. Then in https://github.com/oneapi-src/unified-runtime/blob/main/CMakeLists.txt#L119 it calls https://github.com/oneapi-src/unified-runtime/blob/main/cmake/helpers.cmake#L9, which use the output of `clang-format --version`, which in my case is empty.

Then the empty output make `${cmd_ret}` empty, and without the quotes, it would be a missing argument to `STRING REGEX`(instead of a empty string with quotes).


kbenzie (approved):
Thanks for the fix @yingcong-wu

PR #359, opened 2023-03-15
[ur] Add separate host/device descriptors to 'urUSMSharedAlloc'

### LINKED ISSUE #266: Consider passing both host and device flags/descriptors to urUSMSharedAlloc
Currently we are passing a single flags/descriptor parameter to `urUSMSharedAlloc`:  
https://github.com/oneapi-src/unified-runtime/blob/61cb864b3a73918124bff11f9ad58dafb8c7769a/include/ur_api.h#L2814-L2821

LevelZero on accepts both host and device allocation descriptors:
https://github.com/oneapi-src/level-zero/blob/4ed13f327d3389285592edcf7598ec3cb2bc712e/include/ze_api.h#L4342-L4351

veselypeta:
Closes #266

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -231,9 +231,11 @@ typedef enum ur_structure_type_t {
     UR_STRUCTURE_TYPE_BUFFER_ALLOC_LOCATION_PROPERTIES = 5, ///< ::ur_buffer_alloc_location_properties_t
     UR_STRUCTURE_TYPE_PROGRAM_PROPERTIES = 6,               ///< ::ur_program_properties_t
     UR_STRUCTURE_TYPE_USM_DESC = 7,                         ///< ::ur_usm_desc_t
-    UR_STRUCTURE_TYPE_USM_POOL_DESC = 8,                    ///< ::ur_usm_pool_desc_t
-    UR_STRUCTURE_TYPE_USM_POOL_LIMITS_DESC = 9,             ///< ::ur_usm_pool_limits_desc_t
-    UR_STRUCTURE_TYPE_DEVICE_BINARY = 10,                   ///< ::ur_device_binary_t
+    UR_STRUCTURE_TYPE_USM_HOST_DESC = 8,                    ///< ::ur_usm_host_desc_t
+    UR_STRUCTURE_TYPE_USM_DEVICE_DESC = 9,                  ///< ::ur_usm_device_desc_t
+    UR_STRUCTURE_TYPE_USM_POOL_DESC = 10,                   ///< ::ur_usm_pool_desc_t
+    UR_STRUCTURE_TYPE_USM_POOL_LIMITS_DESC = 11,            ///< ::ur_usm_pool_limits_desc_t
+    UR_STRUCTURE_TYPE_DEVICE_BINARY = 12,                   ///< ::ur_device_binary_t
     /// @cond
     UR_STRUCTURE_TYPE_FORCE_UINT32 = 0x7fffffff
     /// @endcond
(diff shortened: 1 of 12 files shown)


scripts/core/usm.yml
```diff
@@ -255,9 +255,12 @@ params:
     - type: $x_device_handle_t
       name: hDevice
       desc: "[in] handle of the device object"
-    - type: $x_usm_desc_t*
-      name: pUSMDesc
-      desc: "[in][optional] USM memory allocation descriptor"
+    - type: const $x_usm_desc_t*
```
igchor, line 258: If we have separate descriptors for device and host I think we should also refactor `ur_usm_mem_flags_t`. RIght now we have `INITIAL_PLACEMENT_DEVICE` and `INITIAL_PLACEMENT_HOST`. Probably it would be enough to have 
 a single `INITIAL_PLACEMENT` flag - and the placement would be determined whether it is set as part of device or host descriptor (similarly as in L0).
 
 Also, I think some of the flags do not make sense for host:
- `UR_USM_MEM_FLAG_DEVICE_READ_ONLY`
- `UR_USM_MEM_FLAG_WRITE_COMBINED` (L0 does not allow specifying WC for host)

Perhaps we should have a separate type for host/device descriptors?

veselypeta: I like your suggestion. Perhaps this can be achieved by adding separate descriptors for host/device and linking them through the `pNext` parameter

veselypeta: I added a test [here](https://github.com/oneapi-src/unified-runtime/pull/359/files#diff-7edcb8a2de0724f7f91e69ae9a7105611b9344ec3f9d19f90c22f0a12b1a25ebR40) To show how this might work.

igchor: Yes, I think linking the descriptors is a good idea.

include/ur_api.h
```diff
@@ -2116,19 +2118,39 @@ urSamplerCreateWithNativeHandle(
 /// @brief USM memory property flags
 typedef uint32_t ur_usm_flags_t;
 typedef enum ur_usm_flag_t {
-    UR_USM_FLAG_BIAS_CACHED = UR_BIT(0),              ///< Allocation should be cached
-    UR_USM_FLAG_BIAS_UNCACHED = UR_BIT(1),            ///< Allocation should not be cached
-    UR_USM_FLAG_WRITE_COMBINED = UR_BIT(2),           ///< Memory should be allocated write-combined (WC)
-    UR_USM_FLAG_INITIAL_PLACEMENT_DEVICE = UR_BIT(3), ///< Optimize shared allocation for first access on the device
-    UR_USM_FLAG_INITIAL_PLACEMENT_HOST = UR_BIT(4),   ///< Optimize shared allocation for first access on the host
-    UR_USM_FLAG_DEVICE_READ_ONLY = UR_BIT(5),         ///< Memory is only possibly modified from the host, but read-only in all
```
jandres742, line 2124: @veselypeta : UR_USM_FLAG_DEVICE_READ_ONLY was moved to UR_USM_DEVICE_MEM_FLAG_DEVICE_READ_ONLY to ur_usm_device_mem_flags_t, which is in ur_usm_device_desc_t, but ur_usm_device_desc_t is not passed to urUSMDeviceAlloc, we are still passing there ur_usm_desc_t. Same goes for ur_usm_host_desc_t and urUSMHostAlloc.

is that a miss or I'm missing something?

include/ur_api.h
```diff
@@ -2345,12 +2383,12 @@ urUSMDeviceAlloc(
...
 urUSMSharedAlloc(
-    ur_context_handle_t hContext, ///< [in] handle of the context object
-    ur_device_handle_t hDevice,   ///< [in] handle of the device object
-    ur_usm_desc_t *pUSMDesc,      ///< [in][optional] USM memory allocation descriptor
-    ur_usm_pool_handle_t pool,    ///< [in][optional] Pointer to a pool created using urUSMPoolCreate
-    size_t size,                  ///< [in] size in bytes of the USM memory object to be allocated
-    void **ppMem                  ///< [out] pointer to USM shared memory object
+    ur_context_handle_t hContext,  ///< [in] handle of the context object
+    ur_device_handle_t hDevice,    ///< [in] handle of the device object
+    const ur_usm_desc_t *pUSMDesc, ///< [in][optional] Pointer to USM memory allocation descriptor.
```
jandres742, line 2388: @veselypeta : should we have here both, device and host descriptors?

include/ur_api.h
jandres742: or is that ur_usm_device_desc_t and ur_usm_host_desc_t are to be passed as pNext of ur_usm_desc_t?

veselypeta: Yes the idea is to add the actual descriptor with the `pNext` pointer. I included an example for the shared alloc variant [here](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/usm/urUSMSharedAlloc.cpp#L42)

include/ur_api.h
veselypeta: We decided to chain them instead of adding additional parameters to the function. See the example [here](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/usm/urUSMSharedAlloc.cpp#L42)

PR #361, opened 2023-03-15
[ur] Make creation of buffer objects extensible

### LINKED ISSUE #98: Reconsider urMemBufferCreate Signature
# Issue

The `urMemBufferCreate` and `piMemBufferCreate` have different signatures.

`urMemBufferCreate` has the following declaration:
```
urMemBufferCreate(
    ur_context_handle_t hContext,                   ///< [in] handle of the context object
    ur_mem_flags_t flags,                           ///< [in] allocation and usage information flags
    size_t size,                                    ///< [in] size in bytes of the memory object to be allocated
    void* pHost,                                    ///< [in] pointer to the buffer data
    ur_mem_handle_t* phBuffer                       ///< [out] pointer to handle of the memory buffer created
    );
```

`piMemBufferCreate` has the following declaration:
```
__SYCL_EXPORT pi_result piMemBufferCreate(
    pi_context context, pi_mem_flags flags, size_t size, void *host_ptr,
    pi_mem *ret_mem, const pi_mem_properties *properties = nullptr);
```

From the above we observe the follwowing differences:
1. `piMemBufferCreate` accepts a `properties` argument which is missing in `urMemBufferCreate`.

Examining the [usage of `piMemBufferCreate`](https://github.com/intel/llvm/blob/5b13d5b6d364917c8ef3a7b51ffd01c8be580095/sycl/source/detail/memory_manager.cpp#L150) in the DPC++ runtime we can see that it is passed a `properties` argument, so difference 1. is invalid.

# Task

Add a `properties` argument to `urMemBufferCreate` to meet the requirements of the DPC++ runtime.

kbenzie:
Enable extensibility of `urMemBufferCreate` and `urMemBufferPartition`
via `pNext` chained property structures. This also enables existing
functionality in PI relating to the `cl_intel_mem_channel_property` and
`cl_intel_mem_alloc_buffer_location` OpenCL extensions while also
enabling future extensions without ABI changes.

Fixes #98

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -223,12 +223,16 @@ typedef enum ur_result_t {
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Defines structure types
 typedef enum ur_structure_type_t {
-    UR_STRUCTURE_TYPE_CONTEXT_PROPERTIES = 0,   ///< ::ur_context_properties_t
-    UR_STRUCTURE_TYPE_IMAGE_DESC = 1,           ///< ::ur_image_desc_t
-    UR_STRUCTURE_TYPE_PROGRAM_PROPERTIES = 2,   ///< ::ur_program_properties_t
-    UR_STRUCTURE_TYPE_USM_DESC = 3,             ///< ::ur_usm_desc_t
-    UR_STRUCTURE_TYPE_USM_POOL_DESC = 4,        ///< ::ur_usm_pool_desc_t
-    UR_STRUCTURE_TYPE_USM_POOL_LIMITS_DESC = 5, ///< ::ur_usm_pool_limits_desc_t
+    UR_STRUCTURE_TYPE_CONTEXT_PROPERTIES = 0,               ///< ::ur_context_properties_t
+    UR_STRUCTURE_TYPE_IMAGE_DESC = 1,                       ///< ::ur_image_desc_t
+    UR_STRUCTURE_TYPE_BUFFER_PROPERTIES = 2,                ///< ::ur_buffer_properties_t
+    UR_STRUCTURE_TYPE_BUFFER_REGION = 3,                    ///< ::ur_buffer_region_t
+    UR_STRUCTURE_TYPE_BUFFER_CHANNEL_PROPERTIES = 4,        ///< ::ur_buffer_channel_properties_t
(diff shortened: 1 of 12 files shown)


kbenzie:
> Just for my understanding - is this how the API is intended to be used?
> 
> ```c++
>     ur_buffer_alloc_location_properties_t location_properties{
>         UR_STRUCTURE_TYPE_BUFFER_ALLOC_LOCATION_PROPERTIES, nullptr,
>         /* location */ 1};
>     ur_buffer_channel_properties_t chanel_properties{
>         UR_STRUCTURE_TYPE_BUFFER_CHANNEL_PROPERTIES, &location_properties,
>         /* channel */ 1};
>     ur_buffer_properties_t properties{UR_STRUCTURE_TYPE_BUFFER_PROPERTIES,
>                                       &chanel_properties, nullptr};
> 
>     urMemBufferCreate(context, UR_MEM_FLAG_READ_WRITE, 4096, &properties,
>                       &buffer);
> ```

If you're wanting to use all of the properties, yes.

scripts/core/memory.yml
```diff
@@ -238,10 +238,58 @@ returns:
...
+base: $x_base_properties_t
+members:
+    - type: uint32_t
+      name: channel
+      desc: >
+          [in] Identifies the channel/region to which the buffer should be
+          mapped.
+--- #--------------------------------------------------------------------------
+type: struct
+desc: "Buffer alloction location creation properties"
```
fabiomestre, line 270: allocation

include/ur_api.h
```diff
@@ -1506,9 +1510,60 @@ urMemImageCreate(
     ur_mem_handle_t *phMem                 ///< [out] pointer to handle of image object created
 );
 
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Buffer creation properties
+typedef struct ur_buffer_properties_t {
+    ur_structure_type_t stype; ///< [in] type of this structure, must be
+                               ///< ::UR_STRUCTURE_TYPE_BUFFER_PROPERTIES
+    void *pNext;               ///< [in,out][optional] pointer to extension-specific structure
+    void *pHost;               ///< [in][optional] pointer to the buffer data
```
fabiomestre, line 1519: pHost is a pointer to user allocated data. I think putting it inside a "properties" struct is slightly confusing. Generally properties are enums or something equivalent. 

I would personally keep it as a separate parameter. However, it could be that I'm biased by the way OpenCL implements this function.

fabiomestre (approved):

veselypeta (approved):
LGTM

PR #362, opened 2023-03-15
[ur] Unify naming for usm related types.

igchor:
Right now, we have types that start with: `usm`, `usm_mem` and `mem`.

I think we should also rename `ur_mem_advice_t` to have `usm` prefix but I'm not sure if we want `ur_usm_mem_advice_t` or `ur_usm_advice_t`. `usm_mem` duplicates 'memory' but we already have USMMemAdvise functions.

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -796,7 +796,7 @@ class ur_sampler_addressing_mode_t(c_int):
 
 ###############################################################################
 ## @brief USM memory property flags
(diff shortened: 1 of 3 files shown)


kbenzie (approved):
👍 for consistency

veselypeta (approved):

PatKamin (approved):

PR #367, opened 2023-03-17
[UR] Add devie type enumerations to urInit()

### LINKED ISSUE #204: Add missing enumartions for device types
In #200 we removed `ur_platform_init_flags_t` parameter from `urInit` since we handle platform discovery and selection through the loader. In fact the loader does nothing with either of these two input flags, which raises the question if `ur_device_init_flags_t`. 

We should determine if device flags actually have any use or purpose and if not remove this additional parameter.

veselypeta:
Closes #204


martygrant (approved):
LGTM

PR #375, opened 2023-03-27
[ur] Fix -ldl linker issue on common

veselypeta:
I was getting undefined references to `dlopen` and `dlclose`. This is due to the loader code now being moved into common which didn't list against `libdl`.


martygrant (approved):

pbalcer:
Now that libdl and others are integrated into glibc it's very easy to make this type of mistake...

PR #378, opened 2023-03-27
[ur] Fix part of logger tests not running properly + cleanup

PatKamin:
Fix FileSink logger tests checks not running. These checks detected failures in setting up logger with the `setenv()` function. This was due to the fact that the global logger object is a static singleton, therefore it's configuration can be provided only once per test binary run. Replaced global logger in these tests with local logger objects. New tests for validation of logger configured with the env var are not part of this PR.

Cleaned up logger and logger tests code a bit.

### DIFF
--- a/source/common/logger/ur_logger_details.hpp
+++ b/source/common/logger/ur_logger_details.hpp
@@ -27,14 +27,6 @@ class Logger {
         }
     }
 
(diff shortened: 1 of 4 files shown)


pbalcer (approved):
lgtm.

btw - splitting changes into multiple commits is good, but I think generally it would be best to consolidate similar changes to the same stuff in a  single one :P

PR #380, opened 2023-03-28
[CTS] Minor CTS fixes from CUDA adapter

veselypeta:
Some minor CTS fixes that fell out from testing against the CUDA adapter.


kbenzie (approved):
LGTM

aarongreig (approved):

PR #383, opened 2023-03-28
[CTS] Fix event tests to use queue with profiling enabled

### LINKED ISSUE #377: `urEventGetProfilingInfo` tests should use a queue which has profiling enabled.
`urEventGetProfilingInfo` entry-point should be tested with a queue that has profiling enabled. Currently this is not the case: `urQueueTest` and `urQueueTestWithParam<T>` contain queues that do not support profiling. The flag that is required is `UR_QUEUE_FLAG_PROFILING_ENABLE`

veselypeta:
The event testsuite has an entrypoint `urEventGetProfilingInfo` which requires a queue that has the `UR_QUEUE_FLAG_PROFILING_ENABLE` set when it's created.

Closes #377

### DIFF
--- a/test/conformance/testing/include/uur/fixtures.h
+++ b/test/conformance/testing/include/uur/fixtures.h
@@ -209,6 +209,45 @@ template <class T> struct urQueueTestWithParam : urContextTestWithParam<T> {
     ur_queue_handle_t queue;
 };
 
+struct urProfilingQueueTest : urContextTest {
+    void SetUp() override {
+        UUR_RETURN_ON_FATAL_FAILURE(urContextTest::SetUp());
+        ur_queue_property_t props[] = {UR_QUEUE_PROPERTIES_FLAGS,
+                                       UR_QUEUE_FLAG_PROFILING_ENABLE, 0};
+        ASSERT_SUCCESS(
+            urQueueCreate(this->context, this->device, props, &queue));
+    }
+
+    void TearDown() override {
+        if (queue) {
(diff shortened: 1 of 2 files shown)


test/conformance/testing/include/uur/fixtures.h
```diff
@@ -209,6 +209,43 @@ template <class T> struct urQueueTestWithParam : urContextTestWithParam<T> {
     ur_queue_handle_t queue;
 };
 
+struct urProfilingQueueTest : urContextTest {
+    void SetUp() override {
+        UUR_RETURN_ON_FATAL_FAILURE(urContextTest::SetUp());
+        ur_queue_property_t props[] = {UR_QUEUE_PROPERTIES_FLAGS,
+                                       UR_QUEUE_FLAG_PROFILING_ENABLE, 0};
+        urQueueCreate(this->context, this->device, props, &queue);
```
martygrant, line 217: Should there be an EXPECT_SUCCESS in both of the SetUp() methods for creating the queue?

kbenzie (approved):
LGTM

test/conformance/testing/include/uur/fixtures.h
kbenzie: Good spot, I think yes.

veselypeta: Thanks. Fixed

martygrant (approved):

PR #384, opened 2023-03-28
[ur] Make sampler creation extensible

### LINKED ISSUE #373: `ur_sampler_properties_t` enum should not start from 0
The `ur_sampler_properties_t` enum has values starting from 0. This isn't compatible with `urSamplerCreate`, which takes a 0-terminated list of property/value pairs. `UR_SAMPLER_PROPERTIES_NORMALIZED_COORDS` is interpreted as the end of the list.

Similar types that don't have this issue are `ur_device_partition_t` which takes its values from PI instead, and `ur_queue_properties_t` which starts from -1.

kbenzie:
Replace the `pProps` pointer to null-terminated list of key, value pairs
with a pointer `ur_sampler_desc_t` with the relevant required values as
data members. This also enables extensibility of `ur_sampler_handle_t`
creation. Fixes #373.


martygrant (approved):

veselypeta (approved):
LGTM

Issue #340, opened 2023-03-09
Add a ur-trace collector tool for use with xpti tracing layer 


pbalcer:
With #300, UR loader now has built-in tracing functionality that can be leveraged to programmatically intercept any UR call. This is flexible and can be used to accomplish many tasks. However, doing simple things like logging calls to stdout can be quite involved since it requires writing a custom xpti collector.

This task is to create a utility, similar to the ones found in SYCL (https://github.com/intel/llvm/tree/sycl/sycl/tools), that uses the xpti tracing to accomplish a couple of common tasks:

* pretty print function calls (including arguments and return values) to configurable output
* perform rudimentary profiling on functions
* ...

PR #391, opened 2023-03-29
add human-readable (partial) serialization for ur types

pbalcer:
UR defines `_params_t` structures for storing function arguments during tracing. Reading this struct programmatically is easy, but pretty printing it is not.

This patch implements stream insertion (<<) operators for param structs and all other ur types. This lets software easily (semi-)serialize function calls during tracing.

This doesn't support generic propreties (props, e.g., device info). The spec currently does not provide sufficient information to programmatically cast the `void *` prop pointers to the appropriate type specified by `*_info_t` enumerator value.

### DIFF
--- a/scripts/templates/params.hpp.mako
+++ b/scripts/templates/params.hpp.mako
@@ -1,0 +1,229 @@
+<%!
+import re
+from templates import helper as th
+%><%
+    n=namespace
+    N=n.upper()
+
+    x=tags['$x']
+    X=x.upper()
+%>/*
+ *
+ * Copyright (C) 2023 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * @file ${name}.hpp
+ *
+ */
+#ifndef ${X}_PARAMS_HPP
+#define ${X}_PARAMS_HPP 1
+
+#include "${x}_api.h"
+#include <ostream>
+#include <bitset>
+
+namespace ${x}_params {
+
+template <typename T> inline void serializePtr(std::ostream &os, T *ptr);
+
+<%def name="member(iname, itype, loop)">
+    %if iname == "pNext":
+        serializeStruct(os, ${caller.body()});
+    %elif th.type_traits.is_flags(itype):
+        serializeFlag_${itype}(os, ${caller.body()});
+    %elif not loop and th.type_traits.is_pointer(itype):
+        serializePtr(os, ${caller.body()});
+    %elif loop and th.type_traits.is_pointer_to_pointer(itype):
+        serializePtr(os, ${caller.body()});
+    %elif th.type_traits.is_handle(itype):
+        serializePtr(os, ${caller.body()});
+    %else:
+        os << ${caller.body()};
+    %endif
+</%def>
+
+<%def name="line(item, n, params)">
+    <%
+        iname = th._get_param_name(n, tags, item)
+        prefix = "p" if params else ""
+        pname = prefix + iname
+        itype = th._get_type_name(n, tags, obj, item)
+        access = "->" if params else "."
+        deref = "*" if params else ""
+    %>
+    %if n != 0:
+        os << ", ";
+    %endif
+    ## can't iterate over 'void *'...
+    %if th.param_traits.is_range(item) and "void*" not in itype:
+        os << ".${iname} = [";
+        for (size_t i = ${th.param_traits.range_start(item)}; ${deref}(params${access}${pname}) != NULL && i < ${deref}params${access}${prefix + th.param_traits.range_end(item)}; ++i) {
+            if (i != 0) {
+                os << ", ";
+            }
+            <%call expr="member(iname, itype, True)">
+                (${deref}(params${access}${pname}))[i]
+            </%call>
+        }
+        os << "]";
+    %else:
+        os << ".${iname} = ";
+        <%call expr="member(iname, itype, False)">
+            ${deref}(params${access}${pname})
+        </%call>
+    %endif
+</%def>
+
+%for spec in specs:
+%for obj in spec['objects']:
+## ENUM #######################################################################
+%if re.match(r"enum", obj['type']):
+    %if "api_version" in obj['name']:
+    inline std::ostream &operator<<(std::ostream &os, enum ${th.make_enum_name(n, tags, obj)} value) {
+        os << UR_MAJOR_VERSION(value) << "." << UR_MINOR_VERSION(value);
+        return os;
+    }
+    %else:
+    inline std::ostream &operator<<(std::ostream &os, enum ${th.make_enum_name(n, tags, obj)} value) {
+        switch (value) {
+            %for n, item in enumerate(obj['etors']):
+                <%
+                ename = th.make_etor_name(n, tags, obj['name'], item['name'])
+                %>
+                case ${ename}:
+                    os << "${ename}";
+                    break;
+            %endfor
+                default:
+                    os << "unknown enumerator";
+                    break;
+        }
+        return os;
+    }
+    %endif
+    %if "structure_type" in obj['name']:
+    inline void serializeStruct(std::ostream &os, const void *ptr) {
+        if (ptr == NULL) {
+            serializePtr(os, ptr);
+            return;
+        }
+
+        ## structure type enum value must be first
+        enum ${th.make_enum_name(n, tags, obj)} *value = (enum ${th.make_enum_name(n, tags, obj)} *)ptr;
+        switch (*value) {
+            %for n, item in enumerate(obj['etors']):
+                <%
+                ename = th.make_etor_name(n, tags, obj['name'], item['name'])
+                %>
+                case ${ename}: {
+                    const ${th.subt(n, tags, item['desc'])} *pstruct = (const ${th.subt(n, tags, item['desc'])} *)ptr;
+                    serializePtr(os, pstruct);
+                } break;
+            %endfor
+                default:
+                    os << "unknown enumerator";
+                    break;
+        }
+    }
+    %endif
+%if th.type_traits.is_flags(obj['name']):
+inline void serializeFlag_${th.make_type_name(n, tags, obj)}(std::ostream &os, ${th.make_type_name(n, tags, obj)} flag) {
+    uint32_t val = flag;
+    bool first = true;
+    %for n, item in enumerate(obj['etors']):
+        <%
+        ename = th.make_etor_name(n, tags, obj['name'], item['name'])
+        %>
+        if ((val & ${ename}) == (uint32_t)${ename}) {
+            ## toggle the bits to avoid printing overlapping values
+            ## instead of e.g., FLAG_FOO | FLAG_BAR | FLAG_ALL, this will just
+            ## print FLAG_FOO | FLAG_BAR (or just FLAG_ALL, depending on order).
+            val ^= (uint32_t)${ename};
+            if (!first) {
+                os << " | ";
+            } else {
+                first = false;
+            }
+            os << ${ename};
+        }
+    %endfor
+    if (val != 0) {
+        std::bitset<32> bits(val);
+        if (!first) {
+            os << " | ";
+        }
+        os << "unknown bit flags " << bits;
+    } else if (first) {
+        os << "0";
+    }
+}
+%endif
+## STRUCT/UNION ###############################################################
+%elif re.match(r"struct|union", obj['type']):
+inline std::ostream &operator<<(std::ostream &os, const ${obj['type']} ${th.make_type_name(n, tags, obj)} params) {
+    os << "(${obj['type']} ${th.make_type_name(n, tags, obj)}){";
+    <% mlist = obj['members'] %>
+    %for n, item in enumerate(mlist):
+        ${line(item, n, False)}
+    %endfor
+    os << "}";
+    return os;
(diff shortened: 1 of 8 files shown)


pbalcer:
The first part of #340. 

Example output:
```
function_with_args_begin(1) - urInit(.device_flags = UR_DEVICE_INIT_FLAG_GPU | UR_DEVICE_INIT_FLAG_CPU);
function_with_args_end(1) - urInit(.device_flags = UR_DEVICE_INIT_FLAG_GPU | UR_DEVICE_INIT_FLAG_CPU) -> UR_RESULT_SUCCESS;
Platform initialized.
function_with_args_begin(2) - urPlatformGet(.NumEntries = 1, .phPlatforms = [], .pNumPlatforms = 0x7ffe5796f200 (0));
function_with_args_end(2) - urPlatformGet(.NumEntries = 1, .phPlatforms = [], .pNumPlatforms = 0x7ffe5796f200 (1)) -> UR_RESULT_SUCCESS;
function_with_args_begin(3) - urPlatformGet(.NumEntries = 1, .phPlatforms = [nullptr], .pNumPlatforms = nullptr);
function_with_args_end(3) - urPlatformGet(.NumEntries = 1, .phPlatforms = [0x80800001], .pNumPlatforms = nullptr) -> UR_RESULT_SUCCESS;
function_with_args_begin(4) - urPlatformGetApiVersion(.hDriver = 0x80800001, .pVersion = 0x7ffe5796f210 (0.0));
function_with_args_end(4) - urPlatformGetApiVersion(.hDriver = 0x80800001, .pVersion = 0x7ffe5796f210 (0.6)) -> UR_RESULT_SUCCESS;
API version: 0.6
function_with_args_begin(5) - urDeviceGet(.hPlatform = 0x80800001, .DeviceType = UR_DEVICE_TYPE_GPU, .NumEntries = 0, .phDevices = [], .pNumDevices = 0x7ffe5796f220 (0));
function_with_args_end(5) - urDeviceGet(.hPlatform = 0x80800001, .DeviceType = UR_DEVICE_TYPE_GPU, .NumEntries = 0, .phDevices = [], .pNumDevices = 0x7ffe5796f220 (1)) -> UR_RESULT_SUCCESS;
function_with_args_begin(6) - urDeviceGet(.hPlatform = 0x80800001, .DeviceType = UR_DEVICE_TYPE_GPU, .NumEntries = 1, .phDevices = [nullptr], .pNumDevices = nullptr);
function_with_args_end(6) - urDeviceGet(.hPlatform = 0x80800001, .DeviceType = UR_DEVICE_TYPE_GPU, .NumEntries = 1, .phDevices = [0x80800002], .pNumDevices = nullptr) -> UR_RESULT_SUCCESS;
function_with_args_begin(7) - urDeviceGetInfo(.hDevice = 0x80800002, .propName = UR_DEVICE_INFO_TYPE, .propSize = 4, .pPropValue = 0x7ffe5796f230, .pPropSizeRet = nullptr);
function_with_args_end(7) - urDeviceGetInfo(.hDevice = 0x80800002, .propName = UR_DEVICE_INFO_TYPE, .propSize = 4, .pPropValue = 0x7ffe5796f230, .pPropSizeRet = nullptr) -> UR_RESULT_SUCCESS;
function_with_args_begin(8) - urDeviceGetInfo(.hDevice = 0x80800002, .propName = UR_DEVICE_INFO_NAME, .propSize = 1023, .pPropValue = 0x7ffe5796f340, .pPropSizeRet = nullptr);
function_with_args_end(8) - urDeviceGetInfo(.hDevice = 0x80800002, .propName = UR_DEVICE_INFO_NAME, .propSize = 1023, .pPropValue = 0x7ffe5796f340, .pPropSizeRet = nullptr) -> UR_RESULT_SUCCESS;
Found a Null Device gpu.
function_with_args_begin(9) - urTearDown(.pParams = nullptr);
function_with_args_end(9) - urTearDown(.pParams = nullptr) -> UR_RESULT_SUCCESS;
```

test/unit/utils/params.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+template <typename T>
+std::unique_ptr<WrappedParams<typename T::value_type>> createParams();
+
+template <typename T> class ParamsTest : public testing::Test {
+  protected:
+    ParamsTest() : params(createParams<T>()) {}
+    ~ParamsTest() override {}
+
+    std::unique_ptr<WrappedParams<typename T::value_type>> params;
```
igchor, line 27: It might make sense to have a non-template base class so that you don't have to use TYPED_TEST_SUITE but I don't see any benefit in using a template base class (WrappedParams).

Why not just store `T` directly here and in tests, just call `params.get_struct()`/`params.get_expected()`?

source/common/ur_params.hpp
```diff
@@ -0,0 +1,7780 @@
...
+#define UR_PARAMS_HPP 1
+
+#include "ur_api.h"
+#include <ostream>
+
+namespace ur_params {
+
+template <typename T> inline void serializePtr(std::ostream &os, T *ptr);
+
+inline std::ostream &operator<<(std::ostream &os,
```
igchor, line 20: Do you need to add those operators for handles at all? The default implementation for ptrs should work pretty well: it would just print `0` instead of nullptr.

source/common/ur_params.hpp
```diff
@@ -0,0 +1,7780 @@
...
+template <typename T> inline void serializePtr(std::ostream &os, T *ptr) {
+    if (ptr == nullptr) {
+        os << "nullptr";
+    } else if constexpr (std::is_same_v<T, void> ||
+                         std::is_same_v<T, const void>) {
+        os << ptr;
+    } else if constexpr (std::is_same_v<T, void *> ||
+                         std::is_same_v<T, const void *>) {
+        os << (void *)(ptr) << " (";
+        if (*ptr == nullptr) {
```
igchor, line 7404: nit: just call `serializePtr(os, *ptr)`, you can then even support more indirection levels.

source/common/ur_params.hpp
```diff
@@ -0,0 +1,7780 @@
...
+
+    serializePtr(os, *(params->ppHostTimestamp));
+
+    return os;
+}
+
+template <typename T> inline void serializePtr(std::ostream &os, T *ptr) {
+    if (ptr == nullptr) {
+        os << "nullptr";
+    } else if constexpr (std::is_same_v<T, void> ||
```
igchor, line 7398: What if it's volatile ptr? Probably it never is, but it's safer to use https://en.cppreference.com/w/cpp/types/remove_cv and then compare. Or you might use https://en.cppreference.com/w/cpp/types/decay if you want to support arrays as well.

source/common/ur_params.hpp
```diff
@@ -0,0 +1,7780 @@
...
+        break;
+    }
+    return os;
+}
+inline void serializeFlag_ur_queue_flags_t(std::ostream &os,
+                                           ur_queue_flags_t flag) {
+    uint32_t val = flag;
+    bool first = true;
+
+    if ((val & UR_QUEUE_FLAG_OUT_OF_ORDER_EXEC_MODE_ENABLE) ==
```
igchor, line 3239: can this flag cehcking code be made generic? It's seems like a simple loop should be enough

source/common/ur_params.hpp
```diff
@@ -0,0 +1,7780 @@
...
+    os << (params.channel);
+
+    os << "}";
+    return os;
+}
+inline std::ostream &
+operator<<(std::ostream &os,
+           const struct ur_buffer_alloc_location_properties_t params) {
+    os << "(struct ur_buffer_alloc_location_properties_t){";
+
```
igchor, line 2245: Can we get rid of those newlines (or ideally, put all `os << ...` in a single line) to make this file smaller?

scripts/templates/params.hpp.mako
```diff
@@ -0,0 +1,233 @@
...
+    %for n, item in enumerate(obj['params']):
+        ${line(item, n, True)}
+    %endfor
+    return os;
+}
+
+%endfor
+%endfor
+
+template <typename T> inline void serializePtr(std::ostream &os, T *ptr) {
```
igchor, line 196: Do we need to have those helper functions in `mako`? I think it would be better to put generic functions directly to some .hpp file

test/unit/utils/params.cpp
pbalcer: I'm not sure what you mean exactly. I could store just T here (I'd also need the expected string, but that's a minor detail), but that wouldn't help with getting rid of TYPED_TEST_SUITE.
Theoretically, I could change it so that I just store strings here, and that would let me remove TYPED_TEST_SUITE, but I'm not sure how that is better.

source/common/ur_params.hpp
pbalcer: Over bits? But that would require us to cast it back to the appropriate enum type if we want to actually print useful information. And how do you know when to stop the loop? Would we iterate over all the bits in a type?

I don't know. That doesn't seem simpler to me.

source/common/ur_params.hpp
pbalcer: I prioritized mako template readability over cleanliness of the output file. It gets formatted afterward anyway.

I found one newline to remove from this part of the script.

source/common/ur_params.hpp
pbalcer: I refactored this function a bit but had to use some magicks. Take a look ;)

source/common/ur_params.hpp
```diff
@@ -0,0 +1,7779 @@
...
+    case UR_RESULT_ERROR_ADAPTER_SPECIFIC:
+        os << "UR_RESULT_ERROR_ADAPTER_SPECIFIC";
+        break;
+
+    case UR_RESULT_ERROR_UNKNOWN:
+        os << "UR_RESULT_ERROR_UNKNOWN";
+        break;
+    default:
+        os << (value == 0 ? "0" : "invalid enumerator");
```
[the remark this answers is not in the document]
pbalcer: Not necessarily. E.g., `flag_t` enums don't have 0 values. And you can force an enum parameter to accept any value if you cast it.
But overall, I think I agree. I changed it all to "unknown".

source/common/ur_params.hpp
igchor: Pointers should be printed as 0x... anyway I think, but if you want to have 'nullptr' that's fine

source/common/ur_params.hpp
igchor: I mean you could just generate a list of values in mako using `%for n, item in enumerate(obj['etors']):` and then just have a for loop here over those values, but if it's easier to implement in mako this way then it's fine.

source/common/ur_params.hpp
igchor: Looks, good.

source/common/ur_params.hpp
pbalcer: With `is_type_complete_v` I was able to remove operators for handles, and instead use serializePtr.

source/common/ur_params.hpp
pbalcer: so something like:
```
parseFlagsGeneric<T>(T flag, T all_flags[]) {
   for (f : flags) { 
     if ((flag & f) == f) os << f;
   }
}

parseFlags_foo(enum foo_flag flag) {
   enum foo_flag all_etors[] = {FOO_A, FOO_B};
   parseFlagsGeneric(flag, all_etors);
}
```
(or without the template function)

I guess either method will work, I don't think one is better than the other. It's all autogenerated anyway.

test/unit/utils/params.cpp
```diff
@@ -0,0 +1,240 @@
...
+typedef Types<UrInitParamsNoFlags, UrInitParamsInvalidFlags,
+              UrUsmHostAllocParamsEmpty, UrPlatformGetEmptyArray,
+              UrPlatformGetTwoPlatforms, UrUsmHostAllocParamsUsmDesc,
+              UrUsmHostAllocParamsHostDesc>
+    Implementations;
+
+using ::testing::MatchesRegex;
+using namespace ur_params;
+
+TYPED_TEST_SUITE(ParamsTest, Implementations, );
```
PatKamin, line 224: nit: A comma at the end and a whitespace is here on purpose?

pbalcer: yes, otherwise there's a warning about not enough macro arguments.

test/unit/utils/params.cpp
pbalcer: I simplified the tests.

test/unit/utils/params.cpp
```diff
@@ -0,0 +1,192 @@
...
+template <typename T> std::unique_ptr<T> createParams() {
+    return std::make_unique<T>();
+}
+
+template <typename T> class ParamsTest : public testing::Test {
+  protected:
+    ParamsTest() : params(createParams<T>()) {}
+    ~ParamsTest() override {}
+
+    std::unique_ptr<T> params;
```
igchor, line 21: Can't we just store `T` directly, instead of using a unique_ptr?

igchor (approved):
LGTM

test/unit/utils/params.cpp
pbalcer: No, I need T to stay at the same address because I'm using & to an internal variable in a constructor.

PR #393, opened 2023-03-29
Make the YAML details: section more flexible

### LINKED ISSUE #366: Make the YAML `details:` section more flexible
Currently the generator scripts expect the `details:` to be a list of strings:

```yaml
details:
    - "If the event list is empty, it waits for all previously enqueued commands to complete."
    - "It returns an event which can be waited on."
```

Which then gets output as a bullet list in the C++ source as a Doxygen `@details` section:

```cpp
/// @details
///     - If the event list is empty, it waits for all previously enqueued
///       commands to complete.
///     - It returns an event which can be waited on.
```

Given that Doxygen supports Markdown formatting limiting the content of the `details:` section to a bullet list is quite restrictive, esepcially when attempting to describe a more complex concept or behaivour.

Instead, we could take advantage of the Markdown support in Doxygen and change the generator script requirements of `details:` to accept either:

1. A list of strings to maintain the current behaviour.
2. A multi-line string to be inserted verbatim, allowing the use of Markdown syntax for richer documentation.

An example of 2. :

```yaml
details: |
    Below the equivalent bullet list but in Markdown format.

    - If the event list is empty, it waits for all previously enqueued commands
      to complete.
    - It returns an event which can be waited on.

    Additionally, making use of the ability to have separate paragraphs to help
    explain the entity, and even make use of fenced code-blocks containing code
    examples:

    ```cpp
    void foo() {}
    ```
```

Which could then be output as the following Doxygen comment:

```cpp
/// @details
/// Below the equivalent bullet list but in Markdown format.
///
/// - If the event list is empty, it waits for all previously enqueued commands
///   to complete.
/// - It returns an event which can be waited on.
///
/// Additionally, making use of the ability to have separate paragraphs to help
/// explain the entity, and even make use of fenced code-blocks containing code
/// examples:
///
/// ```cpp
/// void foo() {}
/// ```
```

fabiomestre:
Closes #366


martygrant (approved):
LGTM

veselypeta (approved):
LGTM

scripts/templates/helper.py
```diff
@@ -780,23 +780,30 @@ def make_details_lines(namespace, tags, obj):
...
+                            for line in split_line(subt(namespace, tags, val, True), 66):
+                                lines.append("%s %s"%(prologue, line))
+                                prologue = "         "
+                else:
+                    prologue = "    -"
+                    for line in split_line(subt(namespace, tags, item, True), 70):
                         lines.append("%s %s"%(prologue, line))
                         prologue = "     "
+        # If obj['details'] is a string, then assume that it's already correctly formatted using markdown.
+        else:
```
kbenzie, line 803: I feel like we should actually check `obj['details']` is an instance of a string and raise a `TypeError` when its not. That way we have a chance to explain to the user what's expected.

kbenzie (approved):
LGTM, one comment

scripts/templates/helper.py
fabiomestre: That check is being done in the function __validate_details() on line 175. It was missing a couple of parentheses though which I have added in the latest commit.

kbenzie: ah okay 👍

PR #395, opened 2023-03-30
[UR] Generate a bit mask for `flags_t` parameter validation

veselypeta:
The API generator has the ability to check that flags and enums are within range, however this does not apply to `[optional]` parameters. This then means that adapters will have to check if the value is within range. It would be great if we report this range as part of the API specification and avoid having the plugins compute what this value is - which could become outdated as additional enumerations are added/removed and is easily missed.

The adapter can then do the following:
```cpp
if(flags & UR_<ENUM>_MASK){
   return UR_RESULT_ERROR_INVALID_ENUMERATION;
}
```

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -219,6 +219,7 @@ typedef enum ur_result_t {
     /// @endcond
 
 } ur_result_t;
+#define UR_RESULT_MAX_VALUE UR_RESULT_ERROR_UNKNOWN
 
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Defines structure types
@@ -242,6 +243,7 @@ typedef enum ur_structure_type_t {
     /// @endcond
 
 } ur_structure_type_t;
+#define UR_STRUCTURE_TYPE_MAX_VALUE UR_STRUCTURE_TYPE_SAMPLER_DESC
 
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Base for all properties types
@@ -298,6 +300,7 @@ typedef enum ur_device_init_flag_t {
     /// @endcond
 
 } ur_device_init_flag_t;
+#define UR_DEVICE_INIT_FLAGS_MAX_VALUE 0x1f
 
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Initialize the 'oneAPI' driver(s)
@@ -399,6 +402,7 @@ typedef enum ur_platform_info_t {
     /// @endcond
 
 } ur_platform_info_t;
+#define UR_PLATFORM_INFO_MAX_VALUE UR_PLATFORM_INFO_PROFILE
 
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Retrieves various information about platform
@@ -445,6 +449,7 @@ typedef enum ur_api_version_t {
     /// @endcond
 
 } ur_api_version_t;
+#define UR_API_VERSION_MAX_VALUE UR_API_VERSION_CURRENT
 
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Returns the API version supported by the specified platform
@@ -642,6 +647,7 @@ typedef enum ur_device_type_t {
     /// @endcond
 
 } ur_device_type_t;
+#define UR_DEVICE_TYPE_MAX_VALUE UR_DEVICE_TYPE_VPU
 
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Retrieves devices within a platform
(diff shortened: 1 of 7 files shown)


include/ur_api.h
```diff
@@ -219,6 +219,7 @@ typedef enum ur_result_t {
     /// @endcond
 
 } ur_result_t;
+#define UR_RESULT_MAX_VALUE UR_RESULT_ERROR_UNKNOWN
```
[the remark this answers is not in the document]
pbalcer: This specific enum has `UR_RESULT_ERROR_UNKNOWN = 0x7ffffffe,` (which is much larger than the actual last value of `UR_RESULT_ERROR_ADAPTER_SPECIFIC = 67,`. This means that `if (val > UR_RESULT_ERROR_UNKNOWN) {...}` won't actually catch all invalid cases.

Not sure if this is much of a problem for `ur_result_t`, but other enums might have the same issue. Perhaps it would be better to use 0 (or just generally, the beginning of the enum) for special values like `UR_RESULT_ERROR_UNKNOWN` ?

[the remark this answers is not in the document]
veselypeta: tbh I want this mostly just for `flags_t` enumerations - I extended it a bit so that it can support an arbitrary enum, but as you say doesn't work that well if the enum has gaps.

include/ur_api.h
```diff
@@ -1043,6 +1054,7 @@ typedef enum ur_device_exec_capability_flag_t {
     /// @endcond
 
 } ur_device_exec_capability_flag_t;
+#define UR_DEVICE_EXEC_CAPABILITY_FLAGS_MAX_VALUE 0x3
```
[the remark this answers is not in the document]
veselypeta: You're right, there is no such check. I think then considering a bitmask and having only from flag_t values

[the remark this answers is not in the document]
pbalcer: Alternatively, we could add a check to the validation layer to verify if the flags (and other enum types) are correct. On the other hand, having a valid flags value might be useful more broadly.

scripts/parse_specs.py
```diff
@@ -529,15 +530,20 @@ def _generate_meta(d, ordinal, meta):
         if 'enum' == type:
             value = -1
             max_value = -1
+            bit_mask = 0
             meta[type][name]['etors'] = []
             for idx, etor in enumerate(d['etors']):
                 meta[type][name]['etors'].append(etor['name'])
                 value = _get_etor_value(etor.get('value'), value)
+                if type_traits.is_flags(name):
+                    bit_mask |= value
```
veselypeta, line 539: This will generate a bit mask for all flags defined in the spec and will handle if there are gaps.

pbalcer (approved):

include/ur_api.h
```diff
@@ -298,6 +298,7 @@ typedef enum ur_device_init_flag_t {
     /// @endcond
 
 } ur_device_init_flag_t;
+#define UR_DEVICE_INIT_FLAGS_MASK 0xffffffe0
```
kbenzie, line 301: I'd like to see some documentation about what these are for?

scripts/templates/api.h.mako
```diff
@@ -92,6 +92,9 @@ typedef enum ${th.make_enum_name(n, tags, obj)}
     %endfor
 
 } ${th.make_enum_name(n, tags, obj)};
+%if th.type_traits.is_flags(obj['name']):
+${th.make_flags_bitmask(n, tags, obj, meta)}
```
kbenzie, line 96: I guess this is where the tempate documentation for the flags would go.

source/loader/ur_libapi.cpp
```diff
@@ -35,7 +35,7 @@ extern "C" {
 ///     - ::UR_RESULT_ERROR_UNINITIALIZED
 ///     - ::UR_RESULT_ERROR_DEVICE_LOST
 ///     - ::UR_RESULT_ERROR_INVALID_ENUMERATION
-///         + `0x1f < device_flags`
+///         + `0xffffffe0 & device_flags`
```
kbenzie, line 38: Shouldn't this be using the name of the mask macro?

kbenzie (approved):
LGTM, two things

martygrant (approved):

include/ur_api.h
veselypeta: 👍

scripts/templates/api.h.mako
veselypeta: 👍

source/loader/ur_libapi.cpp
veselypeta: Fixed. It's the same value, but makes it more readable. 👍

PR #398, opened 2023-03-31
[UMA] Add umaMemoryProviderSetAttributes to set madvise-like attributes

ldorau:
Add umaMemoryProviderSetAttributes to set madvise-like attributes of the memory space pointed by ptr from the memory provider.

### DIFF
--- a/source/common/unified_memory_allocation/include/uma/memory_provider.h
+++ b/source/common/unified_memory_allocation/include/uma/memory_provider.h
@@ -12,6 +12,10 @@
 #include <uma/base.h>
 #include <uma/memory_provider_ops.h>
 
+#define UMA_ATTR_DONTNEED 4    /* Don't need these pages.  */
+#define UMA_ATTR_HUGEPAGE 14   /* Worth backing with hugepages.  */
+#define UMA_ATTR_NOHUGEPAGE 15 /* Not worth backing with hugepages.  */
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -56,6 +60,17 @@ enum uma_result_t umaMemoryProviderAlloc(uma_memory_provider_handle_t hProvider,
 ///
 enum uma_result_t umaMemoryProviderFree(uma_memory_provider_handle_t hProvider,
                                         void *ptr, size_t size);
+
+///
+/// \brief Set attributes of the memory space pointed by ptr from the memory provider
+/// \param hProvider handle to the memory provider
+/// \param ptr pointer to the allocated memory
+/// \param size size of the allocation
+/// \param attrs attributes of the memory space to be set
+///
+enum uma_result_t
+umaMemoryProviderSetAttributes(uma_memory_provider_handle_t hProvider,
+                               void *ptr, size_t size, int attrs);
 
 ///
 /// \brief Retrieve string representation of the underlying provider specific
--- a/source/common/unified_memory_allocation/include/uma/memory_provider_ops.h
+++ b/source/common/unified_memory_allocation/include/uma/memory_provider_ops.h
@@ -40,6 +40,8 @@ struct uma_memory_provider_ops_t {
     enum uma_result_t (*alloc)(void *provider, size_t size, size_t alignment,
                                void **ptr);
     enum uma_result_t (*free)(void *provider, void *ptr, size_t size);
+    enum uma_result_t (*set_attrs)(void *provider, void *ptr, size_t size,
+                                   int attrs);
     enum uma_result_t (*get_last_result)(void *provider,
                                          const char **ppMessage);
 };
--- a/source/common/unified_memory_allocation/src/memory_provider.c
+++ b/source/common/unified_memory_allocation/src/memory_provider.c
@@ -60,6 +60,12 @@ enum uma_result_t umaMemoryProviderFree(uma_memory_provider_handle_t hProvider,
 }
 
 enum uma_result_t
+umaMemoryProviderSetAttributes(uma_memory_provider_handle_t hProvider,
+                               void *ptr, size_t size, int attrs) {
+    return hProvider->ops.set_attrs(hProvider->provider_priv, ptr, size, attrs);
+}
+
+enum uma_result_t
 umaMemoryProviderGetLastResult(uma_memory_provider_handle_t hProvider,
                                const char **ppMessage) {
     return hProvider->ops.get_last_result(hProvider->provider_priv, ppMessage);
--- a/test/unified_memory_allocation/common/provider.c
+++ b/test/unified_memory_allocation/common/provider.c
@@ -32,6 +32,15 @@ static enum uma_result_t nullFree(void *provider, void *ptr, size_t size) {
     return UMA_RESULT_SUCCESS;
 }
 
+static enum uma_result_t nullSetAttributes(void *provider, void *ptr,
+                                           size_t size, int attrs) {
+    (void)provider;
+    (void)ptr;
+    (void)size;
+    (void)attrs;
+    return UMA_RESULT_SUCCESS;
+}
+
 static enum uma_result_t nullGetLastResult(void *provider, const char **ppMsg) {
     (void)provider;
     (void)ppMsg;
@@ -44,6 +53,7 @@ uma_memory_provider_handle_t nullProviderCreate(void) {
                                             .finalize = nullFinalize,
                                             .alloc = nullAlloc,
                                             .free = nullFree,
+                                            .set_attrs = nullSetAttributes,
                                             .get_last_result =
                                                 nullGetLastResult};
 
@@ -87,6 +97,15 @@ static enum uma_result_t traceFree(void *provider, void *ptr, size_t size) {
     return umaMemoryProviderFree(traceProvider->hUpstreamProvider, ptr, size);
 }
 
+static enum uma_result_t traceSetAttributes(void *provider, void *ptr,
+                                            size_t size, int attrs) {
+    struct traceParams *traceProvider = (struct traceParams *)provider;
+
+    traceProvider->trace("set_attrs");
+    return umaMemoryProviderSetAttributes(traceProvider->hUpstreamProvider, ptr,
+                                          size, attrs);
+}
+
 static enum uma_result_t traceGetLastResult(void *provider,
                                             const char **ppMsg) {
     struct traceParams *traceProvider = (struct traceParams *)provider;
@@ -104,6 +123,7 @@ traceProviderCreate(uma_memory_provider_handle_t hUpstreamProvider,
                                             .finalize = traceFinalize,
                                             .alloc = traceAlloc,
                                             .free = traceFree,
+                                            .set_attrs = traceSetAttributes,
                                             .get_last_result =
                                                 traceGetLastResult};
 
--- a/test/unified_memory_allocation/memoryProviderAPI.cpp
+++ b/test/unified_memory_allocation/memoryProviderAPI.cpp
@@ -31,6 +31,11 @@ TEST_F(test, memoryProviderTrace) {
     ASSERT_EQ(calls["free"], 1);
     ASSERT_EQ(calls.size(), ++call_count);
 
+    ret = umaMemoryProviderSetAttributes(tracingProvider.get(), nullptr, 0, 0);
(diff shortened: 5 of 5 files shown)


source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -82,6 +82,17 @@ enum uma_result_t
...
+///
+/// \brief Set attributes of the memory space pointed by ptr from the memory provider
+/// \param hProvider handle to the memory provider
+/// \param ptr pointer to the allocated memory
+/// \param size size of the allocation
+/// \param attrs attributes of the memory space to be set
+///
+enum uma_result_t
+umaMemoryProviderSetAttributes(uma_memory_provider_handle_t hProvider,
+                                        void *ptr, size_t size, int attrs);
```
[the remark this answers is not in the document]
ldorau: Like that?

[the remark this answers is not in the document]
pbalcer: yes, but we will probably need to narrow down the list. for now it's ok.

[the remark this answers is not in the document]
igchor: Those flags are mostly useful for host memory I belive. Do we want to figure out some possible intersection with USM (and move host-specific/USM-specific flags to appropriate memory provider implementations)? Right now there's not much for [USM](https://github.com/oneapi-src/unified-runtime/blob/dfb0d7541ec5853bfa79368782e48b025b0b253c/include/ur_api.h#L2204)

igchor:
We should probably also expose `protection` flags (none/read/write/rw) - possibly through a separate function.

source/common/unified_memory_allocation/include/uma/memory_provider.h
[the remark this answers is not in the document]
ldorau: A predefined set of attributes added.

Done.

ldorau:
> We should probably also expose `protection` flags (none/read/write/rw) - possibly through a separate function.

Protection flags can be passed via the `params` argument of `umaMemoryProviderCreate`. See https://github.com/intel-innersource/libraries.memory.allocators.qkmalloc/pull/44 for the implementation.

pbalcer:
> > We should probably also expose `protection` flags (none/read/write/rw) - possibly through a separate function.
> 
> Protection flags can be passed via the `params` argument of `umaMemoryProviderCreate`...

Technically the same is true for advise arguments ;) I think we need a dedicated function for both. But that can be done separately. I think the biggest problem here is making this useful enough so that a reasonable implementation can use it without making assumptions about underlying implementation (whether it's on windows or linux, CPU or GPU memory)

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -12,6 +12,31 @@
...
+                                      overrides the coredump filter bits.  */
+#define UMA_ATTR_DODUMP      17    /* Clear the UMA_ATTR_DONTDUMP flag.  */
+#define UMA_ATTR_WIPEONFORK  18    /* Zero memory on fork, child only.  */
+#define UMA_ATTR_KEEPONFORK  19    /* Undo UMA_ATTR_WIPEONFORK.  */
+#define UMA_ATTR_COLD        20    /* Deactivate these pages.  */
+#define UMA_ATTR_PAGEOUT     21    /* Reclaim these pages.  */
+#define UMA_ATTR_POPULATE_READ 22  /* Populate (prefault) page tables
+                                      readable.  */
+#define UMA_ATTR_POPULATE_WRITE 23 /* Populate (prefault) page tables
+                                      writable.  */
```
kswiecicki, line 38: I'm not sure about the intended use of those attributes, but assuming some of the could be used in conjunction we should define those using bit shift `(1 << #)`.

kswiecicki: Also, there's a gap between `4` and `8`.

ldorau: > I'm not sure about the intended use of those attributes, but assuming some of the could be used in conjunction we should define those using bit shift `(1 << #)`.

This assumption is wrong - they will not be used in conjunction.
They are defined exactly like the `MADV_*` flags:
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/mman-linux.h;h=6771f07a0f9a617fcde0eb264d54784788d22e76;hb=HEAD#l68

igchor:
> > > We should probably also expose `protection` flags (none/read/write/rw) - possibly through a separate function.
> > 
> > 
> > Protection flags can be passed via the `params` argument of `umaMemoryProviderCreate`...
> 
> Technically the same is true for advise arguments ;) I think we need a dedicated function for both. But that can be done separately. I think the biggest problem here is making this useful enough so that a reasonable implementation can use it without making assumptions about underlying implementation (whether it's on windows or linux, CPU or GPU memory)

Also, do we want to support only OS/Host/Device memory? In general UMA abstractions (at least memory providers) could be useful for other things as well (allocating some specialized memory from a driver?) that might not support madvice flags. Are we ok with that? Do we have some other use cases at this moment? @vinser52

ldorau:
> In general UMA abstractions (at least memory providers) could be useful for other things as well (allocating some specialized memory from a driver?) that might not support madvice flags. Are we ok with that? 

Those flags do not have to be defined here. They can be implementation-specific.

ldorau:
Formatted according to `make check-generated`

igchor:
LGTM, but please add at least one basic test. You should implement this function for null/trace provider here: https://github.com/oneapi-src/unified-runtime/blob/65ed1ba2b9a8a52e4ea981dc3917a1578b880257/test/unified_memory_allocation/common/provider.c

and add one test case here: https://github.com/oneapi-src/unified-runtime/blob/65ed1ba2b9a8a52e4ea981dc3917a1578b880257/test/unified_memory_allocation/memoryProviderAPI.cpp#L14

vinser52:
Are we adding this functionality to address UR use case? If so, can you put here a code snippet to demonstrate how UR will use this API.
Also, it is not clear to me, should this API be directly used by UR? How pooling works in that case?

ldorau:
> LGTM, but please add at least one basic test. You should implement this function for null/trace provider here: 

@igchor Tests added

ldorau:
@pbalcer @igchor please approve CI builds in this PR

ldorau:
> Are we adding this functionality to address UR use case? If so, can you put here a code snippet to demonstrate how UR will use this API. Also, it is not clear to me, should this API be directly used by UR? How pooling works in that case?

Hi @vinser52, I am adding this functionality to integrate libqkmalloc with UMA.

This is PR integrating libqkmalloc with the current UMA (only `mmap()` and `munmap()` were replaced with `umaMemoryProviderAlloc()` and `umaMemoryProviderFree()` respectively):
https://github.com/intel-innersource/libraries.memory.allocators.qkmalloc/pull/44 (the UMA code is fetched in the CI build)

and this is my branch which adds one commit to the above PR that makes use of `umaMemoryProviderSetAttributes()` (which replaces `madvise()`):
https://github.com/ldorau/libraries.memory.allocators.qkmalloc/commits/devel-UMA

igchor (approved):

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -12,6 +12,31 @@
...
+#define UMA_ATTR_MERGEABLE 12      /* KSM may merge identical pages.  */
+#define UMA_ATTR_UNMERGEABLE 13    /* KSM may not merge identical pages.  */
+#define UMA_ATTR_HUGEPAGE 14       /* Worth backing with hugepages.  */
+#define UMA_ATTR_NOHUGEPAGE 15     /* Not worth backing with hugepages.  */
+#define UMA_ATTR_DONTDUMP 16       /* Explicity exclude from the core dump, */
+                                   /* overrides the coredump filter bits.  */
+#define UMA_ATTR_DODUMP 17         /* Clear the UMA_ATTR_DONTDUMP flag.  */
+#define UMA_ATTR_WIPEONFORK 18     /* Zero memory on fork, child only.  */
+#define UMA_ATTR_KEEPONFORK 19     /* Undo UMA_ATTR_WIPEONFORK.  */
+#define UMA_ATTR_COLD 20           /* Deactivate these pages.  */
```
vinser52, line 33: What deactivate means?

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -57,6 +82,17 @@ enum uma_result_t umaMemoryProviderAlloc(uma_memory_provider_handle_t hProvider,
...
 
+///
+/// \brief Set attributes of the memory space pointed by ptr from the memory provider
+/// \param hProvider handle to the memory provider
+/// \param ptr pointer to the allocated memory
+/// \param size size of the allocation
+/// \param attrs attributes of the memory space to be set
+///
+enum uma_result_t
+umaMemoryProviderSetAttributes(uma_memory_provider_handle_t hProvider,
```
vinser52, line 93: I have a feeling that we are mixing madvise-like attributes (just hints that runtime can ignore) with flags that should be applied or an error should be returned. Is it the right approach? I think we should have a discussion when we introduce such API.

igchor: But can we define a generic set of flags vs attributes for all providers?

source/common/unified_memory_allocation/include/uma/memory_provider.h
ldorau: From https://man7.org/linux/man-pages/man2/madvise.2.html :
```
       MADV_COLD (since Linux 5.4)
              Deactivate a given range of pages.  This will make the
              pages a more probable reclaim target should there be a
              memory pressure.  This is a nondestructive operation.  The
              advice might be ignored for some pages in the range when
              it is not applicable.
```
All attributes are copied from https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/mman-linux.h;h=6771f07a0f9a617fcde0eb264d54784788d22e76;hb=HEAD#l86.
We can narrow down the number of them of course.

ldorau: We need only `MADV_DONTNEED, MADV_HUGEPAGE, MADV_NOHUGEPAGE` in qkmalloc.

source/common/unified_memory_allocation/include/uma/memory_provider.h
ldorau: BTW, we need only `MADV_DONTNEED, MADV_HUGEPAGE, MADV_NOHUGEPAGE` in qkmalloc.

source/common/unified_memory_allocation/include/uma/memory_provider.h
pbalcer: For now, I would define just the few things that you need.

vinser52:
We had a discussion with @igchor regarding this PR on our UMA sync and agreed that we need to bring this topic to UMA sync before we will merge this PR. Because it a new API proposal. Furthermore, this API can also be used by UR and we need to make sure that the proposed API conforms UR's needs.
@igchor agreed to lead this discussion and capture use cases we want to address and options we can consider.

source/common/unified_memory_allocation/include/uma/memory_provider.h
igchor: For MADV_DONTNEED, I think it would be better to implement something like 'purge' in jemalloc. We will need that anyway if we want to implement the 'decay' functionally we discussed some time ago. I'm not sure if there is a good way to implement `purge` for USM/device memory (or we can just call something like zeVirtualMemFree?), but it's probably fine.

As for MADV_HUGEPAGE and MADV_NOHUGEPAGE: I'm not convinced if we should expose those at all. For example, there is no equivalent to MADV_NOHUGEPAGE on Windows, and requesting HUGE_PAGES can only be done during virtual memory reservation, not after. It might be better to just break the abstraction here and on linux, just call `madvise` directly instead of passing it through set_attrs.

ldorau:
Replaced with https://github.com/oneapi-src/unified-runtime/pull/453

PR #403, opened 2023-04-03
[ur] fix existing id registry parsing

pbalcer:
Registry id wasn't being properly read during spec parsing...

### DIFF
--- a/scripts/generate_ids.py
+++ b/scripts/generate_ids.py
@@ -30,10 +30,10 @@ def generate_registry(path, specs):
     try:
         existing_registry = list(util.yamlRead(path))[1]['etors']
         for etor in existing_registry:
-            valid_ids.discard(etor['id'])
+            valid_ids.discard(etor['value'])
             etors[etor['name']] = etor['value']
-    except:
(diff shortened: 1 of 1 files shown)


scripts/generate_ids.py
```diff
@@ -30,10 +30,10 @@ def generate_registry(path, specs):
     try:
         existing_registry = list(util.yamlRead(path))[1]['etors']
         for etor in existing_registry:
-            valid_ids.discard(etor['id'])
+            valid_ids.discard(etor['value'])
             etors[etor['name']] = etor['value']
-    except:
-        print('invalid existing registry, ignoring...')
+    except Exception as e:
```
kbenzie, line 35: Is it possible to be more specific about which exception types we're catching here? `Exception` is the base class of all exceptions so this will also catch things like `KeyboardInterrupt` which can be very unexpected to users.

If there are multiple exceptions a tuple of exception types is accetped syntax, e.g.

```python
    except (KeyError, ValueError) as e:
```

pbalcer: Thanks. I narrowed down the exception types.

PR #405, opened 2023-04-03
[UR] Fail build if doxygen has unresolved references

veselypeta:


scripts/Doxyfile
```diff
@@ -779,7 +779,7 @@ WARN_NO_PARAMDOC       = NO
 # a warning is encountered.
 # The default value is: NO.
 
-WARN_AS_ERROR          = NO
+WARN_AS_ERROR          = FAIL_ON_WARNINGS
```
pbalcer, line 782: I routinely get warnings like below because I use a newer version of doxygen.
```
warning: Tag 'OUTPUT_TEXT_DIRECTION' at line 102 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'CLANG_ASSISTED_PARSING' at line 1057 of file 'Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
warning: Tag 'CLANG_OPTIONS' at line 1065 of file 'Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
warning: Tag 'CLANG_DATABASE_PATH' at line 1075 of file 'Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
```

I'd prefer to leave it as is.

veselypeta: I've changed it so that it will only fail on warnings relating to unresolved references

kbenzie: Yeah, the amount of warnings you get seems to be tied directly to the version of Doxygen in use. Perhaps we could be most strict when running in CI but be more relaxed on local builds?

That might be annoying though, not sure.

veselypeta: Github defined a [`CI`](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) environment variable to true when running the workflow. I've changed it such that it will only fail on CI with unresolved references. Local builds will be unaffected.

kbenzie (approved):

pbalcer (approved):

PR #411, opened 2023-04-04
[ur][common] pretty printing for props

pbalcer:
[ur] introduce enum associated types and use it for props
Many UR functions use enum tags alongside opaque (typically void *)
pointers to construct an algebraic data type. This is useful because
it allows for creating function arguments that can be one of
several types of values. However, this approach, at least in C,
is not very typesafe. It's fairly easy to make a mistake that
can't be easily caught at compile time (and even at runtime).

To help with that a little bit, at least on the spec side of things,
this patch introduces optional 'typed_etor' and 'typename' qualifiers.
An enum with typed etor requires that all its values have specify
a type in their 'desc' field. The typename can be used to tag
opaque pointers with a tag field name.

Typed_etors are checked by codegen scripts, whereas correctness
of typenames is indirectly enforced by the code itself (won't compile
if it's wrong).

[common] use associated type information to serialize info_t types

This patch uses the typed etors and typename information to pretty
print `*_info_t` prop names, which otherwise are opaque pointers.

### DIFF
--- a/scripts/core/kernel.yml
+++ b/scripts/core/kernel.yml
@@ -83,7 +83,7 @@ name: $x_kernel_info_t
 typed_etors: True
 etors:
     - name: FUNCTION_NAME
-      desc: "[char*] Return null-terminated kernel function name."
+      desc: "[char[]] Return null-terminated kernel function name."
     - name: NUM_ARGS
       desc: "[size_t] Return Kernel number of arguments."
     - name: REFERENCE_COUNT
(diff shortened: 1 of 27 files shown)


scripts/core/kernel.yml
```diff
@@ -215,7 +219,7 @@ params:
       desc: "[in] size of the Kernel SubGroup property value"
     - type: "void*"
       name: pPropValue
-      desc: "[in,out][range(0, propSize)][optional] value of the Kernel SubGroup property."
```
igchor, line 218: shouldn't you keep the range here? `range(0, propSize)` is not equivalent to  `typename(propName, propSize)`, right?

pbalcer: This is just a pointer to a byte array of type `void*`. This is true of all pointers, but we don't mark, e.g., `size_t*` as range(0, 8) :) Also, you can't use index operator on 'void *', so this just creates issues in templates.
This range qualifier is useless, as far as I can tell.

igchor: Ok, that makes sense, I didn't fully get what range() was supposed to mean

pbalcer:
This is annoying to rebase and break with almost every UR change, so I'd appreciate a prompt review.

igchor (approved):

kbenzie (approved):
LGTM

PR #414, opened 2023-04-05
[ur] Make queue creation extensible

kbenzie:
Move the optional queue creation properties into the
`ur_queue_properties_t` struct which enables extensibility via `pNext`
chains. Also make use of `pNext` chaining by moving the
`UR_QUEUE_PROPERTY_COMPUTE_INDEX` property, since it is used to enable
the sycl_ext_intel_queue_index extension, into the
`ur_queue_index_properties_t` struct.

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -213,6 +213,8 @@ class ur_structure_type_v(IntEnum):
     USM_POOL_LIMITS_DESC = 11                       ## ::ur_usm_pool_limits_desc_t
     DEVICE_BINARY = 12                              ## ::ur_device_binary_t
     SAMPLER_DESC = 13                               ## ::ur_sampler_desc_t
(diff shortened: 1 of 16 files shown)


include/ur.py
```diff
@@ -1258,7 +1260,7 @@ class ur_queue_info_v(IntEnum):
     DEVICE_DEFAULT = 2                              ## ::ur_queue_handle_t: the current default queue of the underlying
                                                     ## device.
     PROPERTIES = 3                                  ## ::ur_queue_flags_t: the properties associated with
```
kbenzie, line 1262: This should be renamed to `FLAGS`

martygrant (approved):
LGTM

veselypeta (approved):
LGTM

PR #415, opened 2023-04-05
[ur][cts] Fix Windows signed/unsigned mismatch warning

kbenzie:
This warning is raised to an error when develop mode is on, fix the mistmatch by making the integer literal unsigned.


kbenzie:
This recently caused a [failure](https://github.com/oneapi-src/unified-runtime/actions/runs/4619032568/jobs/8167229489?pr=414) but I also reproduced this on the main branch so fixing it separately.

veselypeta (approved):
LGTM

PR #419, opened 2023-04-05
[ur] Add pool descriptor type to common

igchor:
It will be used by a pool manager to create/lookup pools for a given allocation.

It's similar to this code: https://github.com/intel/llvm/blob/928a919f15d9f989bedd5284a3a1e72d29d4a2e3/sycl/plugins/level_zero/pi_level_zero.cpp#L542 but I want to generalize it to be usable by all adapters.

In this PR I kept the original pooling granularity from L0 PI (per-context, per-device, separate pools for shared and sharedReadOnly memory, same pool for sub-devices and sub-sub-devices). This granularity can be customized/changed once we're ready to add pooling to other adapters.

It's a subset of changes from: https://github.com/oneapi-src/unified-runtime/pull/335


source/common/ur_pool_manager.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+    return {UR_RESULT_SUCCESS, sub_devices};
+}
+
+static inline bool
+isSharedAllocationReadOnlyOnDevice(const pool_descriptor &desc) {
+    return desc.type == UR_USM_TYPE_SHARED && desc.deviceReadOnly;
+}
+
+inline bool pool_descriptor::equal(const pool_descriptor &lhs,
```
pbalcer, line 76: It might be useful to have some differentiation (e.g., to prefer local memory or something). And I'm not sure about implementing equals for this - I think it should be a higher-levels decision if multiple ur devices use the same pool (or use each-other separate pools).

source/common/ur_pool_manager.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+inline std::vector<pool_descriptor>
+createPoolDescriptors(ur_usm_pool_handle_t poolHandle,
+                      ur_context_handle_t hContext, size_t deviceCount,
+                      const ur_device_handle_t *phDevices) {
+    std::vector<pool_descriptor> descriptors;
+
+    addPoolDescriptorForHost(poolHandle, hContext, descriptors);
+
+    std::function<void(ur_device_handle_t)> addPoolsForDevicesRec =
```
pbalcer, line 157: auto?

source/common/ur_pool_manager.hpp
```diff
@@ -0,0 +1,179 @@
...
+            if (ret != UR_RESULT_SUCCESS) {
+                throw ret;
+            }
+            for (auto &subDevice : subDevices) {
+                addPoolsForDevicesRec(subDevice);
+            }
+        };
+
+    for (size_t i = 0; i < deviceCount; i++) {
+        addPoolsForDevicesRec(phDevices[i]);
```
pbalcer, line 171: try catch?

source/common/ur_pool_manager.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+    return combine_hashes(0, desc.type, native,
+                          isSharedAllocationReadOnlyOnDevice(desc),
+                          desc.poolHandle);
+};
+
+static inline void
+addPoolDescriptorForHost(ur_usm_pool_handle_t poolHandle,
+                         ur_context_handle_t hContext,
+                         std::vector<pool_descriptor> &descriptors) {
```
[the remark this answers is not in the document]
igchor: I've rewritten this.

source/common/ur_pool_manager.hpp
igchor: I don't think anything other then bad_alloc can be thrown here. It would be better to handle this in upper layer

pbalcer (approved):

source/common/ur_pool_manager.hpp
```diff
@@ -0,0 +1,188 @@
...
+urGetSubDevices(ur_device_handle_t hDevice) {
+    size_t nComputeUnits;
+    auto ret = urDeviceGetInfo(hDevice, UR_DEVICE_INFO_MAX_COMPUTE_UNITS,
+                               sizeof(nComputeUnits), &nComputeUnits, nullptr);
+    if (ret != UR_RESULT_SUCCESS) {
+        return {ret, {}};
+    }
+
+    ur_device_partition_property_t properties[] = {
+        UR_DEVICE_PARTITION_EQUALLY, static_cast<intptr_t>(nComputeUnits), 0};
```
kswiecicki, line 46: It seems like a wrong cast type `static_cast<intptr_t>`.

source/common/ur_pool_manager.hpp
```diff
@@ -0,0 +1,188 @@
...
+        return {ret, {}};
+    }
+
+    std::vector<ur_device_handle_t> devicesAndSubDevices;
+    std::function<void(ur_device_handle_t)> addPoolsForDevicesRec =
+        [&](ur_device_handle_t hDevice) {
+            devicesAndSubDevices.push_back(hDevice);
+            auto [ret, subDevices] = urGetSubDevices(hDevice);
+            if (ret != UR_RESULT_SUCCESS) {
+                throw ret;
```
kswiecicki, line 89: Wouldn't returning this value from `urGetAllDevicesAndSubDevices` be better?
Or should it not be done because `urGetSubDevices` might already have partitioned some devices?

kswiecicki (approved):

source/common/ur_pool_manager.hpp
igchor: Your're right, I should return the status here. Done.

PR #425, opened 2023-04-06
[UR] Fix device tests that skip

### LINKED ISSUE #422: [CTS] Don't use `GTEST_SKIP()` for tests that loop over devices.
In https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/device/urDevicePartition.cpp#L18 we iterate over devices in the platform. Calling `GTEST_SKIP()` here will skip the entire test - we should instead skip the iteration of the loop, but still carry on testing the remaining devices.

veselypeta:
Closes #422


kbenzie (approved):
LGTM

callumfare (approved):

PR #426, opened 2023-04-07
[UR] Refactor checks.h to use overloads in ur_params.hpp

### LINKED ISSUE #421: [CTS] Refactor tests to use `ur_params.hpp` autogenerated ostream overloads
When implementing the CTS we defined overloads for enumerations so that they could be pretty-printed as part of the googtest name - [here](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/testing/include/uur/checks.h#L11).

However, as of #391 these overloads are now generated for all parameter types.

We should remove our own overloads and replace in favour for the definitions in `ur_common.hpp`.

veselypeta:
This PR:
* Use `operator<<` overloads from `ur_params.hpp` in conformance tests
* Move `operator<<` overloads into global namespace to enable argument dependent lookup

Closes #421


kbenzie (approved):
LGTM

martygrant (approved):

fabiomestre (approved):

scripts/templates/params.hpp.mako
```diff
@@ -24,20 +24,20 @@ from templates import helper as th
 #include <bitset>
 
 namespace ${x}_params {
-
 template <typename T> inline void serializePtr(std::ostream &os, T *ptr);
+} // namespace ${x}_params
```
pbalcer, line 28: I initially added the namespace because I wanted to implement these operators for primitive types, but ultimately decided against it in favor of the `serializePtr` function. And I left the namespace there because I thought it might be sometimes useful to selectively opt-in into the operators with `using namespace ...`.
I don't have a strong opinion on this, so I'm fine with the change, but there are some follow-up changes that this makes possible (e.g., no need to forward declare functions). I can do that in another PR.

test/conformance/platform/urInit.cpp
```diff
@@ -10,7 +10,12 @@ INSTANTIATE_TEST_SUITE_P(
                       UR_DEVICE_INIT_FLAG_VPU,
                       /* Combinations */
                       UR_DEVICE_INIT_FLAG_GPU | UR_DEVICE_INIT_FLAG_CPU,
-                      UR_DEVICE_INIT_FLAG_FPGA | UR_DEVICE_INIT_FLAG_VPU));
+                      UR_DEVICE_INIT_FLAG_FPGA | UR_DEVICE_INIT_FLAG_VPU),
+    [](const ::testing::TestParamInfo<ur_device_init_flags_t> &info) {
+        std::stringstream ss;
+        ur_params::serializeFlag_ur_device_init_flags_t(ss, info.param);
```
pbalcer, line 16: I didn't expect the serializeFlag functions to be used outside of the params file ;) It's fine, but I think we can make it nicer with a template like this:
`serializeFlag<ur_device_init_flag_t>(ss, info.param);`

But leave it as-is for now. I'll refactor this alongside other changes.

pbalcer (approved):

scripts/templates/params.hpp.mako
veselypeta: 👍

test/conformance/platform/urInit.cpp
veselypeta: I agree - templates is probably the better solution here 👍

PR #435, opened 2023-04-11
[UR] Fix registry id allocation

veselypeta:
When trying to add a new entry-point the registry was allocating duplicate id's causing it to fail to compile. The reason is that when the existing values are read from the `registry.yml` file as strings and the `valid_ids` list contains `ints`. Therefore, no id's were being removed during this step.

* The fix converts strings to ints, which then correctly removes id's from the `valid_ids` list.


martygrant (approved):

aarongreig (approved):

pbalcer (approved):
well, that's embarrassing ;)

PR #438, opened 2023-04-11
[ur] Improve logger env var handling

PatKamin:
Add tests for validation of the global logger object setup from
an environment variable. Improve logger env var handling.

### DIFF
--- a/cmake/match.cmake
+++ b/cmake/match.cmake
@@ -8,15 +8,25 @@
 
 find_package(Python3 COMPONENTS Interpreter)
 
-if(NOT DEFINED TEST_FILE)
-    message(FATAL_ERROR "TEST_FILE needs to be defined")
+# Process variables passed to the script
+if(NOT DEFINED MODE)
+    message(FATAL_ERROR "MODE needs to be defined. Possible values: 'stdout', 'stderr', 'file'")
+elseif(${MODE} STREQUAL "stdout" OR ${MODE} STREQUAL "stderr")
+    if(NOT DEFINED TEST_FILE)
+        message(FATAL_ERROR "TEST_FILE needs to be defined with a path to a binary to run")
+    else()
+        set(OUT_FILE "_matchtmpfile")
+    endif()
+elseif(${MODE} STREQUAL "file")
+    if(NOT DEFINED OUT_FILE)
+        message(FATAL_ERROR "OUT_FILE needs to be defined with an output file to be verified")
+    endif()
+else()
+    message(FATAL_ERROR "${MODE} mode not recognised. Possible values: 'stdout', 'stderr', 'file'")
 endif()
 if(NOT DEFINED MATCH_FILE)
     message(FATAL_ERROR "MATCH_FILE needs to be defined")
 endif()
-
-set(TEST_OUT "_matchtmpfile")
-
 if(NOT DEFINED TEST_ARGS) # easier than ifdefing the rest of the code
     set(TEST_ARGS "")
 endif()
@@ -24,23 +34,48 @@ endif()
 string(REPLACE "\"" "" TEST_ARGS "${TEST_ARGS}")
 separate_arguments(TEST_ARGS)
 
-execute_process(
-    COMMAND ${TEST_FILE} ${TEST_ARGS}
-    OUTPUT_FILE ${TEST_OUT}
-    RESULT_VARIABLE TEST_RESULT
-)
+if(EXISTS ${OUT_FILE})
+    file(REMOVE ${OUT_FILE})
+endif()
+
+# Run the test binary. Capture the output to the temporary file.
+if(${MODE} STREQUAL "stdout")
+    execute_process(
+        COMMAND ${TEST_FILE} ${TEST_ARGS}
+        OUTPUT_FILE ${OUT_FILE}
+        RESULT_VARIABLE TEST_RESULT
+    )
+elseif(${MODE} STREQUAL "stderr")
+    execute_process(
+        COMMAND ${TEST_FILE} ${TEST_ARGS}
+        ERROR_FILE ${OUT_FILE}
+        RESULT_VARIABLE TEST_RESULT
+    )
+elseif(${MODE} STREQUAL "file")
+    execute_process(
+        COMMAND ${TEST_FILE}
+        RESULT_VARIABLE TEST_RESULT
+    )
+endif()
 
 if(TEST_RESULT)
     message(FATAL_ERROR "Failed: Test ${TEST_FILE} ${TEST_ARGS} returned non-zero (${TEST_RESULT}).")
 endif()
 
+# Compare the output file contents with a match file contents
 execute_process(
-    COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_LIST_DIR}/match.py ${TEST_OUT} ${MATCH_FILE}
+    COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_LIST_DIR}/match.py ${OUT_FILE} ${MATCH_FILE}
     RESULT_VARIABLE TEST_RESULT
 )
 
 if(TEST_RESULT)
-    message(FATAL_ERROR "Failed: The output of ${TEST_FILE} (stored in ${TEST_OUT}) does not match ${MATCH_FILE} (${TEST_RESULT})")
+    file(READ ${OUT_FILE} OUTPUT)
+    file(READ ${MATCH_FILE} MATCH_STRING)
+    message(FATAL_ERROR "Failed: The output of ${OUT_FILE}:\n  ${OUTPUT}\n does not match ${MATCH_FILE}:\n ${MATCH_STRING} (${TEST_RESULT})")
 elseif()
-    message("Passed: The output of ${TEST_FILE} matches ${MATCH_FILE}")
+    message("Passed: The output ${OUT_FILE} matches ${MATCH_FILE}")
+endif()
+
+if(EXISTS ${OUT_FILE})
+    file(REMOVE ${OUT_FILE})
 endif()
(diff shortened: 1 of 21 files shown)


source/common/ur_util.hpp
```diff
@@ -195,28 +197,37 @@ inline std::optional<EnvVarMap> getenv_to_map(const char *env_var_name) {
...
+            throw_wrong_format_map(env_var_name, *env_var);
         }
 
         std::vector<std::string> values_vec;
         std::stringstream values_ss(values);
         std::string value;
+        auto is_quoted = [&value]() {
+            return (value.front() == '\'' && value.back() == '\'') ||
+                           (value.front() == '"' && value.back() == '"')
+                       ? true
```
pbalcer, line 215: no need for `? true : false`.

cmake/match.cmake
```diff
@@ -6,32 +6,64 @@
 # with a known good match file.
 #
 
-if(NOT DEFINED TEST_FILE)
-    message(FATAL_ERROR "TEST_FILE needs to be defined")
-endif()
-if(NOT DEFINED MATCH_FILE)
-    message(FATAL_ERROR "MATCH_FILE needs to be defined")
+# Process variables passed to the script
```
pbalcer, line 9: I'm not sure I like it this way. I'd rather have just one way of building tests with match files. But let's leave it for now.

cmake/match.cmake
```diff
@@ -6,32 +6,64 @@
...
 
-execute_process(
-    COMMAND ${TEST_FILE}
-    OUTPUT_FILE ${TEST_OUT}
-    RESULT_VARIABLE TEST_RESULT
-)
+# In 'nofile' mode, check if the OUT_FILE exists and exit the script
+if(${MODE} STREQUAL "nofile")
+    if(EXISTS ${OUT_FILE})
+        file(REMOVE ${OUT_FILE})
```
pbalcer, line 35: you souldn't remove random files like this.

cmake/match.cmake
```diff
@@ -6,32 +6,64 @@
...
+        message(FATAL_ERROR "Failed: File ${OUT_FILE} should not be created")
+    else()
+        message("Passed: Default logger has been constructed")
```
pbalcer, line 38: make the message generic ;)

source/common/logger/ur_logger.hpp
```diff
@@ -78,37 +78,47 @@ inline Logger create_logger(std::string logger_name) {
...
                    : sink_from_str(logger_name, values[0]);
     } catch (const std::invalid_argument &e) {
-        std::cerr
-            << "Error when creating a logger instance from environment variable"
-            << e.what();
+        std::cerr << "Error when creating a logger instance from the '"
+                  << env_var_name.str() << "' environment variable:\n"
+                  << e.what() << std::endl;
         return Logger(std::make_unique<logger::StderrSink>(logger_name));
     }
```
pbalcer, line 123: can't the rest of the function throw (e.g., `Logger(level, std::move(sink));`) ?

cmake/match.cmake
PatKamin: Moved removal of files to tests set up. However, the output file is not removed after tests, rather before them, so the out file will exist after running the last test.

cmake/match.cmake
PatKamin: Good point :)

source/common/logger/ur_logger.hpp
```diff
@@ -78,37 +78,47 @@ inline Logger create_logger(std::string logger_name) {
...
+        auto map = getenv_to_map(env_var_name.str().c_str());
+        if (!map.has_value()) {
+            return Logger(std::make_unique<logger::StderrSink>(logger_name));
+        }
+
         auto kv = map->find("level");
         if (kv != map->end()) {
             auto value = kv->second.front();
             level = str_to_level(value);
+            map->erase(kv);
```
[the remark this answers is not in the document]
PatKamin: It's for catching wrong, unrecognized parameters, so that there is a strict check for params that can be used for logger setup.

source/common/ur_util.hpp
PatKamin: Indeed, thanks.

source/common/logger/ur_logger.hpp
PatKamin: I assumed it cannot throw as the only possibility of this ctor to do this is when there is no sink. I think that there is guarantee to have a valid sink at this point of the function, since the exception would be already caught when calling `sink_from_str()`.
If I were to check for exceptions in the return statement, I would have to make the whole `create_logger()` funtion a throwable one, and this is sth I wanted to avoid here by making sure that the default logger is properly created.

test/unit/logger/CMakeLists.txt
```diff
@@ -3,4 +3,462 @@
...
+
+    if(EXISTS ${OUT_FILE})
+        file(REMOVE ${OUT_FILE})
+    endif()
+
+    add_test(NAME ${full_name}
+        COMMAND ${TEST_TARGET_NAME}
+    )
+
+    add_test(NAME ${full_name}-validate-output
```
pbalcer, line 31: Are you guaranteed that these tests are going to run in order? Why not just run the test in the match script?

PatKamin:
> Are you guaranteed that these tests are going to run in order? Why not just run the test in the match script? 

There is no guarantee that they run in order, just that the first test is set to finish before this one. Between these two other tests can run. Hence, I restructured a bit tests to be run from a cmake wrapper script using `execute_process()` command.

cmake/match.cmake
```diff
@@ -6,32 +6,50 @@
 # with a known good match file.
 #
 
-if(NOT DEFINED TEST_FILE)
-    message(FATAL_ERROR "TEST_FILE needs to be defined")
+# Process variables passed to the script
+if(NOT DEFINED MODE)
+    message(FATAL_ERROR "MODE needs to be defined. Possible values: 'binary', 'file'")
```
pbalcer, line 11: I think really the difference is what we match against, some specified file or stdout, right? If so, I think it would be better to call these modes 'stdout' and 'file'.

test/unit/logger/logfile_match_wrapper.cmake
```diff
@@ -0,0 +1,50 @@
...
+endif()
+if(DEFINED LOGFILE_MATCH AND NOT DEFINED MATCH_FILE)
+    message(FATAL_ERROR "MATCH_FILE needs to be defined with LOGFILE_MATCH enbled")
+endif()
+
+if(EXISTS ${OUT_FILE})
+    file(REMOVE ${OUT_FILE})
+endif()
+
+execute_process(
```
pbalcer, line 25: I mean, this is exactly what the match.cmake does. So what you are effectively doing is creating a wrapper with the same functionality that is being disabled (by mode 'binary') in the main script?

Again, I'm not sure I fully understand, but if all you are trying to do is match against different output destination (stdout, stderr, file), then it'd be easier to create that argument in the original match script.

test/unit/logger/CMakeLists.txt
```diff
@@ -3,4 +3,437 @@
...
+        COMMAND ${TEST_TARGET_NAME}
+    )
+    
+    set_tests_properties(${full_name}-validate-output PROPERTIES
+        LABELS "unit"
+        ENVIRONMENT ${env_var}\;GTEST_FILTER=*${test_case}\;GTEST_BRIEF=1
     )
+
+    file(READ ${match_file} MATCH_STRING)
```
pbalcer, line 49: I added support for regex matches in #444

source/common/ur_util.hpp
```diff
@@ -107,10 +107,12 @@ static void throw_wrong_format_vec(const char *env_var_name) {
...
 }
 
-static void throw_wrong_format_map(const char *env_var_name) {
+static void throw_wrong_format_map(const char *env_var_name,
+                                   std::string env_var_value) {
     std::stringstream ex_ss;
-    ex_ss << "Wrong format of the " << env_var_name << " environment variable!"
-          << " Proper format is: "
+    ex_ss << "Wrong format of the " << env_var_name
+          << " environment variable: '" << env_var_value << "'\n"
```
kswiecicki, line 114: This wording might be a bit confusing.

source/common/ur_util.hpp
```diff
@@ -107,10 +107,12 @@ static void throw_wrong_format_vec(const char *env_var_name) {
...
+          << "Proper format is: "
              "ENV_VAR=\"param_1:value_1,value_2;param_2:value_1";
```
kswiecicki, line 116: There's `\"` missing at the end of the exception string.

source/common/ur_util.hpp
```diff
@@ -196,23 +198,31 @@ inline std::optional<EnvVarMap> getenv_to_map(const char *env_var_name,
...
         std::string value;
+        auto is_quoted = [&value]() {
+            return (value.front() == '\'' && value.back() == '\'') ||
+                   (value.front() == '"' && value.back() == '"');
+        };
         while (std::getline(values_ss, value, values_delim)) {
-            if (value.empty()) {
-                throw_wrong_format_map(env_var_name);
+            if (value.empty() ||
+                ((value.find(':') != std::string::npos) && !is_quoted())) {
```
kswiecicki, line 220: Since you have an anonymous function for `is_quoted` you might want to create `has_colon` for `(value.find(':') != std::string::npos)`.

source/common/ur_util.hpp
PatKamin: Rephrased a bit

source/common/ur_util.hpp
PatKamin: Added `has_colon()`. Also moved lambdas out of the `while` loop.

kswiecicki (approved):

PR #443, opened 2023-04-14
[uma] add function to retrieve providers

igchor:
This functionality will be used in UR and possibly MPI to query memory properties (that are stored in a memory provider). To make it usable, I think we would also need to add the provider_type (that was initially part of the memory tracking PR) and limit the number of DATA memory providers to 1.

This functionality could also be helpful for lifetime management: the user no longer needs to keep track of memory provider handles to destroy them.

One thing I'm not sure about is whether we want to expose the original memory providers or the wrappers (tracking providers). Returning the original ones makes more sense to me - this way, we don't need to expose any other functions (like getUpstreamProvider).  Also, our memory tracking is done on per-pool basis, if we'd return a memory-tracking provider, what memory pool it would be associated to?

### DIFF
--- a/source/common/unified_memory_allocation/src/memory_provider_internal.h
+++ b/source/common/unified_memory_allocation/src/memory_provider_internal.h
@@ -1,0 +1,24 @@
+/*
+ *
+ * Copyright (C) 2023 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#ifndef UMA_MEMORY_PROVIDER_INTERNAL_H
+#define UMA_MEMORY_PROVIDER_INTERNAL_H 1
+
+#include <uma/memory_provider.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void *umaMemoryProviderGetPriv(uma_memory_provider_handle_t hProvider);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* UMA_MEMORY_PROVIDER_INTERNAL_H */
--- a/source/common/unified_memory_allocation/src/memory_tracker.cpp
+++ b/source/common/unified_memory_allocation/src/memory_tracker.cpp
@@ -10,6 +10,7 @@
 #include <uma/memory_provider.h>
 #include <uma/memory_provider_ops.h>
 
+#include <cassert>
 #include <map>
 #include <mutex>
 #include <shared_mutex>
@@ -171,7 +172,8 @@ enum uma_result_t umaTrackingMemoryProviderCreate(
     uma_memory_provider_handle_t hUpstream, uma_memory_pool_handle_t hPool,
     uma_memory_provider_handle_t *hTrackingProvider) {
     uma_tracking_memory_provider_t params;
-    params.hUpstream = hUpstream, params.hTracker = umaMemoryTrackerGet(),
+    params.hUpstream = hUpstream;
+    params.hTracker = umaMemoryTrackerGet();
     params.pool = hPool;
 
     struct uma_memory_provider_ops_t trackingMemoryProviderOps;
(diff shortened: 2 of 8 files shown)


source/common/unified_memory_allocation/src/memory_provider_internal.h
```diff
@@ -0,0 +1,24 @@
...
+#ifndef UMA_MEMORY_PROVIDER_INTERNAL_H
+#define UMA_MEMORY_PROVIDER_INTERNAL_H 1
+
+#include <uma/memory_provider.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void* umaMemoryProviderGetPriv(uma_memory_provider_handle_t hProvider);
```
igchor, line 18: I don't like that I had to add this function, but I couldn't; come up with a better way - maybe you have some ideas?

The problem is that we need to call a memory tracking provider-specific function that is implemented in memory_tracker.cpp  (getUpstreamProvider). However, this function does not know anything about how `uma_memory_provider_handle_t` is implemented (it's hidden in memory_provider.c) so we need a way to get the internal pointer (tracker-specific).

I think this might be a problem for the user, as well (right now this function is internal but maybe we need to expose it?). What if he wants to call his provider-specific function? He cannot do that if he only has a `uma_memory_provider_handle_t`. There is no way for the user to get a pointer to back his provider class/structure.

pbalcer: I had to spent some time to even understand why we need to get the priv pointer. I agree it's confusing, but I'm not sure there's an alternative - we lost type information about the provider when we create an uma handle. The best approach would be to allow providers to somehow identify themselves (some unique string or id), that then can be used at runtime to match with the correct type - e.g.,:
```
provider providers[] = {a, b, c};

for (p in providers) {
   if (p.implementation_id() == unix_provider.id()) {
          unix_provider_custom_func(p);
   }
}
```
But that's probably too verbose to be useful.

Anyway, I think we should have a get_priv() method. I'd also go a step further and change the provider (and heap) ops - https://github.com/oneapi-src/unified-runtime/blob/main/source/common/unified_memory_allocation/include/uma/memory_provider_ops.h#L21 - to use the handle and internally the providers can use get_priv to get their data. I find the current implementation confusing... We use the variable name 'provider' to mean both the handle and the provider priv data.

igchor: I'm not sure about making provider/pool call get_priv: you'd need to link every pool/provider with uma_core. Right now, we only have a dependency on the uma_base.h header.

Also, passing `priv` to each provider/pool seems very similar to what UR is doing with adapters. Pehaps we should even rename the `priv` to `native` and have `uma_pool/provider_native_handel_t` instead of `void*`?

source/common/unified_memory_allocation/src/memory_tracker.cpp
```diff
@@ -185,4 +187,13 @@ enum uma_result_t umaTrackingMemoryProviderCreate(
     return umaMemoryProviderCreate(&trackingMemoryProviderOps, &params,
                                    hTrackingProvider);
 }
+
+void umaTrackingMemoryProviderGetUpstreamProvider(
+    uma_memory_provider_handle_t hTrackingProvier,
```
PatKamin, line 192: typo: hTrackingProvier -> hTrackingProvider

source/common/unified_memory_allocation/src/memory_tracker.h
```diff
@@ -29,6 +29,10 @@ enum uma_result_t umaTrackingMemoryProviderCreate(
     uma_memory_provider_handle_t hUpstream, uma_memory_pool_handle_t hPool,
     uma_memory_provider_handle_t *hTrackingProvider);
 
+void umaTrackingMemoryProviderGetUpstreamProvider(
+    uma_memory_provider_handle_t hTrackingProvier,
```
PatKamin, line 33: typo: hTrackingProvier -> hTrackingProvider

test/unified_memory_allocation/memoryPoolAPI.cpp
```diff
@@ -185,3 +208,16 @@ TEST_P(poolInitializeTest, errorPropagation) {
...
+    std::array<uma_memory_provider_handle_t, numProviders> providers = {
+        (uma_memory_provider_handle_t)0x1, (uma_memory_provider_handle_t)0x2,
+        (uma_memory_provider_handle_t)0x3, (uma_memory_provider_handle_t)0x4};
+
+    auto [ret, pool] = uma::poolMakeUnique<uma_test::proxy_pool>(
+        providers.data(), numProviders);
+
+    ret = umaPoolGetMemoryProviders(pool.get(), 1, providers.data(), nullptr);
+    ASSERT_EQ(ret, UMA_RESULT_ERROR_INVALID_ARGUMENT);
+}
```
PatKamin, line 223: No newline at EOF

PatKamin (approved):

pbalcer (approved):

PR #447, opened 2023-04-17
Fix error in urMemBufferCreate validation layer

jandres742:
Condition needs to check for pProperties not being null before dereferencing it.


pbalcer (approved):

pbalcer:
It should be possible to create a static check for this type of issue during spec validation (in `parse_spec.py`). Or, better yet, automatically add that NULL-check for optional parameters that contain a validation check. @kswiecicki what do you think?

veselypeta (approved):

PR #450, opened 2023-04-18
Reorder values in `ur_sampler_addressing_mode_t` enum

callumfare:
Reorder the values of the `ur_sampler_addressing_mode_t` enum to match the values used in libclc. For the cuda adapter (and maybe others) this means we can use the enum values directly in the sampler properties field without having to remap them.

### DIFF
--- a/scripts/core/sampler.yml
+++ b/scripts/core/sampler.yml
@@ -25,16 +25,21 @@ desc: "Sampler addressing mode"
 class: $xSampler
 name: $x_sampler_addressing_mode_t
 etors:
-    - name: MIRRORED_REPEAT
-      desc: "Mirrored Repeat"
-    - name: REPEAT
-      desc: "Repeat"
-    - name: CLAMP
-      desc: "Clamp"
-    - name: CLAMP_TO_EDGE
-      desc: "Clamp to edge"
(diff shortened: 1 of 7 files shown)


callumfare:
For reference, this is where the ordering is defined in libclc:
https://github.com/intel/llvm/blob/sycl/libclc/ptx-nvidiacl/libspirv/images/image.cl#L880

pbalcer (approved):

scripts/core/sampler.yml
```diff
@@ -25,16 +25,16 @@ desc: "Sampler addressing mode"
...
     - name: NONE
       desc: "None"
+    - name: CLAMP_TO_EDGE
+      desc: "Clamp to edge"
+    - name: CLAMP
+      desc: "Clamp"
+    - name: REPEAT
+      desc: "Repeat"
+    - name: MIRRORED_REPEAT
+      desc: "Mirrored Repeat"
```
kbenzie, line 37: If the values of these enums matter, we should be explicit about it by adding `value: "<value>"` to each of them.

veselypeta (approved):

kbenzie (approved):
LGTM, one suggestion

PR #452, opened 2023-04-19
[ur] remove unused function callbacks

pbalcer:
The original purpose of these callbacks was manual registration of explicit function callbacks in the tracing layer, similar to how it works with level-zero loader.

However, UR relies on the XPTI framework that allows software to transparently intercept all function calls, making the callbacks redundant.


veselypeta (approved):

kbenzie (approved):
LGTM

PR #453, opened 2023-04-19
[uma] Add purge_[lazy|force] functions

igchor:
for memory provider. They are similar to jemalloc 'purge' hooks. OS memory provider can implement those functions using madvise and MADV_FREE / MADV_DONTNEED.

I've also added a way to query (possible) page sizes for a given pointer so that we can require `ptr` and `size` parameters to `purge` to be page-aligned. I think it's better not to adjust those transparently within the provider. 

I'm not entirely sure how to handle this for transparent huge pages though. Right now the API allows returning an array of supported page sizes (sorted by size) from that function instead of a single value. This way an implementation can either just return an array of all possible page sizes for any given pointer or return a specific page size if it knows what it is. With this approach we can just return {4K, 2M, 1G} or If we'd have separate memory providers for huge pages and for regular pages, we can just return {4K} for the regular-one or {2M, 1G} for huge-pages provider.

@pbalcer also suggested adding a way to query for `recommended` page size so that we can avoid splitting huge pages whenever we call madvise on non-2MB-aligned address (e.g. in purge). Perhaps a better approach would be to just have a `provider::get_minimum_page_size(void *ptr)` (4k for OS memory), and a `provider::get_recommended_page_size(size_t size)`?

This would be quite similar approach to L0: https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=page#zevirtualmemquerypagesize

### DIFF
--- a/source/common/unified_memory_allocation/include/uma/memory_provider.h
+++ b/source/common/unified_memory_allocation/include/uma/memory_provider.h
@@ -83,6 +83,16 @@ umaMemoryProviderGetLastResult(uma_memory_provider_handle_t hProvider,
                                const char **ppMessage);
 
 ///
+/// \brief Retrieve recommended page size for a given allocation size.
+/// \param hProvider handle to the memory provider
+/// \param size allocation size
+/// \param pageSize [out] will be updated with recommended page size.
+/// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure.
+enum uma_result_t
+umaMemoryProviderGetRecommendedPageSize(uma_memory_provider_handle_t hProvider,
+                                        size_t size, size_t *pageSize);
+
+///
 /// \brief Retrieve minumum possible page size used by memory region referenced by given ptr
 /// \param hProvider handle to the memory provider
 /// \param ptr pointer to memory allocated by this memory provider
--- a/source/common/unified_memory_allocation/include/uma/memory_provider.h
+++ b/source/common/unified_memory_allocation/include/uma/memory_provider.h
@@ -82,6 +82,42 @@ enum uma_result_t
 umaMemoryProviderGetLastResult(uma_memory_provider_handle_t hProvider,
                                const char **ppMessage);
 
+///
+/// \brief Retrieve minumum possible page size used by memory region referenced by given ptr
+/// \param hProvider handle to the memory provider
+/// \param ptr pointer to memory allocated by this memory provider
+/// \param pageSize [out] will be updated with page size value.
+/// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure.
+enum uma_result_t
+umaMemoryProviderGetMinPageSize(uma_memory_provider_handle_t hProvider,
+                                void *ptr, size_t *pageSize);
+
+///
+/// \brief Discard physical pages within the virtual memory mapping associated at given addr and size.
+///        This call is asynchronous and may delay puring the pages indefinitely.
+/// \param hProvider handle to the memory provider
+/// \param ptr beginning of the virtual memory range
+/// \param size size of the virtual memory range
+/// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure.
+///         UMA_RESULT_ERROR_INVALID_ALIGNMENT if ptr or size is not page-aligned.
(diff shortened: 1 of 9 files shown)


source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -82,6 +82,45 @@ enum uma_result_t
...
+                              size_t numPageSizes, size_t *pageSizes,
+                              size_t *numPageSizesRet);
+
+///
+/// \brief Discard physical pages within the virtual memory mapping associated at given addr and size.
+///        This call is asynchronous and may delay puring the pages indefinitely.
+/// \param hProvider handle to the memory provider
+/// \param ptr beginning of the virtual memory range
+/// \param size size of the virtual memory range
+/// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure.
```
vinser52, line 106: Should we consider that not all memory providers will support Purge functionality? I think we can add a corresponding return code to notify the caller of this API.

vinser52:
I think you need to update the tracking provider as well.

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -82,6 +82,45 @@ enum uma_result_t
...
+///        This array is sorted in ascending order.
+/// \param hProvider handle to the memory provider
+/// \param ptr pointer to memory allocated by this memory provider
+/// \param numPageSizes [in] number of memory providers to return. If numPageSizes is not equal to or
+///        greater than the number of possible page sizes, UMA_RESULT_ERROR_INVALID_ARGUMENT is returned.
+/// \param pageSizes [out] will be updated with page size values.
+/// \param numPageSizesRet [in] pointer to the actual number of possible page sizes
+/// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure.
+enum uma_result_t
+umaMemoryProviderGetPageSizes(uma_memory_provider_handle_t hProvider, void *ptr,
```
vinser52, line 96: Does the array of possible pageSizes is really needed? Perhaps, just a minimal pageSize is enough because other possible pageSizes should be a multiple of the minimum pageSize.

I am thinking should we also care about proper alignment in the `umaMemoryProviderAlloc` function? And should we at all?

igchor: Yeah, as I think about it more, probably a function to get a minimum page size should be enough for `purge`.

However, a function for recommended page size could also be useful - it could be used to avoid splitting huge pages and also as a hint for the user, what should he pass as an `alignment` in `umaMemoryProviderAlloc ` to (possibly) get huge-pages.

igchor: For transparent huge pages: yes - linux will just split those pages to 4KB ones.
For huge TLB (or windows huge pages): no - madvise will return an error.

But for huge TLB an appropriate option needs to be passed during mmap so the memory provider can either track those pages or we can have a memory provider dedicated to huge pages (in that case, it would always return 2MB or 1GB for minPageSize)

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -82,6 +82,51 @@ enum uma_result_t
 umaMemoryProviderGetLastResult(uma_memory_provider_handle_t hProvider,
                                const char **ppMessage);
 
+///
+/// \brief Retrieve recommended page size for a given allocation size.
+/// \param hProvider handle to the memory provider
+/// \param size allocation size
+/// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure.
+enum uma_result_t
+umaMemoryProviderGetRecommendedPageSize(uma_memory_provider_handle_t hProvider,
```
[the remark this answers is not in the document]
igchor: Done.

vinser52 (approved):
Were these changes reviewed/discussed with folks who are working on qkmalloc? Does it conform to their needs?

ldorau:
> Were these changes reviewed/discussed with folks who are working on qkmalloc? Does it conform to their needs?

@vinser52 Yes, it does. Thanks @vinser52 !

ldorau (approved):
LGTM

PR #454, opened 2023-04-20
[UR][CUDA] Vector addition example

veselypeta:
[WARNING - DO NOT MERGE!]
This PR is based on an early release of unified runtime `weekly-2023-03-31` so that it can be tested against the CUDA adapter which used this version of unified runtime.

### DIFF
--- a/examples/vector_addition/CMakeLists.txt
+++ b/examples/vector_addition/CMakeLists.txt
@@ -4,6 +4,9 @@
 add_executable(vector_addition
     ${CMAKE_CURRENT_SOURCE_DIR}/vector_addition.cpp
 )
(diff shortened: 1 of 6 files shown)


examples/vector_addition/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: MIT
```
omarahmed1111, line 2: Just wanted to give an update that the license of the project have been updated in this https://github.com/oneapi-src/unified-runtime/pull/557. So, we should be able to update all the files license to the llvm one now.

PR #456, opened 2023-04-21
[ur] Fix linking against libstdc++fs

### LINKED ISSUE #451: SYCL builds of UR fail due to undefined reference to std::filesystem
After addition in #439 some SYCL builds of UR are failing with undefined reference to std::filesystem::__cxx11::path::_M_split_cmpts()`

See this comment:: https://github.com/intel/llvm/pull/9053#issuecomment-1510034425

kbenzie:
This patch replaces changes the logic used to specify linking against `libstdc++fs.a` required to use early iterations of `std::filesystem`. Previously CMake checked for the existence of the `<filesystem>` header to enable linking the static library, however in GCC 8 this header exists while still requiring manual linking against the static library. Now we check if we are compiling with `libstdc++` and always manually link against the static library. Fixes #451.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.14.0 FATAL_ERROR)
 project(unified-runtime VERSION 0.6.0)
 
 include(GNUInstallDirs)
-include(CheckIncludeFileCXX)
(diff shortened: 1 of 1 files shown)


martygrant (approved):

veselypeta (approved):
LGTM

PR #458, opened 2023-04-24
[UR] Implement extension mechanism

veselypeta:
This MR:
* Introduces a basic concept of extensions for unified runtime.
* The extensions specification is largely based on L0 [see here](https://spec.oneapi.io/level-zero/latest/core/EXT.html)
* Introduces `urPlatformGetExtensionProperties` to query which extensions are supported.

### DIFF
--- a/scripts/core/EXT.rst
+++ b/scripts/core/EXT.rst
@@ -1,4 +1,7 @@
 <%
+import re
+from templates import helper as th
+%><%
     OneApi=tags['$OneApi']
     x=tags['$x']
     X=x.upper()
@@ -78,7 +81,7 @@ convention and values should be assigned to avoid future compatibility issues.
 Extending Structures
 --------------------
 
-Any structure derived from `${x}_base_desc_t`` or `${x}_base_properties_t`` can be extended using a structure chain
+Any structure derived from ``${x}_base_desc_t`` or ``${x}_base_properties_t`` can be extended using a structure chain
 by adding a pointer to the extended structure in `pNext` member variable. No other method of extending structures is allowed.
 
 A structure chain can contain more than one extension structure, in any order. Therefore, extensions should not
@@ -101,7 +104,13 @@ Adding an extension
 
 List of Standard Extensions
 ===========================
-// TODO - list all standard extensions
+%for name in meta['macro']:
+%if name.endswith('EXT_NAME'):
+
+    - :ref:`${th.subt(namesapce, tags, meta['macro'][name]['values'][0])} <${th.subt(namesapce, tags, name)}>`
+
+%endif
+%endfor
 
 List of Experimental Extensions
 ===============================
--- a/scripts/core/EXT.rst
+++ b/scripts/core/EXT.rst
@@ -16,14 +16,14 @@ in a later version of Unified Runtime, when agreed upon by the Working Group. Al
 optional and are not required to be implemented by any particular adapter, but are expected to 
 be widely available and possibly be required in a future version of the specification. Adapters 
 must report which extensions are supported through the ${x}PlatformGetExtensionProperties query.
-Each extensions may also impose additional restrictions on when it can be used - i.e. a platform
+Each extension may also impose additional restrictions on when it can be used - i.e. a platform
 or device query.
 
 There are two types of extensions defined by this specification:
 
 1. **Standard** -  extension will be included into the current and all future version of the specification.
 2. **Experimental** - extensions require additional experimentation and development, before becoming a standard extension. 
-Applications should not rely on experimental extensions in production.
+   Applications should not rely on experimental extensions in production.
 
 Requirements
 ============
@@ -78,8 +78,8 @@ convention and values should be assigned to avoid future compatibility issues.
 Extending Structures
 --------------------
 
-Any structure derived from `${x}_base_desc_t`` or `${x}_base_properties_t`` can be extended using structure chains. No other
-method of extending structures is allowed.
+Any structure derived from `${x}_base_desc_t`` or `${x}_base_properties_t`` can be extended using a structure chain
+by adding a pointer to the extended structure in `pNext` member variable. No other method of extending structures is allowed.
 
 A structure chain can contain more than one extension structure, in any order. Therefore, extensions should not
 be dependent on their order relative to other extensions and the implementation must be order agnostic. In addition,
--- a/scripts/core/EXT.rst
+++ b/scripts/core/EXT.rst
@@ -15,7 +15,7 @@ without affecting the core specification. Extensions may in future be promoted t
 in a later version of Unified Runtime, when agreed upon by the Working Group. All extensions are 
 optional and are not required to be implemented by any particular adapter, but are expected to 
 be widely available and possibly be required in a future version of the specification. Adapters 
-must report which extensions are supported through the ${x}PlatformGetExtensionProperties query.
+must report which extensions are supported through the ${X}_PLATFORM_INFO_EXTENSIONS platform info query.
 Each extension may also impose additional restrictions on when it can be used - i.e. a platform
 or device query.
 
@@ -29,7 +29,7 @@ Requirements
 ============
 
 - Extensions must use globally unique names for macros, enums, structures and functions
-- Extensions must have globally unique extension names reported from ${x}PlatformGetExtensionProperties
+- Extensions must have globally unique extension names reported from ${X}_PLATFORM_INFO_EXTENSIONS platform info query
 - All extensions must be defined in this specification
 - Extensions must not break backwards compatibility of the core APIs
 - Standard extension versions must be backwards compatible with prior versions
--- a/scripts/core/EXT.rst
+++ b/scripts/core/EXT.rst
@@ -4,7 +4,7 @@
     X=x.upper()
 %>
 ==============
- Extensions
+Extensions
 ==============
 
 Objective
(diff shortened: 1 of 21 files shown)


scripts/core/EXT.rst
```diff
@@ -0,0 +1,91 @@
...
+Applications should not rely on experimental extensions in production.
+
+Requirements
+============
+
+- Extensions must use globally unique names for macros, enums, structures and functions
+- Extensions must have globally unique extension names reported from ${x}PlatformGetExtensionProperties
+- All extensions must be defined in this specification
+- Extensions must not break backwards compatibility of the core APIs
+- Standard extension versions must be backwards compatible with prior versions
```
veselypeta, line 35: I copied this from L0 extension scheme, but perhaps we could use semantic versioning with extensions with major/minor versions?

scripts/core/EXT.rst
```diff
@@ -0,0 +1,91 @@
...
+Objective
+=========
+
+Extensions allow for additional functionality to be added to the Unified Runtime specification
+without affecting the core specification. Extensions may in future be promoted to the core specification
+in a later version of Unified Runtime, when agreed upon by the Working Group. All extensions are 
+optional and are not required to be implemented by any particular adapter, but are expected to 
+be widely available and possibly be required in a future version of the specification. Adapters 
+must report which extensions are supported through the ${x}PlatformGetExtensionProperties query.
+Each extensions may also impose additional restrictions on when it can be used - i.e. a platform
```
EwanC, line 19:
```suggestion
Each extension may also impose additional restrictions on when it can be used - i.e. a platform
```

scripts/core/EXT.rst
```diff
@@ -0,0 +1,91 @@
...
+or device query.
+
+There are two types of extensions defined by this specification:
+
+1. **Standard** -  extension will be included into the current and all future version of the specification.
+2. **Experimental** - extensions require additional experimentation and development, before becoming a standard extension. 
+Applications should not rely on experimental extensions in production.
```
kbenzie, line 26: If this is supposed to be part of the `2.` it should be indented:

```suggestion
   Applications should not rely on experimental extensions in production.
```
```

File: scripts/core/EXT.rst [hunk 4]
```diff
@@ -0,0 +1,91 @@
...
+
+Any existing enumeration may be extended by adding new enumerators. Enumerators must use the extensions naming 
+convention and values should be assigned to avoid future compatibility issues.
+
+
+Extending Structures
+--------------------
+
+Any structure derived from `${x}_base_desc_t`` or `${x}_base_properties_t`` can be extended using structure chains. No other
+method of extending structures is allowed.
```
kbenzie, line 82: Since we refer to these as `pNext` chains in the spec, we should do that here too.

veselypeta: 👍

scripts/core/EXT.rst
```diff
@@ -0,0 +1,91 @@
...
+
+A structure chain can contain more than one extension structure, in any order. Therefore, extensions should not
+be dependent on their order relative to other extensions and the implementation must be order agnostic. In addition,
+the implementation will ignore extended structures that it does not support.
+
+The extension must document the specific structures and functions that may be extended using the structure chain.
+
```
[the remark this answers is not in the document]
veselypeta: Good idea. I'll add a section which will outline the steps.

scripts/core/platform.yml
```diff
@@ -242,3 +242,40 @@ etors:
...
+--- #--------------------------------------------------------------------------
+type: macro
+desc: "Extension name maximum length."
+name: $X_MAX_EXTENSION_NAME_LENGTH
+value: "256"
+--- #--------------------------------------------------------------------------
+type: struct
+desc: "Extension properties."
+name: $x_extension_properties_t
+base: $x_base_properties_t
```
kbenzie, line 254: I don't think this is the expected usage of `_properties_t` which are intended for creating objects with `pNext` chains, not querying info about things.

scripts/core/platform.yml
```diff
@@ -242,3 +242,40 @@ etors:
...
+      name: name[$X_MAX_EXTENSION_NAME_LENGTH]
+      desc: "[in] null-terminated extension name."
+    - type: uint32_t
+      name: version
+      desc: "[in] version of the extension using $X_MAKE_VERSION."
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Retrieve the set of supported extensions."
+class: $xPlatform
+name: GetExtensionProperties
```
kbenzie, line 267: Currently we have a `PlatformGetInfo` query for supported extensions but it returns a list of names in a string with some delimiter. I'm not convinced we need an additional entry point, we could return an array of structs from `PlatformGetInfo` instead of a list of names in a string.

veselypeta: sounds like a good plan 👍

scripts/core/EXT.rst
```diff
@@ -0,0 +1,108 @@
...
+Adding an extension
+===================
+
+* Extend the specification with the functions, enumerations, macros, and structures required by the extension, ensuring that
+  all additions fully comply with the naming and ABI requirements detailed above.
+
+* Document the extension in `EXT_<ext-name>.rst` file based on the `EXT_Template.rst`, ensuring it is added to the list of
+  extensions below.
+
+* The extension will only be accepted if all conditions are met and its addition is agreed upon by the Working Group.
```
AerialMantis, line 99: Does this refer to an extension becoming a standard extension?

scripts/core/EXT.rst
```diff
@@ -0,0 +1,108 @@
...
+
+
+List of Standard Extensions
```
AerialMantis, line 102: Just a thought, maybe these lists should also track the implementation status of the extensions, so whether they are implemented, and then which version of UR supports which version of the extension, particularly for the experimental extensions which may change, so high-level languages know what they can use when updating the UR interface.

scripts/core/EXT.rst
```diff
@@ -0,0 +1,108 @@
...
+----------------------
+
+Any existing enumeration may be extended by adding new enumerators. Enumerators must use the extensions naming 
+convention and values should be assigned to avoid future compatibility issues.
+
+
+Extending Structures
+--------------------
+
+Any structure derived from `${x}_base_desc_t`` or `${x}_base_properties_t`` can be extended using a structure chain
```
EwanC, line 81: think you have an extra backtick on the end of the two types in this line

kbenzie: Given this is reStructureText, I think the start is missing a backtick, if the intent is to have these render as inline monospace.

scripts/core/EXT_Template.rst
```diff
@@ -0,0 +1,42 @@
...
+
+    * List all functions defined by the extension.
+    * ...
+
+
+Unified Runtime Extension Template
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In this section describe in detail the purpose of the extension, along with its
+valid usage.
```
EwanC, line 42: Quite like it when extensions have a version history at the bottom, could we add that to the template?

scripts/core/EXT_Template.rst
```diff
@@ -0,0 +1,42 @@
...
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+%>
+:orphan:
+
+==================================
+Unified Runtime Extension Template
+==================================
+
```
EwanC, line 14: Couple of ideas for sections in this template:
1) Do we want a section to list contributors?
2) Should we specify the version of the UR spec the extension was written against?

scripts/core/EXT_Template.rst
kbenzie: This is useful info, I do wonder if having the ability to view different extension versions side-by-side would also be useful though?

scripts/core/EXT.rst
```diff
@@ -0,0 +1,108 @@
...
+   Applications should not rely on experimental extensions in production.
+
+Requirements
+============
+
+- Extensions must use globally unique names for macros, enums, structures and functions
+- Extensions must have globally unique extension names reported from ${x}PlatformGetExtensionProperties
+- All extensions must be defined in this specification
+- Extensions must not break backwards compatibility of the core APIs
+- Standard extension versions must be backwards compatible with prior versions
```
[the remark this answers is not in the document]
veselypeta: Yes I think that makes sense. Experimental extensions should have to ABI or backwards compatibility requirements.

scripts/core/EXT_Template.rst
veselypeta: That would be really good. @kbenzie I've seen versioned docs before on the Zephyr project [link](https://docs.zephyrproject.org/latest/) - the is a dropdown in the bottom left where you can select the version of the API you want to view. I'm not really sure how this is implemented, but perhaps we can have something like that as well as a changelog (like ewan suggested).

scripts/core/EXT.rst
```diff
@@ -0,0 +1,108 @@
...
+2. **Experimental** - extensions require additional experimentation and development, before becoming a standard extension. 
+   Applications should not rely on experimental extensions in production.
+
+Requirements
+============
+
+- Extensions must use globally unique names for macros, enums, structures and functions
+- Extensions must have globally unique extension names reported from ${x}PlatformGetExtensionProperties
+- All extensions must be defined in this specification
+- Extensions must not break backwards compatibility of the core APIs
```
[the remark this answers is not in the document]
veselypeta: I think an extension can alter the behavour of an existing API. For example you might want to chain an additional structure to the input or use extension specific enums, but no extension usage should remain the same.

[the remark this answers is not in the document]
kbenzie: Hmm, I'm not sure I'm so confortable with extensions that change semantics because we have no way to opt-out with this architecture. If we were able to enable extensions, like Vulkan, maybe I'd be more conformtable with that.

scripts/core/EXT.rst
veselypeta: 👍 fixed

scripts/core/EXT_Template.rst
kbenzie: Looking at the Zephyr page I think that's for the entire page rather than just the extension specs, my gut feeling is you'd need to have a separate Sphinx build for each extension to get something exapctly like that.

veselypeta:
In the WG there was a strong response against implementing extensions in this way - especially standard extensions. Instead we will create a `Contribution Guide` which will outline how anyone will be able to add "experimental features" to extend the API for their purposes. We have not yet decided on how we should then make an "experimental feature" a core feature, but we cannot make any references to the working group as this would inhibit 3rd party contributions.

veselypeta:
Closing this in favour of Experimental Features

Issue #392, opened 2023-03-29
Update UR versions of piext*CreateWithNativeHandle entry points


kbenzie:
The `piext*CreateWithNativeHandle` entry points have an additional parameter `OwnNativeHandle` which is missing from UR versions of these entry points. Additionally in https://github.com/intel/llvm/pull/7196 a new `piextQueueGetNativeHandle2` introduces another parameter `int32_t nativeHandleDesc`.

Additionally, the `ur*CreateWithNativeHandle` entry points do not currently provide any form of extensibility which could require ABI changes or additional entry points to be introduced in future.

As such, update all `ur*CreateWithNativeHandle` entry points to include a pointer to struct parameter, which `pNext` chaining abilities, while also enabling the `OwnNativeHandle` and `nativeHandleDesc` functionality with is already present in PI.

smaslov-intel:
Note that `ur*getNativeHandle` also need to return `nativeHandleDesc`

jandres742:
So we need OwnNativeHandle in all the CreateNativeWithHandle APIs. Below you can see the interfaces I have identified currently need this change, and some extra ones:

urMemCreateWithNativeHandle: needs
    OwnNativeHandle

urKernelCreateWithNativeHandle: needs
    Program
    OwnNativeHandle
Draft PR: https://github.com/oneapi-src/unified-runtime/pull/434

urContextGetNativeHandle: needs
    NumDevices
    Devices
    OwnNativeHandle
Draft PR: https://github.com/oneapi-src/unified-runtime/pull/433

kbenzie:
Okay, these look reasonable to me. I do think that the `bool OwnNativeHandle` should exist in `ur_<object>_native_handle_props_t` structs for all the `ur<Object>CreateWithNativeHandle` entry points to enable `pNext` chaining.

jandres742:
> Okay, these look reasonable to me. I do think that the `bool OwnNativeHandle` should exist in `ur_<object>_native_handle_props_t` structs for all the `ur<Object>CreateWithNativeHandle` entry points to enable `pNext` chaining.

Thanks @kbenzie . Good idea. Please see https://github.com/oneapi-src/unified-runtime/pull/433.

jandres742:
#433 merged for urContextGetNativeHandle.

https://github.com/oneapi-src/unified-runtime/pull/434 updated for urKernelCreateWithNativeHandle

jandres742:
[...]

PR #462, opened 2023-04-25
[ur] Make native memory object creation extensible

### LINKED ISSUE #428: Add entry-point to UR to support `piextMemImageCreateWithNativeHandle`.
intel/llvm#8603 introduces `piextMemImageCreateNativeHandle`.

```cpp
__SYCL_EXPORT pi_result piextMemImageCreateWithNativeHandle(
    pi_native_handle nativeHandle, pi_context context, bool ownNativeHandle,
    const pi_image_format *ImageFormat, const pi_image_desc *ImageDesc,
    pi_mem *img);
```

We should add this entry-point to UR, but this will likely depend on #392 being resolved.

kbenzie:
This patch replaces the `urMemCreateWithNativeHandle` entry point with
two new entry points to handle buffers and images separately;
`urMemBufferCreateWithNativeHandle` and
`urMemImageCreateWithNativeHandle` (fixes #428). Both take a
`pProperites` argument to enable extensibility of native memory handle
creation via `pNext` chaining. The `ur_mem_native_properties_t`
structure includes `isNativeHandleOwned` in line with discussions in
issue #392.


scripts/core/memory.yml
```diff
@@ -434,28 +434,59 @@ params:
...
-      desc: "[in] handle of the context object"
+      desc: "[in] handle of the context object."
+    - type: const $x_mem_native_properties_t*
+      name: pProperties
+      desc: "[in][optional] pointer to native memory creation properties."
     - type: $x_mem_handle_t*
       name: phMem
-      desc: |
-            [out] pointer to the handle of the mem object created.
+      desc: "[out] pointer to handle of image object created."
```
jandres742, line 489: mem object better, no? since it wont be always an image.

kbenzie: Yes, this was a mistake from a previous iteration with two separate entry points. I'll revert to the old wording.

kbenzie: Fixed.

callumfare (approved):

jandres742 (approved):

include/ur_api.h
```diff
@@ -1949,9 +1978,10 @@ urMemGetNativeHandle(
 ///         + `NULL == phMem`
 UR_APIEXPORT ur_result_t UR_APICALL
 urMemCreateWithNativeHandle(
-    ur_native_handle_t hNativeMem, ///< [in] the native handle of the mem.
-    ur_context_handle_t hContext,  ///< [in] handle of the context object
-    ur_mem_handle_t *phMem         ///< [out] pointer to the handle of the mem object created.
+    ur_native_handle_t hNativeMem,                 ///< [in] the native handle to the memory.
+    ur_context_handle_t hContext,                  ///< [in] handle of the context object.
+    const ur_mem_native_properties_t *pProperties, ///< [in][optional] pointer to native memory creation properties.
```
jandres742, line 1983: @kbenzie : since we dont have `urMemImageCreateWithNativeHandle`, then the only way for `urMemCreateWithNativeHandle` to know we are creating a memory or an image is by checking the pNext of properties, and checking for its sType == `UR_STRUCTURE_TYPE_MEM_IMAGE_NATIVE_PROPERTIES`, to confirm it is an image. This means that Properties is not optional when the native handle is an image.

Alternatively, we might need to define `urMemImageCreateWithNativeHandle` to specifically create images.

what do you think?

kbenzie: So my thinking here was to reuse this entry point for both buffer and image native handle creation, on the buffer path the native image properties would be optional. Maybe that idea is flawed from a clarity of usage standpoint and a separate entry point should be the way forwards.

kbenzie: If this were a separate entry point, I'd probably move the non-optional image format/desc to function arguments and reused the `ur_mem_native_properties_t` for `isNativeHandleOwned`.

kbenzie: I've made this change, and renamed `urMemCreateWithNativeHandle` to `urMemBufferCreateWithNativeHandle` to match `urMemImageCreateWithNativeHandle`.

veselypeta (approved):
LGTM

PR #473, opened 2023-04-26
Add platform discovery search paths

kswiecicki:
This patch makes the loader search additional directories when searching for hardcoded adapter libraries.
Additional search paths can be added via `UR_ADAPTERS_SEARCH_PATH` variable that holds a comma-separated list of directory paths (e.g. `UR_ADAPTERS_SEARCH_PATH=/home/kswiecic/dev/sycl_workspace/libs/`).
The loader now also searches the directory it was linked from.

Related issue: https://github.com/oneapi-src/unified-runtime/issues/128


source/loader/ur_adapter_registry.hpp
```diff
@@ -88,9 +83,56 @@ class AdapterRegistry {
...
+
+    std::vector<fs::path> getEnvAdapterSearchPaths() {
+        return getEnvVarValuesToVec<fs::path>("UR_ADAPTERS_SEARCH_PATH", ',');
+    }
+
     void discoverKnownAdapters() {
-        for (const auto &path : knownPlatformNames) {
-            discovered_adapters.emplace_back(path);
+        auto searchPaths = getEnvAdapterSearchPaths();
+        auto loaderPath = getLoaderLibDirectory();
```
bmyates, line 117: Would this result in adapter libraries in loader lib directory to take precedence over libraries discovered in LD_LIBRARY_PATH?

As a user I would expect LD_LIBRARY_PATH to have highest search precedence

source/loader/ur_adapter_registry.hpp
```diff
@@ -88,9 +83,56 @@ class AdapterRegistry {
     static constexpr std::array<const char *, 1> knownPlatformNames{
         MAKE_LIBRARY_NAME("ur_adapter_level_zero", "0")};
 
+    template <typename T, typename = std::enable_if_t<
+                              std::is_constructible_v<T, const std::string &>>>
+    std::vector<T> getEnvVarValuesToVec(const std::string envVar,
```
igchor, line 88: I think we already have a function for parsing env var to a vector in ur_utils.hpp: [getenv_to_vec](https://github.com/oneapi-src/unified-runtime/blob/main/source/common/ur_util.hpp#L132)

You could just map that vector of strings to vector of fs::path by doing something like: 
```
std::vector<fs::path> paths;
std::transform(s.cbegin(), s.cend(), std::back_inserter(paths), [](const std::string &s) { return fs::path(s); });
```

source/loader/ur_adapter_registry.hpp
```diff
@@ -88,9 +82,54 @@ class AdapterRegistry {
...
+        std::vector<fs::path> searchPaths;
+
+        auto searchPathsEnvOpt = getEnvAdapterSearchPaths();
+        if (searchPathsEnvOpt.has_value()) {
+            auto searchPathsEnv = searchPathsEnvOpt.value();
+            std::copy(searchPathsEnv.begin(), searchPathsEnv.end(),
+                      std::back_inserter(searchPaths));
+        }
+
+        auto searchPathsOSOpt = getOSSearchPaths();
```
igchor, line 110: I'm actually not sure we should have the same search order for both linux and Windows. On Windows, `PATH` is expected to be checked last I think (after the current directory): https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order?redirectedfrom=MSDN#search_order_for_desktop_applications

kswiecicki: I've grouped the current directory, environmental variable paths, and system directory under the `getOSSearchPaths(...)` function and resolved the search order there.

source/loader/ur_adapter_registry.hpp
kswiecicki: I've set the highest search precedence to `UR_ADAPTERS_SEARCH_PATH` and made `LD_LIBRARY_PATH` a higher priority than the current loader lib directory.

source/loader/ur_adapter_registry.hpp
kswiecicki: Done.

source/loader/windows/loader_location.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    std::optional<std::string> pathStringsOpt;
+    try {
+        pathStringsOpt = ur_getenv("PATH");
+    } catch (const std::invalid_argument &e) {
+        return std::nullopt;
+    }
+
+    if (pathStringsOpt.has_value()) {
+        std::cout << "PATH: " << *pathStringsOpt << std::endl;
```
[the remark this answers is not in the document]
kswiecicki: I've left that by accident. I'm only logging exception errors.
I could add a log message when we successfully load an adapter with `LibLoader::loadAdapterLibrary(name.c_str())` if you think it's necessary.

source/loader/linux/loader_location.cpp
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    return std::nullopt;
+}
+
+static std::optional<std::vector<fs::path>> getEnvOSSearchPaths() {
+    std::optional<std::string> pathStringsOpt;
+    try {
+        pathStringsOpt = ur_getenv("LD_LIBRARY_PATH");
```
igchor, line 34: I looked at the documentation for ld.so.8 (https://man7.org/linux/man-pages/man8/ld.so.8.html) and I'm actually not sure that interpreting LD_LIBRARY_PATH within UR is a good idea. There are a few special cases that we might not cover by interpreting it ourselves:
- "This variable is ignored in secure-execution mode"
- "Within the pathnames specified in LD_LIBRARY_PATH, the
              dynamic linker expands the tokens $ORIGIN, $LIB, and
              $PLATFORM (or the versions using curly braces around the
              names) as described above in Dynamic string tokens."
- The directory lists can also be separated by using semicolons
              
I think it would be safer to let the linker handle `LD_LIBRARY_PATH`.

igchor: There might be something similiar on Windows as well.

source/loader/ur_adapter_registry.hpp
```diff
@@ -88,9 +88,64 @@ class AdapterRegistry {
...
+        if (searchPathsOSOpt.has_value()) {
+            auto searchPathsOS = searchPathsOSOpt.value();
+            std::copy(searchPathsOS.begin(), searchPathsOS.end(),
+                      std::back_inserter(searchPaths));
+        }
+
+        for (const auto &platformName : knownPlatformNames) {
```
PatKamin, line 138: nit: platformName -> adapterName

test/loader/adapter_registry/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: MIT
+
+set(DUMMY_LIB_DIR ${CMAKE_BINARY_DIR}/bin/${CMAKE_BUILD_TYPE}/adapter-reg)
```
PatKamin, line 4: This has no effect as this variable is set again below.

test/loader/adapter_registry/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+    set(DUMMY_LIB_DIR ${CMAKE_BINARY_DIR}/bin/adapter-reg)
+endif()
+
+if(WIN32)
+    set(DUMMY_LIB_PATH ${DUMMY_LIB_DIR}/ur_adapter_level_zero.dll)
+else()
+    set(DUMMY_LIB_PATH ${DUMMY_LIB_DIR}/libur_adapter_level_zero.so.0)
+endif()
+
+# Create a dummy adapter library in adapter the runtime output directory
```
PatKamin, line 18: in the adapter runtime output directory?

test/loader/adapter_registry/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: MIT
+
+set(DUMMY_LIB_DIR ${CMAKE_BINARY_DIR}/bin/${CMAKE_BUILD_TYPE}/adapter-reg)
+
+if(MSVC)
+    set(DUMMY_LIB_DIR ${CMAKE_BINARY_DIR}/bin/${CMAKE_BUILD_TYPE}/adapter-reg)
```
PatKamin, line 7: Please, check if both MSVC and non-MSVC paths could be resolved with `CMAKE_CURRENT_BINARY_DIR`:
```
${CMAKE_CURRENT_BINARY_DIR}/adapter-reg
```

test/loader/adapter_registry/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+endif()
+
+if(WIN32)
+    set(DUMMY_LIB_PATH ${DUMMY_LIB_DIR}/ur_adapter_level_zero.dll)
+else()
+    set(DUMMY_LIB_PATH ${DUMMY_LIB_DIR}/libur_adapter_level_zero.so.0)
+endif()
+
+# Create a dummy adapter library in adapter the runtime output directory
+file(WRITE ${DUMMY_LIB_PATH} "")
```
PatKamin, line 19: Since this will be created in the UR bin folder, consider removing it after tests finish. You might need to create a cmake script for that.

source/common/ur_util.hpp
```diff
@@ -119,8 +122,8 @@ static void throw_wrong_format_map(const char *env_var_name,
 
 /// @brief Get a vector of values from an environment variable \p env_var_name
 ///        A comma is a delimiter for extracting values from env var string.
-///        Colons and semicolons are not allowed to align with the similar
-///        getenv_to_map() util function and avoid confusion.
+///        Colons and semicolons are allowed only inside quotes to align with
```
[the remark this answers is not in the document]
kswiecicki: Done.

source/loader/linux/loader_location.cpp
kswiecicki: Since that was the case, I've changed the adapter registry to store a sorted list of paths that should be searched when looking for each hardcoded adapter. Eg. 1. Paths from `UR_ADAPTERS_SEARCH_PATH`. 2. Hardcoded library name (in that case the OS searches for this library). 3. UR Loader directory (current binary/library directory).

source/common/ur_util.hpp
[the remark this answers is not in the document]
kswiecicki: Done.

source/loader/linux/loader_location.cpp
```diff
@@ -0,0 +1,81 @@
+/*
+ *
+ * Copyright (C) 2023 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+
+#include <dlfcn.h>
+#include <filesystem>
```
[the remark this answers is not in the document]
kswiecicki: Done.

test/loader/adapter_registry/CMakeLists.txt
kswiecicki: I've removed this part.

test/loader/adapter_registry/CMakeLists.txt
kswiecicki: It's no longer relevant.

test/loader/adapter_registry/CMakeLists.txt
kswiecicki: It's no longer relevant.

source/loader/ur_adapter_registry.hpp
kswiecicki: Done.

source/loader/windows/loader_location.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <filesystem>
+namespace filesystem = std::filesystem;
+#else
+#include <experimental/filesystem>
+namespace filesystem = std::experimental::filesystem;
+#endif #include < filesystem>
+
+#include "ur_loader.hpp"
+
+#define MAX_PATH_LEN_WIN 1024
```
bratpiorka, line 24: why 1024? MAX_PATH is 256 (default) or 32K (need changes in registry)

kswiecicki: That was an arbitrary number that I was going to change in the future but forgot.
I've made the buffer bigger, equal to the maximum possible path length according to https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry.

bratpiorka (approved):

source/loader/ur_adapter_registry.hpp
```diff
@@ -56,41 +59,104 @@ class AdapterRegistry {
...
+            }
+
+            loadPaths.emplace_back(fs::path(adapterName));
+
+            if (loaderLibPathOpt.has_value()) {
+                auto loaderLibPath = loaderLibPathOpt.value();
+                loadPaths.emplace_back(loaderLibPath / adapterName);
+            }
+
+            adaptersLoadPaths.emplace_back(loadPaths);
```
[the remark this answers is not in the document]
kswiecicki: So that I would have a clear division between paths for different hardcoded adapters.

source/loader/ur_adapter_registry.hpp
```diff
@@ -56,41 +59,104 @@ class AdapterRegistry {
...
+        if (pathStringsOpt.has_value()) {
+            std::vector<fs::path> paths;
+            auto pathStrings = pathStringsOpt.value();
+
+            std::transform(pathStrings.cbegin(), pathStrings.cend(),
+                           std::back_inserter(paths),
+                           [](const std::string &s) { return fs::path(s); });
+            paths.erase(std::remove_if(paths.begin(), paths.end(),
+                                       [](const fs::path &path) {
+                                           return !fs::exists(path);
```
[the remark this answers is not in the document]
kswiecicki: Done.

scripts/core/INTRO.rst
```diff
@@ -260,6 +260,15 @@ Specific environment variables can be set to control the behavior of unified run
 
     This environment variable should be used for development and debugging only.
```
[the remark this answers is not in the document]
kswiecicki: Done. I've also added this information to the `UR_ADAPTERS_SEARCH_PATH` notes.

source/loader/ur_adapter_registry.hpp
```diff
@@ -56,41 +59,104 @@ class AdapterRegistry {
...
+        return adaptersLoadPaths.cbegin();
+    }
+
+    std::vector<std::vector<fs::path>>::const_iterator cend() const noexcept {
+        return adaptersLoadPaths.cend();
     }
 
   private:
-    std::vector<std::string> discovered_adapters;
+    std::vector<std::vector<fs::path>> adaptersLoadPaths;
```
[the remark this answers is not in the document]
kswiecicki: Done.

pbalcer (approved):

PR #474, opened 2023-04-26
Expand urEventCreateWithNativeHandle

jandres742:
Add needed members for PI:

OwnNativeHandle

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -219,6 +219,7 @@ class ur_structure_type_v(IntEnum):
     KERNEL_NATIVE_PROPERTIES = 17                   ## ::ur_kernel_native_properties_t
     QUEUE_NATIVE_PROPERTIES = 18                    ## ::ur_queue_native_properties_t
     MEM_NATIVE_PROPERTIES = 19                      ## ::ur_mem_native_properties_t
+    EVENT_NATIVE_PROPERTIES = 20                    ## ::ur_event_native_properties_t
 
 class ur_structure_type_t(c_int):
     def __str__(self):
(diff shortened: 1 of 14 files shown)


jandres742:
@kbenzie , @pbalcer : do you know why the CI is complaining about format errors in a lot of functions and files not related to my change? like:

```
+## @brief Function-pointer for urInit
+if __use_win_types:
+    _urInit_t = WINFUNCTYPE( ur_result_t, ur_device_init_flags_t )
+else:
+    _urInit_t = CFUNCTYPE( ur_result_t, ur_device_init_flags_t )
+
```

jandres742:
I dont get any errors with `python3.exe ./run.py --html` or just make. I haven't been able to run ever other build or generate commands (just config issues on my side).

kbenzie:
Okay, rather than only `python3.exe ./run.py --html` the equivelent to the `generate` target is the follow commands from CMake:

From the `generate-code` target (which should always be available):
https://github.com/oneapi-src/unified-runtime/blob/3bc0f7582453038361d52d26198b3488ed66e6a3/CMakeLists.txt#L229-L230

From the `cppformat` target (which is only enabled when `-DUR_FORMAT_CPP_STYLE=ON` and version 15 of `clang-format` is found, this should be available from the packages installed with [third_party/requirements.txt](https://github.com/oneapi-src/unified-runtime/blob/main/third_party/requirements.txt#L39):
https://github.com/oneapi-src/unified-runtime/blob/3bc0f7582453038361d52d26198b3488ed66e6a3/cmake/helpers.cmake#L30-L33

I think its probably easier to try and fix your local config than run these commands manually.

jandres742:
@kbenzie ready for review. I added a commit addressing the format issues in the CI (I was able to run the generate target locally, but couldn't find reason for that format change).

There are two commits, so it is easy to see the changes associated with the native handle APIs and the changes for the format.

kbenzie (approved):
LGTM

callumfare (approved):
LGTM

Issue #374, opened 2023-03-24
Missing `ur_device_info_t` enum values


callumfare:
There is no equivalent for the following PI enums in UR:
```
PI_DEVICE_INFO_BUILD_ON_SUBDEVICE
PI_EXT_ONEAPI_DEVICE_INFO_MAX_WORK_GROUPS_3D
PI_EXT_INTEL_DEVICE_INFO_MEMORY_BUS_WIDTH
PI_EXT_ONEAPI_DEVICE_INFO_CUDA_ASYNC_BARRIER
```

It's not clear (to me anyway) whether UR has exact equivalents for these:
```
PI_DEVICE_INFO_GPU_HW_THREADS_PER_EU
PI_DEVICE_INFO_GPU_EU_COUNT_PER_SUBSLICE
```

Ideally we need equivalents so pi2ur can map the device info properly. #55 covered this, but the PR that resolved this doesn't seem to have covered all the missing values.

kbenzie:
https://github.com/intel/llvm/pull/9203 will also add `PI_EXT_INTEL_DEVICE_INFO_MEM_CHANNEL_SUPPORT` or similar.

kbenzie:
Fixed by #478.

PR #460, opened 2023-04-24
Change bool to ur_bool_t in some places

bmyates:
Only use ur_bool_t in ur_device_info_t for consistency

Signed-off-by: Brandon Yates <brandon.yates@intel.com>

PR #478, opened 2023-04-27
Added missing enums for ur_device_info_t.

martygrant:
See #374

### DIFF
--- a/test/conformance/device/urDeviceGetInfo.cpp
+++ b/test/conformance/device/urDeviceGetInfo.cpp
@@ -33,7 +33,7 @@ static std::unordered_map<ur_device_info_t, size_t> device_info_size_map = {
     {UR_DEVICE_INFO_MEMORY_CLOCK_RATE, sizeof(uint32_t)},
     {UR_DEVICE_INFO_ADDRESS_BITS, sizeof(uint32_t)},
     {UR_DEVICE_INFO_MAX_MEM_ALLOC_SIZE, sizeof(uint64_t)},
-    {UR_DEVICE_INFO_IMAGE_SUPPORTED, sizeof(bool)},
+    {UR_DEVICE_INFO_IMAGE_SUPPORTED, sizeof(ur_bool_t)},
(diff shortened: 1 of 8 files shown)


veselypeta (approved):
LGTM!
Just one suggestion: could you also add these enums in the CTS test [here](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/device/urDeviceGetInfo.cpp#L8)

test/conformance/device/urDeviceGetInfo.cpp
```diff
@@ -101,6 +104,10 @@ static std::unordered_map<ur_device_info_t, size_t> device_info_size_map = {
     {UR_DEVICE_INFO_BFLOAT16, sizeof(bool)},
     {UR_DEVICE_INFO_MAX_COMPUTE_QUEUE_INDICES, sizeof(uint32_t)},
     {UR_DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS, sizeof(bool)},
+    {UR_DEVICE_INFO_MEMORY_BUS_WIDTH, sizeof(bool)},
+    {UR_DEVICE_INFO_MAX_WORK_GROUPS_3D, sizeof(uint32_t)},
+    {UR_DEVICE_INFO_ASYNC_BARRIER, sizeof(bool)},
+    {UR_DEVICE_INFO_MEM_CHANNEL_SUPPORT, sizeof(bool)},
```
kbenzie, line 110: I beleive all instances of `bool` here should actually be `ur_bool_t`, the are the same size but since we are testing the spec we should use exactly the same type. Might as well fix them all while touching this file.

kbenzie: https://github.com/oneapi-src/unified-runtime/pull/460 was did a similar cleanup.

martygrant: Just copied what was already in the file but that makes sense! Changed them over now, squashed my commits down to make it a little cleaner.

Issue #400, opened 2023-03-31
Enable support for host pipe read/write commands


kbenzie:
In https://github.com/intel/llvm/pull/7468 two enqueue commands were introduced:

* `piextEnqueueReadHostPipe`
* `piextEnqueueWriteHostPipe`

These are only implemented in the OpenCL plugin and make use of the [cl_intel_program_scope_host_pipe](https://github.com/KhronosGroup/OpenCL-Docs/pull/892) vendor extension. All other adapters will fail/crash if these entry points are used.

I'm unsure of the utility of these commands beyond OpenCL devices which support pipes, which if I'm not mistaken is generally seen as an FPGA feature.

UR will need to enable this feature, however its unclear to me if this should be in the core spec or defined as an extension devices can opt into supporting in a similar way to the OpenCL vendor extension.

PR #486, opened 2023-05-01
[UR] Enable support for host pipe read/write commands

martygrant:
See #400

Success testing is not included in this PR as it requires a very specific setup (typically only FPGAs support the pipes feature so a supported device and kernel would be required). A follow up ticket will be created for this.

### DIFF
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -377,8 +377,8 @@ etors:
       desc: "[$x_bool_t] return true if Async Barrier is supported"
     - name: MEM_CHANNEL_SUPPORT
       desc: "[$x_bool_t] return true if specifying memory channels is supported"
-    - name: HOST_PIPE_RW_SUPPORTED
-      desc: "[$x_bool_t] Return true if the device supports reading from and writing to host pipes."
+    - name: HOST_PIPE_READ_WRITE_SUPPORTED
+      desc: "[$x_bool_t] Return true if the device supports enqueing commands to read and write pipes from the host."
 --- #--------------------------------------------------------------------------
 type: function
 desc: "Retrieves various information about device"
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -377,6 +377,8 @@ etors:
       desc: "[$x_bool_t] return true if Async Barrier is supported"
     - name: MEM_CHANNEL_SUPPORT
       desc: "[$x_bool_t] return true if specifying memory channels is supported"
+    - name: HOST_PIPE_RW_SUPPORTED
+      desc: "[$x_bool_t] Return true if the device supports reading from and writing to host pipes."
 --- #--------------------------------------------------------------------------
 type: function
 desc: "Retrieves various information about device"
--- a/scripts/core/registry.yml
+++ b/scripts/core/registry.yml
@@ -211,6 +211,9 @@ etors:
 - name: MEM_GET_NATIVE_HANDLE
   desc: Enumerator for $xMemGetNativeHandle
   value: '68'
+- name: ENQUEUE_READ_HOST_PIPE
+  desc: Enumerator for $xEnqueueReadHostPipe
+  value: '69'
 - name: MEM_GET_INFO
   desc: Enumerator for $xMemGetInfo
   value: '70'
@@ -352,3 +355,6 @@ etors:
 - name: MEM_IMAGE_CREATE_WITH_NATIVE_HANDLE
   desc: Enumerator for $xMemImageCreateWithNativeHandle
   value: '116'
+- name: ENQUEUE_WRITE_HOST_PIPE
+  desc: Enumerator for $xEnqueueWriteHostPipe
+  value: '117'
--- a/test/conformance/enqueue/urEnqueueReadHostPipe.cpp
+++ b/test/conformance/enqueue/urEnqueueReadHostPipe.cpp
@@ -7,66 +7,64 @@ using urEnqueueReadHostPipeTest = uur::urHostPipeTest;
 UUR_INSTANTIATE_DEVICE_TEST_SUITE_P(urEnqueueReadHostPipeTest);
 
 TEST_P(urEnqueueReadHostPipeTest, InvalidNullHandleQueue) {
-    bool blocking = true;
     uint32_t numEventsInWaitList = 0;
     ur_event_handle_t phEventWaitList;
     ur_event_handle_t *phEvent = nullptr;
 
     ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_NULL_HANDLE,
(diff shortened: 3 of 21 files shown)


scripts/core/device.yml
```diff
@@ -785,3 +785,15 @@ etors:
...
+desc: "Return true if the device supports reading from and writing to host pipes."
+class: $xDevice
+name: HostPipeRWSupported
+ordinal: "0"
+details:
+    - "x"
+params:
+    - type: $x_device_handle_t
+      name: hDevice
+      desc: "[in] handle of the device instance"
```
kbenzie, line 799: The way the generator is set up I don't think its possible to specify anything other than `ur_result_t`, so you'd need to have an our parameter to tell the user the answer to this query.

Rather than a new entry point, I think this should be queriable via [urDeviceGetInfo](https://oneapi-src.github.io/unified-runtime/core/api.html#urdevicegetinfo) by adding a new enumeration to [ur_device_info_t](https://oneapi-src.github.io/unified-runtime/core/api.html#_CPPv416ur_device_info_t), e.g. `UR_DEVICE_INFO_HOST_PIPE_RW_SUPPORTED` with an out parameter type of `ur_bool_t`.

martygrant: Ah I was just assuming I could make it return true/false. That's a great suggestion thanks I'll update it to work like that!

scripts/core/registry.yml
```diff
@@ -211,6 +211,9 @@ etors:
 - name: MEM_GET_NATIVE_HANDLE
   desc: Enumerator for $xMemGetNativeHandle
   value: '68'
+- name: ENQUEUE_READ_HOST_PIPE
+  desc: Enumerator for $xEnqueueReadHostPipe
+  value: '69'
```
omarahmed1111, line 216: [NIT suggestion] maybe it is better to move that at the end with `ENQUEUE_WRITE_HOST_PIPE` for more clarity, but not sure of that as that will leave the gap here empty.

test/conformance/enqueue/urEnqueueReadHostPipe.cpp
```diff
@@ -0,0 +1,55 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: MIT
+#include <uur/fixtures.h>
+
+using urEnqueueReadHostPipeTest = uur::urHostPipeTest;
+
+UUR_INSTANTIATE_DEVICE_TEST_SUITE_P(urEnqueueReadHostPipeTest);
+
```
omarahmed1111, line 8: just out of curiosity, is there something that blocks us from adding the success test?

test/conformance/testing/include/uur/fixtures.h
```diff
@@ -192,6 +192,47 @@ struct urQueueTest : urContextTest {
...
+struct urHostPipeTest : urQueueTest {
+    void SetUp() override {
+        UUR_RETURN_ON_FATAL_FAILURE(urQueueTest::SetUp());
+        uur::KernelsEnvironment::instance->LoadSource("foo", 0, il_binary);
+        ASSERT_SUCCESS(urProgramCreateWithIL(
+            context, il_binary->data(), il_binary->size(), nullptr, &program));
+
+        size_t size = 0;
+        ASSERT_SUCCESS(urDeviceGetInfo(
+            device, UR_DEVICE_INFO_HOST_PIPE_RW_SUPPORTED, 0, nullptr, &size));
```
omarahmed1111, line 204: I think if `UR_DEVICE_INFO_HOST_PIPE_RW_SUPPORTED` is not supported we could skip the test with:
```
if (!supported) {
        GTEST_SKIP() << "[Feature tested] is not supported.";
}
```

test/conformance/testing/include/uur/fixtures.h
```diff
@@ -192,6 +192,47 @@ struct urQueueTest : urContextTest {
...
+
+    const char *pipe_symbol = "pipe_symbol";
+    bool blocking = true;
+
+    static const size_t size = 1024;
+    char buffer[size];
+
+    uint32_t numEventsInWaitList = 0;
+    ur_event_handle_t phEventWaitList;
+    ur_event_handle_t *phEvent = nullptr;
```
omarahmed1111, line 233: I think also this parameters will be needed with each test to define a new event and waitEvent, so I think we could leave their definition inside the test itself.

scripts/core/registry.yml
martygrant: So this file gets populated automatically when you run the `generate` target, pretty strange why it's decided to put it there. I just tried removing both and it has added them in the same places. I don't think it's too important as long as they're both present in the ifle.

omarahmed1111: Ah okay, I didn't notice that it is auto-generated, just ignore that.

test/conformance/testing/include/uur/fixtures.h
```diff
@@ -192,6 +192,47 @@ struct urQueueTest : urContextTest {
...
+        }
+        UUR_RETURN_ON_FATAL_FAILURE(urQueueTest::TearDown());
+    }
+
+    std::shared_ptr<std::vector<char>> il_binary;
+    std::string program_name = "foo";
+    ur_program_handle_t program = nullptr;
+
+    const char *pipe_symbol = "pipe_symbol";
+    bool blocking = true;
```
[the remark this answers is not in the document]
martygrant: Sure will move this to the individual tests.

test/conformance/testing/include/uur/fixtures.h
martygrant: Moving these too.

omarahmed1111 (approved):
LGTM.
I prefer the `blocking` parameter as an inlined parameter passed to the entry-point like `TEST_P(urEnqueueUSMMemcpyTest, Blocking)` test, but that's just a personal stylish stuff, feel free to ignore.

scripts/core/device.yml
```diff
@@ -377,6 +377,8 @@ etors:
       desc: "[$x_bool_t] return true if Async Barrier is supported"
     - name: MEM_CHANNEL_SUPPORT
       desc: "[$x_bool_t] return true if specifying memory channels is supported"
+    - name: HOST_PIPE_RW_SUPPORTED
```
kbenzie, line 380: I feel this would be more readable replacing `RW` with `READ_WRITE`.

scripts/core/device.yml
```diff
@@ -377,6 +377,8 @@ etors:
...
+      desc: "[$x_bool_t] Return true if the device supports reading from and writing to host pipes."
```
kbenzie, line 381:
```suggestion
      desc: "[$x_bool_t] Return true if the device supports enqueing commands to read and write pipes from the host."
```

PR #493, opened 2023-05-04
[ur] Print log messages in a thread-safe way.

PatKamin:


source/common/logger/ur_sinks.hpp
```diff
@@ -48,26 +49,41 @@ class Sink {
         flush_level = logger::Level::ERR;
     }
 
+    void flush_buffer() {
+        std::scoped_lock<std::mutex> output_lock(output_mutex);
+        get_buffer().sync();
```
igchor, line 54: does this sync/flush do anything on stringstream?

source/common/logger/ur_sinks.hpp
```diff
@@ -48,26 +49,41 @@ class Sink {
...
+        get_buffer().str("");
+    }
+
   private:
     std::string logger_name;
     bool skip_prefix;
+    std::mutex output_mutex;
+
+    std::stringstream &get_buffer() {
+        static thread_local std::stringstream buffer;
```
igchor, line 67: If this is inside the Sink class, then I think all instances of Sink will share it (including Stdout and File Sinks)- is this what you want?

source/common/logger/ur_sinks.hpp
PatKamin: Seems like it's unnecessary, removed.

source/common/logger/ur_sinks.hpp
```diff
@@ -25,11 +26,15 @@ class Sink {
...
-            *ostream << "[" << level_to_str(level) << "]: ";
+            get_buffer() << "<" << logger_name << ">";
+            get_buffer() << "[" << level_to_str(level) << "]: ";
         }
         format(fmt, std::forward<Args &&>(args)...);
-        *ostream << "\n";
+        get_buffer() << "\n";
+
+        std::scoped_lock<std::mutex> output_lock(output_mutex);
+        *ostream << get_buffer().str();
```
igchor, line 36: Shouldn't this also be under `if (level >= flush_level)`? Otherwise, what's the point of keeping the buffer as thread local if we always write the entire buffer to the stream?

source/common/logger/ur_sinks.hpp
PatKamin: Moved buffer implementation to derived classes, so that every Sink type has unique buffer now. However, still the buffer can be shared between logger instances with the same sink types.

source/common/logger/ur_sinks.hpp
PatKamin: What is achieved with current implementation is much better granularity of writing to the output stream. Now the whole message coming from a single log() call goes to an ostream, not a single character as it was before. This has a great impact on printing messages from multiple threads.

I left this out of the flush_level guard for now, as I couldn't make it work properly in cases where there are messages with level lower than the flush level only (thus, during an application run there is no message firing up such a flush). I find it reasonable to leave it this way, as this is also the current behavior of Logger (flushing generally is maintained by an ostream, additionally we can have a guarantee for a flush for specific level messages).

igchor: Right, but do you need `thread_local` storage, then? If you clear the buffer after each log() anyway, it could be just a local variable, right? Or is it much slower?

source/common/logger/ur_sinks.hpp
```diff
@@ -24,12 +24,15 @@ class Sink {
...
-            *ostream << "[" << level_to_str(level) << "]: ";
+            buffer << "<" << logger_name << ">";
+            buffer << "[" << level_to_str(level) << "]: ";
         }
-        format(fmt, std::forward<Args &&>(args)...);
-        *ostream << "\n";
+        buffer << format(fmt, std::forward<Args &&>(args)...);
+        buffer << "\n";
+
+        *ostream << buffer.str();
```
pbalcer, line 35: you are now doing two allocations, one at each step. I think it would be better to have it all reside in the format function so that we avoid one.

test/unit/logger/logger.cpp
```diff
@@ -113,9 +114,129 @@ TEST_F(LoggerWithFileSink, NestedFilePath) {
...
+TEST_P(FileSinkLoggerMultipleThreads, Multithreaded) {
+    std::vector<std::thread> threads;
+    auto local_logger = logger::Logger(
+        logger::Level::WARN,
+        std::make_unique<logger::FileSink>(logger_name, file_path, true));
+
+    // Messages below the flush level
+    for (int i = 0; i < thread_count; i++) {
+        threads.emplace_back([&]() {
+            for (int j = 0; j < thread_count; ++j) {
```
pbalcer, line 137: hm, `message_count`?

source/common/logger/ur_sinks.hpp
PatKamin: Restructured a bit.

pbalcer: I'm not sure if this is any better. I thought we'd be able to `format("{} ... - ...", actually assemble the prefix here, rest of the args);`. But now that I look at it, I'm not sure if that's possible.
I think we could make the format function optionally take a `ostringstream` instance, and do it like this:
```
std::ostringstream oss;

if (!skip_prefix) {
  oss << "<" << logger_name << ">" << "[" << level_to_str(level) << "]: ";
}

format_oss(oss, fmt, std::forward<Args>(args)...);
{
  scoped_lock();
  *ostream << buffer;
  ...
}
```

PatKamin: Added an ostringstream as a required argument (as format() functions are not used in any other place anyway) so that there is only one object used as a buffer.

pbalcer (approved):

test/unit/logger/logger.cpp
```diff
@@ -113,9 +114,98 @@ TEST_F(LoggerWithFileSink, NestedFilePath) {
...
+//////////////////////////////////////////////////////////////////////////////
+INSTANTIATE_TEST_SUITE_P(
+    ThreadCount, CommonLoggerWithMultipleThreads,
+    ::testing::Values(2, std::thread::hardware_concurrency()));
+TEST_P(CommonLoggerWithMultipleThreads, StdoutMultithreaded) {
+    std::vector<std::thread> threads;
+    auto local_logger =
+        logger::Logger(logger::Level::WARN,
+                       std::make_unique<logger::StdoutSink>("test", true));
+    constexpr int message_count = 50;
```
kswiecicki, line 182: We could make the `message_count` a test suite parameter like the `thread_count`.

PR #499, opened 2023-05-08
[ur] Handle exceptions in each API function

igchor:
We don't catch all possible exceptions (especially std::bad_alloc when using std containers). This change adds checks to each API function and return appropriate status in case of any exception.

### DIFF
--- a/scripts/templates/libapi.cpp.mako
+++ b/scripts/templates/libapi.cpp.mako
@@ -53,7 +53,7 @@ ${th.make_func_name(n, tags, obj)}(
     ${line}
     %endfor
     )
-{
+try {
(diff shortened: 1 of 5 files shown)


pbalcer (commented):
hm, we should do this either for ur_libapi.cpp entry points (the boundary between C and C++) or in the layers/ddi (which would let us catch these exceptions earlier). I see little point in doing both.

igchor:
Ok, I forgot we compile everything into a single shared lib. There is some ur_libapi.cpp-specific code that can throw (e.g. Context initialization) so I think it would be best to put those catches there.

igchor:
Done

pbalcer (approved):

PR #523, opened 2023-05-18
[ci] Add CodeQL scans

PatKamin:
Add CodeQL scans to make it possible to catch issues during the PR refinement as opposed to Coverity scans, which can run on the main branch only.

A sample execution of this workflow: https://github.com/PatKamin/unified-runtime/actions/runs/5011460219/jobs/8982314529

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,72 @@
+name: "CodeQL"
+
+on: [push, pull_request]
+
+jobs:
+  analyze-ubuntu:
+    name: Analyze on Ubuntu
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp', 'python' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+
+    - name: Install pip packages
+      run: pip install -r third_party/requirements.txt
+
+    - name: Configure CMake
+      run: cmake -B ${{github.workspace}}/build -DUR_DEVELOPER_MODE=ON -DUR_BUILD_TESTS=ON -DUR_ENABLE_TRACING=ON -DUR_BUILD_TOOLS=ON
+
+    - name: Build
+      run: cmake --build ${{github.workspace}}/build -j $(nproc)
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
+      with:
+        category: "/language:${{matrix.language}}"
+
+  analyze-windows:
+    name: Analyze on Windows
+    runs-on: windows-latest
+    permissions:
+      security-events: write
+
+    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,81 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v2
```
lukaszstolarczuk, line 43: why autobuild and not a custom one (like in other jobs)? you may have no control of used CMake params, I guess...?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,81 @@
+name: "CodeQL"
```
lukaszstolarczuk, line 1: do you have a preview of this job's execution?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,81 @@
...
+
+on:
+  push:
+    paths:
+      - '**/*.c'
```
lukaszstolarczuk, line 6: why this 'paths' limitations? what about if you change e.g. something in CMake that will affect the compilation...? Is this job too long to be executed for all pushes/PRs?

.github/workflows/codeql.yml
pbalcer: Right, we should build with all optional things enabled (like we do for other CI jobs).

PatKamin: I analyzed params used in cmake.yml and I think that there is no param worth explicit set for CodeQL scan (for linux build at least), hence the Autobuild:
```
CMAKE_C_COMPILER/CMAKE_CXX_COMPILER - I think that building with any compiler should suffice,
UR_ENABLE_TRACING=ON - triggers third-party lib build, no need to scan that,
UR_DEVELOPER_MODE=ON - I assume that triggering additional compiler warnings do not affect CodeQL scan,
CMAKE_BUILD_TYPE - one of Debug/Release should suffice, chose Release, again only additional compiler flags are set in Debug mode,
UR_BUILD_TESTS=ON - worth having enabled, but it's enabled by default anyway,
UR_FORMAT_CPP_STYLE=ON - linting shouldn't affect this workflow
```

.github/workflows/codeql.yml
PatKamin: Here is a sample execution of this workflow: https://github.com/PatKamin/unified-runtime/actions/runs/5011460219/jobs/8982314529

.github/workflows/codeql.yml
PatKamin: Yes, it's quite long, ca. 20 minutes. And it's running for every push, so I decided to limit this workflow somehow to not run on changes to docs.

As to the cmake, you're right - I can just add CMakeLists.txt to this list. What do you think?

.github/workflows/codeql.yml
lukaszstolarczuk: also, I can see there's plan `make` command executed - could we make use of parallel building, like `make -j$(nproc)` ?

PR #525, opened 2023-05-18
[ur] Introduce virtual memory interfaces

### LINKED ISSUE #357: Expose Low-Level Memory Management API (with Virtual Memory reservation)
Rationale (from L0 doc):
If an application needs finer grained control of physical memory consumption for device allocations then it can reserve a range of the virtual address space and map this to physical memory as needed. This provides flexibility for applications to manage large dynamic data structures which can grow and shrink over time while maintaining optimal physical memory usage.

Such API is present in CUDA and level-zero:
https://developer.nvidia.com/blog/introducing-low-level-gpu-virtual-memory-management/
https://spec.oneapi.io/level-zero/latest/core/PROG.html#reserved-device-allocations

One option is to implement API similar to the one in CUDA/level-zero. Alternatively, we could also leverage UMA memory provider abstraction. Memory provider is expected to support virtual memory reservations and physical allocations anyway.

kbenzie:
Fixes #357 by introducing a port of the PI changes from intel/llvm#8954
which add the ability to reserve virtual memory regions separately from
physical memory backing allocations.

* [x] Define interfaces in the spec & header
* [ ] Add conformance tests exercising interfaces


jandres742 (commented):
+1

scripts/core/virtual memory.yml
```diff
@@ -0,0 +1,282 @@
+#
+# Copyright (C) 2021 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
```
omarahmed1111, line 4: Just wanted to give an update that the license of the project have been updated in this https://github.com/oneapi-src/unified-runtime/pull/557. So, we should be able to update all the files license to the llvm one now.

vinser52:
The question just raised in my mind:

For USM we are using Unified Memory Allocation Framework to organize pooling. In the future, we are considering this pooling framework as a common point for all allocations to improve interop with other runtimes. For example, MPI can benefit from it by using our pooling framework to get info about where memory (pointed by virtual address) resides. 
The question is how this API correlated with the above-mentioned.

kbenzie:
@vinser52 while porting the PI changes over to UR I was pondering how this might be more tightly integrated into the USM interfaces. I did wonder if it would be possible to use USM allocations as the backing for the virtual memory reservations but I think there are some complications with the need to track the adapter specific state lifetimes (e.g `CUmemGenericAllocationHandle` in CUDA) with `ur_physical_mem_handle_t`.

Would it make sense for ownership of adapter specific handles to be managed in a `ur_usm_pool_handle_t`?

Then maybe we could be to include a `ur_usm_pool_handle_t` as an optional parameter to `urVirtualMemReserve`.

I'm really not sure if this is feasible though, our hands are somewhat tied due to the CUDA interfaces.

I'd like to hear @steffenlarsen's thoughts on this.

scripts/core/virtual memory.yml
kbenzie: Fixed.

steffenlarsen:
I share @kbenzie's concern regarding this idea, given the relation between this and the CUDA interfaces. I am not sure we can always make the association at reservation time either, as I don't believe there is any guarantee about USM-like behavior of the reserved pointer until it is mapped to the physical memory.

steffenlarsen (approved):
Interfaces look good!

include/ur.py
```diff
@@ -2805,6 +2905,69 @@ class ur_usm_dditable_t(Structure):
         ("pfnPoolGetInfo", c_void_p)                                    ## _urUSMPoolGetInfo_t
     ]
 
+###############################################################################
+## @brief Function-pointer for urVirtualMemGranularityGetInfo
+if __use_win_types:
+    _urVirtualMemGranularityGetInfo_t = WINFUNCTYPE( ur_result_t, ur_context_handle_t, ur_device_handle_t, ur_virtual_mem_granularity_info_t, c_size_t, c_void_p, POINTER(c_size_t) )
+else:
+    _urVirtualMemGranularityGetInfo_t = CFUNCTYPE( ur_result_t, ur_context_handle_t, ur_device_handle_t, ur_virtual_mem_granularity_info_t, c_size_t, c_void_p, POINTER(c_size_t) )
```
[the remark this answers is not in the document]
kbenzie: Okay, I've updated `urVirtualMemGranularityGetInfo` to reflect this change.

igchor:
LGTM, I don't have any good idea for integrating that with the USM interfaces for now.  We'll need to think about how to best expose UMF memory providers from UR - then we could implement the memory reservation using UMF providers and make them discoverable

veselypeta (approved):

kbenzie:
@vinser52 @igchor I've opened #635 to track future work resulting from this PR.

include/ur.py
```diff
@@ -1304,6 +1310,59 @@ def __str__(self):
...
+
+class ur_virtual_mem_granularity_info_t(c_int):
+    def __str__(self):
+        return str(ur_virtual_mem_granularity_info_v(self.value))
+
+
+###############################################################################
+## @brief Virtual memory access mode flags.
+class ur_virtual_mem_access_flags_v(IntEnum):
+    READ_WRITE = UR_BIT(0)                          ## Virtual memory both read and write accessible
```
steffenlarsen, line 1327: Sorry for the late reaction to this, but I just realized that we are losing the ability to set the access mode to "none" here by defining `READ_WRITE` as 0. CUDA considers an access mode of 0 as revoking access. @kbenzie - Could we either have a `NONE` value here or offset the existing ones?

kbenzie: Missed that! Yeah, I'll add none.

steffenlarsen:
@kbenzie - Sorry for the late realization, but I cannot find any analogue for `PI_EXT_ONEAPI_DEVICE_INFO_SUPPORTS_VIRTUAL_MEM` in these. Am I blind or was it overlooked?

kbenzie:
> @kbenzie - Sorry for the (very) late realization, but I cannot find any analogue for `PI_EXT_ONEAPI_DEVICE_INFO_SUPPORTS_VIRTUAL_MEM` in these. Am I blind or was it overlooked?

I do not believe you are blind, I think it was overlooked. I've opened https://github.com/oneapi-src/unified-runtime/pull/737 to add it.

PR #528, opened 2023-05-19
[UR][CTS] Fix device tests

fabiomestre:
- Update spec to show that MAX_WORK_GROUPS_3D returns an array of 3 elements
- Make it clear the urDeviceGetInfo can return UR_RESULT_ERROR_INVALID_ENUMERATION if the adapter does not support the query.
- Fix bug in urDevicePartition test
- Update urDeviceRelease and urDeviceRetain tests to take into account the differences betwen root devices and subdevices.
- Update the specification to make clear that the behaviour of urDeviceRelease and urDeviceRetain changes depending on the type of device.

### DIFF
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -429,7 +429,7 @@ analogue:
     - "**clRetainDevice**"
 details:
     - "Increments the device reference count if `hDevice` is a valid sub-device created by a call to `urDevicePartition`.
-    If `hDevice` is a root level device (e.g. obtained with `urDeviceGet`), the reference count remains unchanged."
+      If `hDevice` is a root level device (e.g. obtained with `urDeviceGet`), the reference count remains unchanged."
     - "It is not valid to use the device handle, which has all of its references released."
     - "The application may call this function from simultaneous threads for the same device."
     - "The implementation of this function should be thread-safe."
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -372,7 +372,7 @@ etors:
             [$x_bool_t] Return 1 if the device doesn't have a notion of a "queue index". Otherwise,
             return the number of queue indices that are available for this device.
     - name: MAX_WORK_GROUPS_3D
-      desc: "[uint32_t] return max 3D work groups"
+      desc: "[size_t[3]] return max 3D work groups"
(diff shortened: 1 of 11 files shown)


martygrant (approved):

scripts/core/device.yml
```diff
@@ -426,6 +428,8 @@ ordinal: "0"
 analogue:
     - "**clRetainDevice**"
 details:
+    - "Increments the device reference count if `hDevice` is a valid sub-device created by a call to `urDevicePartition`.
+    If `hDevice` is a root level device (e.g. obtained with `urDeviceGet`), the reference count remains unchanged."
```
kbenzie, line 432: This indentation looks a bit off

```suggestion
      If `hDevice` is a root level device (e.g. obtained with `urDeviceGet`), the reference count remains unchanged."
```

test/conformance/device/urDeviceGetInfo.cpp
```diff
@@ -105,7 +106,7 @@ static std::unordered_map<ur_device_info_t, size_t> device_info_size_map = {
     {UR_DEVICE_INFO_MAX_COMPUTE_QUEUE_INDICES, sizeof(uint32_t)},
     {UR_DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS, sizeof(ur_bool_t)},
     {UR_DEVICE_INFO_MEMORY_BUS_WIDTH, sizeof(ur_bool_t)},
-    {UR_DEVICE_INFO_MAX_WORK_GROUPS_3D, sizeof(uint32_t)},
+    {UR_DEVICE_INFO_MAX_WORK_GROUPS_3D, sizeof(size_t) * 3u},
```
kbenzie, line 109: `sizeof(size_t[3])` should also work here, then the the type in the test and the spec match up.

kbenzie (approved):

scripts/core/device.yml
fabiomestre: Fixed in the most recent commit

test/conformance/device/urDeviceGetInfo.cpp
fabiomestre: Fixed in the most recent commit

PR #530, opened 2023-05-22
[urtrace] fix coverity issues

pbalcer:


PatKamin:
If it was possible to have `constexpr` logging function, then possibly there would be errors during compilation time. The first attempt to do that was on a version of logger which differed quite a lot from the current one. Worth trying to implement that, however not in this PR for sure.

PatKamin (approved):

kswiecicki (approved):

PR #532, opened 2023-05-22
[loader] don't fail with no platforms present

### LINKED ISSUE #518: urInit with no adapters
Currently `urInit` would return UR_ERROR_UNINITIALIZED when no adapters were discovered. I suggest that we either 1) return a more descriptive error e.g. UR_ERROR_NO_ADAPTERS, or 2) return success, and then return 0 platforms in succeeding calls to `urPlatformGet`, if any.

pbalcer:
Closes #518

### DIFF
--- a/scripts/templates/ldrddi.cpp.mako
+++ b/scripts/templates/ldrddi.cpp.mako
@@ -51,18 +51,12 @@ namespace ur_loader
     %>
 
         %if re.match(r"Init", obj['name']):
-        bool atLeastOneplatformValid = false;
         for( auto& platform : context->platforms )
         {
             if(platform.initStatus != ${X}_RESULT_SUCCESS)
                 continue;
             platform.initStatus = platform.dditable.${n}.${th.get_table_name(n, tags, obj)}.${th.make_pfn_name(n, tags, obj)}( ${", ".join(th.make_param_lines(n, tags, obj, format=["name"]))} );
-            if(platform.initStatus == ${X}_RESULT_SUCCESS)
-                atLeastOneplatformValid = true;
         }
-
-        if(!atLeastOneplatformValid)
-            result=${X}_RESULT_ERROR_UNINITIALIZED;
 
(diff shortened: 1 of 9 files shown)


scripts/templates/ldrddi.cpp.mako
```diff
@@ -233,35 +224,21 @@ ${tbl['export']['name']}(
...
-    if(!atLeastOneplatformValid)
-        result = ${X}_RESULT_ERROR_UNINITIALIZED;
-    else
-        result = ${X}_RESULT_SUCCESS;
-    %endif
-
     if( ${X}_RESULT_SUCCESS == result )
     {
-        if( ( ur_loader::context->platforms.size() > 1 ) || ur_loader::context->forceIntercept )
+        if( ur_loader::context->platforms.size() == 0 || ur_loader::context->platforms.size() > 1 || ur_loader::context->forceIntercept )
```
kswiecicki, line 241: `ur_loader::context->platforms.size() >= 0`

pbalcer: No, this code is here to make direct calls to the adapter if there's only one. I did `... != 1` instead ;)
Thanks. Done.

test/loader/platforms/platforms.cpp
```diff
@@ -0,0 +1,78 @@
+/*
+ *
+ * Copyright (C) 2023 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ */
+#include <cassert>
+#include <cstdlib>
```
PatKamin, line 9: Include either cstdlib or stdlib.h, certainly there is no need to include both headers.

test/loader/platforms/platforms.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include <iostream>
+#include <logger/ur_logger.hpp>
```
PatKamin, line 11: nit: Please, make two separate groups of headers: one group with headers from the standard library and the other with UR headers.

PatKamin (approved):

PR #534, opened 2023-05-23
[ci] Add CI Debug builds

PatKamin:
- add Debug build type builds to a workflow
- use multiple threads to build, as $(nproc) would have no effect on
  Windows, setting the number of cores to 2 as this is the current max
  for Github Actions jobs
- hide message about libbacktrace not found on Windows builds.
- change the flush level of tracing messages in tracing tests to have certainty
for the order of messages

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Copyright (C) 2022-2023 Intel Corporation
 # SPDX-License-Identifier: MIT
 
-cmake_minimum_required(VERSION 3.15.0 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.14.0 FATAL_ERROR)
 project(unified-runtime VERSION 0.6.0)
 
 include(GNUInstallDirs)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Copyright (C) 2022-2023 Intel Corporation
 # SPDX-License-Identifier: MIT
 
-cmake_minimum_required(VERSION 3.14.0 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15.0 FATAL_ERROR)
```
pbalcer, line 4: Would it be possible to do without this change? CMake version is inherited from SYCL.

tools/urtrace/urtrace
```diff
@@ -81,15 +80,16 @@ env['UR_COLLECTOR_ARGS'] = collector_args
 log_collector = ""
 if args.debug:
     log_collector += "level:debug;"
+    log_collector += "flush:debug;"
 else:
     log_collector += "level:info;"
+    log_collector += "flush:info;"
```
pbalcer, line 86: I think having a separate flag to configure flushing would be better.

pbalcer (approved):

kswiecicki (approved):

pbalcer:
please rebase

PatKamin:
> please rebase 

Done

PR #535, opened 2023-05-23
[UMA] add more tests for allocations

DamianDuy:
Basic tests for UMA allocations. More will be added in the following PRs.

### DIFF
--- a/test/unified_malloc_framework/memoryPool.hpp
+++ b/test/unified_malloc_framework/memoryPool.hpp
@@ -5,9 +5,12 @@
 
 #include "pool.hpp"
 
+#include <array>
 #include <cstring>
 #include <functional>
 #include <random>
+#include <string>
+#include <thread>
 
 #ifndef UMF_TEST_MEMORY_POOL_OPS_HPP
 #define UMF_TEST_MEMORY_POOL_OPS_HPP
@@ -30,6 +33,7 @@ struct umfPoolTest : umf_test::test,
     }
 
     umf::pool_unique_handle_t pool;
+    static constexpr int NTHREADS = 5;
 };
 
 struct umfMultiPoolTest : umfPoolTest {
@@ -57,21 +61,43 @@ TEST_P(umfPoolTest, allocFree) {
     umfPoolFree(pool.get(), ptr);
 }
 
-TEST_P(umfPoolTest, pow2AlignedAlloc) {
-#ifdef _WIN32
-    // TODO: implement support for windows
-    GTEST_SKIP();
-#endif
+TEST_P(umfPoolTest, reallocFree) {
+    if (!umf_test::isReallocSupported(pool.get())) {
+        GTEST_SKIP();
+    }
+    static constexpr size_t allocSize = 64;
+    static constexpr size_t multiplier = 3;
+    auto *ptr = umfPoolMalloc(pool.get(), allocSize);
+    ASSERT_NE(ptr, nullptr);
+    auto *new_ptr = umfPoolRealloc(pool.get(), ptr, allocSize * multiplier);
+    ASSERT_NE(new_ptr, nullptr);
+    std::memset(new_ptr, 0, allocSize * multiplier);
+    umfPoolFree(pool.get(), new_ptr);
+}
 
+TEST_P(umfPoolTest, callocFree) {
+    if (!umf_test::isCallocSupported(pool.get())) {
+        GTEST_SKIP();
+    }
+    static constexpr size_t num = 10;
+    static constexpr size_t size = sizeof(int);
+    auto *ptr = umfPoolCalloc(pool.get(), num, size);
+    ASSERT_NE(ptr, nullptr);
+    for (size_t i = 0; i < num; ++i) {
+        ASSERT_EQ(((int *)ptr)[i], 0);
+    }
+    umfPoolFree(pool.get(), ptr);
+}
+
+void pow2AlignedAllocHelper(umf_memory_pool_handle_t pool) {
     static constexpr size_t maxAlignment = (1u << 22);
     static constexpr size_t numAllocs = 4;
-
     for (size_t alignment = 1; alignment <= maxAlignment; alignment <<= 1) {
         std::cout << alignment << std::endl;
         std::vector<void *> allocs;
 
         for (size_t alloc = 0; alloc < numAllocs; alloc++) {
-            auto *ptr = umfPoolAlignedMalloc(pool.get(), alignment, alignment);
+            auto *ptr = umfPoolAlignedMalloc(pool, alignment, alignment);
             ASSERT_NE(ptr, nullptr);
             ASSERT_TRUE(reinterpret_cast<uintptr_t>(ptr) % alignment == 0);
             std::memset(ptr, 0, alignment);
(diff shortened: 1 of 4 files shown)


test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -6,10 +6,13 @@
 #include <cstring>
 #include <functional>
 #include <random>
+#include <thread>
 
 #ifndef UMA_TEST_MEMORY_POOL_OPS_HPP
 #define UMA_TEST_MEMORY_POOL_OPS_HPP
 
+#define NTHREADS 5
```
PatKamin, line 14: This should be a static constexpr defined in the umaPoolTest class.

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -73,6 +76,56 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
     }
 }
 
+TEST_P(umaPoolTest, allocZeroSize) {
+    static constexpr size_t allocSize = 0;
+    auto *ptr = umaPoolMalloc(pool.get(), allocSize);
```
PatKamin, line 81: Should ptr be a nullptr or a valid ptr in this case? Please, add an assert accordingly.

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -73,6 +76,56 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
...
+    umaPoolFree(pool.get(), ptr);
```
PatKamin, line 82: Add an assert that it's a nullptr now.

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -73,6 +76,56 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
...
+}
+
+TEST_P(umaPoolTest, allocOutOfMem) {
+    // test whether memory is kept in a pool accordingly to MaxSize
+    static constexpr size_t allocSize = 16;
+    // MaxSize equals 16 * 1024 * 1024;
+    static constexpr size_t maxAllocSizeMultiplier = 1024 * 1024;
+
+    // allocate until oom
+    auto *ptr = umaPoolMalloc(pool.get(), allocSize);
```
PatKamin, line 98: You could just initialize the ptr to nullptr here and do all of allocations in the loop below.

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -73,6 +76,56 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
...
+    void *ptr = nullptr;
+    umaPoolFree(pool.get(), ptr);
+    ASSERT_EQ(ptr, nullptr);
+}
+
+TEST_P(umaPoolTest, allocOutOfMem) {
+    // test whether memory is kept in a pool accordingly to MaxSize
+    static constexpr size_t allocSize = 16;
+    // MaxSize equals 16 * 1024 * 1024;
+    static constexpr size_t maxAllocSizeMultiplier = 1024 * 1024;
```
PatKamin, line 95: This could be named just `maxAllocs`

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -73,6 +76,56 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
...
+
+    // allocate until oom
+    auto *ptr = umaPoolMalloc(pool.get(), allocSize);
+    ASSERT_NE(ptr, nullptr);
+
+    for (size_t i = 0; i <= maxAllocSizeMultiplier; ++i) {
```
PatKamin, line 101: You allocate more than the max pool size here. Add a check what should happen with the last allocation, ie. should it be in the pool?

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -73,6 +76,56 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
...
+        ptr = umaPoolMalloc(pool.get(), allocSize);
```
PatKamin, line 102: Memory leaks here. You're overwriting the `ptr` pointer with new addresses. Old ones are lost without freeing them. Create a vector of pointers and add each malloc to it. At the end of the test, iterate through all pointers in a vector and free memory.

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -73,6 +76,56 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
...
+
+    ptr = umaPoolMalloc(pool.get(), allocSize);
+    ASSERT_NE(ptr, nullptr);
+}
+
+TEST_P(umaPoolTest, multiThreadedMallocFree) {
+    auto poolMalloc = [](size_t allocSize, uma_memory_pool_handle_t pool) {
+        auto *ptr = umaPoolMalloc(pool, allocSize);
+        ASSERT_NE(ptr, nullptr);
+        umaPoolFree(pool, ptr);
```
PatKamin, line 116: Add an assert that it's a nullptr now.

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -73,6 +76,56 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
...
+TEST_P(umaPoolTest, allocZeroSize) {
+    static constexpr size_t allocSize = 0;
+    auto *ptr = umaPoolMalloc(pool.get(), allocSize);
+    umaPoolFree(pool.get(), ptr);
+}
+
+TEST_P(umaPoolTest, freeNullptr) {
+    void *ptr = nullptr;
+    umaPoolFree(pool.get(), ptr);
+    ASSERT_EQ(ptr, nullptr);
```
igchor, line 88: Can this ever be false? umaPoolFree cannot modify the ptr.

test/unified_memory_allocation/memoryPool.hpp
```diff
@@ -73,6 +76,56 @@ TEST_P(umaPoolTest, pow2AlignedAlloc) {
...
+    }
+
+    umaPoolFree(pool.get(), ptr);
+
+    ptr = umaPoolMalloc(pool.get(), allocSize);
+    ASSERT_NE(ptr, nullptr);
+}
+
+TEST_P(umaPoolTest, multiThreadedMallocFree) {
+    auto poolMalloc = [](size_t allocSize, uma_memory_pool_handle_t pool) {
```
igchor, line 113: It would be better to do multiple allocations in each thread to make sure we test different interleavings - please add some loop here.

Also, there should be some test with different allocation sizes.

test/unified_malloc_framework/memoryPool.hpp
```diff
@@ -57,6 +61,33 @@ TEST_P(umfPoolTest, allocFree) {
...
+
+TEST_P(umfPoolTest, callocFree) {
+    static constexpr size_t num = 10;
+    static constexpr size_t size = sizeof(int);
+    auto *ptr = umfPoolCalloc(pool.get(), num, size);
+    umf_result_t ret = umfPoolGetLastAllocationError(pool.get());
+    if (ret == UMF_RESULT_ERROR_NOT_SUPPORTED) {
+        GTEST_SKIP();
+    }
+    ASSERT_NE(ptr, nullptr);
```
igchor, line 87: you could probably verify the content here

test/unified_malloc_framework/memoryPool.hpp
```diff
@@ -84,6 +115,209 @@ TEST_P(umfPoolTest, pow2AlignedAlloc) {
...
+
+    for (auto provider : retProviders) {
+        if (std::string(umfMemoryProviderGetName(provider)) ==
+            std::string("null")) {
+            GTEST_SKIP();
+        }
+    }
+
+    // test whether memory is kept in a pool accordingly to MaxSize
+    static constexpr size_t allocSize = 16;
```
igchor, line 142: how did you come up with those values? aren't those specific to some memory pool?

test/unified_malloc_framework/memoryPool.hpp
```diff
@@ -84,6 +115,209 @@ TEST_P(umfPoolTest, pow2AlignedAlloc) {
...
+        }
+
+        for (auto allocation : allocations) {
+            umfPoolFree(pool, allocation);
+        }
+    };
+
+    std::vector<std::thread> threads;
+    for (int i = 0; i < NTHREADS; i++) {
+        threads.push_back(std::thread(poolMalloc, allocSize, pool.get()));
```
igchor, line 188: nit: `threads.emplace_back(poolMalloc, allocSize, pool.get());` to avoid copy

test/unified_malloc_framework/memoryPool.hpp
```diff
@@ -84,6 +115,209 @@ TEST_P(umfPoolTest, pow2AlignedAlloc) {
...
+            GTEST_SKIP();
+        }
+    }
+
+    // test whether memory is kept in a pool accordingly to MaxSize
+    static constexpr size_t allocSize = 16;
+    // MaxSize equals 16 * 1024 * 1024;
+    static constexpr size_t maxAllocs = 1024 * 1024;
+
+    // allocate until oom
```
igchor, line 146: I don't see any oom in this test, there should be some check for that I think.

test/unified_malloc_framework/memoryPool.hpp
```diff
@@ -84,6 +115,209 @@ TEST_P(umfPoolTest, pow2AlignedAlloc) {
...
+
+            for (auto &ptr : allocs) {
+                umfPoolFree(pool, ptr);
+            }
+        }
+    };
+
+    std::vector<std::thread> threads;
+    for (int i = 0; i < NTHREADS; i++) {
+        threads.push_back(std::thread(poolPow2AlignedAlloc, maxAlignment,
```
igchor, line 226: same here

test/unified_malloc_framework/memoryPool.hpp
```diff
@@ -84,6 +115,209 @@ TEST_P(umfPoolTest, pow2AlignedAlloc) {
...
+
+TEST_P(umfPoolTest, multiThreadedpow2AlignedAlloc) {
+#ifdef _WIN32
+    // TODO: implement support for windows
+    GTEST_SKIP();
+#endif
+
+    static constexpr size_t maxAlignment = (1u << 22);
+    static constexpr size_t numAllocs = 4;
+    auto poolPow2AlignedAlloc = [](size_t maxAlignment, size_t numAllocs,
```
igchor, line 204: This looks pretty much the same as in pow2AlignedAlloc. Can you avoid code duplication?

test/unified_malloc_framework/memoryPool.hpp
```diff
@@ -84,6 +115,209 @@ TEST_P(umfPoolTest, pow2AlignedAlloc) {
...
+        std::vector<void *> allocations;
+        for (size_t i = 0; i <= 10; ++i) {
+            allocations.emplace_back(umfPoolMalloc(pool, allocSize));
+            ASSERT_NE(allocations.back(), nullptr);
+        }
+
+        for (auto allocation : allocations) {
+            auto *ptr =
+                umfPoolRealloc(pool, allocation, allocSize * multiplier);
+            umf_result_t ret = umfPoolGetLastAllocationError(pool);
```
igchor, line 249: I'm not sure if you should call GTEST_SKIP() from multiple threads. Perhaps it would be better to check the support once, and only then lunch the test? You could even implement some helpers for checking the support, something like this:

```
bool is_realloc_supported(umf_memory_pool_handle_t p) {
    static constexpr size_t allocSize = 1;
    bool supported;
    auto ptr = umfPoolMalloc(pool, allocSize);
    auto ptr2 = umfPoolRealloc(pool, ptr, allocSize * 2);
    
    if (ptr2) supported = true;
    else if (umfPoolGetLastAllocationError(pool) == UMF_RESULT_ERROR_NOT_SUPPORTED) supported = false;
    else throw std::runtime_error("realloc failed with unexpected error");
    
    umfPoolFree(ptr2);
}
```

test/unified_malloc_framework/memoryPool.hpp
```diff
@@ -57,30 +61,243 @@ TEST_P(umfPoolTest, allocFree) {
...
-#endif
+TEST_P(umfPoolTest, reallocFree) {
+    static constexpr size_t allocSize = 64;
+    static constexpr size_t multiplier = 3;
+    auto *ptr = umfPoolMalloc(pool.get(), allocSize);
+    ASSERT_NE(ptr, nullptr);
+    auto *new_ptr = umfPoolRealloc(pool.get(), ptr, allocSize * multiplier);
+    umf_result_t ret = umfPoolGetLastAllocationError(pool.get());
+    if (ret == UMF_RESULT_ERROR_NOT_SUPPORTED) {
+        GTEST_SKIP();
```
igchor, line 72: Shouldn;t you use the `is*supporeted` function here ? E.g. `if (!isReallocSupported(pool.get()) GTEST_SKIP())` ? I think it would make the rest of the test a bit cleaner.

test/unified_memory_allocation/memoryPool.hpp
DamianDuy: Done.

test/unified_memory_allocation/memoryPool.hpp
DamianDuy: It is removed now.

test/unified_memory_allocation/memoryPool.hpp
DamianDuy: Done.

test/unified_malloc_framework/memoryPool.hpp
DamianDuy: Done.

test/unified_malloc_framework/memoryPool.hpp
DamianDuy: Done.

test/unified_malloc_framework/memoryPool.hpp
DamianDuy: Done.

test/unified_malloc_framework/memoryPool.hpp
DamianDuy: Done. I've used it for all the realloc/calloc tests.

igchor (approved):

PR #536, opened 2023-05-23
[UR] Refactor urDevicePartition to use desc struct

veselypeta:
Discussion on the approach can be seen in #401

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -249,7 +249,7 @@ typedef enum ur_structure_type_t {
     UR_STRUCTURE_TYPE_PROGRAM_NATIVE_PROPERTIES = 23,       ///< ::ur_program_native_properties_t
     UR_STRUCTURE_TYPE_SAMPLER_NATIVE_PROPERTIES = 24,       ///< ::ur_sampler_native_properties_t
     UR_STRUCTURE_TYPE_QUEUE_NATIVE_DESC = 25,               ///< ::ur_queue_native_desc_t
-    UR_STRUCTURE_TYPE_DEVICE_PARTITION_DESC = 26,           ///< ::ur_device_partition_desc_t
+    UR_STRUCTURE_TYPE_DEVICE_PARTITION_PROPERTIES = 26,     ///< ::ur_device_partition_properties_t
     /// @cond
     UR_STRUCTURE_TYPE_FORCE_UINT32 = 0x7fffffff
     /// @endcond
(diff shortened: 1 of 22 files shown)


include/ur_api.h
```diff
@@ -1051,18 +1103,23 @@ typedef enum ur_device_partition_t {
...
-    const ur_device_partition_property_t *pProperties, ///< [in] null-terminated array of <$_device_partition_t enum, value> pairs.
-    uint32_t NumDevices,                               ///< [in] the number of sub-devices.
-    ur_device_handle_t *phSubDevices,                  ///< [out][optional][range(0, NumDevices)] array of handle of devices.
-                                                       ///< If NumDevices is less than the number of sub-devices available, then
-                                                       ///< the function shall only retrieve that number of sub-devices.
-    uint32_t *pNumDevicesRet                           ///< [out][optional] pointer to the number of sub-devices the device can be
-                                                       ///< partitioned into according to the partitioning property.
+    ur_device_handle_t hDevice,                    ///< [in] handle of the device to partition.
+    const ur_device_partition_desc_t *pProperties, ///< [in] Array of partition descriptors.
+    size_t DescCount,                              ///< [in] Number of descriptors pointed to by `pProperties`.
```
kbenzie, line 1116: I think the array and count should be in the properties struct and rather than having each `ur_device_partition_desc_t` have `stype` and `pNext` don't call it a desc. Then we both have extensibility and a simple array query for the properties.

```cpp
struct ur_device_partition_property_t {
    ur_device_partition_t type;
    ur_device_partition_value_t value;
};

struct ur_device_partition_properties_t {
    ur_structure_type_t stype;
    void* pNext;
    const ur_device_partition_property_t *pProperties;
    size_t DescCount;
};

UR_APIEXPORT ur_result_t UR_APICALL urDevicePartition(
    ur_device_handle_t hDevice,
    const ur_device_partition_properties_t *pProperties,
    uint32_t NumDevices,
    ur_device_handle_t *phSubDevices,
    uint32_t *pNumDevicesRet);
```

fabiomestre:
I agree with Bennie's suggestion. But looks good to me otherwise :+1:

include/ur_api.h
veselypeta: I've implemented your suggestion 👍

pbalcer (approved):

include/ur_api.h
```diff
@@ -851,8 +852,8 @@ typedef enum ur_device_info_t {
     UR_DEVICE_INFO_PREFERRED_INTEROP_USER_SYNC = 74,            ///< [::ur_bool_t] prefer user synchronization when sharing object with
                                                                 ///< other API
     UR_DEVICE_INFO_PARENT_DEVICE = 75,                          ///< [::ur_device_handle_t] return parent device handle
-    UR_DEVICE_INFO_PARTITION_PROPERTIES = 76,                   ///< [::ur_device_partition_property_t[]] Returns an array of partition
-                                                                ///< types supported by the device
+    UR_DEVICE_INFO_PARTITION_PROPERTIES = 76,                   ///< [::ur_device_partition_t[]] Returns an array of partition types
```
kbenzie, line 855: I think this type should be `ur_device_partition_property_t`

kbenzie (approved):
LGTM - one comment

include/ur_api.h
kbenzie: Oh, okay. I don't think this should have `_PROPERTIES` for UR then, `UR_DEVICE_INFO_SUPPORTED_PARTITIONS` maybe?

Given these changes, I'd expect `UR_DEVICE_INFO_PARTITION_PROPERTIES` to be returning the array of `ur_device_partition_property_t`s used to create a paritioned device. There isn't a query for that at the moment, is there?

veselypeta: The query that returns `ur_device_partition_property_t[]` is `UR_DEVICE_INFO_PARTITION_TYPE` which will return the partition properties passed when to the subdevice when it was created with `urDeviceParititon`. I find that these seem to be confusingly named in OpenCL and we've copied the style into UR. Maybe from your suggestion we could do the following:

`UR_DEVICE_INFO_PARTITION_PROPERTIES` -> `UR_DEVICE_INFO_SUPPORTED_PARTITIONS`
`UR_DEVICE_INFO_PARTITION_TYPE` -> `UR_DEVICE_INFO_PARTITION_PROPERTIES`

veselypeta: I'm going to leave this as is for now, but I've created #552 to update the naming of these enumerations.

PR #539, opened 2023-05-24
[ci] Test tracing using libbacktrace

PatKamin:
Requires #534 

- add libbacktrace builds in CI
- fixes UR build issue when libbacktrace is used

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,20 +1,21 @@
-name: CMake
+name: Build and test
 
 on: [push, pull_request]
-
-env:
-  BUILD_TYPE: Release
 
 jobs:
   ubuntu-build:
     name: Build - Ubuntu
     strategy:
       matrix:
-        config:
-          - {os: 'ubuntu-20.04', compiler: {c: gcc, cxx: g++}}
-          - {os: 'ubuntu-22.04', compiler: {c: gcc, cxx: g++}}
-          - {os: 'ubuntu-22.04', compiler: {c: clang, cxx: clang++}}
-    runs-on: ${{matrix.config.os}}
+        os: ['ubuntu-20.04', 'ubuntu-22.04']
+        build_type: [Debug, Release]
+        compiler: [{c: gcc, cxx: g++}]
+        include:
+          - os: 'ubuntu-22.04'
+            build_type: Release
+            compiler: {c: clang, cxx: clang++}
+
+    runs-on: ${{matrix.os}}
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


.github/workflows/cmake.yml
```diff
@@ -33,6 +33,16 @@ jobs:
     - name: Install pip packages
       run: pip install -r third_party/requirements.txt
 
+    - name: Install libbacktrace
+      if: matrix.libbacktrace == '-DVAL_USE_LIBBACKTRACE_BACKTRACE=ON'
+      run: |
+        git clone https://github.com/ianlancetaylor/libbacktrace.git
+        pushd libbacktrace
```
pbalcer, line 40: Are these standard gnu utilities or are integrated into sh? :D

Up to you, but I'd rather use `cd libbacktrace` and `cd ..`

source/loader/layers/validation/backtrace_libbacktrace.cpp
```diff
@@ -42,14 +42,14 @@ int backtrace_cb(void *data, uintptr_t pc, const char *filename, int lineno,
     } else if (function != NULL) {
         backtraceLine << "(" << function << ") ";
     } else {
-        backtraceLine << "(????????) ";
+        backtraceLine << "(???????\?) ";
```
kswiecicki, line 45: Add a comment explaining that we escape the `?` sign to avoid a trigraph compiler warning.

.github/workflows/cmake.yml
```diff
@@ -1,20 +1,26 @@
...
+        compiler: [{c: gcc, cxx: g++}]
+        libbacktrace: ['-DVAL_USE_LIBBACKTRACE_BACKTRACE=OFF']
+        include:
+          - os: 'ubuntu-22.04'
+            build_type: Release
+            compiler: {c: clang, cxx: clang++}
+          - os: 'ubuntu-22.04'
+            build_type: Release
+            compiler: {c: gcc, cxx: g++}
+            libbacktrace: '-DVAL_USE_LIBBACKTRACE_BACKTRACE=ON'
```
kswiecicki, line 21: Should the `libbacktrace` be built only with gcc/g++ compilers?

PatKamin: Added clang check.

.github/workflows/cmake.yml
PatKamin: It's a bash builtin, works on Ubuntu on which these tests are run :) Although the same will be achieved with `cd` and probably it's more portable if someone would like to run this snippet in some other shell...

pbalcer (approved):

PR #541, opened 2023-05-25
[ur] fix minor coverity issues

DamianDuy:
- fix copy instead of move issue

### DIFF
--- a/test/conformance/source/environment.cpp
+++ b/test/conformance/source/environment.cpp
@@ -195,7 +195,7 @@ void DevicesEnvironment::TearDown() {
 KernelsEnvironment *KernelsEnvironment::instance = nullptr;
 
 KernelsEnvironment::KernelsEnvironment(int argc, char **argv,
-                                       std::string kernels_default_dir)
+                                       const std::string &kernels_default_dir)
     : DevicesEnvironment(argc, argv),
       kernel_options(parseKernelOptions(argc, argv, kernels_default_dir)) {
     instance = this;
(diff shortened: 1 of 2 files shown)


PatKamin:
Update template file `valddi.cpp.mako`.

source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -3613,7 +3613,8 @@ __urdlllocal ur_result_t UR_APICALL urEnqueueMemBufferReadRect(
             return UR_RESULT_ERROR_INVALID_SIZE;
         }
 
-        if (bufferSlicePitch != 0 && bufferSlicePitch % bufferRowPitch != 0) {
```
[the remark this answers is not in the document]
DamianDuy: Done.

pbalcer (approved):

test/conformance/source/environment.cpp
```diff
@@ -197,7 +197,8 @@ KernelsEnvironment *KernelsEnvironment::instance = nullptr;
 KernelsEnvironment::KernelsEnvironment(int argc, char **argv,
                                        std::string kernels_default_dir)
     : DevicesEnvironment(argc, argv),
-      kernel_options(parseKernelOptions(argc, argv, kernels_default_dir)) {
+      kernel_options(
+          parseKernelOptions(argc, argv, std::move(kernels_default_dir))) {
```
igchor, line 201: nit: What coverity issue does it fix exactly? I think a better fix would be to change the parameter type to `const std::string & kernels_default_dir` - it shows the intention better

DamianDuy: It is a ```copy_instead_of_move``` issue. Done.

igchor (approved):

PR #543, opened 2023-05-25
[UR] clarify sampler spec and add CTS

veselypeta:
This MR:
  * Makes `pPropSizeRet` & `pPropValue` optional parameters in `urSamplerGetInfo`
  * Adds basic CTS tests for the sampler set of entry points

### DIFF
--- a/scripts/core/sampler.yml
+++ b/scripts/core/sampler.yml
@@ -164,6 +164,11 @@ returns:
     - $X_RESULT_ERROR_INVALID_VALUE
     - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
     - $X_RESULT_ERROR_OUT_OF_RESOURCES
+    - $X_RESULT_ERROR_INVALID_SIZE:
+      - "`propSize == 0 && pPropValue != NULL`"
+    - $X_RESULT_ERROR_INVALID_NULL_POINTER:
+      - "`propSize != 0 && pPropValue == NULL`"
+      - "`pPropValue == NULL && pPropSizeRet == NULL`"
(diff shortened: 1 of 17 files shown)


scripts/core/sampler.yml
```diff
@@ -155,10 +155,10 @@ params:
       desc: "[in] size in bytes of the sampler property value provided"
     - type: "void*"
       name: pPropValue
-      desc: "[out][typename(propName, propSize)] value of the sampler property"
+      desc: "[out][typename(propName, propSize)][optional] value of the sampler property"
     - type: "size_t*"
       name: pPropSizeRet
-      desc: "[out] size in bytes returned in sampler property value"
+      desc: "[out][optional] size in bytes returned in sampler property value"
```
kbenzie, line 161: While these are optional, they still have valid usage requirements that we should define:

```cpp
if (propSize == 0 && pPropValue != nullptr) {
  return UR_RESULT_ERROR_INVALID_SIZE;
}
if (propSize != 0 && pPropValue == nullptr) {
  return UR_RESULT_ERROR_INVALID_NULl_POINTER;
}
if (pPropValue == nullptr && pPropSizeRet == nullptr) {
  return UR_RESULT_ERROR_INVALID_NULl_POINTER;
}
```

I wonder if this also needs applied to all the other `*GetInfo` entry points...

veselypeta: I have fixed this for sampler, but I've created a separate issue to update all remaining `GetInfo` style entry-points with these checks - #549

kbenzie (approved):

callumfare (approved):

PR #544, opened 2023-05-26
[coverity] Fix Coverity issues

PatKamin:
I assumed that nullptr checks in UR API are not necessary. If given handle ptr is not an array such checks do not exist, which is inconsistent. Might add nullptr checks if required.

### DIFF
--- a/source/common/ur_singleton.hpp
+++ b/source/common/ur_singleton.hpp
@@ -45,7 +45,7 @@ template <typename singleton_tn, typename key_tn> class singleton_factory_t {
     /// the params are forwarded to the ctor of the singleton
     /// the first parameter must be the unique identifier of the instance
     template <typename... Ts> singleton_tn *getInstance(Ts &&...params) {
-        auto key = getKey(std::forward<Ts>(params)...);
+        auto key = getKey(params...);
 
         if (key == 0) { // No zero keys allowed in map
             return static_cast<singleton_tn *>(0);
--- a/test/unified_memory_allocation/common/pool.hpp
+++ b/test/unified_memory_allocation/common/pool.hpp
@@ -79,7 +79,7 @@ struct proxy_pool : public pool_base {
 
         memset(ptr, 0, num * size);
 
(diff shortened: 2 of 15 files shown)


test/unified_memory_allocation/common/pool.hpp
```diff
@@ -79,7 +79,11 @@ struct proxy_pool : public pool_base {
 
         memset(ptr, 0, num * size);
 
-        EXPECT_EQ(ret, UMA_RESULT_SUCCESS);
+        try {
+            EXPECT_EQ(ret, UMA_RESULT_SUCCESS);
```
pbalcer, line 83: Not the biggest fan of this approach, especially since it's repeated multiple times.
Maybe it would be worth defining something like `EXPECT_EQ_NOEXCEPT` ?

Do these functions need to be `noexcept`?

PatKamin: Yes, they need to be noexcept to compile. Added a macro.

pbalcer:
@igchor please take a look at the uma test changes. I'm not convinced...

source/common/ur_singleton.hpp
```diff
@@ -45,7 +45,7 @@ template <typename singleton_tn, typename key_tn> class singleton_factory_t {
     /// the params are forwarded to the ctor of the singleton
     /// the first parameter must be the unique identifier of the instance
     template <typename... Ts> singleton_tn *getInstance(Ts &&...params) {
-        auto key = getKey(std::forward<Ts>(params)...);
+        auto key = getKey(params...);
```
igchor, line 48: Why did you remove std::forward? It does not change anything with the current implementation but we might have some overloads in the future perhaps.

test/unified_memory_allocation/common/pool.hpp
igchor: Do ASSERT_* macros also throw? ASSERT might be a better fit here anywhere.

Also, do we actually run the gtest with throw_on_failure? Can we disable it?

igchor (approved):

PR #553, opened 2023-05-30
Add kernel execution tests.

aarongreig:


test/conformance/kernel/urKernelSetArgPointer.cpp
```diff
@@ -33,27 +41,39 @@ TEST_P(urKernelSetArgPointerTest, SuccessHost) {
     ASSERT_NE(allocation, nullptr);
 
     ASSERT_SUCCESS(urKernelSetArgPointer(kernel, 0, allocation));
+    ASSERT_SUCCESS(urKernelSetArgValue(kernel, 1, sizeof(data), &data));
```
aarongreig, line 44: As far as I can tell from the spec and the IR you get from the fill_usm test program, these *should* work - but none of them do with the cuda plugin. I'm still looking into it but the test code is sound I think

test/conformance/kernel/urKernelSetArgPointer.cpp
```diff
@@ -62,20 +82,26 @@ TEST_P(urKernelSetArgPointerTest, SuccessShared) {
...
+                                        nullptr, nullptr));
+    Launch1DRange(array_size);
+    ValidateAllocation();
 }
 
 struct urKernelSetArgPointerNegativeTest : urKernelSetArgPointerTest {
     // Get any valid allocation we can to test validation of the other parameters.
     void SetUpAllocation() {
-        bool host_supported = false;
+        ur_device_usm_access_capability_flags_t host_supported = false;
```
kbenzie, line 96: `= false` might compile but it should really be `0`. same goes for all the other instances of this check in this file.

test/conformance/kernel/urKernelSetExecInfo.cpp
```diff
@@ -56,7 +56,7 @@ struct urKernelSetExecInfoUSMPointersTest : uur::urKernelTest {
 UUR_INSTANTIATE_KERNEL_TEST_SUITE_P(urKernelSetExecInfoUSMPointersTest);
 
 TEST_P(urKernelSetExecInfoUSMPointersTest, SuccessHost) {
-    bool host_supported = false;
+    ur_device_usm_access_capability_flags_t host_supported = false;
```
kbenzie, line 59: `= false` might compile but it should really be `0`. same goes for all the other instances of this check in this file.

test/conformance/kernel/urKernelSetArgPointer.cpp
kbenzie: What's not working about this on the CUDA adapter?

aarongreig: So I'm finding it works fine if I do this
```
ASSERT_SUCCESS(urKernelSetArgPointer(kernel, 0, &allocation));
```
instead of just passing the `allocation` pointer. This makes sense looking at how the plugin implements passing the arg through to be stored in its kernel struct; but it doesn't match the UR spec, the pi header or the analogous CL entry point. Currently trying to figure out if this is de-facto how the sycl runtime is using the entry point from the original plugin and we just need to update the spec to reflect that, or if there's something weirder happening

kbenzie: Ah, yeah. We deffo want consistency with this. I wonder if that's a holder over from PI which is a holder over from OpenCL... Do OpenCL implenmetations of `clSetKernelArg` do `&mem_object`?

aarongreig: Alright what I've just pushed is fully working. I think I've gathered enough insight into how the runtime and various plugins are using this stuff to justify changing the signature of `urKernelSetArgPointer` to take `void**` instead of just `void*`. Should be a nfc but it better represents how the entry point is supposed to be used in conjunction with the various USM alloc entry points.

kbenzie (approved):

scripts/generate_kernel_header.py
```diff
@@ -0,0 +1,108 @@
+"""
+ Copyright (C) 2022 Intel Corporation
+
+ SPDX-License-Identifier: MIT
+"""
```
omarahmed1111, line 5: Just wanted to give an update that the license of the project have been updated in this https://github.com/oneapi-src/unified-runtime/pull/557. So, we should be able to update all the files license to the llvm one now.

omarahmed1111 (approved):
LGTM, just small notes on license update

Issue #555, opened 2023-05-30
Change ur_bool_t to uint32_t


jandres742:
pi_bool on sycl is uint32_t, so better to match it. We have seen already errors where due to the difference in size, a false value returned by the adapter is interpreted as true by the sycl runtime.

kbenzie:
https://github.com/intel/llvm/pull/9672 was merged and the follow up will be to look into making `pi_bool` match `ur_bool_t` instead.

PR #556, opened 2023-05-30
Change ur_bool_t to uint32_t

jandres742:
This to match the boolean type used in SYCL, where pi_bool is also uint32_t.

Resolves: #555


kbenzie:
We actually have a pi2ur workaround for this size mismatch called `fixupReturnValues()`.

jandres742:
> fixupReturnValues

thanks @kbenzie . This error came from a value UR L0 is returning in piDeviceGetInfo, and there we are calling fixupInfoValueTypes. So maybe the WA is not working completely. Should we proceed with this PR so we can eventually remove the WA or would you prefer to keep the WA?

kbenzie:
Lets discuss this in the WG call shortly.

kbenzie:
Closing for reason described in #555

PR #557, opened 2023-05-31
Change project license from MIT to apache 2.0 with llvm exceptions

omarahmed1111:
The script for changing the license headers in project files is added in the PR in file: [scripts/change_license.py](https://github.com/oneapi-src/unified-runtime/pull/557/files#diff-7e5636e96d862b768f9dbd738e2d5126c5d92d6ee387a3d21c5ce9b4c0092d2b)


LICENSE.TXT
```diff
@@ -0,0 +1,278 @@
+==============================================================================
+The LLVM Project is under the Apache License v2.0 with LLVM Exceptions:
```
alycm, line 2:
```suggestion
The Unified Runtime is under the Apache License v2.0 with LLVM Exceptions:
```

veselypeta (approved):
LGTM!

LICENSE.TXT
```diff
@@ -0,0 +1,278 @@
...
+terms. All such code will be identified clearly using at least one of two
+mechanisms:
+1) It will be in a separate directory tree with its own `LICENSE.txt` or
+   `LICENSE` file at the top containing the specific license and restrictions
+   which apply to that software, or
+2) It will contain specific license and restriction terms at the top of every
+   file.
+
+==============================================================================
+Legacy LLVM License (https://llvm.org/docs/DeveloperPolicy.html#legacy):
```
alycm, line 237: I don't think that we need to include this legacy license, right?  It has no relevance to this repo.

LICENSE.TXT
```diff
@@ -0,0 +1,278 @@
...
+software that is licensed under the GPLv2 ("Combined Software") and if a
+court of competent jurisdiction determines that the patent provision (Section
+3), the indemnity provision (Section 9) or other Section of the License
+conflicts with the conditions of the GPLv2, you may retroactively and
+prospectively choose to deem waived or otherwise exclude such Section(s) of
+the License, but only in their entirety and only with respect to the Combined
+Software.
+
+==============================================================================
+Software from third parties included in the Unified Runtime Project:
```
alycm, line 225: Currently there is no code under other licenses.

So probably least confusing thing to do is remove this section?  It's not part of vanilla Apache, it's an LLVM extra bit.

It is however neutral in terms of actual meaning, so open to thoughts.

scripts/change_license.py
```diff
@@ -0,0 +1,44 @@
+"""
```
alycm, line 1: Feels like we'll never need this script again?  Why commit it?

alycm: To run on pending PRs?

alycm (commented):
Fine in principal, some questions that we can address very quickly.

alycm (commented):
Fine in principal, some questions that we can address very quickly.

scripts/change_license.py
pbalcer: It might be useful to run something similar in the CI that checks (or fixes) whether a correct license is present in the files being committed.

omarahmed1111: Yeah, I was meant to keep it for any pending PRs as I saw one of them is adding new file. but I feel now that we could only mention that change on any current PR adding new files and it should be simple change. so, I think I will take that approach.

LICENSE.TXT
omarahmed1111: I was thinking maybe when we move the adapters here, we will also fetch `l0` and `cuda` headers but I am not sure if that fetch for this headers is regarded as third party software.

LICENSE.TXT
omarahmed1111: yup, deleted that

scripts/change_license.py
omarahmed1111: > It might be useful to run something similar in the CI that checks (or fixes) whether a correct license is present in the files being committed.

Yeah, good point, I think we could add a check to CI to check for license using the regex in the script.

LICENSE.TXT
omarahmed1111: There is also some third-party python software installed for the repo.

kbenzie: I think fetching dependencies as part of the build process does not need to come under this license file but @alycm should confirm if I'm correct or not.

alycm: Yes, fetching is different as the code isn't in the repo.  But!  Lets just leave the wording as is then just to make sure we're covered, it doesn't hurt.

alycm (approved):

PR #560, opened 2023-06-01
[Experimental] Add command-buffers

Bensuo:
- Adds support for command-buffers experimental feature
- Add command-buffer desc type to common.yml
- Add device info query for command-buffer support

Porting the PI command-buffer extension added in https://github.com/intel/llvm/pull/9375 which is currently under review. The functions implemented do not cover all possible command types, only those that have actually been implemented in the linked PR.

Notes:
- As there's no way to extend existing enums from the command-buffer YAML file I've modified those directly where necessary.

### DIFF
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -251,6 +251,8 @@ etors:
       desc: "Invalid host pointer"
     - name: ERROR_INVALID_USM_SIZE
       desc: "Invalid USM size"
+    - name: ERROR_INVALID_COMMAND_BUFFER_EXP
+      desc: "Invalid Command-Buffer"
     - name: ERROR_OBJECT_ALLOCATION_FAILURE
       desc: "Objection allocation failure"
     - name: ERROR_ADAPTER_SPECIFIC
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -253,6 +253,10 @@ etors:
       desc: "Invalid USM size"
     - name: ERROR_INVALID_COMMAND_BUFFER_EXP
       desc: "Invalid Command-Buffer"
+    - name: ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_EXP
+      desc: "Sync point is not valid for the command-buffer"
+    - name: ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_WAIT_LIST_EXP
+      desc: "Sync point wait list is invalid"
     - name: ERROR_OBJECT_ALLOCATION_FAILURE
       desc: "Objection allocation failure"
     - name: ERROR_ADAPTER_SPECIFIC
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -318,6 +318,8 @@ etors:
       desc: $x_queue_native_desc_t
     - name: DEVICE_PARTITION_PROPERTIES
       desc: $x_device_partition_properties_t
+    - name: EXP_COMMAND_BUFFER_DESC
+      desc: $x_exp_command_buffer_desc_t
 --- #--------------------------------------------------------------------------
 type: struct
 desc: "Base for all properties types"
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -251,17 +251,20 @@ etors:
       desc: "Invalid host pointer"
     - name: ERROR_INVALID_USM_SIZE
       desc: "Invalid USM size"
-    - name: ERROR_INVALID_COMMAND_BUFFER_EXP
-      desc: "Invalid Command-Buffer"
-    - name: ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_EXP
-      desc: "Sync point is not valid for the command-buffer"
-    - name: ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_WAIT_LIST_EXP
-      desc: "Sync point wait list is invalid"
     - name: ERROR_OBJECT_ALLOCATION_FAILURE
       desc: "Objection allocation failure"
     - name: ERROR_ADAPTER_SPECIFIC
       desc: "An adapter specific warning/error has been reported and can be retrieved
              via the urGetLastResult entry point."
+    - name: ERROR_INVALID_COMMAND_BUFFER_EXP
+      value: "0x1000"
+      desc: "Invalid Command-Buffer"
+    - name: ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_EXP
+      value: "0x1001"
+      desc: "Sync point is not valid for the command-buffer"
+    - name: ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_WAIT_LIST_EXP
+      value: "0x1002"
+      desc: "Sync point wait list is invalid"
     - name: ERROR_UNKNOWN
       value: "0x7ffffffe"
       desc: "Unknown or internal error"
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -383,8 +383,6 @@ etors:
       desc: "[$x_bool_t] Return true if the device supports enqueing commands to read and write pipes from the host."
     - name: MAX_REGISTERS_PER_WORK_GROUP
       desc: "[uint32_t] The maximum number of registers available per block."
-    - name: EXP_COMMAND_BUFFER_SUPPORT
-      desc: "[$x_exp_command_buffer_support_level_t] Level of experimental command-buffer support."
 --- #--------------------------------------------------------------------------
 type: function
 desc: "Retrieves various information about device"
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -383,6 +383,8 @@ etors:
       desc: "[$x_bool_t] Return true if the device supports enqueing commands to read and write pipes from the host."
     - name: MAX_REGISTERS_PER_WORK_GROUP
       desc: "[uint32_t] The maximum number of registers available per block."
+    - name: EXP_COMMAND_BUFFER_SUPPORT
+      desc: "[$x_exp_command_buffer_support_level_t] Level of experimental command-buffer support."
 --- #--------------------------------------------------------------------------
 type: function
 desc: "Retrieves various information about device"
--- a/scripts/core/exp-command-buffer.yml
+++ b/scripts/core/exp-command-buffer.yml
@@ -134,7 +134,7 @@ params:
     - type: "void*"
       name: pDst
       desc: "[in] Location the data will be copied to."
-    - type: "void*"
+    - type: "const void*"
       name: pSrc
       desc: "[in] The data to be copied."
     - type: "size_t"
@@ -236,6 +236,9 @@ params:
     - type: $x_exp_command_buffer_handle_t
       name: hCommandBuffer
       desc: "[in] handle of the command-buffer object."
+    - type: $x_queue_handle_t
+      name: hQueue
+      desc: "[in] the queue to submit this command-buffer for execution."
     - type: uint32_t
       name: numEventsInWaitList
       desc: "[in] size of the event wait list"
--- a/scripts/core/exp-command-buffer.yml
+++ b/scripts/core/exp-command-buffer.yml
@@ -3,19 +3,6 @@ type: header
 desc: "Intel $OneApi Command-Buffer Extension APIs"
 ordinal: "0"
 --- #--------------------------------------------------------------------------
-type: enum
-desc: "Command buffer support level for device info queries."
-class: $xCommandBuffer
-name: $x_exp_command_buffer_support_level_t
-etors:
-    - name: UNSUPPORTED
-      desc: "No support for command-buffers."
-      value: "0"
-    - name: EMULATED
-      desc: "Emulated support for command-buffers."
-    - name: NATIVE
-      desc: "Native support for command-buffers."
---- #--------------------------------------------------------------------------
 type: struct
 desc: "Command-Buffer Descriptor Type"
 name: $x_exp_command_buffer_desc_t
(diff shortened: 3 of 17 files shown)


scripts/core/device.yml
```diff
@@ -381,6 +381,8 @@ etors:
       desc: "[$x_bool_t] Return true if the device supports enqueing commands to read and write pipes from the host."
     - name: MAX_REGISTERS_PER_WORK_GROUP
       desc: "[uint32_t] The maximum number of registers available per block."
+    - name: EXP_COMMAND_BUFFER_SUPPORT
+      desc: "[$x_exp_command_buffer_support_level_t] Level of experimental command-buffer support."
```
EwanC, line 385: I think before we discussed just using the existing extension device query which returns a list of strings, and then reporting our extension in there forn now. Then coming up with a more thought out reporting mechanism later, if we want to return version or support for individual capabilities within the optional feature.

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,253 @@
...
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Stop recording on a command-buffer object such that no more commands can be appended and make it ready to enqueue."
+class: $xCommandBuffer
+name: FinalizeExp
+params:
+    - type: $x_exp_command_buffer_handle_t
+      name: hCommandBuffer
+      desc: "[in] handle of the command-buffer object"
+returns:
```
EwanC, line 89: Could also return `ERROR_INVALID_COMMAND_BUFFER_EXP` if command-buffer has already been finalize (when we have a spec can write that up more formally)

Should probably add this error to all the entry-points.

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,253 @@
...
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
+    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Append a kernel execution command to a command-buffer object"
+class: $xCommandBuffer
+name: AppendKernelExp
```
EwanC, line 96: This name entry-point name has "Append" in it, but none of the memcpy commands do which are also appending. Preference for consistently using "Append" for all command adding entry-points or dropping it.

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,253 @@
...
+name: FinalizeExp
+params:
+    - type: $x_exp_command_buffer_handle_t
+      name: hCommandBuffer
+      desc: "[in] handle of the command-buffer object"
+returns:
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
+    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
+--- #--------------------------------------------------------------------------
+type: function
```
EwanC, line 93: We should copy the `-returns` from the core entry points and add invalid comment-

scripts/core/exp-command-buffer.yml
Bensuo: Done!

scripts/core/exp-command-buffer.yml
Bensuo: I've updated the returns across all our functions and made sure they line up with the core entry points, thanks!

scripts/core/device.yml
kbenzie: Yeah I think this makes the most sense in the short term so it doesn't extend the `ur_device_info_t`. Here's a straw man suggestion for the naming convention in the extension string: `ur_exp_<feature>`. I'll open a ticket about clarifying this in the contibution guide.

Bensuo: I've removed the extension to `ur_device_info_t` for now and the associated enum for the support level.

EwanC (approved):
Looks good from my perspective

scripts/core/common.yml
```diff
@@ -249,6 +249,12 @@ etors:
       desc: "Invalid host pointer"
     - name: ERROR_INVALID_USM_SIZE
       desc: "Invalid USM size"
+    - name: ERROR_INVALID_COMMAND_BUFFER_EXP
+      desc: "Invalid Command-Buffer"
+    - name: ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_EXP
+      desc: "Sync point is not valid for the command-buffer"
+    - name: ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_WAIT_LIST_EXP
+      desc: "Sync point wait list is invalid"
```
omarahmed1111, line 257: I think we could add `EXP_*` at the beginning rather than at the end.

omarahmed1111 (commented):
We need to add also `COMMAND_BUFFER_SUPPORT_EXP` in `device.yml` `x_device_info_t` to support this device info query.
Edit: this was discussed in this [comment](https://github.com/oneapi-src/unified-runtime/pull/560#discussion_r1217790283)

scripts/core/common.yml
kbenzie, line 257: I feel like these should be towards the end of the `etors` list, with only `ERROR_UNKNOWN` below. I also think they should have explicit values, starting at `0x1000` to separate them from the range of core spec results.

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,300 @@
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Command-Buffer Extension APIs"
+ordinal: "0"
```
kbenzie, line 4: I'd recommend using `99` for the ordinal.

I'll need to document this.

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,300 @@
...
+name: AppendKernelExp
+params:
+    - type: $x_exp_command_buffer_handle_t
+      name: hCommandBuffer
+      desc: "[in] handle of the command-buffer object"
+    - type: $x_kernel_handle_t
+      name: hKernel
+      desc: "[in] kernel to append"
+    - type: uint32_t
+      name: WorkDim
```
kbenzie, line 100:
```suggestion
      name: workDim
```

although I realised this may have been copied from [urEnqueueKernelLaunch](https://oneapi-src.github.io/unified-runtime/core/api.html#urenqueuekernellaunch) in which case I need to go canonicalise them following the contribution guide.

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,300 @@
...
+      desc: "[in] Global work size to use when executing kernel."
+    - type: "const size_t*"
+      name: pLocalWorkSize
+      desc: "[in] Local work size to use when executing kernel."
+    - type: uint32_t
+      name: numSyncPointsInWaitList
+      desc: "[in] The number of sync points in the provided dependency list."
+    - type: "const $x_exp_command_buffer_sync_point_t*"
+      name: pSyncPointWaitList
+      desc: "[in] A list of sync points that this command depends on."
```
kbenzie, line 116: Are sync-point wait lists optional?

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,300 @@
...
+    - type: "$x_exp_command_buffer_sync_point_t*"
+      name: pSyncPoint
+      desc: "[out] sync point associated with this command"
```
kbenzie, line 119: Are returned sync-points optional?

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,300 @@
...
+      desc: "[in] handle of the command-buffer object"
+returns:
+    - $X_RESULT_ERROR_INVALID_COMMAND_BUFFER_EXP
+    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Append a kernel execution command to a command-buffer object"
+class: $xCommandBuffer
+name: AppendKernelExp
```
kbenzie, line 91: This is analagous to [urEnqueueKernelLaunch](https://oneapi-src.github.io/unified-runtime/core/api.html#urenqueuekernellaunch), right?

If so I think it should be named as below for consistency:

```suggestion
name: AppendKernelLaunchExp
```

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,300 @@
...
+      name: SrcRowPitch
+      desc: "[in] Row pitch of the source memory."
+    - type: "size_t"
+      name: SrcSlicePitch
+      desc: "[in] Slice pitch of the source memory."
+    - type: "size_t"
+      name: DstRowPitch
+      desc: "[in] Row pitch of the destination memory."
+    - type: "size_t"
+      name: DstSlicePitch
```
kbenzie, line 247: Some more instances of names that should be camelCase in this `params:` list.

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,300 @@
+--- #--------------------------------------------------------------------------
```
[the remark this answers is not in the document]
kbenzie: > [NIT] more of a stylish thing so feel free to ignore, but I think I prefer the file name without the exp part, and we could add a comment to the file to explain that this is an exp feature.

@omarahmed1111 that's as defined in the [Contribution Guide](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#experimental-features) and wihtout the `exp-` prefix it would be very hard to distinguish between the core and experimental features in the generator scripts.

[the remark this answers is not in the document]
omarahmed1111: Ah okay, that make sense. ignore that then

scripts/core/exp-command-buffer.yml
Bensuo: These should all be fixed now.

scripts/core/exp-command-buffer.yml
Bensuo: They should be yes, this is now changed.

scripts/core/exp-command-buffer.yml
Bensuo: They should be yes, this is now changed.

scripts/core/exp-command-buffer.yml
Bensuo: Fixed, thanks!

scripts/core/exp-command-buffer.yml
Bensuo: Good point, this is fixed now.

scripts/core/exp-command-buffer.yml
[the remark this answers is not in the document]
Bensuo: I've added the license to this file now.

scripts/core/common.yml
Bensuo: Good idea, I've made this change now.

scripts/core/exp-command-buffer.yml
Bensuo: Done, thanks!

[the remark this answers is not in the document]
Bensuo: As suggested in [this comment](https://github.com/oneapi-src/unified-runtime/pull/560/files/412ebda36c5648609d749da03f520a7be3fed2e4#r1221630727) I've set the ordinal to `99` now.

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,300 @@
...
+type: header
+desc: "Intel $OneApi Command-Buffer Extension APIs"
```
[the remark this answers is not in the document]
Bensuo: Good point, I've updated the text here as suggested.

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,307 @@
+#
+# Copyright (C) 2021 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
```
omarahmed1111, line 4: We have updated the project license to `Apache-2.0 WITH LLVM-exception`,
```suggestion
#
# Copyright (C) 2022 Intel Corporation
#
# Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
# See LICENSE.TXT
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

Bensuo: Apologies this should be fixed now!

Bensuo:
> @Bensuo please rebase, run the `generate` target, and commit the changes to fix the GitHub Actions.

PR should be up-to-date now.

omarahmed1111 (approved):

kbenzie (approved):

scripts/core/exp-command-buffer.yml
```diff
@@ -0,0 +1,309 @@
...
+      name: pSyncPoint
+      desc: "[out][optional] sync point associated with this command"
+returns:
+    - $X_RESULT_ERROR_INVALID_COMMAND_BUFFER_EXP
+    - $X_RESULT_ERROR_INVALID_KERNEL
+    - $X_RESULT_ERROR_INVALID_WORK_DIMENSION
+    - $X_RESULT_ERROR_INVALID_WORK_GROUP_SIZE
+    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_EXP
+    - $X_RESULT_ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_WAIT_LIST_EXP
```
EwanC, line 136: Should this have a trailing colon? `$X_RESULT_ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_WAIT_LIST_EXP:` just looking at how some of the other errors with sub-bullets are structured

kbenzie: yes, good catch!

Bensuo: Fixed, thanks!

PR #568, opened 2023-06-05
Move USM allocator implementation from intel/llvm

igchor:
and implement UMA memory pool on top of it. This PR is an updated version of https://github.com/oneapi-src/unified-runtime/pull/217

This is the first step for generalizing memory pooling across adapters (@bmyates @jandres742). The idea is that each adapter will implement an UMA memory provider, which can be used to construct a DisjointPool instance. DisjointPool itself implements a generic UMA memory pool interface so it can be replaced in the future by a different implementation.


source/common/uma_pools/disjoint_pool.cpp
```diff
@@ -0,0 +1,948 @@
...
+#include <list>
+#include <memory>
+#include <mutex>
+#include <shared_mutex>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+// TODO: replace with logger?
```
pbalcer, line 25: Yes, but in a separate PR. We will need to align with making the same change in the adapters so things are consistent.

source/common/uma_pools/disjoint_pool.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include <atomic>
+#include <memory>
+
+#include "../uma_helpers.hpp"
+
+namespace usm {
+
+// Configuration for specific USM allocator instance
+class DisjointPoolConfig {
```
pbalcer, line 20: might be better to move this to `disjoint_pool_config.hpp`

pbalcer (approved):

source/common/uma_pools/disjoint_pool_config.cpp
```diff
@@ -0,0 +1,270 @@
...
+    Configs[DisjointPoolMemType::SharedReadOnly].MaxPoolableSize = 4_MB;
+    Configs[DisjointPoolMemType::SharedReadOnly].Capacity = 4;
+    Configs[DisjointPoolMemType::SharedReadOnly].SlabMinSize = 2_MB;
+};
+
+DisjointPoolAllConfigs parseDisjointPoolConfig(const std::string &config,
+                                               bool trace) {
+    DisjointPoolAllConfigs AllConfigs;
+
+    // TODO: replace with UR ENV var parser and avoid creating a copy of 'config'
```
kswiecicki, line 67: Those config parameters will need an entry in docs.

source/common/uma_pools/disjoint_pool_config.hpp
```diff
@@ -0,0 +1,57 @@
...
+// SlabMinSize:     Minimum allocation size requested from USM.
+//                  Default 64KB host and device, 2MB shared.
+//
+// Example of usage:
+// "1;32M;host:1M,4,64K;device:1M,4,64K;shared:0,0,2M"
+DisjointPoolAllConfigs parseDisjointPoolConfig(const std::string &config,
+                                               bool trace = 1);
+} // namespace usm
+
+#endif
```
kswiecicki, line 57: eof

source/common/uma_pools/disjoint_pool.cpp
```diff
@@ -0,0 +1,948 @@
...
+    // Get a slab to be used for chunked allocations.
+    decltype(AvailableSlabs.begin()) getAvailSlab(bool &FromPool);
+
+    // Get a slab that will be used as a whole for a single allocation.
+    decltype(AvailableSlabs.begin()) getAvailFullSlab(bool &FromPool);
+};
+
+class DisjointPool::AllocImpl {
+    // It's important for the map to be destroyed last after buckets and their
+    // slabs This is because slab's destructor removes the object from the map.
```
kswiecicki, line 262: In that case, it might be a good idea to create a destructor for `AllocImpl` that destroys every slab in `KnownSlabs` before its lifecycle ends.

test/unified_memory_allocation/uma_pools/disjoint_pool.cpp
```diff
@@ -0,0 +1,44 @@
...
+    config.SlabMinSize = 4096;
+    config.MaxPoolableSize = 4096;
+    config.Capacity = 4;
+    config.MinBucketSize = 64;
+    return config;
+}
+
+static auto makePool() {
+    auto providerUnique =
+        uma::memoryProviderMakeUnique<uma_test::provider_malloc>().second;
```
kswiecicki, line 28: You could assert the returned `uma_result_t`.

test/unified_memory_allocation/uma_pools/disjoint_pool.cpp
```diff
@@ -0,0 +1,44 @@
...
+    auto provider = providerUnique.release();
+    auto pool =
+        uma::poolMakeUnique<usm::DisjointPool>(&provider, 1, poolConfig())
+            .second;
```
kswiecicki, line 32: Same here.

source/common/uma_pools/disjoint_pool.hpp
igchor: I'm not sure. I wanted to keep the disjoint_pool implementation self-contained (which also means this configuration) but independent on how many pools we create.

disjoint_pool_config only provides defaults and parses string for a particular use-case (USM allocations with separate pools per device, host, shared, sharedReadOnly). I think I should name it better tough - perhaps disjoint_pool_config_parser or something.

kswiecicki (approved):

PR #573, opened 2023-06-06
Add UR_DEVICE_INFO_DEVICE_IP_VERSION property

dm-vodopyanov:
This patch extends `ur_device_info_t` with the new property `UR_DEVICE_INFO_DEVICE_IP_VERSION`. 
This property is needed to query the device architecture ID value as described in the [Level Zero spec](https://github.com/oneapi-src/level-zero-spec/pull/23). 
The property will be used in the implementation of host API of [sycl_ext_oneapi_device_architecture](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_device_architecture.asciidoc) extension

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -925,6 +925,9 @@ typedef enum ur_device_info_t {
     UR_DEVICE_INFO_HOST_PIPE_READ_WRITE_SUPPORTED = 111,        ///< [::ur_bool_t] Return true if the device supports enqueing commands to
                                                                 ///< read and write pipes from the host.
     UR_DEVICE_INFO_MAX_REGISTERS_PER_WORK_GROUP = 112,          ///< [uint32_t] The maximum number of registers available per block.
+    UR_DEVICE_INFO_DEVICE_IP_VERSION = 113,                     ///< [uint32_t] Device IP version. The meaning of the device IP version is
+                                                                ///< implementation-defined, but newer devices should have a higher version 
+                                                                ///< than older devices.
     /// @cond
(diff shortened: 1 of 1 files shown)


dm-vodopyanov:
> This file is automatically generated from a YAML description of the API. Please read the [Contributing Guide](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html) about how to make these changes.

Thanks, generated

scripts/core/device.yml
```diff
@@ -383,6 +383,8 @@ etors:
       desc: "[$x_bool_t] Return true if the device supports enqueing commands to read and write pipes from the host."
     - name: MAX_REGISTERS_PER_WORK_GROUP
       desc: "[uint32_t] The maximum number of registers available per block."
+    - name: DEVICE_IP_VERSION
```
kbenzie, line 386: The generator actually prefixed `UR_DEVICE_INFO_` to the enumerator name so this ends up becoming `UR_DEVICE_INFO_DEVICE_IP_VERSION` which is an unnecessary repetition of `DEVICE`.

```suggestion
    - name: IP_VERSION
```

kbenzie (changes requested):
One small fixup, otherwise looking good!

scripts/core/device.yml
dm-vodopyanov: Applied in https://github.com/oneapi-src/unified-runtime/pull/573/commits/53eb2104392c07fbdfa3a6425592164a977bdf1b

kbenzie (approved):

veselypeta (approved):

PR #576, opened 2023-06-06
Add queue properties to support immediate and batched submissions.

rdeodhar:
This change adds two new queue properties which will be used to handle the corresponding SYCL queue properties for immediate and batched submissions. Short-running kernels are best collected into batches and submitted together.  Long-running kernels are better submitted immediately.

In batched mode, submission is separated from the actual start of execution on the GPU. Multiple submissions can be collected and then issued together.

In immediate mode submission of an operation and start of execution happen together.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -4000,6 +4000,12 @@ typedef enum ur_queue_flag_t {
     UR_QUEUE_FLAG_DISCARD_EVENTS = UR_BIT(4),                ///< Events will be discarded
     UR_QUEUE_FLAG_PRIORITY_LOW = UR_BIT(5),                  ///< Low priority queue
     UR_QUEUE_FLAG_PRIORITY_HIGH = UR_BIT(6),                 ///< High priority queue
+    UR_QUEUE_FLAG_SUBMISSION_BATCHED = UR_BIT(7),            ///< Do not necessarily submit the operation as
+                                                             ///< soon as enqueue completes. Collect operations
+                                                             ///< and then submit at a later time in a batch.
+                                                             ///< Batch size will be determined automatically.
+    UR_QUEUE_FLAG_SUBMISSION_IMMEDIATE = UR_BIT(8),          ///< Submit immediately after the enqueue operation
+                                                             ///< completes.
     /// @cond
     UR_QUEUE_FLAG_FORCE_UINT32 = 0x7fffffff
     /// @endcond
--- a/test/conformance/queue/urQueueCreate.cpp
+++ b/test/conformance/queue/urQueueCreate.cpp
@@ -88,3 +88,17 @@ TEST_P(urQueueCreateTest, InvalidQueueProperties) {
     ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_QUEUE_PROPERTIES,
                      urQueueCreate(context, device, &props, &queue));
 }
+
+TEST_P(urQueueCreateTest, InvalidQueueProperties) {
+  ur_queue_handle_t queue = nullptr;
+
+  // It should be an error to specify both batched and immediate submission
+  ur_queue_properties_t props = {
+      /*.stype =*/UR_STRUCTURE_TYPE_QUEUE_PROPERTIES,
+      /*.pNext =*/nullptr,
+      /*.flags =*/UR_QUEUE_FLAG_SUBMISSION_BATCHED |
+          UR_QUEUE_FLAG_SUBMISSION_IMMEDIATE,
+  };
+  ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_QUEUE_PROPERTIES,
+                   urQueueCreate(context, device, &props, &queue));
+}
--- a/source/common/ur_params.hpp
+++ b/source/common/ur_params.hpp
@@ -7093,6 +7093,14 @@ inline std::ostream &operator<<(std::ostream &os, enum ur_queue_flag_t value) {
     case UR_QUEUE_FLAG_PRIORITY_HIGH:
         os << "UR_QUEUE_FLAG_PRIORITY_HIGH";
         break;
+
+    case UR_QUEUE_FLAG_SUBMISSION_BATCHED:
+      os << "UR_QUEUE_FLAG_SUBMISSION_BATCHED";
+      break;
+
+    case UR_QUEUE_FLAG_SUBMISSION_IMMEDIATE:
+      os << "UR_QUEUE_FLAG_SUBMISSION_IMMEDIATE";
+      break;
     default:
         os << "unknown enumerator";
         break;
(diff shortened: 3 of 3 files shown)


include/ur_api.h
```diff
@@ -4000,6 +4000,12 @@ typedef enum ur_queue_flag_t {
     UR_QUEUE_FLAG_DISCARD_EVENTS = UR_BIT(4),                ///< Events will be discarded
     UR_QUEUE_FLAG_PRIORITY_LOW = UR_BIT(5),                  ///< Low priority queue
     UR_QUEUE_FLAG_PRIORITY_HIGH = UR_BIT(6),                 ///< High priority queue
+    UR_QUEUE_FLAG_SUBMISSION_BATCHED = UR_BIT(7),            ///< Do not necessarily submit the operation as
```
jandres742, line 4003: Do not necessarily submit -> Opportunistically submit the operation ...

rdeodhar: OK, done.

kbenzie (changes requested):
These files are automatically generated from a YAML description of the API. Please read the [Contributing Guide](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html) for infomation about how to go about making changes.

kbenzie:
> How do these changes get tested? I am mainly concerned with the code factoring done in ur_params.hpp

To validate these changes with `ur_params.hpp` you should point the [FetchContent](https://github.com/intel/llvm/blob/sycl/sycl/plugins/unified_runtime/CMakeLists.txt#L6-L13) in the intel/llvm repo at this fork & branch, then run the `check-sycl-e2e` tests targetting the L0 backend.

rdeodhar:
I've read the "contributing" document and have made the primary change in queue.yaml.

Running "make generate-code" leads to a doxygen not found error. Is doxygen expected to be in the path?

rdeodhar:
I tried to redo this PR by changing only queue.yml and urQueueCreate.cpp but there are extensive changes in many files generated and quite a few of the changes appear to be formatting changes. I may not be using the right cmake and make invocations. Can you post the recommended cmake and make invocations I should use? I have read the Contributing Guide but it lists many alternative cmake and make invocations. I may not be using the right ones.

kbenzie:
> I've read the "contributing" document and have made the primary change in queue.yaml.
> 
> Running "make generate-code" leads to a doxygen not found error. Is doxygen expected to be in the path?

Yes, `doxygen` should be present on the `PATH`. I'll check if this is mentioned in the docs and if not make sure its added in future.

> I tried to redo this PR by changing only queue.yml and urQueueCreate.cpp but there are extensive changes in many files generated and quite a few of the changes appear to be formatting changes. I may not be using the right cmake and make invocations. Can you post the recommended cmake and make invocations I should use? I have read the Contributing Guide but it lists many alternative cmake and make invocations. I may not be using the right ones.

If you're seeing formatting changes that would suggest to me that `clang-format` version 15 was not found by make, this would also mean that the `generate` target is not defined. I think that's backed up by your previous comment you mention the `generate-code` target rather than the `generate` target.

Since you mention `make` can I assume you're using Linux?

If so, this is the workflow I recommend. First install `doxygen` with the system package manager. Then invoke the following commands at the root of the repository:

```bash
# create a python virtual environment
python3 -m venv .local
# then activate the python virtual environment
source .local/bin/activate
# install the python packages, this convienetly also includes clang-format version 15
pip install -r third_party/requirements.txt
# configure a build directory with cmake
cmake . -Bbuild -DCMAKE_BUILD_TYPE=Release
```

With that you should have configured a working build environment and can go ahead with building:

```
# activate the python virtual environment
# this is only required once for new terminal sessions
source .local/bin/activate
# run the generate target
make -C build generate
```

At this point, you should commit all of the files both the ones you've edited and those changed by the generator scripts.

I hope this helps. I've also opened #582 to make sure we improve the instructions for first setup moving forwards.

rdeodhar:
Thanks for the instructions. We work on shared Ubuntu machines where we don't have root access. So I had to use a simplified version of the steps. 
I ensured doxygen and clang-format (15.0.1) were in the path. Then just this:

cmake . -Bbuild -DCMAKE_BUILD_TYPE=Release
make -C build generate

Things look much better.

kbenzie:
> We work on shared Ubuntu machines where we don't have root access.

Interesting, are you able to run containers (`docker`/`podman`) on your shared Ubuntu machines? If so, we could potentially provide a container image with all the dependencies installed to ease this process.

test/conformance/queue/urQueueCreate.cpp
```diff
@@ -78,13 +78,17 @@ TEST_P(urQueueCreateTest, InvalidValueProperties) {
...
+        /*.flags =*/UR_QUEUE_FLAG_PRIORITY_HIGH | UR_QUEUE_FLAG_PRIORITY_LOW |
+            UR_QUEUE_FLAG_SUBMISSION_BATCHED |
+            UR_QUEUE_FLAG_SUBMISSION_IMMEDIATE,
     };
+    // It should be an error to specify both low/high priorities
+    ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_QUEUE_PROPERTIES,
+                     urQueueCreate(context, device, &props, &queue));
+    // It should be an error to specify both batched and immediate submission
     ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_QUEUE_PROPERTIES,
                      urQueueCreate(context, device, &props, &queue));
```
kbenzie, line 93: This is asserting the same thing twice, instead the `prop.flags` in each call to `urQueueCreate` should be testing the flags the comments actually talk about:

```suggestion
    // It should be an error to specify both low/high priorities
    ur_queue_properties_t props = {
        /*.stype =*/UR_STRUCTURE_TYPE_QUEUE_PROPERTIES,
        /*.pNext =*/nullptr,
        /*.flags =*/UR_QUEUE_FLAG_PRIORITY_HIGH | UR_QUEUE_FLAG_PRIORITY_LOW,
    };
    ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_QUEUE_PROPERTIES,
                     urQueueCreate(context, device, &props, &queue));
    // It should be an error to specify both batched and immediate submission
    props.flags =
        UR_QUEUE_FLAG_SUBMISSION_BATCHED | UR_QUEUE_FLAG_SUBMISSION_IMMEDIATE;
    ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_QUEUE_PROPERTIES,
                     urQueueCreate(context, device, &props, &queue));
```

kbenzie (changes requested):

test/conformance/queue/urQueueCreate.cpp
rdeodhar: I changed the test to separately test low/high and batched/immediate slightly differently than the suggestion to avoid clobbering an existing queue object.

kbenzie: Thanks, if you could also run `clang-format` on this file that should fix the failed GitHub Actions.

rdeodhar: Formatting change has been made.

scripts/core/queue.yml
```diff
@@ -62,6 +62,12 @@ etors:
     - name: PRIORITY_HIGH
       value: "$X_BIT(6)"
       desc: "High priority queue"
+    - name: SUBMISSION_BATCHED
+      value: "$X_BIT(7)"
+      desc: "Enqueue and submit in a batch later, as determined by adaptor or driver"
+    - name: SUBMISSION_IMMEDIATE
+      value: "$X_BIT(8)"
+      desc: "Enqueue and submit immediately"
```
aarongreig, line 70: I think these could do with a remark in the descriptions on whether they are performance hints with no semantic impact on API usage, or otherwise if they're expected to affect/be affected by stuff like the `QueueFlush` entry point

rdeodhar: Changed "desc" field to say these properties are hints to the implementation and do not affect queue semantics.

veselypeta (approved):

scripts/core/queue.yml
aarongreig: thanks, you'll need to run the generate target again so the changes are reflected in the headers etc.

rdeodhar: Ok, done.

PR #578, opened 2023-06-06
Add definitions in support of USM Import/Release functions.

rdeodhar:
This change adds the definitions necessary for two new functions: USMImport and USMRelease.

USMImport converts host system memory into something akin to USM host memory.
USMRelease undoes the effect of a previous USMImport.

This functionality is used by SYCL to import/release host memory to speed up host <-> device data transfers.


kbenzie:
@rdeodhar following up after this was discussed in the WG call. Given that these additions are intended to expose experimental Level Zero functionality they should also be defined as a [Experimental Feature](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#experimental-features) in Unified Runtime, #560 is a practical example.

@gmlueck also noted that we should only define the API based on what is currently supported by Level Zero rather than the semantics that will be added at a later date. This maps well to our definition for an experimental feature, which do not require long term API/ABI stability.

PR #580, opened 2023-06-07
[ur] Actually fail when RST validation finds an error

kbenzie:
Following on from #569 and #579 with the goal of failing the build when
the generation scripts have errors, this patch introduces exceptions to
the `_generate_valid_rst` where messages containing `error :` were being
printed.


veselypeta (approved):

martygrant (approved):

scripts/generate_docs.py
```diff
@@ -128,14 +130,15 @@ def _generate_valid_rst(fin, fout, namespace, tags, ver, rev, meta):
                     symbol_type = _find_symbol_type(symbol, meta)
                     if not symbol_type:
                         print("%s(%s) : error : symbol '%s' not found"%(fin, iline+1, symbol))
-                        continue
+                        raise Exception()
```
omarahmed1111, line 133: [Suggestion] maybe we could `continue` here and mark a flag to fail, then check that flag and raise and exception if this flag was set that there is failure. This way we print all symbol problems

kbenzie: I'll see if that's feasible.

kbenzie: Done and this is a much nicer user experience.

PR #588, opened 2023-06-09
[UR] Assign registry IDs from largest value

### LINKED ISSUE #587: Only append to the highest function ID in registry.yml
As noted in https://github.com/oneapi-src/unified-runtime/pull/584#discussion_r1223606810 when the sequence of enumerations of function ID's in `scripts/core/registry.yml` contains a gap, the script will fill that gap.

XPTI expects function ID's to be globally unique over time, so reusing ID's of functions which have been removed is not ideal. Instead we should only incremenet the highest ID when adding new functions.

veselypeta:
Closes #587

### DIFF
--- a/scripts/generate_ids.py
+++ b/scripts/generate_ids.py
@@ -37,27 +37,22 @@ def generate_registry(path, specs):
     except (TypeError, IndexError, KeyError) as e:
         raise Exception('invalid existing registry... ' + str(e))
 
-    updated = False
-
     new_registry = list()
     for fname in functions:
         etor_name = util.to_snake_case(fname)
         id = etors.get(etor_name)
         if id is None:
-            updated = True
             max_reg_entry += 1
             id = max_reg_entry
         new_registry.append({'name': util.to_snake_case(fname), 'desc': 'Enumerator for $x'+fname, 'value': str(id)})
 
-    if updated is False:
-        return
 
-    print("Updating registry %s"%path)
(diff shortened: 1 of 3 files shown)


scripts/generate_ids.py
```diff
@@ -56,7 +57,7 @@ def generate_registry(path, specs):
     ids = new_registry
     ids = sorted(ids, key=lambda x: int(x['value']))
     wrapper = { 'name': ENUM_NAME, 'type': 'enum', 'desc': 'Defines unique stable identifiers for all functions' , 'etors': ids}
-    header = {'type': 'header', 'desc': quoted('Intel$OneApi Unified Rutime function registry'), 'ordinal': quoted(9)}
+    header = {'type': 'header', 'desc': quoted('Intel$OneApi Unified Runtime function registry'), 'ordinal': quoted(9)}
```
kbenzie, line 60:
```suggestion
    header = {'type': 'header', 'desc': quoted('Intel $OneApi Unified Runtime function registry'), 'ordinal': quoted(9)}
```

I was going to comment on the other places but this should also update the registry.yml, right?

veselypeta: Sometimes :(
The registry is only updated if it detects changes (a new function is added). I think we should just regenerate it every time - performance isn't really an issue here so no sense to early return really.

kbenzie (approved):

pbalcer (approved):

PR #593, opened 2023-06-12
[UR] Fix documentation build script

### LINKED ISSUE #586: Document generation doesn't fail build when errors occur
To generate the docs the script generator calls a `sphinx-build` command to generate the HTML docs. If the `sphinx-build` command is unavailable in the system the build will complete without errors with no documentation being generated. We should raise an error when this situation occurs. The script also uses `os.system()` to run the command, we should instead change this to use `subprocess.run()`

veselypeta:
Closes #586

### DIFF
--- a/scripts/generate_docs.py
+++ b/scripts/generate_docs.py
@@ -245,9 +245,12 @@ def generate_html(dstpath):
     sourcepath = os.path.join(dstpath, "source")
 
     print("Generating HTML...")
-    cmdline = "sphinx-build -M html %s ../docs"%sourcepath
-    print(cmdline)
-    os.system(cmdline)
+    result = subprocess.run(["sphinx-build", "-M", "html", sourcepath, "../docs" ], stderr=subprocess.PIPE)
+    if result.returncode != 0:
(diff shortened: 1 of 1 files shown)


scripts/generate_docs.py
```diff
@@ -257,9 +257,13 @@ def generate_html(dstpath):
     sourcepath = os.path.join(dstpath, "source")
 
     print("Generating HTML...")
-    cmdline = "sphinx-build -M html %s ../docs"%sourcepath
-    print(cmdline)
-    os.system(cmdline)
+    proc = subprocess.Popen(["sphinx-build", "-M", "html", sourcepath, "../docs" ], stderr=subprocess.PIPE)
```
pbalcer, line 260: hmm, this might be simpler:
```
result = subprocess.run(f"sphinx-build -M html {sourcepath} ../docs", stderr=subprocess.PIPE)
if result.returncode != 0:
   ...
```

veselypeta: Done. 👍

pbalcer (approved):

PR #594, opened 2023-06-12
[Exp][Bindless Images] Experimental Bindless Images

isaacault:
Experimental support for bindless images in Unified Runtime.

This feature is to add support for the [experimental implementation of bindless images extension #9665](https://github.com/intel/llvm/pull/9665) in DPC++ after [Port CUDA plugin to Unified Runtime #9512](https://github.com/intel/llvm/pull/9512) is merged.

### DIFF
--- a/include/ur_ddi.h
+++ b/include/ur_ddi.h
@@ -1214,6 +1214,331 @@ typedef ur_result_t(UR_APICALL *ur_pfnGetQueueProcAddrTable_t)(
     ur_queue_dditable_t *);
 
 ///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesUnsampledImageHandleDestroyExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesUnsampledImageHandleDestroyExp_t)(
+    ur_context_handle_t,
+    ur_exp_image_handle_t);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesSampledImageHandleDestroyExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesSampledImageHandleDestroyExp_t)(
+    ur_context_handle_t,
+    ur_exp_image_handle_t);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesImageAllocateExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesImageAllocateExp_t)(
+    ur_context_handle_t,
+    const ur_image_format_t *,
+    const ur_image_desc_t *,
+    ur_exp_image_mem_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesImageFreeExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesImageFreeExp_t)(
+    ur_context_handle_t,
+    ur_exp_image_mem_handle_t);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesUnsampledImageCreateExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesUnsampledImageCreateExp_t)(
+    ur_context_handle_t,
+    ur_exp_image_mem_handle_t,
+    const ur_image_format_t *,
+    const ur_image_desc_t *,
+    ur_mem_handle_t *,
+    ur_exp_image_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesSampledImageCreateExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesSampledImageCreateExp_t)(
+    ur_context_handle_t,
+    ur_exp_image_mem_handle_t,
+    const ur_image_format_t *,
+    const ur_image_desc_t *,
+    ur_sampler_handle_t,
+    ur_mem_handle_t *,
+    ur_exp_image_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesImageCopyExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesImageCopyExp_t)(
+    ur_context_handle_t,
+    void *,
+    void *,
+    const ur_image_format_t *,
+    const ur_image_desc_t *,
+    ur_exp_image_copy_flags_t,
+    uint32_t,
+    const ur_event_handle_t *,
+    ur_event_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesImageGetInfoExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesImageGetInfoExp_t)(
+    ur_exp_image_mem_handle_t,
+    ur_image_info_t,
+    void *,
+    size_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesMipmapGetLevelExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesMipmapGetLevelExp_t)(
+    ur_context_handle_t,
+    ur_exp_image_mem_handle_t,
+    uint32_t,
+    ur_exp_image_mem_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesMipmapFreeExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesMipmapFreeExp_t)(
+    ur_context_handle_t,
+    ur_exp_image_mem_handle_t);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesImportOpaqueFDExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesImportOpaqueFDExp_t)(
+    ur_context_handle_t,
+    size_t,
+    uint32_t,
+    ur_exp_interop_mem_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesMapExternalArrayExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesMapExternalArrayExp_t)(
+    ur_context_handle_t,
+    const ur_image_format_t *,
+    const ur_image_desc_t *,
+    ur_exp_interop_mem_handle_t,
+    ur_exp_image_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesReleaseInteropExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesReleaseInteropExp_t)(
+    ur_context_handle_t,
+    ur_exp_interop_mem_handle_t);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesImportExternalSemaphoreOpaqueFDExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesImportExternalSemaphoreOpaqueFDExp_t)(
+    ur_context_handle_t,
+    uint32_t,
+    ur_exp_interop_semaphore_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesDestroyExternalSemaphoreExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesDestroyExternalSemaphoreExp_t)(
+    ur_context_handle_t,
+    ur_exp_interop_semaphore_handle_t);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesWaitExternalSemaphoreExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesWaitExternalSemaphoreExp_t)(
+    ur_queue_handle_t,
+    ur_exp_interop_semaphore_handle_t,
+    uint32_t,
+    const ur_event_handle_t *,
+    ur_event_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesSignalExternalSemaphoreExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesSignalExternalSemaphoreExp_t)(
+    ur_queue_handle_t,
+    ur_exp_interop_semaphore_handle_t,
+    uint32_t,
+    const ur_event_handle_t *,
+    ur_event_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Table of BindlessImagesExp functions pointers
+typedef struct ur_bindless_images_exp_dditable_t {
+    ur_pfnBindlessImagesUnsampledImageHandleDestroyExp_t pfnUnsampledImageHandleDestroyExp;
+    ur_pfnBindlessImagesSampledImageHandleDestroyExp_t pfnSampledImageHandleDestroyExp;
+    ur_pfnBindlessImagesImageAllocateExp_t pfnImageAllocateExp;
+    ur_pfnBindlessImagesImageFreeExp_t pfnImageFreeExp;
+    ur_pfnBindlessImagesUnsampledImageCreateExp_t pfnUnsampledImageCreateExp;
+    ur_pfnBindlessImagesSampledImageCreateExp_t pfnSampledImageCreateExp;
+    ur_pfnBindlessImagesImageCopyExp_t pfnImageCopyExp;
+    ur_pfnBindlessImagesImageGetInfoExp_t pfnImageGetInfoExp;
+    ur_pfnBindlessImagesMipmapGetLevelExp_t pfnMipmapGetLevelExp;
+    ur_pfnBindlessImagesMipmapFreeExp_t pfnMipmapFreeExp;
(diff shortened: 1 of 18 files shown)


scripts/core/common.yml
```diff
@@ -109,6 +109,26 @@ desc: "Handle of memory object which can either be buffer or image"
...
+type: handle
+desc: "Handle of interop memory"
+class: $xBindlessImages
+name: "$x_exp_interop_mem_handle_t"
+--- #--------------------------------------------------------------------------
+type: handle
+desc: "Handle of interop semaphore"
+class: $xBindlessImages
+name: "$x_exp_interop_semaphore_handle_t"
+--- #--------------------------------------------------------------------------
```
kbenzie, line 131: These should instead be defined at the top of `scripts/core/exp-bindless-images.yml`.

Its not always possible, such as extending existing enums, but the idea is to avoid touching the core yaml files as much as possible when defining an experimental extension.

scripts/core/sampler.yml
```diff
@@ -107,6 +107,42 @@ returns:
...
+      desc: "[in] pointer to the sampler description"
+    - type: float
+      name: minMipmapLevelClamp
+      desc: "[in] minimum mipmap level from which we can sample, minimum value being 0"
+    - type: float
+      name: maxMipmapLevelClamp
+      desc: "[in] maximum mipmap level from which we can sample, maximum value being the number of levels"
+    - type: float
+      name: maxAnistropy
+      desc: "[in] anisotropic ratio used when samplling the mipmap with anisotropic filtering"
```
kbenzie, line 134: I don't think we need a new `urSamplerCreateExp` entry point in order to specify these extra arguments because the `pDesc` argument is extensible using a `pNext` chain.

For an example of this pattern take a look at [urMemBufferCreate](https://oneapi-src.github.io/unified-runtime/core/api.html#urmembuffercreate) where the [ur_buffer_channel_properties_t](https://oneapi-src.github.io/unified-runtime/core/api.html#structur__buffer__channel__properties__t) and [ur_buffer_alloc_location_properties_t](https://oneapi-src.github.io/unified-runtime/core/api.html#structur__buffer__alloc__location__properties__t) structs can be optionally specificed as part of a `pNext` chain.

This new struct should also be defined in the `source/core/exp-bindless-images.yml` file leaving no changes to this file.

scripts/core/usm.yml
```diff
@@ -331,6 +331,62 @@ returns:
...
+      - "`pUSMDesc && pUSMDesc->align != 0 && ((pUSMDesc->align & (pUSMDesc->align-1)) != 0)`" # alignment must be power of two
+      - "If `align` is greater that the size of the largest data type supported by `hDevice`."
+    - $X_RESULT_ERROR_INVALID_USM_SIZE:
+      - "`widthInBytes == 0`"
+      - "`size` is greater than $X_DEVICE_INFO_MAX_MEM_ALLOC_SIZE."
+    - $X_RESULT_ERROR_INVALID_OPERATION:
+      - "If `UR_DEVICE_INFO_USM_SINGLE_SHARED_SUPPORT` and `UR_DEVICE_INFO_USM_CROSS_SHARED_SUPPORT` are both false."
+    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
+--- #--------------------------------------------------------------------------
```
kbenzie, line 388: This should be defined in the `source/core/exp-bindless-images.yml` file.

scripts/core/device.yml
```diff
@@ -383,6 +383,38 @@ etors:
...
+    - name: MIPMAP_LEVEL_REFERENCE_SUPPORT_EXP
+      desc: "[$x_bool_t] returns true if the device supports using images created from individual mipmap levels"
+    - name: INTEROP_MEMORY_IMPORT_SUPPORT_EXP
+      desc: "[$x_bool_t] returns true if the device supports importing external memory resources"
+    - name: INTEROP_MEMORY_EXPORT_SUPPORT_EXP
+      desc: "[$x_bool_t] returns true if the device supports exporting internal memory resources"
+    - name: INTEROP_SEMAPHORE_IMPORT_SUPPORT_EXP
+      desc: "[$x_bool_t] returns true if the device supports importing external semaphore resources"
+    - name: INTEROP_SEMAPHORE_EXPORT_SUPPORT_EXP
+      desc: "[$x_bool_t] returns true if the device supports exporting internal event resources"
```
kbenzie, line 417: Each of these enumerations should also defined a value starting at an offset to avoid interacting with the core enumeration value range. In the command buffer experimental feature they started at `0x1000` but will need to verify we don't overlap that range as well.

scripts/core/event.yml
```diff
@@ -75,6 +75,10 @@ etors:
       desc: Event created by $xEnqueueWriteHostPipe
     - name: COMMAND_BUFFER_ENQUEUE_EXP
       desc: Event created by $xCommandBufferEnqueueExp
+    - name: INTEROP_SEMAPHORE_WAIT_EXP
+      desc: Event created by $xBindlessImagesWaitExternalSemaphoreExp
+    - name: INTEROP_SEMAPHORE_SIGNAL_EXP
+      desc: Event created by $xBindlessImagesSignalExternalSemaphoreExp
```
kbenzie, line 81: These should defined `value:`s at an offset, e.g. `0x1000`, to avoid conflicting with the core spec enumeration range.

kbenzie (changes requested):

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
+      name: pImageDesc
+      desc: "[in] pointer to image description"
+    - type: $x_exp_image_mem_handle_t*
+      name: phImageMem
+      desc: "[out] pointer to handle of image memory allocated"
+returns:
+    - $X_RESULT_ERROR_INVALID_CONTEXT
+    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR:
+      - "`pImageDesc && UR_MEM_TYPE_IMAGE1D_BUFFER < pImageDesc->type`"
```
veselypeta, line 91: I wonder if this is needed - I think the api generator will generate this check anyway.

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
+etors:
+  - name: NEAREST
+    desc: "Filter mode nearest."
+  - name: LINEAR
+    desc: "Filter mode linear."
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Destroy bindless unsampled image handles"
+class: $xBindlessImages
+name: UnsampledImageHandleDestroyExp
```
veselypeta, line 31: I wonder if the `Handle` is redundant in the function name?

include/ur_ddi.h
```diff
@@ -1485,6 +1701,46 @@ typedef ur_result_t(UR_APICALL *ur_pfnGetUSMProcAddrTable_t)(
     ur_api_version_t,
     ur_usm_dditable_t *);
 
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urUSMPitchedAllocExp
+typedef ur_result_t(UR_APICALL *ur_pfnUSMPitchedAllocExp_t)(
+    ur_context_handle_t,
```
JackAKirk, line 1707: I don't think that you need the `ur_context_handle_t` argument, you get the native `cuContext` from the device argument. I think this is true of all adapters that will support this feature.
I don't think unused arguments should be included in the interfaces.

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
+    - type: $x_context_handle_t
+      name: hContext
+      desc: "[in] handle of the context object"
+    - type: $x_exp_image_handle_t
+      name: hImage
+      desc: "[in] pointer to handle of image object to destroy"
+returns:
+    - $X_RESULT_ERROR_INVALID_CONTEXT
+    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_INVALID_MEM_OBJECT
```
veselypeta, line 45:
```suggestion
```
There is no mem object passed to this function.

include/ur_ddi.h
```diff
@@ -747,6 +784,185 @@ typedef ur_result_t(UR_APICALL *ur_pfnGetMemProcAddrTable_t)(
...
+    const ur_image_format_t *,
+    const ur_image_desc_t *,
+    ur_exp_interop_mem_handle_t,
+    ur_exp_image_handle_t *);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function-pointer for urBindlessImagesReleaseInteropExp
+typedef ur_result_t(UR_APICALL *ur_pfnBindlessImagesReleaseInteropExp_t)(
+    ur_context_handle_t,
```
JackAKirk, line 889: Will this interface work when we have multiple devices per `ur_context_handle_t`? Shouldn't there be a device argument instead?

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
```
veselypeta, line 64: same as above

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
+      name: phImageMem
+      desc: "[out] pointer to handle of image memory allocated"
+returns:
+    - $X_RESULT_ERROR_INVALID_CONTEXT
+    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR:
+      - "`pImageDesc && UR_MEM_TYPE_IMAGE1D_BUFFER < pImageDesc->type`"
+    - $X_RESULT_ERROR_INVALID_IMAGE_SIZE
+    - $X_RESULT_ERROR_INVALID_OPERATION
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
```
veselypeta, line 94:
```suggestion
```
This will be added automatically by the api generator

include/ur_ddi.h
JackAKirk: Basically when this is merged https://github.com/intel/llvm/pull/6446
you might have bugs if the interface is taking context when it should be taking device. Also more generally it could lead to bugs in backends other than cuda/hip.

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
+      name: phMem
+      desc: "[out] pointer to handle of image object created"
+    - type: $x_exp_image_handle_t*
+      name: phImage
+      desc: "[out] pointer to handle of image object created"
+returns:
+    - $X_RESULT_ERROR_INVALID_CONTEXT
+    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR:
+      - "`pImageDesc && UR_MEM_TYPE_IMAGE1D_BUFFER < pImageDesc->type`"
```
veselypeta, line 144: Maybe can be removed.

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
```
veselypeta, line 182: same as above

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
+    - $X_RESULT_ERROR_INVALID_IMAGE_SIZE
+    - $X_RESULT_ERROR_INVALID_SAMPLER
+    - $X_RESULT_ERROR_INVALID_OPERATION
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
```
veselypeta, line 186:
```suggestion
```

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
+      name: phImage
+      desc: "[out] pointer to handle of image object created"
+returns:
+    - $X_RESULT_ERROR_INVALID_CONTEXT
+    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR:
+      - "`pImageDesc && UR_MEM_TYPE_IMAGE1D_BUFFER < pImageDesc->type`"
+    - $X_RESULT_ERROR_INVALID_IMAGE_SIZE
+    - $X_RESULT_ERROR_INVALID_OPERATION
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
```
veselypeta, line 147:
```suggestion
```

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
+            must be complete. 
+    - type: $x_event_handle_t*
+      name: phEvent
+      desc: |
+            [out][optional] return an event object that identifies this particular command instance.
+returns:
+    - $X_RESULT_ERROR_INVALID_CONTEXT
+    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR:
+      - "`pImageDesc && UR_MEM_TYPE_IMAGE1D_BUFFER < pImageDesc->type`"
```
veselypeta, line 234: Can perhaps be removed.

scripts/core/exp-bindless-images.yml
```diff
@@ -0,0 +1,473 @@
...
+      name: paramInfoName
+      desc: "[in] queried info name"
+    - type: void*
+      name: pParamValue
+      desc: "[out] returned query value"
+    - type: "size_t*"
+      name: pParamValueSize
+      desc: "[out] returned query value size"
+returns:
+    - $X_RESULT_ERROR_INVALID_VALUE
```
veselypeta, line 257: We have a standard set of return codes for `GetInfo` type queries - you can pretty much copy paste them from [here](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/device.yml#L419)

scripts/core/registry.yml
```diff
@@ -376,9 +376,18 @@ etors:
 - name: COMMAND_BUFFER_FINALIZE_EXP
   desc: Enumerator for $xCommandBufferFinalizeExp
   value: '123'
+- name: BINDLESS_IMAGES_UNSAMPLED_IMAGE_HANDLE_DESTROY_EXP
+  desc: Enumerator for $xBindlessImagesUnsampledImageHandleDestroyExp
+  value: '124'
```
veselypeta, line 381: After you rebase - I would recommend reverting this file back to it's state in main and rerunning the generator. we have made changes to the way the registry values are allocated and we will no longer back-fill gaps.

scripts/core/sampler.yml
```diff
@@ -107,6 +107,42 @@ returns:
     - $X_RESULT_ERROR_OUT_OF_RESOURCES
 --- #--------------------------------------------------------------------------
 type: function
+desc: "Create a sampler object in a context"
+class: $xSampler
+name: CreateExp
```
veselypeta, line 112: I would recommend using the `pNext` chaining to pass additional arguments to `urSamplerCreate`

veselypeta (commented):
Looks good. I've added some suggestions about the return codes - I think some of them will be automatically added by the API generator, if not feel free to include them. Also please disregard my naming changes if you disagree - it's your experimental feature after all 👍

include/ur_ddi.h
kbenzie: I may be wrong but I think its included to enable the future work of supporting Level Zero.

JackAKirk: Just looked at L0 PI and you're likely right (Although I'm not certain in this specific case).
They use context only to prevent sycl UB of devices in different contexts copying memory to each other.
Point is l0 plugin seems to be a lot stricter on this than the last time I looked at it. Since level_zero seems to have fully adopted this context constraint probably there is nothing we can do about it now, even though noone ever came up with a use case for introducing this constraint (which really defines context) in any of these backends.

scripts/core/exp-bindless-images.yml
isaacault: Removed.

scripts/core/exp-bindless-images.yml
isaacault: Removed.

scripts/core/exp-bindless-images.yml
isaacault: Will treat all of these comments as the one above. Leaving for now, will remove if it is unnecessary.

scripts/core/exp-bindless-images.yml
isaacault: Thanks! Done.

scripts/core/registry.yml
isaacault: Done.

scripts/core/sampler.yml
isaacault: Done. The new struct is `$x_exp_sampler_mip_properties_t`

scripts/core/sampler.yml
isaacault: Done. The new struct is `$x_exp_sampler_mip_properties_t` in `source/core/exp-bindless-images.yml`

scripts/core/usm.yml
isaacault: Done. Removed from here.

scripts/core/exp-bindless-images.yml
isaacault: Removed.

scripts/core/exp-bindless-images.yml
veselypeta: No you're right in this case - ignore all my previous comments 👍

veselypeta (approved):
LGTM. Just one tiny thing 👍

kbenzie (approved):

scripts/core/device.yml
kbenzie: > Perhaps worth considering a document to register/reserve ranges? Maybe unnecessary though.

yeah, I'll add it to the list of improvements for this process.

include/ur_ddi.h
JackAKirk: L0 usm allocation functions appear to take and use a context argument. So I guess they will expect this new alloc function to also have the context.

JackAKirk (approved):

PR #599, opened 2023-06-12
[UR] Pointer to handle arguments require a range() check

veselypeta:
A pointer to handle argument requires that a `[range(start, end)]` check is included as part of it's description. This is required for the loader to do some local conversion of it's types. This also leads to some confusion as without this check the script generator will match a different part of the description resulting in other parts of the description being included into the generated files.

### DIFF
--- a/scripts/templates/helper.py
+++ b/scripts/templates/helper.py
@@ -1148,6 +1148,9 @@ def get_loader_prologue(namespace, tags, obj, meta):
             fty_name = re.sub(r"(\w+)_handle_t", r"\1_factory", tname)
 
             if type_traits.is_pointer(item['type']):
+                if not param_traits.is_range(item):
+                    print(item)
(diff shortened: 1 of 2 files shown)


scripts/templates/helper.py
```diff
@@ -1148,6 +1148,9 @@ def get_loader_prologue(namespace, tags, obj, meta):
             fty_name = re.sub(r"(\w+)_handle_t", r"\1_factory", tname)
 
             if type_traits.is_pointer(item['type']):
+                if not param_traits.is_range(item):
```
pbalcer, line 1151: I think this check should happen alongside other similar spec validation functions in `__validate_params` from `parse_spec.py`.

veselypeta: 👍

pbalcer (approved):

PR #602, opened 2023-06-12
[uma] add a function to query name of a provider

igchor:
To allow interacting with the underlying implementation directly. This will be needed to support returning native error codes (name can be used to figure out how to interpret those errors) and possibly to support provider-specific queries.

### DIFF
--- a/source/common/unified_memory_allocation/include/uma/memory_provider_ops.h
+++ b/source/common/unified_memory_allocation/include/uma/memory_provider_ops.h
@@ -49,6 +49,7 @@ struct uma_memory_provider_ops_t {
                                            size_t *pageSize);
     enum uma_result_t (*purge_lazy)(void *provider, void *ptr, size_t size);
     enum uma_result_t (*purge_force)(void *provider, void *ptr, size_t size);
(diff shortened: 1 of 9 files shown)


source/common/unified_memory_allocation/include/uma/memory_provider_ops.h
```diff
@@ -49,6 +49,7 @@ struct uma_memory_provider_ops_t {
                                            size_t *pageSize);
     enum uma_result_t (*purge_lazy)(void *provider, void *ptr, size_t size);
     enum uma_result_t (*purge_force)(void *provider, void *ptr, size_t size);
+    void (*name)(void *provider, const char **ppName);
```
vinser52, line 52: maybe better to rename `name` -> `get_name`?

vinser52 (approved):

source/common/unified_memory_allocation/include/uma/memory_provider_ops.h
igchor: Right, it's better for consistency with other getter functions. Done.

PR #605, opened 2023-06-13
[UR] Remove references to Level Zero

veselypeta:
Remove the last remaining references to level zero.

### DIFF
--- a/scripts/generate_docs.py
+++ b/scripts/generate_docs.py
@@ -203,18 +203,6 @@ def generate_rst(docpath, section, namespace, tags, ver, rev, specs, meta):
             meta=meta,
             specs=specs)
 
-
-"""
-Entry-point:
-    generate reference JSON file
-"""
-def generate_ref(dstpath, ref):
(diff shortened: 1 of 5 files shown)


aarongreig (approved):

scripts/generate_docs.py
```diff
@@ -213,7 +213,7 @@ def generate_ref(dstpath, ref):
     util.removePath(refpath)
     util.makePath(refpath)
 
-    util.jsonWrite(os.path.join(refpath, "level_zero.json"), ref)
```
pbalcer, line 216: You are only modifying the function that writes out this json file. If there's no corresponding reader and this is unused, I suggest we just remove this whole thing.

veselypeta: You are correct. This file is not used or read by anything. I'm not 100% sure why it exists, but it's places in `/docs/ref` perhaps for debugging reasons or as a reference? 
If you feel it's not worthwhile then we can remove.

PR #610, opened 2023-06-14
End-to-end code generation example

kurapov-peter:
I was able to come up with an end-to-end pipeline of code generation and execution that typically occurs in a JIT engine (our use case is the analytical query compilation in [HDK](https://github.com/intel-ai/hdk/tree/main)). It creates a simple function (adding +1 to an array and storing the result in a second one) via llvm api, converts it to spirv, and submits it to the runtime. 

The example is a bit complex in terms of dependencies: it uses llvm api, the spirv translator library, and intel gpu stack for execution. I found managing those in conda was the easiest way (I'm a bit biased though :) ). Nevertheless, I'm attaching the exact environment in `scripts/deps.yml` for reproduction simplicity (doesn't include the toolchain since any dev env for UR would already have it). Use as:
```bash
conda env create -f scripts/deps.yml
conda activate examples
```
Configure with:
```bash
cmake -DUR_BUILD_ADAPTER_L0=on ..
```

Ideally, we would want to be able to have the same flow for all the devices (including nvidia for which we would also convert to spirv if needed).

Corresponding L0-specific examples can be found [here](https://github.com/kurapov-peter/L0Snippets).

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -6,7 +6,9 @@
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 
 add_subdirectory(hello_world)
-add_subdirectory(codegen)
+if(UR_BUILD_EXAMPLE_CODEGEN)
+    add_subdirectory(codegen)
+endif()
 if(UR_ENABLE_TRACING)
     add_subdirectory(collector)
 endif()
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -6,6 +6,7 @@
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 
 add_subdirectory(hello_world)
+add_subdirectory(codegen)
 if(UR_ENABLE_TRACING)
     add_subdirectory(collector)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,7 @@ option(UR_BUILD_ADAPTER_L0 "build level 0 adapter from SYCL" OFF)
 option(UR_BUILD_ADAPTER_OPENCL "build opencl adapter from SYCL" OFF)
 option(UR_BUILD_ADAPTER_CUDA "build cuda adapter from SYCL" OFF)
 option(UR_BUILD_ADAPTER_HIP "build hip adapter from SYCL" OFF)
+option(UR_BUILD_EXAMPLE_CODEGEN "Build the codegen example." OFF)
 
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
--- a/examples/codegen/CMakeLists.txt
+++ b/examples/codegen/CMakeLists.txt
@@ -1,0 +1,39 @@
+# Copyright (C) 2022 Intel Corporation
+# Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+# See LICENSE.TXT
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+set(TARGET_NAME codegen)
+
+find_package(LLVM REQUIRED CONFIG)
+message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
+message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
+llvm_map_components_to_libnames(llvm_libs support core irreader bitwriter)
+
+find_package(PkgConfig REQUIRED)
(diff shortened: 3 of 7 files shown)


kbenzie:
I've converted this to draft for now until we resolve getting access to hardware.

kbenzie:
@kurapov-peter which versions of LLVM is this known to work with?

kbenzie:
> > @kurapov-peter which versions of LLVM is this known to work with?
> 
> This one is for llvm 10..14 (I used 14 to test it). I can update to support higher version if you like.

Ah okay, that probably explains why LLVM 16 wasn't working for me locally.

I feel like the CMake for finding the necessary packages will need to be more forgiving before we merge this, maybe disable building example if LLVM/SPIRV-LLVM-Translator aren't found instead of an error.

No rush though, I was mainly looking at this locally because we've introduced some changes to how UR is initalized recently and wanted to see if things still work with some modifications.

kurapov-peter:
@kbenzie, sure, will do.

kurapov-peter:
@kbenzie, I tested with llvm-16, should work now.

> I feel like the CMake for finding the necessary packages will need to be more forgiving before we merge this, maybe disable building example if LLVM/SPIRV-LLVM-Translator aren't found instead of an error.

Done. I had to add extra parenthesis for the `LLVMSPIRVLib_FOUND`. Funny :)

kbenzie:
> Done. I had to add extra parenthesis for the `LLVMSPIRVLib_FOUND`. Funny :)

Thanks. Yeah that's very odd - CMake 🤷 

I've been testing this rebased on a more recent commit on the adapters branch, this requires some changes to the loader/platform initialization which I'm happy to contribute. I am however seeing a SEGFAULT in the exit handlers which I'm still debugging, so will hold off until I've figured out what's going on there.

kbenzie:
> I am however seeing a SEGFAULT in the exit handlers which I'm still debugging, so will hold off until I've figured out what's going on there.

I reinstalled my Level Zero driver and this went away. I've been successfully running the example in Debug mode but get SEGFAULT's in Release or RelWithDebInfo modes.

I've created https://github.com/kurapov-peter/unified-runtime/pull/1 to update the example code with the adapter handles.

kurapov-peter:
Sounds good! I wasn't able to rebase without conflicts, so will need to resolve them first. Should I just merge the patch or do I need to rebase the current branch on top of the latest adapters branch first?

kbenzie:
> I reinstalled my Level Zero driver and this went away. I've been successfully running the example in Debug mode but get SEGFAULT's in Release or RelWithDebInfo modes.

I figure these out. There's no dependency on building the adapters for the examples so a full build is required, then the SEGFAULT's went away.

> Sounds good! I wasn't able to rebase without conflicts, so will need to resolve them first. 

We have some guidance on dealing with merge conflicts in the [Forks and Pull Requests](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#forks-and-pull-requests) section of the contirbution guide. Hopefully that will help resolving those.

> Should I just merge the patch or do I need to rebase the current branch on top of the latest adapters branch first?

Whatever works for you, if you manage to rebase on the adapters branch its probably easiest to cherry-pick the top commit from my patch branch and ignore the merge in from the adapters branch in my PR.

examples/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 
 add_subdirectory(hello_world)
+add_subdirectory(codegen)
```
veselypeta, line 9: Can we perhaps have this only enabled through a cmake option. I think we don't need to have to build each time.
```
option(UR_ENABLE_CODEGEN "Build the codegen example." OFF)
```

```suggestion
if(UR_ENABLE_CODEGEN)
    add_subdirectory(codegen)
endif()
```

kurapov-peter:
@veselypeta, I've rebased to the latest `adapters` state and cherry-picked the commit for API changes from https://github.com/kurapov-peter/unified-runtime/pull/1. I also added the missing copyrights and added the unlicense license to the original repo.

examples/CMakeLists.txt
kurapov-peter: Sure. Do you want an option per example or a generic `UR_ENABLE_EXAMPLES`?

pbalcer: My suggestion right would be to add `UR_BUILD_EXAMPLE_CODEGEN` just for this example to be consistent with other cmake options like this (`UR_BUILD_TOOLS`, `UR_BUILD_ADAPTER_...` and so on).

veselypeta (approved):
LGTM!

pbalcer (approved):
I added a few minor comments. I don't want to put more work on you, so I'm OK with leaving these as todo for later.

kbenzie:
> I added a few minor comments. I don't want to put more work on you, so I'm OK with leaving these as todo for later.

I've added #859 to track this.

PR #623, opened 2023-06-15
[UR] Refactor experimental feature docs to match template

### LINKED ISSUE #615: Refactor existing experimental features to match template.
We currently already have two experimental features in the unified runtime:
* Command Buffers
* Bind-less Images

These were added without using the template script introduced in #607 and hence do not match exactly the style that this sets out. We should update these feature documents to match for uniformity.

veselypeta:
Closes #615


scripts/core/EXP-COMMAND-BUFFER.rst
```diff
@@ -1,14 +1,14 @@
...
     X=x.upper()
 %>
+
 .. _experimental-command-buffer:
 
-==============
-Command-Buffer
-==============
+================================================================================
+Command Buffer
```
EwanC, line 10: I've tried to consistently hyphenate "command-buffer" everywhere in the spec (both here & OpenCL land), so would prefer to keep this hyphenated unless there's a reason do so causes issues with the docs build

EwanC (approved):
one minor question, but LGTM

scripts/core/EXP-COMMAND-BUFFER.rst
veselypeta: Yep. No reason it can't be hyphenated. By default the template script with de-hyphenate, but you're free to choose any name. I'll fix this. 👍

kbenzie (approved):

veselypeta:
Isaac is happy with these changes so will merge.

PR #625, opened 2023-06-15
[ur] optimize align allocation in DisjointPool

igchor:
Do not increase size by Alignment - 1 if the requested Alignment is smaller or equal to coarse-grain allocation alignemnt - in such case, Bucket for the allocation will already be properly aligned.


source/common/uma_pools/disjoint_pool.cpp
```diff
@@ -742,9 +758,15 @@ void *DisjointPool::AllocImpl::allocate(size_t Size, size_t Alignment,
...
-    // (using umaMemoryProviderGetMinPageSize)?
-    size_t AlignedSize = (Size > 1) ? (Size + Alignment - 1) : Alignment;
+    size_t AlignedSize;
+    if (Alignment <= ProviderMinPageSize) {
+        // This allocation will be served from a Bucket which size is multiple
+        // of Alignment and Slab address is aligned to ProviderMinPageSize
+        // so the bucket will be properly aligned.
+        AlignedSize = (Size > 1) ? AlignUp(Size, Alignment) : Alignment;
+    } else {
+        AlignedSize = Size + Alignment - 1;
```
pbalcer, line 768: I'm not exactly sure what's the purpose of adding alignment to the size here? Can't we just always `AlignUp`? This should always be called with `size` being a multiple of `alignment` anyway (which is what `AlignUp` enforces).

igchor: Consider a case where `size = alignment = 1MB`. Slabs are only aligned to `ProviderMinPageSize` (let's say 4KB) so aligning up to Alignment is not enough (and we're allocating from Slabs up to 2GB allocations).

When we're allocating directly from the memory provider (line 775) we don't need to do any adjustments, we just leave it to the provider.

pbalcer: This is potentially *very* wasteful if there are many such allocations. We just need to create a bucket that offsets the beginning of the slab. Or we can just create an aligned slab, which should be doable if alignment is page-aligned. And even if it's not, at worst we'd waste 4kb per slab.
Let's leave it as-is for now, but might be worth investigating if this is a common case and worth optimizing.

But please add a comment here explaining this.

igchor: Done. Just one clarification - the 2GB I mentioned is actually just the upper limit for `poolableSize`. By default `poolableSize` is around a few MB (so allocation bigger than a few MBs will bypass the pooling).

PR #626, opened 2023-06-16
[UR] Allow for extending enums

### LINKED ISSUE #606: Add support for extending existing enumerations in experimental feature yaml files
Currently if an experimental feature needs to extend an existing enum type, e.g. `ur_device_info_t`, the `scripts/core/device.yml` file must be edited. This conflicts with the goal of keeping experimental features self contained within their own `scripts/core/exp-<feature>.yml` file.

Add support to the generator scripts for extending an existing enumeration *after* it has already been defined. 

This should also take into consideration the Mako tempaltes which generate validation of the enum value ranges.

veselypeta:
Method to extend existing enums:

* Introduces a new yaml syntax - `enum` can now include and `extend` field to indicate that this extends an existing enum
* When parsing this new type, it inserts all the etors into already existing etor
* We have to manually refresh the meta data structure - to ensure downstream changes as picked up such as validation/loader etc.

Closes #606


scripts/parse_specs.py
```diff
@@ -568,7 +573,7 @@ def _generate_meta(d, ordinal, meta):
                 if bit_mask != 0:
                     meta[type][name]['bit_mask'] = hex(ctypes.c_uint32(~bit_mask).value)
             else:
-                meta[type][name]['max'] = d['etors'][idx]['name']
+                meta[type][name]['max'] = d['etors'][max_index]['name']
```
veselypeta, line 576: small bug fix

scripts/core/exp-bindless-images.yml
```diff
@@ -23,6 +23,78 @@ desc: "Handle of interop semaphore"
 class: $xBindlessImages
 name: "$x_exp_interop_semaphore_handle_t"
 --- #--------------------------------------------------------------------------
+type: enum_exp
```
kbenzie, line 26: What about still calling this `type: enum` but having an additional flag like `extend: true`?

kbenzie:
Once the design is done it should be documented in `scripts/core/CONTRIB.rst` and the `YaML.md`.

scripts/core/exp-bindless-images.yml
veselypeta: I like this idea 👍  - makes a few thing simpler

scripts/core/CONTRIB.rst
```diff
@@ -216,6 +216,14 @@ implement your experimental feature.
     $ python scripts/add_experimental_feature.py <name-of-your-experimental-feature>
 
 
+Experimental features *must* not make any changes to the core YaML files and 
+*must* be described entirely in their own YaML file. Sometimes, however 
+experimental feature require extending enumerations of the core specification. 
+If this is necessary should set the `extend` field to true and list the required
```
kbenzie, line 222: double backticks for monospace in rst files

scripts/core/CONTRIB.rst
```diff
@@ -216,6 +216,14 @@ implement your experimental feature.
     $ python scripts/add_experimental_feature.py <name-of-your-experimental-feature>
 
 
+Experimental features *must* not make any changes to the core YaML files and 
+*must* be described entirely in their own YaML file. Sometimes, however
```
kbenzie, line 220: Are there deffo no cases now where core yaml files would need to be touched?

kbenzie (approved):
LGTM - minor comments

scripts/core/CONTRIB.rst
kbenzie: Just want to ensure these uses of *must* are not too strong.

pbalcer: Extending existing structs/unions (e.g., adding a new partition method)?. But experimental features should probably avoid doing things like that in the first place.

scripts/core/CONTRIB.rst
pbalcer, line 222:
```suggestion
If this is necessary, create a new enum with the `extend` field set to true and list the required
```

scripts/parse_specs.py
```diff
@@ -797,6 +802,31 @@ def _generate_ref(specs, tags, ref):
...
+    if obj.get('class'):
+        meta['class'][obj['class']]['enum'].remove(obj['name'])
+        
+    del meta['enum'][obj['name']]
+    ## re-generate meta
+    meta = _generate_meta(obj, header['ordinal'], meta)
+
+
+def handle_enum_extensions(d, specs, header, meta):
+    matching_enum = [obj for s in specs for obj in s['objects'] if obj['type'] == 'enum' and d['name'] == obj['name']][0]
```
pbalcer, line 816: Does this assume that the matching enum already exists? If so, it might be necessary to ensure that enum extensions are always processed last.

scripts/core/CONTRIB.rst
veselypeta: Yeah I think extending existing structs is not allowed for experimental features. Hopefully if this is needed it can be done with the `pNext` chain. Unions I'm less sure about.

veselypeta:
Another idea I had. Instead of requiring every extended enum to provide a unique value - we could just specify a starting range.
i.e.
```yml
type: 'enum'
name: $x_device_info_t
desc: "a desc"
extended: true
offset: "0x1000"
etors:
  - name: "ETOR_1"
    desc: "desc"
  - name: "ETOR_2"
    desc: "desc 2"
```

So enums will automatically be allocated sequentially

scripts/core/exp-bindless-images.yml
```diff
@@ -24,6 +24,81 @@ class: $xBindlessImages
 name: "$x_exp_interop_semaphore_handle_t"
 --- #--------------------------------------------------------------------------
 type: enum
+extend: true
```
pbalcer, line 27: One of my comments somehow got lost. So again:

This should include `typed_etors: True` to ensure validation checks run. In general, some/most of the fields, except for `etors` need to be repeated exactly in these extend enum types.

pbalcer:
> Another idea I had. Instead of requiring every extended enum to provide a unique value - we could just specify a starting range. i.e.
> 
> ```yaml
> type: 'enum'
> name: $x_device_info_t
> desc: "a desc"
> extended: true
> offset: "0x1000"
> etors:
>   - name: "ETOR_1"
>     desc: "desc"
>   - name: "ETOR_2"
>     desc: "desc 2"
> ```
> 
> So enums will automatically be allocated sequentially

Do the specific values matter? If not, would it be possible to, for example, automatically allocate ranges for new extensions from the top of the enum's range? e.g.,
```
enum FOO {
   FOO_OPTION_A = 1,
   FOO_OPTION_B = 2,
    ....

   EXT2_OPTION_A,
   EXT2_OPTION_B,
   EXT1_OPTION_A,
   EXT1_OPTION_B,
   MAX = UINT32_MAX,
};
```

veselypeta:
> Do the specific values matter? If not, would it be possible to, for example, automatically allocate ranges for new extensions from the top of the enum's range? e.g.,

I don't think so, but I think that assigning values could be problematic. Possibly adding a new experimental feature might change the order in which they are allocated and modify the range.

scripts/core/exp-bindless-images.yml
veselypeta: Done.

veselypeta:
I've decided not to introduce `offset` or automatically allocating experimental feature ranges yet, instead enforcing that each etor must have a unique value. I feel like we can easily iterate on this and add this at a later point.

pbalcer (approved):

scripts/core/CONTRIB.rst
kbenzie: Extending a struct/union implies an ABI change which I think we should avoid for experimental features. That's how I've been thinking about this.

Issue #500, opened 2023-05-09
Improve error reporting errors up through the stack


kbenzie:
Currently UR adapters make a best effort to map driver specific errors to `ur_result_t` enumerations and return those to the parallel language runtime on top of UR. This approach is problematic as it obscures details about how an adapter is using a driver from the user, details which are very often necessary to determine how to resolve a given error condition.

This purpose of this issue is to track proposed solutions, the decision making process, and to determine the next steps to enable improved error reporting.

This is blocking progress on the design of #68 and is related to #471.

alycm:
- Already have a UR_BACKEND_SPECIFIC_ERROR enum to indicate when string can be returned
- Still need to get error codes back from UR for native backend error codes, ideally not by parsing the error string.

JackAKirk:
L0 already uses UR_BACKEND_SPECIFIC_ERROR consistently. The pi_cuda and pi_hip mostly don't use it. Once the PI is ported to UR it might be nice if UR did something like this in the cuda/hip cases: https://github.com/intel/llvm/pull/8303/files#diff-7525901710934f7bdb2ad36238c4b67163f112d3bd233db7af0b0078b5b01e80R5920

The format for setPluginSpecificMessage is open for suggestions. We could make it a bit more verbose like this: https://github.com/intel/llvm/blob/484cf252246a958b089a8e94e35b14bd791a213c/sycl/plugins/cuda/pi_cuda.cpp#L181
But there is a 256 char limit to be aware of.

igchor:
@pbalcer  @vinser52 I think we might also need something similar for UMA (a way to return native error code). When using UMA memory provider on top of L0 we lose L0-specific error codes and L0 adapter might want to know exactly, why the allocation failed.

vinser52:
> @pbalcer @vinser52 I think we might also need something similar for UMA (a way to return native error code). When using UMA memory provider on top of L0 we lose L0-specific error codes and L0 adapter might want to know exactly, why the allocation failed.

As I remember each memory provider has a `get_last_error` interface. Isn't it what is needed?

igchor:
> > @pbalcer @vinser52 I think we might also need something similar for UMA (a way to return native error code). When using UMA memory provider on top of L0 we lose L0-specific error codes and L0 adapter might want to know exactly, why the allocation failed.
> 
> As I remember each memory provider has a `get_last_error` interface. Isn't it what is needed?

[...]

PR #631, opened 2023-06-16
[Exp][usm-p2p] Initial usm-p2p UR extension.

JackAKirk:
This PR adds new functions and enum to UR corresponding to the new PI functions introduced in this PR (which will be updated to take into account the UR migration): https://github.com/intel/llvm/pull/8303

FYI note that currently "P2P" is being translated to "P2_P". I was thinking of getting around this by switching to "usm-ptop", but this is not ideal since I don't think I am allows the case variation: "PtoP"?


include/ur.py
```diff
@@ -1902,6 +1902,9 @@ class ur_function_v(IntEnum):
     BINDLESS_IMAGES_WAIT_EXTERNAL_SEMAPHORE_EXP = 148   ## Enumerator for ::urBindlessImagesWaitExternalSemaphoreExp
     BINDLESS_IMAGES_SIGNAL_EXTERNAL_SEMAPHORE_EXP = 149 ## Enumerator for ::urBindlessImagesSignalExternalSemaphoreExp
     PLATFORM_GET_LAST_ERROR = 150                   ## Enumerator for ::urPlatformGetLastError
+    USM_P2_P_ENABLE_PEER_ACCESS_EXP = 151           ## Enumerator for ::urUsmP2PEnablePeerAccessExp
```
JackAKirk, line 1905: "P2P" is being translated to "P2_P" here and in various other places. I was thinking of getting around this by switching to "usm-ptop", but this is not ideal since I don't think I am allows the case variation: "PtoP"?

scripts/core/EXP-USM-P2P.rst
```diff
@@ -0,0 +1,70 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* ${x}UsmP2PEnablePeerAccessExp
+* ${x}UsmP2PDisablePeerAccessExp
+* ${x}UsmP2PPeerAccessGetInfoExp
+
+Changelog
+--------------------------------------------------------------------------------
+.. comment:
+    When making a change to an experimental feature, increment the version and 
+    provide a brief description of the change in the table below.
```
veselypeta, line 56: You should remove these comments - same below

scripts/core/exp-usm-p2p.yml
```diff
@@ -0,0 +1,113 @@
...
+desc: "Intel $OneApi Unified Runtime Experimental APIs for Usm P2P"
+ordinal: "99"
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Supported peer info"
+class: $xUsmP2P
+name: $x_exp_peer_info_t
+typed_etors: True
+etors:
+  - name: PI_PEER_ACCESS_SUPPORTED
```
veselypeta, line 21: If `PI` refers to Plugin Interface we should remove this reference.

scripts/core/exp-usm-p2p.yml
```diff
@@ -0,0 +1,113 @@
...
+#
+# Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+# See LICENSE.TXT
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Unified Runtime Experimental APIs for Usm P2P"
```
veselypeta, line 12:
```suggestion
desc: "Intel $OneApi Unified Runtime Experimental APIs for USM P2P"
```

scripts/core/exp-usm-p2p.yml
```diff
@@ -0,0 +1,113 @@
...
+type: function
+desc: "Enable access to peer device memory"
+class: $xUsmP2P
+name: EnablePeerAccessExp
+ordinal: "0"
+analogue:
+    - "**cuCtxEnablePeerAccess**"
+details:
+  - "Enables the command device to access and write device memory allocations located on the peer device, provided that a P2P link between the two devices is available."
+  - "When Peer Access is successfully enabled, P2P memory accesses are guaranteed to be allowed on the peer device until `DisablePeerAccessExp` is called."
```
veselypeta, line 35:
```suggestion
  - "When Peer Access is successfully enabled, P2P memory accesses are guaranteed to be allowed on the peer device until `${x}UsmP2PDisablePeerAccessExp` is called."
```

scripts/core/exp-usm-p2p.yml
```diff
@@ -0,0 +1,113 @@
...
+  - "Queries the peer access capabilities from the command device to the peer device according to the query `propName`."
+
+params:
+    - type: $x_device_handle_t
+      name: commandDevice
+      desc: "[in] handle of the command device object"
+    - type: $x_device_handle_t
+      name: peerDevice
+      desc: "[in] handle of the peer device object"
+
```
veselypeta, line 92:
```suggestion
```

scripts/core/exp-usm-p2p.yml
```diff
@@ -0,0 +1,113 @@
...
+type: function
+desc: "Disable access to peer device memory"
+class: $xUsmP2P
+name: PeerAccessGetInfoExp
+ordinal: "0"
+analogue:
+    - "**cuDeviceGetP2PAttribute**"
+details:
+  - "Queries the peer access capabilities from the command device to the peer device according to the query `propName`."
+
```
veselypeta, line 84:
```suggestion
```

scripts/core/exp-usm-p2p.yml
JackAKirk: Thanks, I've corrected the naming.

scripts/core/exp-usm-p2p.yml
```diff
@@ -0,0 +1,113 @@
...
+            If propSize is not equal to or greater than the real number of bytes needed to return the info
+            then the $X_RESULT_ERROR_INVALID_SIZE error is returned and pPropValue is not used.
+    - type: "size_t*"
+      name: pPropSizeRet
+      desc: |
+            [out][optional] pointer to the actual size in bytes of the queried propName.
+returns:
+    - $X_RESULT_SUCCESS
+    - $X_RESULT_ERROR_ADAPTER_SPECIFIC:
+      - "Returned if a native error code is returned by the adapter driver function querying the property. Consult the UR documentation for details on retrieving the corresponding native error code."
```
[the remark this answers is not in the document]
JackAKirk: Thanks, I've added these return types.
I think that it is important to have RESULT_ERROR_ADAPTER_SPECIFIC also for the general device/platform getinfo functions, so we can swap code like this that throws away backend specific errors by just calling an assert: https://github.com/intel/llvm/blob/1a6f7a51502bfe6bf177d7302cd1f08de4a8ec02/sycl/plugins/unified_runtime/ur/adapters/cuda/context.cpp#L70

See e.g. here to see the list of native errors that ` cuDeviceGetAttribute` can throw that are currently being thrown away by the cuda adapter: https://docs.nvidia.com/cuda/archive/10.2/cuda-driver-api/group__CUDA__DEVICE.html#group__CUDA__DEVICE_1g9c3e1414f0ad901d3278a4d6645fc266
Note also that any cu function can return errors from previous async cu driver calls. It is important to preserve all such native errors.
This is consistent with what the L0 plugin was doing in the past and is what we agreed to do generally here https://github.com/oneapi-src/unified-runtime/issues/500.
DPC++ already automatically retrieves the native backend error and passes it to a sycl::exception when `PI_ERROR_PLUGIN_SPECIFIC_ERROR` is returned from a PI call: we should check that it can do the same with `UR_RESULT_ERROR_ADAPTER_SPECIFIC` : actually I suspect that this is not the case and is something that we will need to fix.

include/ur.py
pbalcer: We've merged a fix. You can now rebase, and this issue will be gone.

scripts/core/exp-usm-p2p.yml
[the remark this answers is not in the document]
veselypeta: I've created #636 to track this 👍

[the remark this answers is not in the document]
JackAKirk: Yeah it was broke. I fixed it here: https://github.com/intel/llvm/pull/9975

JackAKirk:
@pbalcer Thanks, it is looking good now.

scripts/core/exp-usm-p2p.yml
kbenzie, line 1: There's inconsistent indentation, the project uses 4 space indents but there are 2 space indents in various places. Please make them consistently 4 spaces.

JackAKirk: OK thanks, will do.

kbenzie:
@JackAKirk looks like you need to commit files after running the `generate` target.

kbenzie (approved):

veselypeta (approved):
LGTM!

JackAKirk:
> Let me know when you're happy to merge and I can hit the button (I'm assuming that you don't have the permissions) +1

Cheers, I just have a little query then I'm happy for this to be merged.

Currently "ur_api.cpp" has these generated functions returning `UR_SUCCESS`. By default we want e.g.
```
  std::ignore = command_device;
  std::ignore = peer_device;

  setErrorMessage("piextEnablePeerAccess not "
                  "implemented in hip backend",
                  PI_ERROR_PLUGIN_SPECIFIC_ERROR);
  return PI_ERROR_PLUGIN_SPECIFIC_ERROR;
```

So what is the procedure to do this. I guess in the intel/llvm PR I should define the function in each adapter. But I just wanted to check that it isn't the case that there will be a default definition that some adapters might revert to?
I think this isn't the case but I just want to check I shouldn't e.g. update a correct default definition. Thanks

veselypeta:
> 

If your question is about asking UR if the extension is supported. I believe we currently do this through the device query `UR_DEVICE_INFO_EXTENSIONS` which will return a `char[]` of implemented extensions. However, you could also just the entry-point and if it's not been implemented - UR will return `UR_RESULT_ERROR_UNINITIALIZED`

JackAKirk:
> > 
> 
> If your question is about asking UR if the extension is supported. I believe we currently do this through the device query `UR_DEVICE_INFO_EXTENSIONS` which will return a `char[]` of implemented extensions. However, you could also just the entry-point and if it's not been implemented - UR will return `UR_RESULT_ERROR_UNINITIALIZED`.
> 
> This is something we should have documented so I will add this to the documentation.

The point is that previously we could have a portable implementation of e.g. this function:
https://github.com/intel/llvm/pull/8303/files#diff-e5a5ec910bab410f31617da9dfb56ee9e435769e7d6a830e85094c07b80c5454R216

even when a particular backend did not yet implement P2P in the PI/UR: In such case the Adapter should return an error stating it isn't yet implemented. Note that the query functions had a default where the P2P_infos simply return false if the Enable/Disable functions are not yet implemented for that adapter. This behaviour was requested by Greg in this PR https://github.com/intel/llvm/pull/8303.

veselypeta:
> In such case the Adapter should return an error stating it isn't yet implemented.

I think you can implement this in the pi2ur layer. If the adapter doesn't support it, then you can set the PI error as you say.

JackAKirk:
> > In such case the Adapter should return an error stating it isn't yet implemented.
> 
> I think you can implement this in the pi2ur layer. If the adapter doesn't support it, then you can set the PI error as you say.

OK cool, I'm just wondering what the long term solution is to such a situation in the long term when pi2ur is removed. But in any case I don't think anything is to be changed in this PR so happy for it to be merged now!
Thanks for the clarifications.

veselypeta:
I'm not sure long-term in terms of the SYCL-RT, but in UR I don't think we should be returning a `UR_ADAPTER_SPECIFIC` error if the adapter doesn't support the entry-point. I think you should have to query support before making the call.

JackAKirk:
> I'm not sure long-term in terms of the SYCL-RT, but in UR I don't think we should be returning a `UR_ADAPTER_SPECIFIC` error if the adapter doesn't support the entry-point. I think you should have to query support before making the call.

So the advantage of returning UR_ADAPTER_SPECIFIC is that the sycl runtime looks for the PI equivalent of this error automatically via plugin::check: if it is returned then it passes the last message to sycl::exception. This allows the relevant information (that the function isn't yet implemented for that adapter) to be passed to the runtime automatically without any special backend specific checks.
I suppose equivalent behaviour could be implemented differently, but currently that is how it does it. Ideally any replacement could be done in such a way that it doesn't require the runtime to do anything other than plugin::check.

veselypeta:
Maybe this is a topic for discussion at the WG call @kbenzie ?

JackAKirk:
> Maybe this is a topic for discussion at the WG call @kbenzie ?

The answer seems to be that atm people are using (see https://github.com/intel/llvm/pull/9992/files):

```
sycl::detail::ur::die("___ not implemented"
```
etc, which uses `std::cerr`. I think this is fine personally, although technically sycl spec/ extensions mandate functions should be throwing `sycl::exception`. I don't think it matters. I will just copy the above.

PR #637, opened 2023-06-20
[UR] Add CUDA-specific queue flags

callumfare:


omarahmed1111 (approved):

scripts/core/queue.yml
```diff
@@ -68,6 +68,12 @@ etors:
     - name: SUBMISSION_IMMEDIATE
       value: "$X_BIT(8)"
       desc: "Hint: enqueue and submit immediately. No change in queue semantics. Implementation chooses submission mode."
+    - name: CUDA_USE_DEFAULT_STREAM
+      desc: "Use the default stream. Only meaningful for CUDA. Other platforms may ignore this flag."
+      value: "$X_BIT(9)"
+    - name: CUDA_SYNC_WITH_DEFAULT
```
kbenzie, line 74: maybe drop the `CUDA` part? we didn't add them for the L0 specific flags directly above

kbenzie (approved):

scripts/core/queue.yml
callumfare: I thought about dropping it but I wasn't sure because these flags refer to streams which are a CUDA specific thing. In theory other adapters could implement the L0 batching flags.

callumfare: Actually I suppose the HIP adapter could theoretically implement these too, so renaming it makes sense

PR #640, opened 2023-06-20
[uma] Improve reporting of pool/provider specific errors

igchor:
Initially I wanted to implement the same `get_last_error` function for both memory pool and memory provider. However, I realized that the use cases are a bit different:
- for memory provider, we want to get the native code (e.g. to propagate from L0-provider to L0-adapter)
- for memory pool, we are mostly interested with UMA error code (why the allocation/de-allocation failed) - not necessarily with a native error code.

Becasue of that the interface for memory pool and provider is different.

Also, there is one problem with the current approach - return value from `umaPoolGetLastAllocationError` is meaningful only if the last allocation operation (malloc, aligned_malloc, etc.) failed (returned NULL). This means that it cannot be used to report errors from memory_pool::free() since that function does not return anything (and so, we don't know if it failed and whether it's safe to call `umaPoolGetLastAllocationError`).

If we want to make it work `free` we could make the return value well-defined in all cases (e.g. UMA_RESULT_SUCCESS if the previous operation succeeded). This would require any pool implementation to clear last error code on every operation. We could probably provide some callback in memory_pool_ops so that the pool-implementation does not need to do that manually and we'd just call it from memory_pool.c (in implementation of umaPoolMalloc, etc). @pbalcer @vinser52 do you think that would make sense?

### DIFF
--- a/source/common/uma_pools/disjoint_pool.cpp
+++ b/source/common/uma_pools/disjoint_pool.cpp
@@ -74,6 +74,10 @@ static size_t AlignUp(size_t Val, size_t Alignment) {
     return (Val + Alignment - 1) & (~(Alignment - 1));
 }
 
+struct MemoryProviderError {
+    uma_result_t code;
+};
+
 DisjointPoolConfig::DisjointPoolConfig()
     : limits(std::make_shared<DisjointPoolConfig::SharedLimits>()) {}
 
@@ -371,7 +375,13 @@ Slab::~Slab() {
     } catch (std::exception &e) {
         std::cout << "DisjointPool: unexpected error: " << e.what() << "\n";
     }
-    memoryProviderFree(bucket.getMemHandle(), MemPtr);
+
+    try {
+        memoryProviderFree(bucket.getMemHandle(), MemPtr);
+    } catch (MemoryProviderError &e) {
+        std::cout << "DisjointPool: error from memory provider: " << e.code
+                  << "\n";
+    }
 }
 
 // Return the index of the first available chunk, SIZE_MAX otherwise
@@ -717,7 +727,7 @@ void Bucket::printStats(bool &TitlePrinted, const std::string &Label) {
     }
 }
 
-void *DisjointPool::AllocImpl::allocate(size_t Size, bool &FromPool) {
+void *DisjointPool::AllocImpl::allocate(size_t Size, bool &FromPool) try {
     void *Ptr;
 
     if (Size == 0) {
@@ -742,10 +752,13 @@ void *DisjointPool::AllocImpl::allocate(size_t Size, bool &FromPool) {
     }
 
     return Ptr;
+} catch (MemoryProviderError &e) {
+    uma::getPoolLastStatusRef<DisjointPool>() = e.code;
+    return nullptr;
 }
 
 void *DisjointPool::AllocImpl::allocate(size_t Size, size_t Alignment,
-                                        bool &FromPool) {
+                                        bool &FromPool) try {
     void *Ptr;
 
     if (Size == 0) {
@@ -789,6 +802,9 @@ void *DisjointPool::AllocImpl::allocate(size_t Size, size_t Alignment,
     }
 
     return AlignPtrUp(Ptr, Alignment);
+} catch (MemoryProviderError &e) {
+    uma::getPoolLastStatusRef<DisjointPool>() = e.code;
+    return nullptr;
 }
 
 Bucket &DisjointPool::AllocImpl::findBucket(size_t Size) {
@@ -803,7 +819,7 @@ Bucket &DisjointPool::AllocImpl::findBucket(size_t Size) {
     return *(*It);
 }
 
-void DisjointPool::AllocImpl::deallocate(void *Ptr, bool &ToPool) {
+void DisjointPool::AllocImpl::deallocate(void *Ptr, bool &ToPool) try {
     auto *SlabPtr = AlignPtrDown(Ptr, SlabMinSize());
 
     // Lock the map on read
@@ -846,6 +862,8 @@ void DisjointPool::AllocImpl::deallocate(void *Ptr, bool &ToPool) {
     // to some slab with an entry in the map. So we find a slab
     // but the range checks fail.
     memoryProviderFree(getMemHandle(), Ptr);
+} catch (MemoryProviderError &e) {
+    uma::getPoolLastStatusRef<DisjointPool>() = e.code;
 }
 
 void DisjointPool::AllocImpl::printStats(bool &TitlePrinted,
@@ -938,7 +956,7 @@ void DisjointPool::free(void *ptr) {
 }
 
 enum uma_result_t DisjointPool::get_last_allocation_error() {
-    return UMA_RESULT_ERROR_UNKNOWN;
+    return uma::getPoolLastStatusRef<DisjointPool>();
 }
 
 DisjointPool::DisjointPool() {}
--- a/source/common/uma_pools/disjoint_pool.cpp
+++ b/source/common/uma_pools/disjoint_pool.cpp
@@ -332,8 +332,7 @@ static void *memoryProviderAlloc(uma_memory_provider_handle_t hProvider,
     void *ptr;
     auto ret = umaMemoryProviderAlloc(hProvider, size, alignment, &ptr);
     if (ret != UMA_RESULT_SUCCESS) {
-        // TODO: Set last error appropriately
-        throw std::runtime_error("umaMemoryProviderAlloc");
+        throw MemoryProviderError{ret};
     }
     return ptr;
 }
@@ -342,8 +341,7 @@ static void memoryProviderFree(uma_memory_provider_handle_t hProvider,
                                void *ptr) {
     auto ret = umaMemoryProviderFree(hProvider, ptr, 0);
     if (ret != UMA_RESULT_SUCCESS) {
-        // TODO: introduce custom exceptions? PI L0 relies on those
-        throw std::runtime_error("memoryProviderFree");
+        throw MemoryProviderError{ret};
     }
 }
 
@@ -939,9 +937,7 @@ void DisjointPool::free(void *ptr) {
     return;
 }
 
-enum uma_result_t DisjointPool::get_last_result(const char **ppMessage) {
-    // TODO: implement and return last error, we probably need something like
-    // https://github.com/oneapi-src/unified-runtime/issues/500 in UMA
+enum uma_result_t DisjointPool::get_last_allocation_error() {
     return UMA_RESULT_ERROR_UNKNOWN;
 }
 
--- a/source/common/unified_memory_allocation/include/uma/memory_pool.h
+++ b/source/common/unified_memory_allocation/include/uma/memory_pool.h
@@ -104,28 +104,23 @@ void umaPoolFree(uma_memory_pool_handle_t hPool, void *ptr);
 void umaFree(void *ptr);
 
 ///
-/// \brief Retrieve string representation of the underlying pool specific
-///        result reported by the last API that returned
-///        UMA_RESULT_ERROR_POOL_SPECIFIC or NULL ptr (in case of allocation
-///        functions). Allows for a pool independent way to
-///        return a pool specific result.
+/// \brief Retrieve uma_result_t representing the error of the last failed allocation
+///        operation in this thread (malloc, calloc, realloc, aligned_malloc).
 ///
 /// \details
-///     - The string returned via the ppMessage is a NULL terminated C style
-///       string.
-///     - The string returned via the ppMessage is thread local.
-///     - The memory in the string returned via the ppMessage is owned by the
-///       adapter.
-///     - The application may call this function from simultaneous threads.
-///     - The implementation of this function should be lock-free.
+/// * Implementations *must* store the error code in thread-local
+///   storage prior to returning NULL from the allocation functions.
+///
+/// * If the last allocation/de-allocation operation succeeded, the value returned by
+///   this function is unspecified.
+///
+/// * The application *may* call this function from simultaneous threads.
+///
+/// * The implementation of this function *should* be lock-free.
 /// \param hPool specified memory hPool
-/// \param ppMessage [out] pointer to a string containing provider specific
-///                  result in string representation.
-/// \return UMA_RESULT_SUCCESS if the result being
-///         reported is to be considered a warning. Any other result code
-///         returned indicates that the adapter specific result is an error.
-enum uma_result_t umaPoolGetLastResult(uma_memory_pool_handle_t hPool,
-                                       const char **ppMessage);
+/// \return Error code desciribng the failure of the last failed allocation operation.
+///         The value is undefined if the previous allocation was successful.
+enum uma_result_t umaPoolGetLastAllocationError(uma_memory_pool_handle_t hPool);
 
 ///
 /// \brief Retrieve memory pool associated with a given ptr.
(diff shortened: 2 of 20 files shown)


igchor:
As an example, in case there is a provider-specific error on allocation, the code would look like this:


```c++
void* ptr = umaPoolMalloc(hPool, size);
if (!ptr) {
    uma_result_t ret = umaPoolGetLastAllocationError(hPool);
    if (ret == UMA_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC) {
        const char *msg;
        int32_t native_error;
        umaMemoryProviderGetLastNativeError(getProvider(hPool), &msg, &native_error);
        // cast native_error to ze_result_t and handle it
    } else {
        ...
    }
}
```

pbalcer (approved):
Principle of least surprise. We should not be inventing new behaviors here IMHO. How free behaves with respect to errors is well defined - it doesn't directly return anything and preserves errno (or `umaPoolGetLastAllocationError` in this case). A user can take no useful action when free fails (assuming that anyone would check that error in the first place?).
Callbacks might make sense.

igchor:
> Principle of least surprise. We should not be inventing new behaviors here IMHO. How free behaves with respect to errors is well defined - it doesn't directly return anything and preserves errno (or `umaPoolGetLastAllocationError` in this case). A user can take no useful action when free fails (assuming that anyone would check that error in the first place?). Callbacks might make sense.

I was also thinking about integrating this with UR. Currently urUSMFree() can return errors such as `DEVICE_LOST`. I don't think there is any other reasonable action that the user can take, other than just resetting the device (or the application). Still, it might be useful to get this error sooner rather than later (on free, rather than on the next allocation) so I think we should have some way of communicating this error.

Callbacks might be a good idea in a general case: I guess it should be even possible to implement a C++ API that would use e.g. exceptions on top of that.

kswiecicki (approved):

source/common/uma_pools/disjoint_pool.cpp
```diff
@@ -67,6 +67,10 @@ static void *AlignPtrUp(void *Ptr, const size_t Alignment) {
     return static_cast<char *>(AlignedPtr) + Alignment;
 }
 
+struct MemoryProviderError {
+    uma_result_t code;
```
DamianDuy, line 71: LGTM
nit: perhaps a better name for this variable would be retCode? I think it may be more descriptive.

pbalcer (approved):
Let's leave it as is for now. We can always return to the problem with `free` later.
Please rebase.

igchor:
> Let's leave it as is for now. We can always return to the problem with `free` later. Please rebase.

Done.

source/common/uma_pools/disjoint_pool.cpp
```diff
@@ -373,7 +375,13 @@ Slab::~Slab() {
     } catch (std::exception &e) {
         std::cout << "DisjointPool: unexpected error: " << e.what() << "\n";
     }
-    memoryProviderFree(bucket.getMemHandle(), MemPtr);
+
+    try {
+        memoryProviderFree(bucket.getMemHandle(), MemPtr);
+    } catch (MemoryProviderError &e) {
+        std::cout << "DisjointPool: error from memory provider: " << e.code
```
vinser52, line 382: is it the right way just to print to std::cout?

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -67,23 +67,25 @@ enum uma_result_t umaMemoryProviderFree(uma_memory_provider_handle_t hProvider,
...
-///     - The string returned via the ppMessage is thread local.
-///     - The memory in the string returned via the ppMessage is owned by the
-///       adapter.
-///     - The application may call this function from simultaneous threads.
-///     - The implementation of this function should be lock-free.
+/// * Implementations *must* store the message and error code in thread-local
+///   storage prior to returning UMA_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC.
 ///
+/// * The message and error code storage is will only be valid if a previously
+///   called entry-point returned UMA_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC.
```
vinser52, line 74: I think there is a misprint is the comment, e.g. `... storage is will only ...`.
The right statement should be something like this:
```
The message and error code storage is valid only if a previously
///   called entry-point returned UMA_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC.
```

source/common/unified_memory_allocation/include/uma/memory_provider.h
```diff
@@ -67,23 +67,25 @@ enum uma_result_t umaMemoryProviderFree(uma_memory_provider_handle_t hProvider,
...
 /// \param ppMessage [out] pointer to a string containing provider specific
 ///        result in string representation
-/// \return UMA_RESULT_SUCCESS if the result being reported is to be considered
-///         a warning. Any other result code returned indicates that the
-///         adapter specific result is an error.
-enum uma_result_t
-umaMemoryProviderGetLastResult(uma_memory_provider_handle_t hProvider,
-                               const char **ppMessage);
+/// \param pError [out] pointer to an integer where the adapter specific error code will be stored
+void umaMemoryProviderGetLastNativeError(uma_memory_provider_handle_t hProvider,
```
vinser52, line 86: what if some pool uses several memory providers, e.g. one for data and another one for metadata?
what will be the user flow?

source/common/unified_memory_allocation/include/uma/memory_pool.h
```diff
@@ -104,28 +104,23 @@ void umaPoolFree(uma_memory_pool_handle_t hPool, void *ptr);
...
-/// \param ppMessage [out] pointer to a string containing provider specific
-///                  result in string representation.
-/// \return UMA_RESULT_SUCCESS if the result being
-///         reported is to be considered a warning. Any other result code
-///         returned indicates that the adapter specific result is an error.
-enum uma_result_t umaPoolGetLastResult(uma_memory_pool_handle_t hPool,
-                                       const char **ppMessage);
+/// \return UMA_RESULT_SUCCESS if the previous function completed successfully, error code
+///         desciribng the failure otherwise.
+enum uma_result_t umaPoolGetLastAllocationError(uma_memory_pool_handle_t hPool);
```
vinser52, line 123: Will it imply additional overhead? It means thread-local variable should be updated on every allocation.

Also, I am confused by the name. Do we restrict only to `Allocation` errors? What about other methods of the heap manager, e.g. `malloc_usable_size` or anything we can add in the future?

vinser52:
> I was also thinking about integrating this with UR. Currently urUSMFree() can return errors such as DEVICE_LOST

I think it is a memory provider-specific error. You will not notice this error until coarse grain allocation/deallocation.

source/common/uma_pools/disjoint_pool.cpp
igchor: We need to integrate UR logger into the disjoint pool (or implement a separate logger for UMF if we'll want to move it to separate repo) first. This is on the TODO list - there are multiple 'cout' calls int the DisjointPool implementation currently.

source/common/unified_memory_allocation/include/uma/memory_provider.h
igchor: That's a good point. User might need to know which (metadata or data) provider failed. I think we need to extend error codes to specify DATA_PROVIDER_ERROR and METADATA_PROVIDER_ERROR.

source/common/unified_memory_allocation/include/uma/memory_pool.h
igchor: Not on every allocation, just on allocation error - I just updated the comment, which incorrectly stated that the function can return OK if the previous allocation succeeded. The idea is that this function should only be called after a failed allocation function.

Yes, the reason for restricting it only to allocation errors is what I described in the PR description - for functions like free or malloc_usable_size we don't know if they failed so we cannot safely use this function. For other functions that we might add in the future, I assume they would be UMF-specific (not like allocation/free functions where we want to keep similar signature to libs functions) so we can just return UMF_STATUS

source/common/unified_memory_allocation/include/uma/memory_provider.h
vinser52: I do not like the idea of introducing provider type-specific errors, especially if we consider changing the way how we specify provider type in the pool.

We can think of storing the handle to the provider that failed and remove `hProvider` argument from the `umaMemoryProviderGetLastNativeError` function.

In general, I would like to see some code-snippet that demonstrates how these API suppose to be used by client (e.g. UR).

igchor:
> > I was also thinking about integrating this with UR. Currently urUSMFree() can return errors such as DEVICE_LOST
> 
> I think it is a memory provider-specific error. You will not notice this error until coarse grain allocation/deallocation.

Yes, of course - but you probably need to know that the deallocation failed (due to whatever reason) so that you know that you need to query memory provider for the actual reason.

source/common/unified_memory_allocation/include/uma/memory_provider.h
igchor: Done.

source/common/unified_memory_allocation/include/uma/memory_provider.h
igchor: I've extended the error codes in https://github.com/oneapi-src/unified-runtime/pull/641

igchor: If we assume that `umaMemoryProviderGetLastNativeError` can only be called by the memory pool user, then we could probably then we could probably replace the `hProvider` handle with a `pool` handle. I don't think we can remove the handle entirely because we need to know which function to call (from memory_provider_ops that is associated with that handle). 

Still I think `umaMemoryProviderGetLastNativeError` might also be useful **inside** a memory pool implementation (e.g. when a pool always uses some specific metadata provider).

The code would look similar to what I proposed in the comment above (just with extra switch for metadata/data provider):

void* ptr = umaPoolMalloc(hPool, size);
if (!ptr) {
    uma_result_t ret = umaPoolGetLastAllocationError(hPool);
    uma_memory_provider_handle_t provider;
    if (ret == UMA_RESULT_ERROR_DATA_MEMORY_PROVIDER_SPECIFIC) {
        provider = umaPoolGetDataMemoryProvider(hPool):
    } else if (ret == UMA_RESULT_ERROR_METADATA_MEMORY_PROVIDER_SPECIFIC {
        provider = umaPoolGetMetadataMemoryProvider(hPool):
    } else {
        ....
    }
    const char *msg;
    int32_t native_error;
    umaMemoryProviderGetLastNativeError(provider, &msg, &native_error);
    // cast native_error to ze_result_t and handle it
}

vinser52: > Still I think umaMemoryProviderGetLastNativeError might also be useful inside a memory pool implementation (e.g. when a pool always uses some specific metadata provider)

I agree with that.

But I would suggest to simplify the code snippet you provided above to the following:

```
void* ptr = umaPoolMalloc(hPool, size);
if (!ptr) {
    uma_result_t ret = umaPoolGetLastAllocationError(hPool);
    uma_memory_provider_handle_t provider;
    if (ret == UMA_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC) {
        const char *msg;
        int32_t native_error;
        // Why the pool internally cannot remember the failed memory provider?
        umaMemoryProviderGetLastNativeError(hPool, &msg, &native_error);
        // cast native_error to ze_result_t and handle it
    }
}
```

Why the pool internally cannot remember the failed memory provider? Such an approach does not require to introduce `UMA_RESULT_ERROR_DATA_MEMORY_PROVIDER_SPECIFIC` and `UMA_RESULT_ERROR_METADATA_MEMORY_PROVIDER_SPECIFIC` error codes.

Also, another question came to my mind. What about external heap manager implementations (like jemalloc), does such approach requires any changes in the heap mgr implementation itself?

igchor: > Why the pool internally cannot remember the failed memory provider?

Well, it can remember it, but where would you pass that provider handle if `umaMemoryProviderGetLastNativeError` no longer accepts it? I wouldn't want to introduce some provider interface specifically for use inside the memory pool implementation. 

> Also, another question came to my mind. What about external heap manager implementations (like jemalloc), does such approach requires any changes in the heap mgr implementation itself?

For jemalloc we would just record the appropriate error in callbacks (.e.g `hooks.alloc`) - that's the place where we would use a memory provider. I don't think jemalloc would use metadata memory provider but that's fine.

igchor: Also, I think it is useful to know which provider failed: whether it's metadata or data provider. If you just call `umaMemoryProviderGetLastNativeError(hPool, &msg, &native_error);` as you suggest you don't know which provider failed and you don't know how to cast the native_error (since those 2 providers can have different native error codes).

vinser52: I got your point. But the only thing I want to fix is to get rid of these two error codes (`UMA_RESULT_ERROR_DATA_MEMORY_PROVIDER_SPECIFIC` and `UMA_RESULT_ERROR_METADATA_MEMORY_PROVIDER_SPECIFIC`) and have only one `UMA_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC`.

vinser52: The reason I want to avoid it is related to [this comment](https://github.com/oneapi-src/unified-runtime/pull/641/files#r1244132297). The proposed pool creation API is very specific to current USM needs.
Imagine we will change the pool creation API in the future then ERROR handling code by the UMF clients has to be changed as well.

igchor: But then how would you know which provider failed? (i.e. which provider to pass to the `umaMemoryProviderGetLastNativeError`)?

igchor: If we decide not to merge https://github.com/oneapi-src/unified-runtime/pull/641 we need some other way of specyfing which memory provider is the one that failed. The simplest approach would be just to have only one provider per pool (as we had originally).

igchor: Anyway, this PR does not introduce UMA_RESULT_ERROR_DATA_MEMORY_PROVIDER_SPECIFIC nor UMA_RESULT_ERROR_METADATA_MEMORY_PROVIDER_SPECIFIC so I think this discussion can continue on this thread: https://github.com/oneapi-src/unified-runtime/pull/641/files#r1244132297

Can this PR be merged or do you have any other comments?

vinser52: In general, to interpret the native error code you should know the concrete implementation referred by the memory provider handle.

> Also, I think it is useful to know which provider failed: whether it's metadata or data provider.

And it does not matter whether it DATA or METADATA memory provider. You just need to know which memory provider failed. For example, imagine the following situation:

1. You have two memory providers (different implementations - different ops structure) and pass them to `umaPoolCreate` function. 
2. `umaPoolMalloc` call failed because one of the memory providers failed. Does it really matter whether failed memory provider was used as DATA or METADATA? One obvious solution could be just to have `umaPoolGetLastFailedProvider`  function to get failed provider. It will help to avoid `if-else if-...-else` block.

But an even better idea is to look how similar problem solved in STL: [`std::error_code`](https://en.cppreference.com/w/cpp/error/error_code)
Also, it would be useful to understand how clients are going to use this native error? Do we have a code snippet from UR perspective?

igchor: Done. I've implemented `umaGetLastFailedProvider` as a standalone function that does not accept a `pool` handle so that it can be implemented in memory_provider.c (as opposed to in each memory pool separately).

vinser52 (approved):

pbalcer (approved):

PR #641, opened 2023-06-20
[uma] Replace list of providers with explicit data and metadata providers.

igchor:
Existing API exposes list of memory provider but does not specify type/purpose of those providers. This makes it challenging to connect a memory allocation from memory pool with a memory provider (in case, someone wants to query properties of that memory provider or needs to get a provider-specific error message, as here: https://github.com/oneapi-src/unified-runtime/pull/640).

There was a discussion on https://github.com/oneapi-src/unified-runtime/pull/315 about passing some extra information along providers to initialize() (e.g. type) but I don't think we have a use case for that right now - passing data and metadata providers seems simpler. In future, if we'll need to pass more then 1 metadata and 1 data provider (e.g. to let pool choose) we might revert the change for initialize(). However, for querying - I think we should always return only 1 data and 1 metadata provider (those that are actually used).

### DIFF
--- a/source/common/unified_memory_allocation/include/uma/memory_pool.h
+++ b/source/common/unified_memory_allocation/include/uma/memory_pool.h
@@ -25,17 +25,16 @@ struct uma_memory_pool_ops_t;
 ///
 /// \brief Creates new memory pool.
 /// \param ops instance of uma_memory_pool_ops_t
-/// \param providers array of memory providers that will be used for coarse-grain allocations.
-///        Should contain at least one memory provider.
-/// \param numProvider number of elements in the providers array
+/// \param data_provider memory provider that should be used for coarse-grain data allocation
+/// \param metadata_provider [optional] memory provider that should be used for metadata allocations
 /// \param params pointer to pool-specific parameters
 /// \param hPool [out] handle to the newly created memory pool
 /// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure.
 ///
 enum uma_result_t umaPoolCreate(struct uma_memory_pool_ops_t *ops,
-                                uma_memory_provider_handle_t *providers,
-                                size_t numProviders, void *params,
-                                uma_memory_pool_handle_t *hPool);
+                                uma_memory_provider_handle_t data_provider,
+                                uma_memory_provider_handle_t metadata_provider,
+                                void *params, uma_memory_pool_handle_t *hPool);
 
 ///
 /// \brief Destroys memory pool.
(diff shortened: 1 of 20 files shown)


pbalcer (approved):
In the future, we should avoid the temptation of architecting things we don't actually need yet... ;)

source/common/unified_memory_allocation/include/uma/memory_pool.h
```diff
@@ -25,17 +25,16 @@ struct uma_memory_pool_ops_t;
...
 /// \param params pointer to pool-specific parameters
 /// \param hPool [out] handle to the newly created memory pool
 /// \return UMA_RESULT_SUCCESS on success or appropriate error code on failure.
 ///
 enum uma_result_t umaPoolCreate(struct uma_memory_pool_ops_t *ops,
-                                uma_memory_provider_handle_t *providers,
-                                size_t numProviders, void *params,
-                                uma_memory_pool_handle_t *hPool);
+                                uma_memory_provider_handle_t data_provider,
+                                uma_memory_provider_handle_t metadata_provider,
```
vinser52, line 36: it is a non-flexible API. As I remember in the past you had an idea of passing a list of memory providers with the lists of types. Have you considered it as an alternative.

source/common/unified_memory_allocation/include/uma/memory_provider_ops.h
```diff
@@ -49,7 +49,7 @@ struct uma_memory_provider_ops_t {
                                            size_t *pageSize);
     enum uma_result_t (*purge_lazy)(void *provider, void *ptr, size_t size);
     enum uma_result_t (*purge_force)(void *provider, void *ptr, size_t size);
-    void (*get_name)(void *provider, const char **ppName);
+    const char *(*get_name)(void *provider);
```
vinser52, line 52: nit: looks like a change unrelated to this PR.

source/common/unified_memory_allocation/include/uma/memory_pool.h
igchor: Yes, as I mentioned in the PR description we might want to accept a list of provides with types. However, for that we would need to define some properties for memory providers (so that the pool can choose the appropriate one) and we don't have those right now. I feel like it's better to start with a simpler API and evolve it over time if needed.

source/common/unified_memory_allocation/include/uma/memory_provider_ops.h
igchor: yes, that's why I've put it into a separate commit, it's just a small improvement

source/common/unified_memory_allocation/include/uma/memory_pool.h
vinser52: My current feeling (and your comment confirms that) is that it is not the final API and furthermore it is specific to `disjoint_pool` implementation.
So why we cannot keep with the current approach when we pass just the list of memory providers and rely on the order of providers in that list?

test/unified_memory_allocation/memoryProviderAPI.cpp
```diff
@@ -77,7 +76,8 @@ INSTANTIATE_TEST_SUITE_P(
     providerInitializeTest, providerInitializeTest,
     ::testing::Values(UMA_RESULT_ERROR_OUT_OF_HOST_MEMORY,
                       UMA_RESULT_ERROR_POOL_SPECIFIC,
-                      UMA_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC,
+                      UMA_RESULT_ERROR_METADATA_MEMORY_PROVIDER_SPECIFIC,
+                      UMA_RESULT_ERROR_DATA_MEMORY_PROVIDER_SPECIFIC,
```
vinser52, line 80: I think we need a test with memory provider which always returns the error. Should it be part of this PR or we are adding tests in separate PRs?

igchor: You mean `memory pool` that returns this error? Those two errors are supposed to be returned by `umaPoolGetLastAllocationError` (from my second PR) to indicate that the error is provider-specific.

There is some overlap between those 2 PR so I guess once one of the is merged, I will rebase (and I can add those tests) to the second one.

source/common/unified_memory_allocation/include/uma/memory_pool.h
igchor: We **don't** strictly need the interface to pass multiple memory providers to the memory pool right now. Even for metadata provider, it was mostly a 'future-looking' feature - the idea was that we could use it in DisjointPool (or any other) pool implementation, but we don't really use it right now (we just use regular new/delete).

As Piotr said, we should probably avoid implementing things we don't need to, perhaps for now, just having a single memory provider is enough - even without the metadata one (since we don;t have any implementation that would use it.)

I can revert the whole change that introduced the list of memory providers and just keep a single provider as it was originally. I would actually prefer that over defining which provider should be on position 0 or 1 in the list, etc. as that can change over time and we might silently break some (probably PoC) implementation silently - with function signature change, there would at least be a compilation error.

igchor: So, I think it's just a question: do we think that having separate data and metadata memory provider is useful (and generic enough) or we'd rather wait for an actual implementation that will want to leverage that distinction.

igchor:
Also, regarding a list of memory providers - I've been thinking and perhaps a better interface might be just to expose a separate function like `provider get_preferred_provider(providers...)` in a memory pool API instead of passing the list to the `initialize` method (in case we'll need to select a provider based on some properties in future). This would simplify the most common case (just passing a single provider) and allow a memory pool to choose the best one based on some properties. But this is orthogonal to the data vs metadata provider discussion, I belive.

PR #646, opened 2023-06-23
[UR] Add documentation to experimental feature support

### LINKED ISSUE #619: Document how experimental support is reported from adapters
We should decide and document on how adapters will report which experimental features they support. Some options are listed below:

* use the existing platform/device extension string queries, define a naming convention that won't conflict with OpenCL or PI extension naming
* use a new enum defining the current set of experimental features (may be empty) and a new entry point to query for experimental feature support, this is more work but may result in a better experience for developers

veselypeta:
This PR:
* Clarifies `UR_RESULT_ERROR_UNINITIALIZED` to include unimplemented entry-points
* Adds documentation to each experimental feature about support.

Closes #619

### DIFF
--- a/scripts/templates/exp_feat.rst.mako
+++ b/scripts/templates/exp_feat.rst.mako
@@ -70,6 +70,16 @@ Changelog
 | 1.0       | Initial Draft           |
 +-----------+------------------------+
 
+Support
+--------------------------------------------------------------------------------
+
+Adapters which support this experimental feature *must* return the valid string 
+defined in ``${"${X}"}_${"_".join(name.split("-")).upper()}_EXTENSION_STRING_EXP`` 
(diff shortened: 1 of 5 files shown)


scripts/templates/exp_feat.rst.mako
```diff
@@ -70,6 +70,16 @@ Changelog
 | 1.0       | Initial Draft           |
 +-----------+------------------------+
 
+Support
+--------------------------------------------------------------------------------
+
+Adapters which support this experimental feature *should* return the valid
```
kbenzie, line 76: I think this should be *must* instead of *should*.

kbenzie (approved):

omarahmed1111 (approved):

pbalcer (approved):

PR #648, opened 2023-06-23
[UR] Automatically update ur_structure_type_t

### LINKED ISSUE #563: [ur] Automatically generate `ur_structure_type_t`
Following on from discussion about #546 and how to make it easier for external parties to make contributions it was noted that it should be possible to completed automate making changes to the `ur_structure_type_t` enumeration in a similar way that the `ur_function_t` enumeration is automatically generated for XPTI integration.

veselypeta:
Closes #563


scripts/core/structure.yml
kbenzie, line 1: Why can't this be new section inside `registry.yml` rather than a new file?

veselypeta: The difficulty is that `registry.yml` is parsed last to ensure that all declared functions have already been parsed. `ur_strcture_type_t` is referenced directly and thus must be parsed first. The simplest thing was to make it a separate file and parse it first - if there are any new additions that result from parsing the remaining files we then update the file.

kbenzie: My concern with this is people already get confused when there are conflicts in `registry.yml` and adding another file that could potentially have conflicts is not a nice workflow.

Could we treat `registry.yml` specially instead to ensure its read when we need it?

veselypeta:
Closing in favour of #670

PR #649, opened 2023-06-23
[UR][CTS] Add urMemImageGetInfo tests and missing spec return values

### LINKED ISSUE #231: Adds CTS tests for `urMemImageGetInfo` entry-point
We should add conformance tests for `urMemImageGetInfo`.

Required Coverage:

callumfare:
Resolves #231

### DIFF
--- a/test/conformance/memory/urMemImageGetInfo.cpp
+++ b/test/conformance/memory/urMemImageGetInfo.cpp
@@ -2,3 +2,85 @@
 // Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
 // See LICENSE.TXT
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#include <uur/fixtures.h>
+
+using urMemImageGetInfoTest = uur::urMemImageTestWithParam<ur_image_info_t>;
+
+static std::unordered_map<ur_image_info_t, size_t> image_info_size_map = {
+    {UR_IMAGE_INFO_FORMAT, sizeof(ur_image_format_t)},
+    {UR_IMAGE_INFO_ELEMENT_SIZE, sizeof(size_t)},
+    {UR_IMAGE_INFO_ROW_PITCH, sizeof(size_t)},
+    {UR_IMAGE_INFO_SLICE_PITCH, sizeof(size_t)},
+    {UR_IMAGE_INFO_WIDTH, sizeof(size_t)},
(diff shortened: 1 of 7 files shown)


test/conformance/memory/urMemImageGetInfo.cpp
```diff
@@ -2,3 +2,83 @@
...
+TEST_P(urMemImageGetInfoTest, Success) {
+    ur_image_info_t info = getParam();
+    size_t size;
+    ASSERT_SUCCESS(urMemImageGetInfo(image, info, 0, nullptr, &size));
+    ASSERT_NE(size, 0);
+
+    if (const auto expected_size = image_info_size_map.find(info);
+        expected_size != image_info_size_map.end()) {
+        ASSERT_EQ(expected_size->second, size);
+    }
```
veselypeta, line 37:
```cpp
else {
    FAIL() << "<some error message>";
}
```

veselypeta (approved):
LGTM - just one suggestion

test/conformance/memory/urMemImageGetInfo.cpp
```diff
@@ -2,3 +2,83 @@
...
+    size_t info_size = 0;
+    ASSERT_EQ_RESULT(
+        urMemImageGetInfo(image, UR_IMAGE_INFO_FORMAT, 0, &info_size, nullptr),
+        UR_RESULT_ERROR_INVALID_SIZE);
+}
+
+TEST_P(urMemImageGetInfoTest, InvalidSizeSmall) {
+    size_t info_size = 0;
+    ASSERT_EQ_RESULT(urMemImageGetInfo(image, UR_IMAGE_INFO_FORMAT,
+                                       sizeof(info_size) - 1, &info_size,
```
aarongreig, line 68: `UR_IMAGE_INFO_FORMAT` is an int so this wont always be too small to hold it

aarongreig (approved):

test/conformance/memory/urMemImageGetInfo.cpp
callumfare: I thought `UR_IMAGE_INFO_FORMAT` was a struct with two enums (so presumably 64-bit total), but either way I've changed it to be int as there's no harm in making it smaller

aarongreig: you're right, got confused reading the header and thought it was another enum

PR #654, opened 2023-06-27
[UR] Install DPCXX

### LINKED ISSUE #650: Enqueue, program and kernel CTS tests aren't built by default
Since https://github.com/oneapi-src/unified-runtime/commit/dfc8b08ef479c59c669dcfcf219a2cac99112f3c these test suites aren't built by default as they require a DPC++ build to handle device code for some tests. This is specified by the `UR_DPCXX` CMake option.

We're not building with this option in CI which means we missed a regression in the urKernelSetArg tests after this recent commit - https://github.com/oneapi-src/unified-runtime/commit/6da20017f64f48813b98b68068b4f730420ec939

There's also no warning if the CMake variable isn't set, or any mention of it the documentation, so it's easy to not build these tests without realizing.

veselypeta:
Closes #650

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -55,6 +55,12 @@ jobs:
         make
         sudo make install        
         cd ..
+    
+    - name: Download DPC++
+      run: |
+        sudo apt install libncurses5
+        wget -O ${{github.workspace}}/dpcpp_compiler.tar.gz https://github.com/intel/llvm/releases/download/sycl-nightly%2F20230626/dpcpp-compiler.tar.gz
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -55,6 +55,12 @@ jobs:
         make
         sudo make install        
         cd ..
+    
+    - name: Download DPC++
+      run: |
+        sudo apt install libncurses5
```
aarongreig, line 61: what's introducing this dependency? out of curiosity

aarongreig (approved):

pbalcer (approved):

PR #655, opened 2023-06-27
[Exp][Bindless Images] Add BI revision 4 functionality

przemektmalon:
- Added bindless images revision 4 API changes
- Fixed some errors from revision 3

### DIFF
--- a/include/ur.py
+++ b/include/ur.py
@@ -590,31 +590,29 @@ class ur_device_info_v(IntEnum):
                                                     ## bindless images backed by USM
     BINDLESS_IMAGES_2D_USM_SUPPORT_EXP = 0x2002     ## [::ur_bool_t] returns true if the device supports the creation of 2D
                                                     ## bindless images backed by USM
-    BINDLESS_IMAGES_3D_USM_SUPPORT_EXP = 0x2003     ## [::ur_bool_t] returns true if the device supports the creation of 3D
-                                                    ## bindless images backed by USM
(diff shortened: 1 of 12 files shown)


veselypeta (approved):

callumfare (approved):

include/ur.py
```diff
@@ -590,31 +590,29 @@ class ur_device_info_v(IntEnum):
                                                     ## bindless images backed by USM
     BINDLESS_IMAGES_2D_USM_SUPPORT_EXP = 0x2002     ## [::ur_bool_t] returns true if the device supports the creation of 2D
                                                     ## bindless images backed by USM
-    BINDLESS_IMAGES_3D_USM_SUPPORT_EXP = 0x2003     ## [::ur_bool_t] returns true if the device supports the creation of 3D
-                                                    ## bindless images backed by USM
-    IMAGE_PITCH_ALIGN_EXP = 0x2004                  ## [uint32_t] returns the required alignment of the pitch between two
+    IMAGE_PITCH_ALIGN_EXP = 0x2003                  ## [uint32_t] returns the required alignment of the pitch between two
```
alycm, line 593: Are we okay with these numbers changing because the feature is experimental?

In general can we catch ABI changes like this automatically?

alycm: Obviously this PR is fully of ABI changes, with all the function signature changes.  So the enum change is no different to those.  Just in general it's easier to miss.

veselypeta: Experimental features are free to break their own ABI/API - we have some checks to ensure that the core API is unaffected.

PR #659, opened 2023-06-29
[UR][CTS] Add urEnqueueMemImageRead/Write tests

callumfare:


test/conformance/enqueue/urEnqueueMemImageWrite.cpp
```diff
@@ -2,3 +2,143 @@
 // Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
 // See LICENSE.TXT
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#include <uur/fixtures.h>
+
+using urEnqueueMemImageWriteTest = uur::urMemImageQueueTest;
+UUR_INSTANTIATE_DEVICE_TEST_SUITE_P(urEnqueueMemImageWriteTest);
+
+TEST_P(urEnqueueMemImageWriteTest, Success1D) {
+    std::vector<uint32_t> output(width * 4, 42);
```
veselypeta, line 11: Is this `*4` because of the four channels?

veselypeta (approved):

test/conformance/enqueue/urEnqueueMemImageWrite.cpp
callumfare: It is, I'll add a comment to make that clearer

test/conformance/testing/include/uur/fixtures.h
```diff
@@ -498,6 +498,77 @@ struct urMemBufferQueueTest : urQueueTest {
...
+                              1,                            // image depth
+                              1,                            // array size
+                              0,                            // row pitch
+                              0,                            // slice pitch
+                              0,                            // mip levels
+                              0};                           // num samples
+
+    ur_image_desc_t desc3D = {UR_STRUCTURE_TYPE_IMAGE_DESC, // stype
+                              nullptr,                      // pNext
+                              UR_MEM_TYPE_IMAGE2D,          // mem object type
```
aarongreig, line 561: should this be `IMAGE3D`?

callumfare: It should, good catch! Fixed now, presumably the depth was just ignored before.

test/conformance/enqueue/urEnqueueMemImageWrite.cpp
```diff
@@ -2,3 +2,143 @@
...
+    ur_rect_offset_t origin{0, 0, 0};
+    ASSERT_SUCCESS(urEnqueueMemImageWrite(queue, image1D, true, origin, region,
+                                          0, 0, output.data(), 0, nullptr,
+                                          nullptr));
+}
+
+TEST_P(urEnqueueMemImageWriteTest, Success2D) {
+    std::vector<uint32_t> output(width * height * 4, 42);
+    ur_rect_region_t region{width, height, 1};
+    ur_rect_offset_t origin{0, 0, 0};
```
aarongreig, line 22: could `output`, `region` and `origin` live in the fixture?

aarongreig (approved):

test/conformance/enqueue/urEnqueueMemImageWrite.cpp
callumfare: It's possible, but `output` should actually `input` for these write tests (I'll rename it), so to avoid confusion they probably shouldn't share the same variable in the fixture. Region and origin change between tests so I'm not sure that's doable nicely

aarongreig: ah they are different! my eyes glazed over after a reading them a few times in the negative tests below - maybe they could live in a local specialization of the fixtures and get given the 
```
std::vector<uint32_t> input(width * 4, 42);
ur_rect_region_t region{width, 1, 1};
ur_rect_offset_t origin{0, 0, 0};
```
common set of inputs? that would cut down on some repetition

PR #667, opened 2023-07-03
[UR] Remove clang-format requirement for make generate-code

lukaszstolarczuk:
'make generate-code' should not require clang-format to be installed. It required the removal of a doubled functionality from run.py (it is covered via 'make cppformat' target).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,14 +238,14 @@ install(
 
 set(API_JSON_FILE ${PROJECT_BINARY_DIR}/unified_runtime.json)
 
-if(UR_FORMAT_CPP_STYLE)
-    # Generate source from the specification
-    add_custom_target(generate-code USES_TERMINAL
-        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/scripts
-        COMMAND ${Python3_EXECUTABLE} run.py --api-json ${API_JSON_FILE} --clang-format=${CLANG_FORMAT}
-        COMMAND ${Python3_EXECUTABLE} json2src.py --api-json ${API_JSON_FILE} ${PROJECT_SOURCE_DIR}
-    )
+# Generate source from the specification
+add_custom_target(generate-code USES_TERMINAL
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/scripts
+    COMMAND ${Python3_EXECUTABLE} run.py --api-json ${API_JSON_FILE}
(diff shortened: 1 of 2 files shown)


pbalcer:
This is essentially a revert of #616. Maybe we should change the formatting in `run.py` to be optional so that it can be skipped if `clang-format` is not available.

kbenzie:
> This is essentially a revert of #616.

Indeed, which was introduced to fix spec rendering issues as a result of Doxygen incorrectly parsing comment blocks.

> Maybe we should change the formatting in `run.py` to be optional so that it can be skipped if `clang-format` is not available.

Is it really so burdensom to have `clang-format` available?

veselypeta:
I agree - we can't correctly generate the spec without using clang-format. Perhaps we could build in the full clang-format step into the the script instead of running it as a separate command?

pbalcer:
I'm OK with keeping things as they are right now. But we've seen people struggle with this one thing time and time again. Another suggestion: we will probably need to containerize our CI system (I've asked @lukaszstolarczuk to look into that). We could reuse the same images in an instruction how to contribute to the spec using a prebuilt system as an alternative for developers that don't want to create their own environment.

kbenzie:
> Another suggestion: we will probably need to containerize our CI system (I've asked @lukaszstolarczuk to look into that). We could reuse the same images in an instruction how to contribute to the spec using a prebuilt system as an alternative for developers that don't want to create their own environment.

I agree with this direction.

lukaszstolarczuk:
> This is essentially a revert of #616. 

Ups, didn't know about this one 😄 

My idea for this change came from Readme, where we have a strict distinction between 'generate' and 'generate-code' commands... Since both of them requires clang-format anyway, perhaps we should just remove the second command from Readme...?

> > Another suggestion: we will probably need to containerize our CI system (I've asked @lukaszstolarczuk to look into that). We could reuse the same images in an instruction how to contribute to the spec using a prebuilt system as an alternative for developers that don't want to create their own environment.
> 
> I agree with this direction.

That'll be my main focus, sure thing.

kbenzie:
> My idea for this change came from Readme, where we have a strict distinction between 'generate' and 'generate-code' commands... Since both of them requires clang-format anyway, perhaps we should just remove the second command from Readme...?

Looks like the README was not updated when the fix was merged, that's unfortunate. I think it makes sense to just document one command, yes.

PR #670, opened 2023-07-03
[UR] Automatically generate ur_structure_type_t

veselypeta:
This PR automatically generates $x_structure_type_t enumerations.


kbenzie (approved):

scripts/parse_specs.py
```diff
@@ -914,6 +969,9 @@ def parse(section, version, tags, meta, ref):
             })
 
     specs = sorted(specs, key=lambda s: s['header']['ordinal'])
+    _inline_extended_structs(specs, meta)
+    registry = [f for f in files if f.endswith('registry.yml')][0]
+    _generate_structure_type_t(specs, meta, registry)
```
pbalcer, line 974: I'm not sure if I understand why this can't simply be a part of the `generate_registry` function. Right now it's called almost immediately after.

kbenzie: Perhaps an artifact of a previous iteration of this change when it was generating a separate file?

veselypeta: To generate the structure type we need to inline the extended structs, which happens after `generate_registry`, but since registry does not depend on this we could run these steps together after `_inline_extended_structs`

pbalcer (approved):

Issue #284, opened 2023-02-20
Enable adapter specific testing


kbenzie:
As part of porting PI plugins to UR we also need to enable testing of adapter specific features. The most obvious of these is to test native handles to make sure integration and interop with the underlaying adapter interface is working as expected.

This is gated on adapter code moving into this repository.

* Determine where the adapter specific code should reside
* Document this in the readme
* Implement an example of what this should look like, either with the null adapter or when the first real adapter is moved into this repository.

kbenzie:
As discussed [here](https://github.com/oneapi-src/unified-runtime/pull/680#discussion_r1251895193) more extensive testing of `urPlatformGetLastError` is an ideal candidate for adapter specific testing as this relies on knowledge of adapter implementation details in order to actually trigger an adapter specific error so that the message and error code returned from `urPlatformGetLastError` can be validated.

veselypeta:
Completed as part of 
#790 & #763

PR #680, opened 2023-07-04
[UR][CTS] Add CTS tests for urPlatformGetLastError

### LINKED ISSUE #671: Add CTS tests for `urPlatformGetLastError`
We should add conformance tests for `urPlatformGetLastError`.

Required Coverage:

-  At least one successful (i.e. called with correct arguments and assert on expected result, not applicable to `CreateWithNativeHandle` entry points)
-  If this a query entry-point or one which accepts flags etc : we should tests every possible enumeration (where possible).
-  A test case for each returned error code (exclude disaster situations i.e. OUT_OF_MEMORY etc)

callumfare:
Closes #671


test/conformance/platform/urPlatformGetLastError.cpp
```diff
@@ -0,0 +1,33 @@
...
+struct urPlatformGetLastErrorTest : uur::platform::urPlatformTest {
+    int32_t error;
+    const char *message = nullptr;
+};
+
+TEST_F(urPlatformGetLastErrorTest, Success) {
+    // We can't reliably generate a UR_RESULT_ERROR_ADAPTER_SPECIFIC error to
+    // test the functionality of this entry point, so instead do a minimal
+    // smoke test and check that calling this doesn't just crash unexpectedly
+    urPlatformGetLastError(platform, &message, &error);
```
kbenzie, line 17: I don't think this test is valid because the spec states:

> * The message and error code storage is will only be valid if a previously called entry-point returned [UR_RESULT_ERROR_ADAPTER_SPECIFIC](https://oneapi-src.github.io/unified-runtime/core/api.html#ur__api_8h_1a70996548e8016a2761523966dd1a470ba0618033f629dd69b011f6e69b5d92fff).

Not sure we can actually validate this for adapter implementations without adding an extra entry point to artificually trigger an adapter specific error. I don't think that's something we should add to the spec.

We could potentially test that `urPlatformGetLastError` works in the loader with a mock adapter implementation but we don't have that machinery set up yet that I'm aware of.

Another option might be to have adapter specific testing for this where we can assume knowledge of an adapters impleentation details. We also don't have that machinery set up yet though.

callumfare: To me that line in the spec makes it sounds like the call to `urPlatformGetLastError` will succeed but the returned values will be undefined, in which case we could check that it still returns UR_RESULT_SUCCESS. But that's maybe not too intuitive for users.

If that's not possible/desirable then I think adapter-specific tests make the most sense, we could create an issue to come back to this once the framework for that is in place.

kbenzie: > To me that line in the spec makes it sounds like the call to `urPlatformGetLastError` will succeed but the returned values will be undefined, in which case we could check that it still returns UR_RESULT_SUCCESS. But that's maybe not too intuitive for users.

Okay, so your interpretation would mean the user *must not dereference the returned pointers*. Given the other valid usage statements are stipulating the of thread-local storage I think that's reasonable. Does suggest there is ambiguity in the valid usage statement though.

callumfare: One other thing I'm wondering is whether UR_RESULT_SUCCESS is always the desired result. The previous version of this entry point (urGetLastResult) said:

> The entry point will return UR_RESULT_SUCCESS if the result being reported is to be considered a warning. Any other result code returned indicates that the adapter specific result is an error.

Was this deliberately removed when it was last updated? If not then we can't check the result as its value is undefined, otherwise checking for success should be ok.

veselypeta (approved):

test/conformance/platform/urPlatformGetLastError.cpp
kbenzie: I'll make a note on #284 suggesting `urPlatformGetLastError` as an ideal candidate for adapter specific testing.

PR #681, opened 2023-07-04
Implement loader config object, allow programatic layer management.

aarongreig:


aarongreig:
this could do with some dedicated unit testing but I'm not really sure where, would feel out of place in the CTS since no adapter code will ever get called

source/loader/layers/tracing/ur_tracing_layer.cpp
```diff
@@ -28,13 +28,14 @@ context_t::context_t() {
     xptiFrameworkInitialize();
 
     call_stream_id = xptiRegisterStream(CALL_STREAM_NAME);
+    traceEnvEnabled = getenv_tobool("UR_ENABLE_TRACING");
```
aarongreig, line 31: This is a rock-and-a-hard-place type situation. There is already an XPTI env var that you need to set to enable the tracing used by this layer (actually there are several) - adding this new env var is our rock (detracts from user experience, seems generally redundant).
So the obvious thing would be to treat the XPTI one as our layer-enabling env var instead of adding the new one - but if we do that the API mechanism becomes redundant. Enabling the XPTI variable is always needed for the layer to work - you'd need to set it even if you wanted to enable your layers via api. This is our hard place, this solution makes the API redundant, and might lead to users making unnecessary API calls. At best it would be weird to document.
There is also a third, un-named, unpleasant location: we could just not report this tracing layer through the api mechanism and leave it as a kind of power-user option. This would preserve the current behaviour for existing users without being potentially misleading by reporting its existence through the loader config mechanism. That kind of sucks though because we have this functionality so we should encourage people to use it. It would also feel weird to have a layer that's fundamentally inconsistent with the layer mechanism we're trying to set up.

veselypeta:
> this could do with some dedicated unit testing but I'm not really sure where, would feel out of place in the CTS since no adapter code will ever get called

There are dedicated loader tests under `test/loader` which would seem reasonable?

pbalcer:
> > this could do with some dedicated unit testing but I'm not really sure where, would feel out of place in the CTS since no adapter code will ever get called
> 
> There are dedicated loader tests under `test/loader` which would seem reasonable?

Yup. I'd recommend just creating a new folder in that location for the config. These tests are less structured than the CTS ones because of the larger variety of test scenarios (e.g., launching the same process multiple times with different environment variables, with expecting predefined output), so just do whatever you think makes the most sense, either normal gtest scenarios or tests that rely on the `match` mechanism (like most of the existing layer tests).

source/loader/ur_adapter_registry.hpp
```diff
@@ -106,8 +106,9 @@ class AdapterRegistry {
     // to load the adapter.
     std::vector<std::vector<fs::path>> adaptersLoadPaths;
 
-    static constexpr std::array<const char *, 1> knownAdapterNames{
-        MAKE_LIBRARY_NAME("ur_adapter_level_zero", "0")};
+    static constexpr std::array<const char *, 2> knownAdapterNames{
+        MAKE_LIBRARY_NAME("ur_adapter_level_zero", "0"),
+        MAKE_LIBRARY_NAME("ur_adapter_cuda", "0")};
```
pbalcer, line 111: This should be a separate PR.

aarongreig:
cool, will add something in there

source/loader/layers/validation/ur_validation_layer.hpp
```diff
@@ -20,19 +20,22 @@ namespace ur_validation_layer {
...
     ur_dditable_t urDdiTable = {};
 
     context_t();
     ~context_t();
 
-    bool isEnabled() override { return enableValidation; };
-    ur_result_t init(ur_dditable_t *dditable) override;
+    bool isAvailable() const override { return true; }
+    std::vector<std::string> getNames() const override {
+        return {"UR_FULL_VALIDATION_LAYER", "UR_PARAMETER_VALIDATION_LAYER",
```
pbalcer, line 34: Maybe these names should be exactly the same as env variables? Or we could get rid of existing env variables and replace it with this:
`UR_ENABLE_LAYERS=UR_FULL_VALIDATION_LAYER,UR_TRACING_LAYER`.

I'd also invert the names: `UR_LAYER_VALIDATION_FULL`, `UR_LAYER_TRACING`.

source/loader/layers/tracing/ur_tracing_layer.cpp
pbalcer: I had the same dilemma. I imagine most users will leverage this layer indirectly through tools like `urtrace` or just `sycl`, so the ergonomics of this are not that important. It's already complicated and requires at least some understanding of the XPTI framework.
So I'm fine with adding the new env variable, provided it's well documented.

source/loader/layers/validation/ur_validation_layer.hpp
aarongreig: yeah the `UR_LAYERS_ENABLE=list of names` approach is way nicer, it let me clean up a bunch of stuff

source/loader/ur_adapter_registry.hpp
aarongreig: added that while testing and forgot, created #687

source/loader/layers/tracing/ur_tracing_layer.cpp
aarongreig: Cool, added some documentation of the new env var mechanism and all the layers in INTRO.rst

aarongreig:
Added a CTS style unit test suite in test/loader/loader_config/ (and it immediately caught a bug!)

source/loader/ur_lib.cpp
```diff
@@ -42,20 +80,102 @@ __urdlllocal ur_result_t context_t::Init(ur_device_init_flags_t device_flags) {
...
+    }
+    hLoaderConfig->incrementReferenceCount();
+    return UR_RESULT_SUCCESS;
+}
+
+ur_result_t urLoaderConfigRelease(ur_loader_config_handle_t hLoaderConfig) {
+    if (!hLoaderConfig) {
+        return UR_RESULT_ERROR_INVALID_NULL_HANDLE;
+    }
+    if (hLoaderConfig->decrementReferenceCount() == 0) {
```
pbalcer, line 114: This is a race, another thread might be concurrently calling release, causing double free on this object. You need to use `fetch_sub` here with `memory_order_acq_rel` (and `fetch_add` with the same flag when you increment).

Unless we don't care about thread safety here?

source/loader/ur_lib.cpp
```diff
@@ -12,25 +12,63 @@
...
-context_t::context_t() {}
+context_t::context_t() {
+    for (auto l : layers) {
+        if (l->isAvailable()) {
+            for (auto &layerName : l->getNames()) {
+                availableLayers += layerName + ";";
+            }
+        }
+    }
+    availableLayers.pop_back();
```
pbalcer, line 31: Is this to remove the ending `;`?

I was confused at first. Can you please add a comment?

source/loader/ur_lib.cpp
```diff
@@ -12,25 +12,63 @@
...
 
 ///////////////////////////////////////////////////////////////////////////////
 context_t::~context_t() {}
 
+bool context_t::layerExists(const std::string &layerName) const {
+    return availableLayers.find(layerName) != std::string::npos;
+}
+
+void context_t::parseEnvEnabledLayers() {
+    auto maybeEnableEnvVar = ur_getenv("UR_ENABLE_LAYERS");
```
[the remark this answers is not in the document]
aarongreig: that's way nicer, thanks for pointing that out

source/loader/ur_lib.cpp
aarongreig: totally didn't consider thread safety first time around, I've made a pass at making ref count atomic

pbalcer (approved):

test/conformance/source/environment.cpp
```diff
@@ -54,6 +59,7 @@ uur::PlatformEnvironment::PlatformEnvironment(int argc, char **argv)
         error = "urInit() failed";
         return;
     }
+    urLoaderConfigRelease(config);
```
kbenzie, line 62: Looks like `config` leaks if there's an error in `urInit()`.

test/conformance/source/environment.cpp
```diff
@@ -43,8 +43,13 @@ std::ostream &operator<<(std::ostream &out,
 uur::PlatformEnvironment::PlatformEnvironment(int argc, char **argv)
     : platform_options{parsePlatformOptions(argc, argv)} {
     instance = this;
+
+    ur_loader_config_handle_t config;
+    urLoaderConfigCreate(&config);
+    urLoaderConfigEnableLayer(config, "UR_LAYER_FULL_VALIDATION");
```
kbenzie, line 49: Should be bail out if these calls results in an error?

kbenzie (approved):
Couple minor comments

test/conformance/source/environment.cpp
aarongreig: yeah all should have been checked, the way we handle errors in this area isn't ideal for this many API calls happening.. I've made an attempt to have checks for everything while making sure stuff gets freed if we need to bail

test/conformance/source/environment.cpp
```diff
@@ -43,8 +43,23 @@ std::ostream &operator<<(std::ostream &out,
...
+    ur_loader_config_handle_t config;
+    if (urLoaderConfigCreate(&config) == UR_RESULT_SUCCESS) {
+        if (urLoaderConfigEnableLayer(config, "UR_LAYER_FULL_VALIDATION")) {
+            urLoaderConfigRelease(config);
+            error = "Failed to enable validation layer";
+            return;
+        }
+    } else {
+        error = "Failed to create loader config handle";
+        return;
```
kbenzie, line 56: Why not throw exceptions?

aarongreig: just following how the error reporting works elsewhere in these environment constructors

kbenzie: okay, I'll chock this up to future work then.

PR #684, opened 2023-07-04
[UR][CTS] Add tests for urPlatformGetNaitiveHandle and urPlatformCreateWithNativeHandle

### LINKED ISSUE #673: Add CTS tests for `urPlatformCreateWithNativeHandle`
We should add conformance tests for `urPlatformCreateWithNativeHandle`.

Required Coverage:

-  At least one successful (i.e. called with correct arguments and assert on expected result, not applicable to `CreateWithNativeHandle` entry points)
-  If this a query entry-point or one which accepts flags etc : we should tests every possible enumeration (where possible).
-  A test case for each returned error code (exclude disaster situations i.e. OUT_OF_MEMORY etc)

callumfare:
Resolves #673 and #674

### DIFF
--- a/test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
+++ b/test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
@@ -1,0 +1,16 @@
+// Copyright (C) 2022-2023 Intel Corporation
+// Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+// See LICENSE.TXT
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "fixtures.h"
+
+using urPlatformCreateWithNativeHandleTest = uur::platform::urPlatformTest;
+
+TEST_F(urPlatformCreateWithNativeHandleTest, InvalidNullPointerPlatform) {
+    ur_native_handle_t native_handle = nullptr;
+    ASSERT_SUCCESS(urPlatformGetNativeHandle(platform, &native_handle));
+    ASSERT_EQ_RESULT(
+        UR_RESULT_ERROR_INVALID_NULL_POINTER,
+        urPlatformCreateWithNativeHandle(native_handle, nullptr, nullptr));
+}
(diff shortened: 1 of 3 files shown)


test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
```diff
@@ -0,0 +1,16 @@
...
+// See LICENSE.TXT
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "fixtures.h"
+
+using urPlatformCreateWithNativeHandleTest = uur::platform::urPlatformTest;
+
+TEST_F(urPlatformCreateWithNativeHandleTest, InvalidNullPlatformPointer) {
+    ur_native_handle_t native_handle = nullptr;
+    ASSERT_SUCCESS(urPlatformGetNativeHandle(platform, &native_handle));
```
veselypeta, line 12: Don't change this as part of this PR - but I think we could have some logic here to skip this test if the entry-point return `UR_RESULT_ERROR_UNSUPPORTED_FEATURE` lots of adapters do not support these entry-points.

veselypeta: We actually do something similar is `urDeviceGetInfo` tests

veselypeta (approved):

test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
callumfare: Yeah that makes sense, I could go over all the GetNative/CreateWithNative entry points in a separate PR to make that consistent

test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2022-2023 Intel Corporation
+// Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+// See LICENSE.TXT
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "fixtures.h"
+
+using urPlatformCreateWithNativeHandleTest = uur::platform::urPlatformTest;
+
+TEST_F(urPlatformCreateWithNativeHandleTest, InvalidNullPlatformPointer) {
```
aarongreig, line 10: nitpick but `NullPointerPlatform` is more consistent with other tests like this

PR #689, opened 2023-07-05
Add BINDLESS_IMAGES_SHARED_USM_SUPPORT to ur_device_info_t

isaacault:
Add `UR_DEVICE_INFO_BINDLESS_IMAGES_SHARED_USM_SUPPORT` to `ur_device_info_t`. 

Note that this differs from `UR_DEVICE_INFO_USM_SYSTEM_SHARED_SUPPORT` in that the query is specifically for bindless image support on shared USM.


include/ur.py
```diff
@@ -587,33 +587,35 @@ class ur_device_info_v(IntEnum):
                                                     ## version than older devices.
     BINDLESS_IMAGES_SUPPORT_EXP = 0x2000            ## [::ur_bool_t] returns true if the device supports the creation of
                                                     ## bindless images
-    BINDLESS_IMAGES_1D_USM_SUPPORT_EXP = 0x2001     ## [::ur_bool_t] returns true if the device supports the creation of 1D
+    BINDLESS_IMAGES_SHARED_USM_SUPPORT = 0x2001     ## [::ur_bool_t] returns true if the device supports the creation of
```
alycm, line 590: Not `_EXP`?

scripts/core/exp-bindless-images.yml
```diff
@@ -32,47 +32,50 @@ etors:
     - name: BINDLESS_IMAGES_SUPPORT_EXP
       value: "0x2000"
       desc: "[$x_bool_t] returns true if the device supports the creation of bindless images"
-    - name: BINDLESS_IMAGES_1D_USM_SUPPORT_EXP
+    - name: BINDLESS_IMAGES_SHARED_USM_SUPPORT
       value: "0x2001"
+      desc: "[$x_bool_t] returns true if the device supports the creation of bindless images backed by shared USM"
```
alycm, line 37: Not sure if this needs spec clarification or if it's just a lack on knowledge on my part.

Is this `BINDLESS_IMAGES_SHARED_USM_SUPPORT` mutually exclusive with `BINDLESS_IMAGES_[1,2]D_USM_SUPPORT_EXP`?

I.e. if I want to create a 2D image backed by shared USM do I need to check both of `BINDLESS_IMAGES_SHARED_USM_SUPPORT` and `BINDLESS_IMAGES_2D_USM_SUPPORT_EXP`?

My initial thought was "why aren't there 1D/2D versions of the shared USM query?" but then I thought maybe that "shared" and "[1,2]D" are independent support properties.

include/ur.py
isaacault: Whoops! Fixed.

scripts/core/exp-bindless-images.yml
isaacault: > if I want to create a 2D image backed by shared USM do I need to check both of `BINDLESS_IMAGES_SHARED_USM_SUPPORT` and `BINDLESS_IMAGES_2D_USM_SUPPORT_EXP`?

Correct. 

The former (and new proposed) query is for support of bindless images on USM memory accessible by both host and device, as specified by the SYCL 2020 `usm::alloc::shared` type, for example. The latter is for support of 2D bindless images backed by USM, which could be limited to just being accessible on the device/host if supported. 

Does that clarify? I'm not sure I'm explaining it any better here. I would welcome any suggestions to update the `desc` with.

kbenzie (commented):
You'll need to update the changelog.

scripts/core/exp-bindless-images.yml
alycm: Yes, it does clarify and I think it was just a me understanding thing rather than a spec problem.

There could be situations where the image specific USM support flag is required? I.e. support for creating images on shared USM will be different from the presence of shared USM?

kbenzie (approved):

scripts/core/exp-bindless-images.yml
isaacault: > support for creating images on shared USM will be different from the presence of shared USM?

@alycm It may be. This would really depend on the adapter in question. Including this flag allows for bindless image support to be stretched to the limits of the adapter being used, rather than, say, having to entirely omit bindless images backed by USM merely because there isn't support for them on shared USM.

alycm: If it's a hypothetical requirement why just have `SHARED`?  The device properties have two levels of shared plus system.

* `USM_SINGLE_SHARED_SUPPORT`
* `USM_CROSS_SHARED_SUPPORT`
* `USM_SYSTEM_SHARED_SUPPORT`

alycm: I think your query matches `SYSTEM`.

isaacault: @alycm I've created [691](https://github.com/oneapi-src/unified-runtime/issues/691) in response.

alycm: Thank you @isaacault !

alycm (approved):
LGTM.

PR #693, opened 2023-07-05
[umf] create/destroy memory tracker in lib ctor/dtor

igchor:
instead of using a static variable.

In C++ no guarantee is made for initialization order of static variables across translation units.

This is problematic when an application (e.g. SYCL) wants to do some cleanup in it's library destructor. Such application might want to specify priority for it's destructor (to make sure it's executed before any other library desturctor) but it doesn't seem to work for static variables.

### DIFF
--- a/source/common/unified_malloc_framework/src/memory_tracker.cpp
+++ b/source/common/unified_malloc_framework/src/memory_tracker.cpp
@@ -18,6 +18,10 @@
 #include <shared_mutex>
 #include <stdlib.h>
 
+#ifdef _WIN32
+#include <windows.h>
+#endif
+
 // TODO: reimplement in C and optimize...
 struct umf_memory_tracker_t {
     enum umf_result_t add(void *pool, const void *ptr, size_t size) {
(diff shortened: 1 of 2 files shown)


kswiecicki (approved):
It fixed the issue with memory tracker destruction in llvm.

vinser52 (approved):

source/common/unified_malloc_framework/src/memory_tracker.cpp
```diff
@@ -84,11 +88,30 @@ umfMemoryTrackerRemove(umf_memory_tracker_handle_t hTracker, const void *ptr,
 
 extern "C" {
 
-umf_memory_tracker_handle_t umfMemoryTrackerGet(void) {
-    static umf_memory_tracker_t tracker;
-    return &tracker;
+#if defined(_WIN32) && defined(UMF_SHARED_LIBRARY)
```
pbalcer, line 91: it's fine for now while we have only one global, but eventually we might want to consolidate this somehow and create windows/linux abstractions (like in the loader).

pbalcer (approved):

PR #694, opened 2023-07-06
[UR] Extend testing of virtual memory

### LINKED ISSUE #669: [CTS] Extend testing of Virtual Memory entry-points with Read/Write commands.
#590 Introduced basic smoke testing of the Virtual Memory interfaced in UR. We should extending testing of these entry-points to include performing read/write operations to virtual memory pages to validate this functionality. We will likely require an adapter implementation which has these entry-points supported in order to extend this testing.

veselypeta:
This tests that we can you a virtual memory pointer instead of a USM pointer. In this test we enqueue a kernel with a virtual memory pointer instead of a USM pointer and using several `urUSM` entry points.

Closes #669

### DIFF
--- a/test/conformance/enqueue/urEnqueueKernelLaunch.cpp
+++ b/test/conformance/enqueue/urEnqueueKernelLaunch.cpp
@@ -126,6 +126,15 @@ struct urEnqueueKernelLaunchWithVirtualMemory : uur::urKernelExecutionTest {
     void SetUp() override {
         program_name = "fill_usm";
         UUR_RETURN_ON_FATAL_FAILURE(uur::urKernelExecutionTest::SetUp());
+
+        ur_bool_t virtual_memory_support = false;
+        ASSERT_SUCCESS(urDeviceGetInfo(
+            device, UR_DEVICE_INFO_VIRTUAL_MEMORY_SUPPORT, sizeof(ur_bool_t),
+            &virtual_memory_support, nullptr));
+        if (!virtual_memory_support) {
+            GTEST_SKIP() << "Virtual memory is not supported.";
+        }
+
         ASSERT_SUCCESS(urVirtualMemGranularityGetInfo(
             context, device, UR_VIRTUAL_MEM_GRANULARITY_INFO_MINIMUM,
             sizeof(granularity), &granularity, nullptr));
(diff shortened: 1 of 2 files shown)


test/conformance/enqueue/urEnqueueKernelLaunch.cpp
```diff
@@ -120,3 +120,82 @@ TEST_P(urEnqueueKernelLaunch3DTest, Success) {
...
+    size_t global_size = alloc_size / sizeof(uint32_t);
+
+    ASSERT_SUCCESS(urKernelSetArgPointer(kernel, 0, nullptr, virtual_ptr));
+
+    ur_event_handle_t kernel_evt;
+    ASSERT_SUCCESS(urEnqueueKernelLaunch(queue, kernel, work_dim,
+                                         &global_offset, &global_size, nullptr,
+                                         0, nullptr, &kernel_evt));
+
+    // Is this copy necessary??
```
fabiomestre, line 190: I think the copy is needed. I had a look at the level zero spec and it says:

`If an application needs finer grained control of physical memory consumption for device allocations then it can reserve a range of the virtual address space and map this to physical memory as needed.`

So, my understanding is that virtual memory behaves in a similar way as USM device allocations. And in USM device allocations we need to do an explicit copy to the host.

test/conformance/enqueue/urEnqueueKernelLaunch.cpp
```diff
@@ -120,3 +120,81 @@ TEST_P(urEnqueueKernelLaunch3DTest, Success) {
...
+
+    std::vector<uint32_t> data(global_size);
+    ASSERT_SUCCESS(urEnqueueUSMMemcpy(queue, true, data.data(), virtual_ptr,
+                                      alloc_size, 1, &kernel_evt, nullptr));
+
+    ASSERT_SUCCESS(urQueueFinish(queue));
+
+    // verify fill worked
+    for (size_t i = 0; i < data.size(); i++) {
+        ASSERT_EQ(i, data.at(i));
```
aarongreig, line 198: does this pass? it doesn't match what I'd expect `fill_usm` to do and we aren't setting the second kernel arg

kbenzie:
The [UR_DEVICE_INFO_VIRTUAL_MEMORY_SUPPORT](https://oneapi-src.github.io/unified-runtime/core/api.html#_CPPv4N16ur_device_info_t37UR_DEVICE_INFO_VIRTUAL_MEMORY_SUPPORTE) device query was added since this was opened. All tests using [virtual memory](https://oneapi-src.github.io/unified-runtime/core/api.html#virtual-memory) should be guarded on this feature being supported or skipped otherwise.

test/conformance/enqueue/urEnqueueKernelLaunch.cpp
veselypeta: Yes you're right. I've fixed the test to now check that all the values are set to the input argument.

aarongreig (approved):

kbenzie (approved):

PR #699, opened 2023-07-07
[Exp] Added const qualifier to some parameters.

rdeodhar:
The context and host pointer parameters now have const qualifier.

### DIFF
--- a/scripts/core/exp-usm-import-release.yml
+++ b/scripts/core/exp-usm-import-release.yml
@@ -10,7 +10,7 @@ name: ImportExp
 details:
     - "Import memory into USM"
 params:
-    - type: "const $x_context_handle_t"
+    - type: $x_context_handle_t
       name: hContext
       desc: "[in] handle of the context object"
     - type: "const void*"
@@ -30,7 +30,7 @@ name: ReleaseExp
 details:
     - "Release memory from USM"
 params:
-    - type: "const $x_context_handle_t"
+    - type: $x_context_handle_t
       name: hContext
       desc: "[in] handle of the context object"
     - type: "const void*"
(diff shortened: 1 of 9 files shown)


scripts/core/exp-usm-import-release.yml
```diff
@@ -10,10 +10,10 @@ name: ImportExp
 details:
     - "Import memory into USM"
 params:
-    - type: $x_context_handle_t
+    - type: "const $x_context_handle_t"
```
kbenzie, line 13: I don't think the context should be `const` qualified the reason being that an adapter may wish to store data in the context relating to the action being performed in the `urUSMImportExp`/`urUSMExportExp` entry points, making this argument `const` means that becomes impossible. Making this `const` is also inconsistent with the rest of the API.

scripts/core/exp-usm-import-release.yml
```diff
@@ -30,10 +30,10 @@ name: ReleaseExp
 details:
     - "Release memory from USM"
 params:
-    - type: $x_context_handle_t
+    - type: "const $x_context_handle_t"
```
kbenzie, line 33: Also don't think this should be `const` qualified, see above.

kbenzie (changes requested):

scripts/core/exp-usm-import-release.yml
rdeodhar: Ok, removed that const on the UR context.
Note that at the SYCL level all USM APIs (such as malloc_device) that take a queue or context parameter declare them const. But its fine to have the underlying context modifiable.

scripts/core/exp-usm-import-release.yml
rdeodhar: OK

rdeodhar: I'm closing this PR because I don't think these changes are needed. There is a remapping, including casting, between SYCL objects and UR objects anyway.

PR #701, opened 2023-07-10
Set global compile flags for UR targets only

PatKamin:
Currently, some of compile flags are set globally for the whole project and third-party libraries. Wrap all Unified Runtime CMake targets so that compile and linking flags can be added globally but without affecting third-party repositories' builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,17 +52,7 @@ if(MSVC)
     set(CUSTOM_COMMAND_BINARY_DIR ${CUSTOM_COMMAND_BINARY_DIR}/$<CONFIG>)
 endif()
 
-# CXX flags setup
 if(NOT MSVC)
-    add_compile_options(-fPIC -Wall -Wpedantic
-        $<$<CXX_COMPILER_ID:GNU>:-fdiagnostics-color=always>
-        $<$<CXX_COMPILER_ID:Clang,AppleClang>:-fcolor-diagnostics>)
-    if (CMAKE_BUILD_TYPE STREQUAL "Release")
-        add_compile_options(-D_FORTIFY_SOURCE=2)
-    endif()
-    if(UR_DEVELOPER_MODE)
-        add_compile_options(-Werror -fno-omit-frame-pointer -fstack-protector-strong)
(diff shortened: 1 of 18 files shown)


cmake/helpers.cmake
```diff
@@ -40,6 +40,54 @@ function(add_cppformat name)
...
+        umf_test_helpers
+        ur_adapter_null
+        ur_collector
+        ur_loader
+        usm_test-usmPoolManager
+        validation_test-leaks
+        validation_test-leaks_mt
+        validation_test-parameters
+    )
+endmacro()
```
kbenzie, line 73: This list is going to be a pain to maintain.

Alternatively, why not wrap `add_executable`/`add_library` to set the UR specific options? This is a pattern used by LLVM and a number of other projects.

CMakeLists.txt
```diff
@@ -119,22 +86,6 @@ if(UR_ENABLE_TRACING)
...
-endif()
-
-if(UR_USE_MSAN)
-    message(WARNING "MemorySanitizer requires that all code is built with
-        its instrumentation, otherwise false positives are possible.
-        See https://github.com/google/sanitizers/wiki/MemorySanitizerLibcxxHowTo#instrumented-libc
-        for details")
-    add_sanitizer_flag(memory)
-endif()
-
```
kbenzie, line 137: I'd argue that we _do_ want third party libraries compiled with sanitizer flags, sanitizers work best when all code has the instrumentation compiled in.

PatKamin:
> This list is going to be a pain to maintain.

Alternatively, why not wrap `add_executable`/`add_library` to set the UR specific options? This is a pattern used by LLVM and a number of other projects. 

I agree, added wrapper functions which adds UR-specific compile flags to targets.

PatKamin:
> I'd argue that we _do_ want third party libraries compiled with sanitizer flags, sanitizers work best when all code has the instrumentation compiled in. 

Leaving sanitizer flags to be set globally.

kbenzie (approved):

pbalcer (approved):

PR #703, opened 2023-07-10
[Bindless Images][Exp] Add opaque file descriptor handle

isaacault:
Add `ur_exp_file_descriptor_handle_t` to support platform dependent file descriptors.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -6601,6 +6601,10 @@ typedef struct ur_exp_interop_mem_handle_t_ *ur_exp_interop_mem_handle_t;
 typedef struct ur_exp_interop_semaphore_handle_t_ *ur_exp_interop_semaphore_handle_t;
 
 ///////////////////////////////////////////////////////////////////////////////
+/// @brief Handle of file descriptor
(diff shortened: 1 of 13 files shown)


include/ur_api.h
```diff
@@ -6600,6 +6600,10 @@ typedef struct ur_exp_interop_mem_handle_t_ *ur_exp_interop_mem_handle_t;
 /// @brief Handle of interop semaphore
 typedef struct ur_exp_interop_semaphore_handle_t_ *ur_exp_interop_semaphore_handle_t;
 
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Handle of file descriptor
+typedef struct ur_exp_file_descriptor_handle_t_ *ur_exp_file_descriptor_handle_t;
```
jandres742, line 6605: so by defining it as handle, each adapter is free to defining as it sees fit, correct?

alycm: Is that the intent?  A file handle doesn't really vary per adapter, it varies per host OS.

isaacault:
Closing in favour of #705

PR #705, opened 2023-07-11
[UR] Support platform dependant typedefs

### LINKED ISSUE #704: Allow for platform dependant type defintions
Currently in UR we have do not have a mechanism to define types diferently for different platforms i.e. Windows/Linux. This is something that would be helpful in implementing for example a file descriptor type which differs across platforms and would be important to implement bindless images experimental features. See discussion [here](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/common.yml#L39)
We can do something similar with macros currently using a `condition: "defined(_WIN32)"` field in the `yaml` syntax which will include an additional gaurd for the definition, however the is nothing for other yaml types. The most obvious place would be to add this to `typedef` which would allow us to define types differently for different platforms - a potential way this could look would be:

```yaml
type: typedef
desc: "file descriptor type:
name: $x_file_descriptor_t
value:
  - condition: "defined(_WIN32)"
    value: HANDLE
alt_value: int
```

veselypeta:
This PR will enable platform dependent types to be defined as part of the UR specification. The supported platform types are: "Windows", "Linux" & "Darwin". This is useful to model platform specific types in UR such as a file descriptor.

#### TODO
- [x] Document new syntax on `typedef` value in YaML.md
- [x] Introduce validation checks for new yaml syntax
- [x] Clean up code generation
- [x] See if `windows.h` has to be included since we refer to `HANDLE`

Closes #704

### DIFF
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -323,5 +323,5 @@ name: $x_file_descriptor_t
 value:
     - if: Windows
       value: HANDLE
-    - else: ""
+    - else:
       value: int
--- a/scripts/core/common.yml
+++ b/scripts/core/common.yml
@@ -316,3 +316,12 @@ members:
     - type: uint64_t
       name: depth
       desc: "[in] scalar (scalar)"
+--- #--------------------------------------------------------------------------
+type: typedef
+desc: "File Descriptor type."
+name: $x_file_descriptor_t
+value:
+    - if: Windows
+      value: HANDLE
+    - else: ""
+      value: int
--- a/scripts/templates/api.h.mako
+++ b/scripts/templates/api.h.mako
@@ -32,6 +32,9 @@ from templates import helper as th
 // standard headers
 #include <stdint.h>
 #include <stddef.h>
+#if defined(_WIN32)
+#include "windows.h"
+#endif
 %endif
 
 #if defined(__cplusplus)
@@ -75,14 +78,13 @@ extern "C" {
 %if isinstance(obj['value'], list):
 <% first = True%>\
 %for cond in obj['value']:
-%if cond.get('if') is not None:
+%if 'if' in cond:
 ${"#if" if first else "#elif"} ${th.convert_platform_to_define(cond['if'])}
 <% first = False %>\
-typedef ${th.subt(n, tags, cond['value'])} ${th.make_type_name(n, tags, obj)};
-%elif cond.get('else') is not None:
+%elif 'else' in cond:
 #else
-typedef ${th.subt(n, tags, cond['value'])} ${th.make_type_name(n, tags, obj)};
 %endif
+typedef ${th.subt(n, tags, cond['value'])} ${th.make_type_name(n, tags, obj)};
 %endfor
 #endif
 %else:
--- a/scripts/templates/api.h.mako
+++ b/scripts/templates/api.h.mako
@@ -72,7 +72,22 @@ extern "C" {
 %endif
 ## TYPEDEF ####################################################################
 %elif re.match(r"typedef", obj['type']):
+%if isinstance(obj['value'], list):
+<% first = True%>\
+%for cond in obj['value']:
+%if cond.get('if') is not None:
+${"#if" if first else "#elif"} ${th.convert_platform_to_define(cond['if'])}
+<% first = False %>\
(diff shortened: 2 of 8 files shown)


scripts/core/common.yml
```diff
@@ -316,3 +316,12 @@ members:
...
       desc: "[in] scalar (scalar)"
+--- #--------------------------------------------------------------------------
+type: typedef
+desc: "File Descriptor type."
+name: $x_file_descriptor_t
+value:
+    - if: Windows
+      value: HANDLE
+    - else: ""
+      value: int
```
isaacault, line 327: nit:
```suggestion
      value: uint32_t
```

veselypeta: Are you sure? `open()` will return an `int` as a file descriptor. `-1` is often used to indicate an invalid file descriptor.

isaacault: Hmm, that seems sound to me. I stand corrected!

scripts/core/common.yml
```diff
@@ -316,3 +316,12 @@ members:
...
       name: depth
       desc: "[in] scalar (scalar)"
+--- #--------------------------------------------------------------------------
+type: typedef
+desc: "File Descriptor type."
+name: $x_file_descriptor_t
+value:
+    - if: Windows
+      value: HANDLE
+    - else: ""
```
kbenzie, line 326: I don't think the `""` is strictly necessary from a yaml spec point of view, were the scripts complaining about it not having a value?

kbenzie: ah, I see, the value of `else:` is `None` so the checks are not doing the correct thing, I'll make some suggestions.

scripts/templates/api.h.mako
```diff
@@ -72,7 +72,22 @@ extern "C" {
 %endif
 ## TYPEDEF ####################################################################
 %elif re.match(r"typedef", obj['type']):
+%if isinstance(obj['value'], list):
+<% first = True%>\
+%for cond in obj['value']:
+%if cond.get('if') is not None:
```
kbenzie, line 78:
```suggestion
%if 'if' in cond:
```

scripts/templates/api.h.mako
```diff
@@ -72,7 +72,22 @@ extern "C" {
...
+${"#if" if first else "#elif"} ${th.convert_platform_to_define(cond['if'])}
+<% first = False %>\
+typedef ${th.subt(n, tags, cond['value'])} ${th.make_type_name(n, tags, obj)};
+%elif cond.get('else') is not None:
```
kbenzie, line 82:
```suggestion
%elif 'else' in cond:
```

scripts/templates/api.py.mako
```diff
@@ -50,8 +51,21 @@ ${th.make_macro_name(n, tags, obj)} = ${th.subt(n, tags, obj['value'])}
 %endif
 ## TYPEDEF ####################################################################
 %elif re.match(r"typedef", obj['type']):
+%if isinstance(obj['value'], list):
+<% first = True%>\
+%for cond in obj['value']:
+%if cond.get('if') is not None:
```
kbenzie, line 57:
```suggestion
%if 'if' in cond:
```

scripts/templates/api.py.mako
```diff
@@ -50,8 +51,21 @@ ${th.make_macro_name(n, tags, obj)} = ${th.subt(n, tags, obj['value'])}
...
+${"if" if first else "elif"} platform.system() == "${cond['if']}":<% first = False %>
+%elif cond.get('else') is not None:
```
kbenzie, line 59:
```suggestion
%elif 'else' in cond:
```

scripts/core/common.yml
kbenzie:
```suggestion
    - else:
```

kbenzie (changes requested):

scripts/templates/api.h.mako
```diff
@@ -32,6 +32,11 @@ from templates import helper as th
 // standard headers
 #include <stdint.h>
 #include <stddef.h>
+#if defined(_WIN32)
+#define NOMINMAX
+#include "windows.h"
```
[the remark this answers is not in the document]
kbenzie: At the very least we should `#define WIN32_LEAN_AND_MEAN` before including `windows.h`.

PR #709, opened 2023-07-12
[umf] add error translation function to helpers

kswiecicki:
Simple helper function for UMF error translation.

### DIFF
--- a/source/common/umf_helpers.hpp
+++ b/source/common/umf_helpers.hpp
@@ -15,6 +15,7 @@
 #include <umf/memory_pool_ops.h>
 #include <umf/memory_provider.h>
 #include <umf/memory_provider_ops.h>
+#include <ur_api.h>
 
 #include <functional>
 #include <memory>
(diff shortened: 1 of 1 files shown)


source/common/umf_helpers.hpp
```diff
@@ -160,6 +161,39 @@ template <typename Type> umf_result_t &getPoolLastStatusRef() {
     return last_status;
 }
 
+/// @brief translates UMF return values to UR.
+/// This function assumes that the native error of
+/// the last failed memory provider is ur_result_t.
+ur_result_t umf2urResult(umf_result_t umfResult) {
```
igchor, line 167: probably this function should be 'inline' - to fix the compilation error you got.

source/common/umf_helpers.hpp
```diff
@@ -160,6 +161,39 @@ template <typename Type> umf_result_t &getPoolLastStatusRef() {
...
+    if (umfResult == UMF_RESULT_SUCCESS) {
```
igchor, line 168: nit: why not have this under the switch with the rest?

kswiecicki: I was inspired by the `ur2piResult()` on sycl repo and didn't bother changing it.
I've moved it to the switch statement now.

veselypeta (approved):

igchor (approved):

PR #710, opened 2023-07-12
[UMF] add handling unsupported functions for disjoint pool

DamianDuy:
This change is needed to skip tests for a given operation if the pool doesn't support it.

### DIFF
--- a/source/common/umf_pools/disjoint_pool.cpp
+++ b/source/common/umf_pools/disjoint_pool.cpp
@@ -908,13 +908,13 @@ void *DisjointPool::malloc(size_t size) { // For full-slab allocations indicates
 
 void *DisjointPool::calloc(size_t, size_t) {
     // Not supported
-    assert(false);
+    umf::getPoolLastStatusRef<DisjointPool>() = UMF_RESULT_ERROR_NOT_SUPPORTED;
     return NULL;
 }
 
 void *DisjointPool::realloc(void *, size_t) {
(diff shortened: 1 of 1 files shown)


source/common/umf_pools/disjoint_pool.cpp
```diff
@@ -934,8 +934,6 @@ void *DisjointPool::aligned_malloc(size_t size, size_t alignment) {
 
 size_t DisjointPool::malloc_usable_size(void *) {
     // Not supported
-    assert(false);
```
vinser52, line 937: why you just deleted this line instead of replace with `umf::getPoolLastStatusRef<DisjointPool>() = UMF_RESULT_ERROR_NOT_SUPPORTED;`?

igchor (approved):

source/common/umf_pools/disjoint_pool.cpp
DamianDuy: I didn't want to use it here, since this function is not an allocation function. And if it returns zero, we can tell that it's not supported.

pbalcer: `malloc_usable_size` returns 0 for null pointers and zero-length allocations are technically a thing. So you might not necessarily be able to disambiguate between a valid call and an error - but do we need to? I'd argue that we don't.

kswiecicki (approved):

Issue #714, opened 2023-07-13
Clarify or fix `urInit` and `urTeardown` behavior


pbalcer:
During the discussion on #220, it was established that current global state initialization and destruction with the use of `urInit` and `urTearDown` are, at best poorly documented, and at worst, they might prevent the correct use of Unified Runtime in two independent modules in the same process.

The current documentation for `urInit` states that `The application may call this function multiple times with different flags or environment variables enabled.` and `The implementation of this function must be thread-safe for scenarios where multiple libraries may initialize the adapter(s) simultaneously.`. These two requirements can only be fulfilled together if the `different flags or environment variables` in all calls after the first one are entirely ignored. This might be misleading, especially if some library that uses UR doesn't take into account that it might be used in a process where UR is already initialized - this might be a relevant consideration for the design of the `ONEAPI_DEVICE_SELECTOR` (#220) feature or the loader config implementation (#681). In other words, software currently has no guarantees whether `urInit` does anything, but the documentation makes it seem as if it does.

The `urTearDown` function documentation doesn't specify whether it can be called multiple times. Current adapter implementations cannot. This creates a problem `for scenarios where multiple libraries may initialize the adapter(s) simultaneously.`, because those libraries would require an external mechanism to coordinate calling `urTearDown`. This might not be practical. 

In previous discussions, a proposal was made to create a separate reference-counted object on which `urInit` and `urTearDown` operate. In addition to helping clarify the global state lifetime, it would also address several other API issues as outlined in https://github.com/oneapi-src/unified-runtime/issues/220#issuecomment-1625154663.

pbalcer:
Addressed in #715, with a few small follow up in changes in #728.

PR #715, opened 2023-07-13
Implement adapter instance handles 

callumfare:
Corresponding changes to the adapters, pi2ur, etc are in https://github.com/intel/llvm/pull/10349/

Summary of new/changed entry points:

- **urInit** Adapters no longer perform global state setup in urInit. The loader may still perform setup (layer config etc), so the entry point is still needed, but it might be possible to combine it with the below.
- **urAdapterGet** Individual adapters return exactly 1 adapter handle, the loader returns however many adapters are known. Each returned adapter increments the reference count of the adapter by 1. When the reference count goes from 0 to 1, adapters may perform setup of global state (which previously happened in urInit). The underlying object pointed to by the `ur_adapter_handle_t` should be a single static object in the adapter.
- **urPlatformGet** Now explicitly takes the adapters to use as arguments. Multiple adapters can be used in the same call. This allows all platforms available to be queried as before, or more fine selection of platforms from certain adapters.
- **urAdapterRetain/Release** Increment or decrement the adapter reference count. When the reference count drops to 0, adapters may perform teardown of global state (which previously happened in urTearDown).
- **urTearDown** Adapters no longer perform global state teardown in urTearDown. This entry point may be kept if useful for the loader, otherwise it could be removed.
- **urAdapterGetLastError** Previously urPlatformGetLastError. No changes other than it takes an adapter handle rather than platform handle, removing the need to always have a platform handle when a UR entry point is called to be able to process an adapter-specific error.
- **urAdapterGetInfo** Can be used to query the backend of the adapter (which is just a duplicate of `UR_PLATFORM_BACKEND`).

Addresses #714 by allowing better sharing of UR adapters when multiple modules in an application share the UR libraries. They may independently call `urAdapterGet` safely, or may explicitly pass adapter handles around (retaining/releasing as necessary).

Provides some more programmatic control over which adapters are used by allowing users to query which backend is supported by each adapter, and only fetching platforms from certain adapters.

Fixes https://github.com/intel/llvm/issues/10066 by allowing `piPluginGetLastError` to be implementable in pi2ur.

### DIFF
--- a/scripts/core/platform.yml
+++ b/scripts/core/platform.yml
@@ -13,7 +13,7 @@ desc: "Intel $OneApi Unified Runtime APIs for Platform"
 ordinal: "1"
 --- #--------------------------------------------------------------------------
 type: function
-desc: "Retrieves all available platforms"
+desc: "Retrieves all available platforms for the given adapters"
 class: $xPlatform
 name: Get
 decl: static
@@ -24,6 +24,12 @@ details:
     - "Multiple calls to this function will return identical platforms handles, in the same order."
     - "The application may call this function from simultaneous threads, the implementation must be thread-safe"
 params:
+    - type: "$x_adapter_handle_t*"
+      name: "phAdapters"
+      desc: "[in][range(0, NumAdapters)] array of adapters to query for platforms."
+    - type: "uint32_t"
+      name: "NumAdapters"
+      desc: "[in] number of adapters pointed to by phAdapters"
     - type: "uint32_t"
       name: NumEntries
       desc: |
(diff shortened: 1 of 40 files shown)


scripts/core/platform.yml
```diff
@@ -24,6 +24,12 @@ details:
     - "Multiple calls to this function will return identical platforms handles, in the same order."
     - "The application may call this function from simultaneous threads, the implementation must be thread-safe"
 params:
+    - type: "$x_adapter_handle_t*"
+      name: "phAdapters"
+      desc: "[in][range(0, NumAdapters)] array of adapters to query for platforms."
+    - type: "uint32_t"
+      name: "NumAdapters"
+      desc: "[in] number of adapters pointed to by phAdapters"
```
kbenzie, line 32: Feels odd to be passing in all the adapters here as a list, its a different approach to `urDeviceGet` only taking a single platform. I can see why this is a less intrusive way to do it though.

scripts/core/runtime.yml
```diff
@@ -64,3 +174,184 @@ params:
...
+      desc: "The backend is Level Zero"
+    - name: OPENCL
+      value: "2"
+      desc: "The backend is OpenCL"
+    - name: CUDA
+      value: "3"
+      desc: "The backend is CUDA"
+    - name: HIP
+      value: "4"
+      desc: "The backend is HIP"
```
kbenzie, line 357: This is missing `NATIVE_CPU` which just got merged.

pbalcer (approved):

scripts/core/runtime.yml
```diff
@@ -64,3 +174,184 @@ params:
...
+      desc: >
+          [out] pointer to an integer where the adapter specific error code
+          will be stored.
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Supported adapter info"
+class: $x
+name: $x_adapter_info_t
+typed_etors: True
+etors:
```
pbalcer, line 292: `REFERENCE_COUNT` and a test ?

pbalcer (commented):
In a separate PR we might also want to update `urInit` and `urTearDown` documentation.

scripts/core/runtime.yml
```diff
@@ -174,3 +174,184 @@ params:
       desc: "[in] pointer to tear down parameters"
 returns:
     - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
+--- #--------------------------------------------------------------------------
```
veselypeta, line 177: Maybe in future we can move this into it's own `adapter.yml`

veselypeta (approved):
LGTM

scripts/core/platform.yml
callumfare: Yeah, I did it this way to keep the changes as unobtrusive as possible, especially since `urPlatformGet` is in pretty much every example and test.

It also means fewer changes in the loader, which already had a special case for `urPlatformGet` where it looped over every adapter in the loader context. This PR just changes that to the adapters passed in.

If consistency with the other entry points is important I could look into changing this.

kbenzie: I don't think its a blocker for this work but perhaps something to look into at a later point.

scripts/core/runtime.yml
callumfare: Added `REFERENCE_COUNT` and tests for this and all the other new entry points

scripts/core/runtime.yml
callumfare: Made a note to revisit this in #728

PR #716, opened 2023-07-13
[Bindless][Exp] Add bindless images interop resource handles

przemektmalon:
- Added new structs `ur_exp_interop_memory_desc_t` and `ur_exp_interop_semaphore_desc_t`.
- Updated `urBindlessImagesImportOpaqueFDExp` and `urBindlessImagesImportExternalSemaphoreOpaqueFDExp` to use the new structs.

### DIFF
--- a/scripts/core/exp-bindless-images.yml
+++ b/scripts/core/exp-bindless-images.yml
@@ -86,6 +86,18 @@ etors:
     - name: EXP_SAMPLER_MIP_PROPERTIES
       desc: $x_exp_sampler_mip_properties_t
       value: "0x2000"
+    - name: EXP_INTEROP_MEMORY_DESC
+      desc: $x_exp_interop_memory_desc_t
+      value: "0x2001"
+    - name: EXP_INTEROP_SEMAPHORE_DESC
+      desc: $x_exp_interop_semaphore_desc_t
+      value: "0x2002"
+    - name: EXP_FILE_DESCRIPTOR
+      desc: $x_exp_file_descriptor_t
+      value: "0x2003"
+    - name: EXP_WIN32_HANDLE
+      desc: $x_exp_win32_handle_t
+      value: "0x2004"
 --- #--------------------------------------------------------------------------
 type: enum
 extend: true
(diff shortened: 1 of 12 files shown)


kbenzie (approved):

scripts/core/exp-bindless-images.yml
```diff
@@ -133,6 +163,20 @@ members:
...
+base: $x_base_desc_t
+members: []
+--- #--------------------------------------------------------------------------
+type: struct
+desc: "Describes an interop semaphore resource descriptor"
+class: $xBindlessImages
+name: $x_exp_interop_semaphore_desc_t
+base: $x_base_desc_t
+members: []
+--- #--------------------------------------------------------------------------
```
veselypeta, line 179: I though we were not going to use empty structs and instead type the parameter as `void *` to avoid this chaining?
so we have something like:

```cpp
ur_result_t urBindlessImagesImportOpaqueFDexp(ur_context_handle_t ctx, ur_device_handle_t dev, size_t size, void * pDesc, ur_exp_interop_mem_handle_t *retHandle);

ur_exp_file_descriptor_t fd{
  UR_STRUCTURE_EXP_FILE_DESCRIPTOR,
  nullptr,
  FILE
};

ur_exp_interop_mem_handle_t handle;
urBindlessImagesImportOpaqueFDexp(context, device, size, &fd, &handle);

```

scripts/core/exp-bindless-images.yml
```diff
@@ -518,9 +562,9 @@ params:
     - type: size_t
       name: size
       desc: "[in] size of the external memory"
-    - type: uint32_t
-      name: fileDescriptor
-      desc: "[in] the file descriptor"
+    - type: $x_exp_interop_memory_desc_t*
```
veselypeta, line 565:
```suggestion
    - type: void*
```

scripts/core/exp-bindless-images.yml
```diff
@@ -597,11 +641,11 @@ params:
     - type: $x_device_handle_t
       name: hDevice
       desc: "[in] handle of the device object"
-    - type: uint32_t
-      name: fileDescriptor
-      desc: "[in] the file descriptor"
+    - type: $x_exp_interop_semaphore_desc_t*
```
veselypeta, line 644:
```suggestion
    - type: void*
```

veselypeta (approved):

scripts/core/exp-bindless-images.yml
kbenzie: Oh, that's not how I understood the converstation. I thought we only agreed to `void*` instead of `HANDLE`.

kbenzie: That would be a departure to the rest of UR's extensibilty approach, I think the PR as it stands is the most consistent approach.

PR #724, opened 2023-07-17
benie/unsupported native handles

kbenzie:
- [ur] Update CTS to match native handle nocheck
- [ur] Allow native handles to be unsupported


scripts/core/memory.yml
```diff
@@ -503,6 +509,9 @@ params:
     - type: $x_mem_handle_t*
       name: phMem
       desc: "[out] pointer to handle of image memory object created."
+returns:
```
aarongreig, line 512: not an issue with these changes but am I missing something or do we have an `ImageCreateWithNativeHandle` with no corresponding `ImageGetNativeHandle`

kbenzie: [urMemGetNativeHandle](https://oneapi-src.github.io/unified-runtime/core/api.html#urmemgetnativehandle) should suffice for both buffers and images, the reason we have separate `urMem{Buffer,Image}CreateWithNative` handle entry-points is to provide the additional information relevant to each.

aarongreig: aahhh gotcha

martygrant (approved):

PR #734, opened 2023-07-19
Add first Docker image recipe

lukaszstolarczuk:
This image can be used in CI or as a development container.

// tested locally, it works like a charm.

### DIFF
--- a/.github/docker/README.md
+++ b/.github/docker/README.md
@@ -1,0 +1,42 @@
+# Content
+
+Dockerfiles and scripts placed in this directory are intended to be used as
+development process vehicles and part of continuous integration process.
+
+Images built out of those recipes may be used with Docker or podman as
+development environment.
+
+# How to build docker image
+
+To build docker image on local machine execute:
+
+```sh
+docker build -t ur:ubuntu-22.04 -f ./ubuntu-22.04.Dockerfile .
+```
+
+To set any build time variable (e.g., an optional ARG from docker recipe), add to the command (after `build`), e.g.:
+
+```sh
+ --build-arg SKIP_DPCPP_BUILD=1
+```
+
+One other example of using these extra build arguments are proxy settings. They are required for accessing network
+(e.g., to download dependencies within docker), if a host is using a proxy server. Example usage:
+
+```sh
+ --build-arg https_proxy=http://proxy.com:port --build-arg http_proxy=http://proxy.com:port 
+```
+
+# How to use docker image
+
+To run docker container (using the previously built image) execute:
+
+```sh
+docker run --shm-size=4G -v /your/workspace/path/:/opt/workspace:z -w /opt/workspace/ -it ur:ubuntu-22.04 /bin/bash
+```
+
+To set (or override) any docker environment variable, add to the command (after `run`):
+
+```sh
+ -e ENV_VARIABLE=VALUE
+```
--- a/.github/docker/ubuntu-22.04.Dockerfile
+++ b/.github/docker/ubuntu-22.04.Dockerfile
@@ -1,0 +1,70 @@
+# Copyright (C) 2023 Intel Corporation
+# Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+# See LICENSE.TXT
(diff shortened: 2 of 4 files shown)


.github/docker/README.md
```diff
@@ -0,0 +1,35 @@
...
+```sh
+ --build-arg SKIP_DPCPP_BUILD=1
+```
+
+# How to use docker image
+
+To run docker container (using the previously built image) execute:
+
+```sh
+docker run --network=bridge --shm-size=4G -v /your/workspace/path/:/opt/workspace:z -w /opt/workspace/ -it ur:ubuntu-22.04 /bin/bash
```
kbenzie, line 28: Is `--network=bridge` necessary for using a proxy network?

.github/docker/ubuntu-22.04.Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+
+# Install libbacktrace
+COPY install_libbacktrace.sh install_libbacktrace.sh
+RUN ./install_libbacktrace.sh
+
+# Add and switch to a new (non-root) 'user'
+ENV USER user
+ENV USERPASS pass
+RUN useradd -m $USER -g sudo -p `mkpasswd $USERPASS`
+USER $USER
```
kbenzie, line 71: I personally use `podman` instead of `docker` which when used in rootless mode matches the uid/gid of the containers root user to that of the current host user which is nice when mounting volumes in the host user home directory as any modified files have the corrent owner. This is obviously not how `docker` works, since all containers are run as the host root user.

Having the container image setup a user breaks the ease of use of podman running in rootless mode. Would it be possible to make user creation optional somehow?

kbenzie: Another approach I've seen with `docker` is to explicitly map the containers root user to the current host users uid/gid by passing `--user $UID:$GID` to `docker run`.

.github/docker/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Images built out of those recipes may by used with Docker or podman as
+development environment.
+
+# How to build docker image
+
+To build docker image on local machine execute:
+
+```sh
+docker build --build-arg https_proxy=http://proxy.com:port --build-arg http_proxy=http://proxy.com:port -t ur:ubuntu-22.04 -f ./ubuntu-22.04.Dockerfile .
```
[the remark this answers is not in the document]
lukaszstolarczuk: This is just a readme file with example commands - many of these parameters are optional (even image tagging) 😉 I believe many of us, who use this instruction, are located behind corporate proxy servers. Nonetheless, I rephrased it to be optional.

.github/docker/README.md
lukaszstolarczuk: actually, I removed it, 'bridge' is the default value

.github/docker/ubuntu-22.04.Dockerfile
lukaszstolarczuk: It seems like a nice thing, but... I can't make it work locally (using `docker`). I guess it's time to switch to podman 😉 

Anyway, I removed the `USER` command, since we can switch to `user` in CI, using your proposed `--user` param.

kbenzie: > It seems like a nice thing, but... I can't make it work locally (using docker).

That's a shame.

> Anyway, I removed the USER command, since we can switch to user in CI, using your proposed --user param.

Just tried out the Dockerfile with `podman` and its working nicely now. Thanks :)

kbenzie (approved):

.github/docker/README.md
```diff
@@ -0,0 +1,42 @@
+# Content
+
+Dockerfiles and scripts placed in this directory are intended to be used as
+development process vehicles and part of continuous integration process.
+
+Images built out of those recipes may by used with Docker or podman as
```
kswiecicki, line 6: nit: [...] may be

kswiecicki (approved):

.github/docker/README.md
lukaszstolarczuk: thx!

pbalcer (approved):

PR #739, opened 2023-07-21
[umf] Make pool tracking optional

kswiecicki:


igchor:
LGTM, alternatively we could consider adding a compile-time flag that would disable tracking + we could put `umfPoolByPtr` and `umfFree` under ifdef - right now, if someone calls `umfFree` it will result in a memory leak (if return value is not checked).

pbalcer:
Yea, I think I'd prefer a compile-time option (that disables this functionality by default).

PR #740, opened 2023-07-21
Implement ONEAPI_DEVICE_SELECTOR in UR

Wee-Free-Scot:
Add urDeviceGetSelected and implement to return only those devices specified by the ONEAPI_DEVICE_SELECTOR (for the platform given as input).

Add urPlatformGetSelected and implement to return only those platforms specified by the ONEAPI_DEVICE_SELECTOR.


pbalcer:
With the introduction of config in #681 we have several functions in the API that are loader-only and are excluded from the [ddi](https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_ddi.h). Should this be the same? If so, it could also be a good idea to create a separate loader-specific `.yml` spec file and improve the way the entry point it would contain are excluded from the ddi ([right now](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/templates/ddi.h.mako#L34C13-L34C13) the scripts look whether the name contains the word 'loader'...).

Wee-Free-Scot:
@pbalcer what are the criteria for loader-only?

ONEAPI_DEVICE_SELECTOR is general to all the devices exposed by UR and applies to all backends -- so it seems incorrect to limit it so that sometimes it takes effect (UR with loader) and other times it does not (UR without loader). Does that make sense?

pbalcer:
Using adapters standalone is not a supported usage model. I think (@kbenzie ?). If it is, it's currently broken (e.g., the loader config symbols would be missing).

My understanding is that we want to implement this functionality at the loader level precisely so that it applies uniformly to all downstream adapters. If we treat the `urPlatformGetSelected` and `urDeviceGetSelected` just like all other APIs, each adapter would have to look at the environment variable individually, and then we'd have to stitch that together at the loader layer - that seems more complicated.

kbenzie:
> Using adapters standalone is not a supported usage model. I think (@kbenzie ?).

That is correct.

> My understanding is that we want to implement this functionality at the loader level precisely so that it applies uniformly to all downstream adapters. If we treat the `urPlatformGetSelected` and `urDeviceGetSelected` just like all other APIs, each adapter would have to look at the environment variable individually, and then we'd have to stitch that together at the loader layer - that seems more complicated.

This matches my expectations.

Wee-Free-Scot:
@pbalcer 
> create a separate loader-specific .yml spec file and improve the way the entry point it would contain are excluded from the ddi

Is that something I could/should do as part of this PR or does that add another dependency?

For now, in this PR, should I just insert "loader" into the API names?

pbalcer:
> @pbalcer
> 
> > create a separate loader-specific .yml spec file and improve the way the entry point it would contain are excluded from the ddi
> 
> Is that something I could/should do as part of this PR or does that add another dependency?
> 
> For now, in this PR, should I just insert "loader" into the API names?

No, if everyone agrees that this makes sense, I can do that refactoring in a separate PR.

pbalcer:
Done in #770

include/ur_api.h
```diff
@@ -1317,6 +1318,46 @@ urDeviceGet(
...
+///     - ::UR_RESULT_ERROR_UNINITIALIZED
+///     - ::UR_RESULT_ERROR_DEVICE_LOST
+///     - ::UR_RESULT_ERROR_ADAPTER_SPECIFIC
+///     - ::UR_RESULT_ERROR_INVALID_NULL_HANDLE
+///         + `NULL == hPlatform`
+///     - ::UR_RESULT_ERROR_INVALID_ENUMERATION
+///         + `::UR_DEVICE_TYPE_VPU < DeviceType`
+///     - ::UR_RESULT_ERROR_INVALID_VALUE
+UR_APIEXPORT ur_result_t UR_APICALL
+urDeviceGetSelected(
```
jandres742, line 1346: why do we need this in addition to urDeviceGet? Main issue here is that libraries would have duplicated code to discover devices and it would be confusing which one to use. It seems that urDeviceGetSelected is for when ONEAPI_DEVICE_SELECTOR is used, but a library developer doesnt know what the user will use at the end.

I was thinking this logic would be added directly to urDeviceGet, so we have one single entry for device discovery. This is similar to what L0 does, where only zeDeviceGet discovers devices, whether the affinity mask is present or not

Wee-Free-Scot: MPI must always ignore the ONEAPI_DEVICE_SELECTOR, SYCL must never ignore the ONEAPI_DEVICE_SELECTOR, hence "get me a list of all devices" will give two different answers depending on which SW component is asking for the list.

One of the reasons for creating ONEAPI_DEVICE_SELECTOR, despite the previous existence of ZE_AFFINITY_MASK was that Level Zero only has one entry-point that always applies ZE_AFFINITY_MASK but that causes trouble for MPI (and similar SW components).

jandres742: thanks @Wee-Free-Scot . 

Maybe you are referring to the fact that MPI needs to have always visibility of all the devices, while SYCL only of those exposed by ONEAPI_DEVICE_SELECTOR, right?

jandres742: talked with @Wee-Free-Scot.
intention is to have urDeviceGet for full system device visibility (like what MPI needs), and urDeviceGetSelected for filtered device visibility (like what SYCL and OpenMP needs).

In that sense, changes are looking good.

source/ur_api.cpp
```diff
@@ -673,6 +673,50 @@ ur_result_t UR_APICALL urDeviceGet(
...
+    ///< will be returned.
+    ur_device_handle_t *
+        phDevices, ///< [out][optional][range(0, NumEntries)] array of handle of devices.
+    ///< If NumEntries is less than the number of devices available, then only
+    ///< that number of devices will be retrieved.
+    uint32_t *pNumDevices ///< [out][optional] pointer to the number of devices.
+    ///< pNumDevices will be updated with the total number of selected devices
+    ///< available for the given platform.
+) {
+    ur_result_t result = UR_RESULT_SUCCESS;
```
Wee-Free-Scot, line 716: What is going on here? Is this just a hangover from the previous non-loader version?

CMakeLists.txt
```diff
@@ -150,7 +150,7 @@ if(UR_FORMAT_CPP_STYLE)
         message(STATUS "Found clang-format: ${CLANG_FORMAT} (version: ${CLANG_FORMAT_VERSION})")
 
         set(CLANG_FORMAT_REQUIRED "15.0")
-        if(NOT (CLANG_FORMAT_VERSION VERSION_EQUAL CLANG_FORMAT_REQUIRED))
+        if(NOT (CLANG_FORMAT_VERSION VERSION_GREATER_EQUAL CLANG_FORMAT_REQUIRED))
```
pbalcer, line 153: clang-format is not backward compatible in terms of its output, so a newer version can (and often does) output something different than the previous one. Since we want to enforce formatting at CI level and to avoid PRs randomly changing styling depending on the author's clang-version, we require one specific clang-format version. If it's different than what you have in your system, the best approach we found is to use pyenv with the [correct version](https://github.com/oneapi-src/unified-runtime/blob/main/third_party/requirements.txt#L40) installed:
https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#build-environment

source/loader/ur_lib.cpp
```diff
@@ -206,4 +212,579 @@ urLoaderConfigSetCodeLocationCallback(ur_loader_config_handle_t hLoaderConfig,
...
+            "\\*\\.\\*" // C++ and regex escapes, literal '*.*'
+            "|"
+            "[[:digit:]]+\\.[[:digit:]]+\\.[[:digit:]]+" // '<num>.<num>.<num>'
+            "|"
+            "[[:digit:]]+\\.[[:digit:]]+\\.\\*" // '<num>.<num>.*'
+            "|"
+            "[[:digit:]]+\\.\\*\\.\\*" // '<num>.*.*'
+            "|"
+            "\\*\\.\\*\\.\\*" // C++ and regex escapes, literal '*.*.*'
+        ")$", std::regex_constants::icase);
```
kbenzie, line 360: We recently removed use of `std::regex` from the OpenCL adapter in #1039 due to profiling with [PIN](https://www.intel.com/content/www/us/en/developer/articles/tool/pin-a-dynamic-binary-instrumentation-tool.html) showed it to produce incredibly innefficient code. That was only for the OpenCL adapter, this would be on the critical path which makes me a bit uncomfortable.

Wee-Free-Scot: That (don't use `std::regex`) is good to know, thanks.

Currently, that regex pattern is only referenced in this comment:
line 491 `// TODO -- use regex validation_pattern to catch all other syntax errors in the ODS string`

Without this catch-all, some syntax errors will make it into the code that assumes there are no syntax errors, which is not great.

I would not expect that this code would ever be on the critical path of a performance critical loop. I would not expect users to query the list of devices (directly or indirectly) in a performance critical situation. Do you anticipate user code that does this?

To mitigate such user code (if it is a design goal), the parsing of the env var (including creating and using the regex) could be done only once and the outcome could be cached in the loader. The current code does not implement that, but it could.

kbenzie: > Currently, that regex pattern is only referenced in this comment:
line 491 // TODO -- use regex validation_pattern to catch all other syntax errors in the ODS string

Is the regex only intended to be used during validation or as part of the parsing of the ODS string?

> To mitigate such user code (if it is a design goal), the parsing of the env var (including creating and using the regex) could be done only once and the outcome could be cached in the loader. The current code does not implement that, but it could.

I think this could be acceptable.

test/conformance/device/urDeviceGetSelected.cpp
```diff
@@ -0,0 +1,249 @@
+// Copyright (C) 2022-2023 Intel Corporation
```
fabiomestre, line 1:
```suggestion
// Copyright (C) 2024 Intel Corporation
```

fabiomestre:
Should we document the expected behaviour of the environment variable in the UR specification? I have concerns that someone trying to implement an adapter for UR wouldn't be able to find all the information needed to implement urDeviceGetSelected() from the UR spec alone.

Also have some concerns about the naming. I think that UR should strive to be agnostic regarding the upstream runtimes. Adding support for a variable with "ONEAPI" in the name goes against that in my opinion. I think we could easily map ONEAPI_DEVICE_SELECTOR to a UR specific variable while keeping the same syntax for simplicity. However, I know there have been some discussions about the naming of ONEAPI_DEVICE_SELECTOR before, so there might be a good reason not to do this.

kbenzie:
> Also have some concerns about the naming. I think that UR should strive to be agnostic regarding the upstream runtimes. Adding support for a variable with "ONEAPI" in the name goes against that in my opinion. I think we could easily map ONEAPI_DEVICE_SELECTOR to a UR specific variable while keeping the same syntax for simplicity. However, I know there have been some discussions about the naming of ONEAPI_DEVICE_SELECTOR before, so there might be a good reason not to do this.

This has come up a few times, the naming is not going to change since its a product stack naming choice and applies to all of oneAPI, not just UR. Also note that this repo exists in the oneapi-src GitHub org, so its not out of place here.

Wee-Free-Scot:
> I have concerns that someone trying to implement an adapter for UR wouldn't be able to find all the information needed to implement urDeviceGetSelected() from the UR spec alone.

No adapter implements this API -- it is loader only. The implementation in loader uses urDeviceGet from the adapter(s), then collates and selects the information that it will pass on to the client.

kbenzie:
I thought which occurred to me recently @Wee-Free-Scot. When we merge this, how would it interact with the implementation of `ONEAPI_DEVICE_SELECTOR` which already exists in the SYCL RT?

Would it not matter because the SYCL RT isn't using the UR loader yet?

Wee-Free-Scot:
There are at least two "wouldn't matter" arguments:
1. SYCL RT does not use URT yet, so the new API will never get called.
2. URT provides both urDeviceGet (unadulterated list of devices passed through from the underlying adapter) and urDeviceGetSelected (modified list of devices applying ONEAPI_DEVICE_SELECTOR), so future versions of SYCL RT can chose when to call the new API.

I would expect an incremental approach to implementation usage in SYCL RT:
1. current -- gets list of devices without URT, runs SYCL RT code to apply ODS
2. intermediate -- gets list of devices from urDeviceGet, runs SYCL RT code to apply ODS, regression test vs, (1)
3. target -- gets list of devices from urDeviceGetSelected, bypasses SYCL RT code to apply ODS, regression test vs, (2)

kbenzie:
Makes sense. Do we have regression tests for *2. intermediate*/*3. target* at the moment?

kbenzie:
> In terms of regression testing for SYCL RT, we'd need to involve the appropriate people when we know who will do the implementation of future implementations (2) and (3).

Given my team will be responsible for replacing PI with UR in the SYCL RT, that could be us.

kbenzie:
I'm testing this in my system which has both a NVIDIA and an AMD GPU. The topology looks like this:

```
.
└── adapters
    ├── 0 hip
    │   └── platforms
    │       └── 0
    │           └── devices
    │               └── RX6700XT
    └── 1 cuda
        └── platforms
            └── 0
                └── devices
                    └── RTX3060Ti
```

I've modified the `urinfo` tool to use `urDeviceGetSelected` instead of `urDeviceGet` for getting quick results. This is without `ONEAPI_DEVICE_SELECTOR` set, both GPU's are listed as expected:

```
$ build-r/bin/urinfo
DEBUG: getenv_to_map parsed env var and failed to produce a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
DEBUG: getenv_to_map parsed env var and failed to produce a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
DEBUG: getenv_to_map parsed env var and failed to produce a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
DEBUG: getenv_to_map parsed env var and failed to produce a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
[adapter(0,hip):platform(0):device(0,gpu)] AMD HIP BACKEND, AMD Radeon RX 6700 XT gfx1031 [HIP 60032.83]
[adapter(1,cuda):platform(0):device(0,gpu)] NVIDIA CUDA BACKEND, NVIDIA GeForce RTX 3060 Ti 8.6 [CUDA 12.2]
```

And when set to `cuda:gpu`:

```
$ ONEAPI_DEVICE_SELECTOR=cuda:gpu build-r/bin/urinfo
ninja: Entering directory `build-r'
ninja: no work to do.
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
WARNING: ignoring term with irrelevant backend
error: urDeviceGetSelected(platform, UR_DEVICE_TYPE_ALL, 0, nullptr, &numDevices) failed: UR_RESULT_ERROR_INVALID_VALUE
```

And set to `hip:gpu`:

```
$ ONEAPI_DEVICE_SELECTOR=hip:gpu build-r/bin/urinfo
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
WARNING: ignoring term with irrelevant backend
error: urDeviceGetSelected(platform, UR_DEVICE_TYPE_ALL, 0, nullptr, &numDevices) failed: UR_RESULT_ERROR_INVALID_VALUE
```

What I think is happening here is these GPU's are provided by different adapters but the currently implementation of `urDeviceGetSelected` is only considering the devices within the platform it is given.

Wee-Free-Scot:
Only looking for devices matching both the input parameters and the ODS env var is the intent of the code.
I was expecting the new API to be a drop-in replacement for the existing UR function, which takes a platform handle.
I was anticipating that the caller/client would already have code for loading each adapter and calling urDeviceGet using each of the platform handles in turn.
If platform==level_zero and ODS==cuda:gpu then the list of selected devices is supposed to be empty.
The error looks wrong -- it should return an empty list, i.e. numDevices:=0 and return value UR_SUCCESS.

kbenzie:
> The error looks wrong -- it should return an empty list, i.e. numDevices:=0 and return value UR_SUCCESS.

Okay, I made this change:

```diff
benie@bench git diff source/loader/ur_lib.cpp
diff --git a/source/loader/ur_lib.cpp b/source/loader/ur_lib.cpp
index 17489efc..64a0f3bb 100644
--- a/source/loader/ur_lib.cpp
+++ b/source/loader/ur_lib.cpp
@@ -564,7 +564,7 @@ ur_result_t urDeviceGetSelected(ur_platform_handle_t hPlatform,

     if (acceptDeviceList.size() == 0 && discardDeviceList.size() == 0) {
         // nothing in env var was understood as a valid term
-        return UR_RESULT_ERROR_INVALID_VALUE;
+        return UR_RESULT_SUCCESS;
     } else if (acceptDeviceList.size() == 0) {
         // no accept terms were understood, but at least one discard term was
         // we are magnanimous to the user when there were bad/ignored accept terms
```

And am now getting these results:

```
$ ONEAPI_DEVICE_SELECTOR=cuda:gpu build-run urinfo
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
WARNING: ignoring term with irrelevant backend
No devices found platform 0.
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
[adapter(1,cuda):platform(0):device(0,gpu)] NVIDIA CUDA BACKEND, NVIDIA GeForce RTX 3060 Ti 8.6 [CUDA 12.2]
```

```
$ ONEAPI_DEVICE_SELECTOR=hip:gpu buildr-/bin/urinfo
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
DEBUG: size of acceptDeviceList = 1
DEBUG: size of discardDeviceList = 0
DEBUG: In ApplyFilter, if block case 1, matches = 1
DEBUG: getenv_to_map parsed env var and produced a map
DEBUG: termType isAcceptFilter
WARNING: ignoring term with irrelevant backend
No devices found platform 0.
[adapter(0,hip):platform(0):device(0,gpu)] AMD HIP BACKEND, AMD Radeon RX 6700 XT gfx1031 [HIP 60032.83]
```

Which is looking much better.

Wee-Free-Scot:
The ODS=cuda:gpu run looks good, but the ODS=hip:gpu run gives contradictory output "No devices found platform 0" followed immediately by details of the 1 device in that zero-length list! Without the debug messages, it looks fine, but how it gets the answer is probably significant.

Wee-Free-Scot:
Is there threading involved here? Are the output lines only partially-ordered? Does it check platform not-0 before it checks platform 0 and then mislabel one of the two final output lines?

kbenzie:
> The ODS=cuda:gpu run looks good, but the ODS=hip:gpu run gives contradictory output "No devices found platform 0" followed immediately by details of the 1 device in that zero-length list! Without the debug messages, it looks fine, but how it gets the answer is probably significant.

They both print `No devices found platform 0.` but at different points, that's coming from `urinfo` so would need to update what it prints.

> Is there threading involved here? Are the output lines only partially-ordered? Does it check platform not-0 before it checks platform 0 and then mislabel one of the two final output lines?

There's no threading, its this function [urinfo::app::enumerateDevices()](https://github.com/oneapi-src/unified-runtime/blob/main/tools/urinfo/urinfo.cpp#L63-L108) but using `urDevicesGetSelected` instead of `urDevicesGet`.

kbenzie (approved):

PR #744, opened 2023-07-24
[umf] add synchronous umf*Memcpy functions

igchor:
which allow to copy memory between different pools/providers.

The idea is to expose `memcpy_from` and `memcpy_to` in provider_ops. When we copy between 2 different memory providers we can try to use `memcpy_from` on the `src provider` and if that fails, try `memcpy_to` on the `dst provider`.  Alternatively, we could have some API to query whether we can perform a copy between 2 providers but that wouldn't change the logic much.

One open question is how `memcpy_to` / `memcpy_from` should be implemented (how to detect whether we can actually copy the memory from/to other provider). The simplest implementation could just call `get_name` and support some predefined set of providers. We could also introduce some kind of type/tag that we could associate with the provider but I'm not sure if it's useful - in most cases I would assume that provider only supports copying between other providers of the same type + USM providers could support copying between host memory and USM.

### DIFF
--- a/source/common/umf_helpers.hpp
+++ b/source/common/umf_helpers.hpp
@@ -131,6 +131,8 @@ auto memoryProviderMakeUnique(Args &&...args) {
     UMF_ASSIGN_OP(ops, T, purge_lazy, UMF_RESULT_ERROR_UNKNOWN);
     UMF_ASSIGN_OP(ops, T, purge_force, UMF_RESULT_ERROR_UNKNOWN);
     UMF_ASSIGN_OP(ops, T, get_name, "");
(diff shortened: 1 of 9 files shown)


pbalcer:
For now, I think we should just do the simplest possible thing that could conceivably work. Eventually, we might want to pick the appropriate copying method depending on the topology, HMAT information, and maybe even some micro-benchmarking... But that's still far out.

vinser52:
We discussed it with @igchor yesterday and agreed that this PR is a placeholder for architectural discussion for data movement API and how it should look like.
I understand the use case for GPU->Host and Host->GPU memory copying when a driver is needed but for Host->Host we already have MPI (that does a lot of microarchitectural optimizations), do we want to compete or re-use them? we need to discuss at least with them because they have much more expertise in that domain.

PR #752, opened 2023-07-28
[ur] Fix adapters linker script path on Linux

kbenzie:
Use an absolute path instead of just the filename for the
`ur_adapter_<name>.map` files to resolve `/usr/bin/ld: cannot open
linker script file` errors.


pbalcer (approved):
That's odd, it works on my system and in CI.

aarongreig (approved):

kbenzie:
> That's odd, it works on my system and in CI.

I use `ninja` instead of `make` so wondered if it was that but I also see this when using `make` so it must be something else but I'm drawing a blank.

pbalcer:
> > That's odd, it works on my system and in CI.
> 
> I use `ninja` instead of `make` so wondered if it was that but I also see this when using `make` so it must be something else but I'm drawing a blank.

I don't think this change will break anything, so no point in wasting time. Lgtm.

PR #757, opened 2023-07-31
[ur] fix ctad-maybe-unsupported warning

igchor:
that appeared on intel/llvm CI


igchor:
@kbenzie @pbalcer do we want to enable this warning in UR repo as well?

PR #759, opened 2023-08-01
[ADAPTERS] update to latest from sycl

pbalcer:
There's no nightly tag yet that contains the latest UR update on the SYCL side, so I'm creating this as draft with `web/sycl-latest-good` tag.


pbalcer:
I also tried adding the `native_cpu` backend, but it still has non-trivial dependencies to SYCL.

kbenzie (approved):

veselypeta (approved):

PR #761, opened 2023-08-01
[UR] Verify that every source file includes an appropriate license 

veselypeta:
This PR:
* Adds a simple check in CMake to check that every source file includes an appropriate comment listing the copyright and license notice.
* The license should appear at the top of each file, so Cmake only reads the first 300 bytes in which it should find the keyword "SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception"

### DIFF
--- a/cmake/helpers.cmake
+++ b/cmake/helpers.cmake
@@ -146,14 +146,3 @@ function(FetchContentSparse_Declare name GIT_REPOSITORY GIT_TAG GIT_DIR)
         WORKING_DIRECTORY ${content-build-dir})
     FetchContent_Declare(${name} SOURCE_DIR ${content-build-dir}/${GIT_DIR})
 endfunction()
-
-
-# Checks that every input file includes an appropriate license notice.
-function(VerifyLicence)
-    foreach(file ${ARGN})
-        file(READ ${file} file_contents LIMIT 300)
-        if(NOT "${file_contents}" MATCHES "SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception")
-            message(FATAL_ERROR "${file} does not contain an appropriate license comment.")
-        endif()
-    endforeach()
-endfunction()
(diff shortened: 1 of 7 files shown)


cmake/helpers.cmake
```diff
@@ -128,3 +128,14 @@ function(FetchContentSparse_Declare name GIT_REPOSITORY GIT_TAG GIT_DIR)
...
     FetchContent_Declare(${name} SOURCE_DIR ${content-build-dir}/${GIT_DIR})
 endfunction()
+
+
+# Checks that every input file includes an appropriate license notice.
+function(VerifyLicence)
+    foreach(file ${ARGN})
+        file(READ ${file} file_contents LIMIT 300)
+        if(NOT "${file_contents}" MATCHES "SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception")
+            message(FATAL_ERROR "${file} does not contain an appropriate license comment.")
```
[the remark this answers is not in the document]
veselypeta: I've changed to be a target `verify-licenses` - which will run in CI

pbalcer (approved):

scripts/verify_license.py
```diff
@@ -0,0 +1,28 @@
...
+ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""
+
+import sys
+import argparse
+
+def verify_file_has_license(file):
+    with open(file, 'r') as in_file:
+        input = in_file.read(300)
```
kbenzie, line 15: nit: the `input` variable shadows the Python [`input()`](https://docs.python.org/3/library/functions.html#input) builtin, `contents` might be a better name.

kbenzie (approved):
1 minor comment, not a blocker

scripts/verify_license.py
veselypeta: 👍

PR #766, opened 2023-08-02
Add UR_DEVICE_INFO_ESIMD_SUPPORT

sarnex:
This will be used by DPC++ PI to determine if a device supports the DPC++ ESIMD extension.


kbenzie:
Hi @sarnex, thanks for the contribution. Is this change required for 2024.0 at all?

sarnex:
@kbenzie Hi, sure, thanks, this is my first contribution to UR so hopefully I did it right. 

This is not required for 2024.0. 

I'll convert the PR from a draft to a real PR once CI passes.

kbenzie:
> @kbenzie Hi, sure, thanks, this is my first contribution to UR so hopefully I did it right.

Looks like you've followed the contribition guide which is much appriciated 🙂 

> This is not required for 2024.0.

That's good to know, thanks.

> I'll convert the PR from a draft to a real PR once CI passes.

We'll probably hold off merging a little until we've got all our 2024.0 ducks in a row.

sarnex:
CI passed so this should be ready for review!

scripts/core/exp-esimd.yml
```diff
@@ -0,0 +1,23 @@
...
+--- #--------------------------------------------------------------------------
+type: enum
+extend: true
+typed_etors: true
+desc: "Extension enums to $x_device_info_t to support ESIMD."
+name: $x_device_info_t
+etors:
+    - name: ESIMD_SUPPORT_EXP
+      value: "0x2010"
+      desc: "[$x_bool_t] returns true if the device supports ESIMD"
```
kbenzie, line 23: Do you expect there will be any other additions relating to device ESIMD support in future, such as additional entry-points or extra handle creation data?

sarnex: Thanks for the review! I don't know of any plans or possible features in ESIMD that would require any UR changes. I think it's very unlikely we will add anything else, but if there is something to do now that would prevent some nightmare in the future if we do end up adding something, I'm happy to do it.

kbenzie: Okay, since there's no other addidions currently planned I think this device info query could probably be added directly to `scripts/core/device.yml` instead of using the experimental feature mechanism. So far we've only used it for features which are likley to have a lot of API churm before becoming stable.

sarnex: Okay thanks, let me make that change.

kbenzie (approved):

veselypeta (approved):
LGTM

Issue #738, opened 2023-07-20
Add CTS tests that pass a pool handle to the USM alloc entry points.


aarongreig:
Potentially this could include checking whether `UR_USM_POOL_FLAG_ZERO_INITIALIZE_BLOCK` is respected but at the very least we should add a basic smoke test for each of the allocs.

PR #768, opened 2023-08-02
Add parameterization to USM alloc tests for testing pools.

aarongreig:
Also fix a couple of minor issues in tests:
* check correct code for invalid size in GetMemAllocInfo test
* fix shadowing in urUSMPoolTestWithParam fixture
* ensure we allocate with a pool for UR_USM_ALLOC_INFO_POOL GetMemAllocInfo test

Addresses #738

### DIFF
--- a/test/conformance/testing/CMakeLists.txt
+++ b/test/conformance/testing/CMakeLists.txt
@@ -4,7 +4,11 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 add_ur_library(ur_testing STATIC 
-    source/utils.cpp)
+    source/utils.cpp
+    source/fixtures.cpp)
 target_include_directories(ur_testing PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
-target_link_libraries(ur_testing PRIVATE gtest_main unified-runtime::headers)
+target_link_libraries(ur_testing PRIVATE
+    gtest_main
+    ${PROJECT_NAME}::common
+    ${PROJECT_NAME}::headers)
 add_library(${PROJECT_NAME}::testing ALIAS ur_testing)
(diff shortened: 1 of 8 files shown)


test/conformance/testing/CMakeLists.txt
```diff
@@ -4,7 +4,11 @@
...
 
 add_ur_library(ur_testing STATIC 
-    source/utils.cpp)
+    source/utils.cpp
+    source/fixtures.cpp)
 target_include_directories(ur_testing PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
-target_link_libraries(ur_testing PRIVATE gtest_main unified-runtime::headers)
+target_link_libraries(ur_testing PRIVATE
+    gtest_main
+    ${PROJECT_NAME}::common
```
omarahmed1111, line 12: Just a question as I am a little confused by this, but which common files should this link?

test/conformance/testing/include/uur/fixtures.h
```diff
@@ -839,11 +842,16 @@ struct urUSMDeviceAllocTestWithParam : urQueueTestWithParam<T> {
 
     void TearDown() override {
         ASSERT_SUCCESS(urUSMFree(this->context, ptr));
+        if (pool) {
+            ASSERT_SUCCESS(urUSMPoolRelease(pool));
+        }
```
omarahmed1111, line 847: [NIT] I guess we could check for `use_pool` and clear it to false here in addition to calling `urUSMPoolRelease`. I think we could also assert for a case where `use_pool` is true while `pool` is nullptr as that would mean something gone wrong.

omarahmed1111 (approved):
LGTM

veselypeta (approved):

test/conformance/testing/CMakeLists.txt
aarongreig: it's because fixtures.hpp includes [checks.h](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/testing/include/uur/checks.h#L11) which includes [ur_params.hpp](https://github.com/oneapi-src/unified-runtime/blob/main/source/common/ur_params.hpp). Previously fixtures.hpp wasn't actually included in the `ur_testing` target, it isn't used in utils.cpp

test/conformance/testing/include/uur/fixtures.h
aarongreig: stuck an assertion on use_pool in there

test/conformance/testing/CMakeLists.txt
omarahmed1111: Ah, that make sense, Thanks!

igchor (approved):

PR #784, opened 2023-08-08
[UR] Check for if handle when type is complete

veselypeta:
This PR:

* cherry-picks the commit from #776  into the adapters branch


veselypeta:
Actually this won't work because of the generator check. It might be possible to merge upto the latest commit provided it does not affect the cuda/hip adapters...

pbalcer:
I'm worried it might conflict with a lot of things... If this is blocking you,  for now I'd suggest just leaving `is_type_complete_v` with a comment. We can fix it later once sycl pulls in the newer ur tag.

Issue #789, opened 2023-08-08
Remove `oprator<<` overload for definitions of handle types


veselypeta:
In #790 we introduce overloads for the definite objects required by unified runtime. This has been fixed in #776 , but cannot be updated in the adapters branch yet until the code in intel/llvm is updated to reflect a more recent tag of unified runtime. When `adapters` is ready to merge the changes from #776 we should delete the overloads introduced in #790 .

PR #763, opened 2023-08-02
[ADAPTERS] HIP specific CTS

veselypeta:
Adds HIP specific CTS for native handle entry-points

1. urContextGetNativehandle
2. urDevicGetNativeHandle
3. urEventGetNativeHandle

PR #790, opened 2023-08-08
[UR][CTS][CUDA] Port CUDA specific PI unittests to UR

veselypeta:
This PR:

* Ports all useful CUDA specific unittests from intel/llvm into UR
* Generates implementations for the overloads for the complete unified runtime types. #789 will be used to track to ensure this change is reverted until we merge in `main`.


test/conformance/adapters/cuda/cuda_urDeviceCreateWithNativeHandle.cpp
```diff
@@ -7,7 +7,7 @@
 
 using urCudaDeviceCreateWithNativeHandle = uur::urPlatformTest;
 
-TEST_F(urCudaDeviceCreateWithNativeHandle, Success) {
+TEST_F(urCudaDeviceCreateWithNativeHandle, DISABLED_Success) {
```
omarahmed1111, line 10: As I understand this test `urDeviceCreateWithNativeHandle` so why we need `cuda_` prefix in the file name? and why we need `cuda` part in the test name, I mean can't we name the test `urDeviceCreateWithNativeHandle` only as testing this entry would mean we already will pass it a cuda handle to test it?

veselypeta: You're right about the `cuda_` prefixes in filenames. It's on my list of stuff to do - see [here](https://github.com/oneapi-src/unified-runtime/pull/763#discussion_r1282042701).

The problem with removing `cuda` from the test name is that there already is a test with the name `urDeviceCreateWithNativeHandleTest` as a generic conformance test which it will conflict with.

test/conformance/adapters/cuda/context_tests.cpp
```diff
@@ -0,0 +1,212 @@
...
+                                     nullptr, &buffer));
+
+    // check that context is now the active cuda context
+    ASSERT_SUCCESS_CUDA(cuCtxGetCurrent(&current));
+    ASSERT_EQ(current, context->get());
+
+    ASSERT_SUCCESS(urQueueRelease(queue));
+    ASSERT_SUCCESS(urContextRelease(context));
+
+    // check that the context was cleaned up, the old context will be restored
```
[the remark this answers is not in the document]
veselypeta: Yeah that should be removed. The pi unittests included a check to see if the current cuda context was no longer set, but this is no longer the case as the context is managed by a device.

veselypeta:
I feel like having separate targets for adapter specific tests would be easier to manage, especially once we start adding more adapters. I think that #ifdefing tests would introduce lots of complicated CMake to conditionally link in each adapter and we would have to build every configuration to test every adapter.

test/conformance/adapters/cuda/cuda_fixtures.h
```diff
@@ -19,6 +19,15 @@ struct ResultCuda {
...
     CUresult value;
 };
+
+inline void clearCudaContext() {
+    CUcontext context = nullptr;
+    do {
+        cuCtxSetCurrent(nullptr);
+        cuCtxGetCurrent(&context);
+    } while (context != nullptr);
+}
```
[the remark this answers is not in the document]
veselypeta: Turns out this was not necessary so I've removed it from the tests. 👍

fabiomestre (approved):

test/conformance/adapters/cuda/kernel_tests.cpp
```diff
@@ -0,0 +1,270 @@
...
+// See LICENSE.TXT
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "kernel.hpp"
+#include <uur/fixtures.h>
+
+using cudaKernelTest = uur::urQueueTest;
+UUR_INSTANTIATE_DEVICE_TEST_SUITE_P(cudaKernelTest);
+
+inline constexpr size_t NumberOfImplicitArgsCUDA = 1;
```
callumfare, line 12: Nit: A comment that explains that this argument is used for the implicit global offset could be helpful

callumfare (approved):
LGTM! For what it's worth I agree with keeping these tests separate from the others, since they reference adapter internal details they're semantically different from the main conformance testing.

PR #793, opened 2023-08-09
Make urInit and urTearDown loader-only

callumfare:
- Make `urInit` and `urTeardown` loader-only, rename them to `urLoaderInit` and `urLoaderTearDown`
- Split `runtime.yml` into `loader.yml` and `adapter.yml`
- Add teardown method to layers so they can still cleanly tidy up any state, as they will no longer intercept the loader-only `urLoaderTearDown`.
- Correct the validation layer's tracking of adapter handles, as `urAdapterGet` may either create the initial reference or increment it. 
- Track the number of in-use adapters in the validation layer. When no more adapters are in use, report all leaked handles. This allows reporting of leaks even if a user does not call `urLoaderTeardown`


veselypeta (approved):

scripts/parse_specs.py
```diff
@@ -872,6 +872,7 @@ def parse(section, version, tags, meta, ref):
     specs = []
 
     files = util.findFiles(path, "*.yml")
+    files.insert(0, files.pop(files.index(os.path.join(path, 'common.yml'))))
```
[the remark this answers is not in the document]
callumfare: Yeah that's definitely cleaner, thanks!

examples/collector/collector.cpp
```diff
@@ -50,16 +50,17 @@ std::ostream &operator<<(std::ostream &os,
 
 /**
  * A map of functions that format the parameters and arguments for each UR function.
- * This example only implements a handler for one function, `urInit`, but it's
```
pbalcer, line 53: this patch makes it impossible to trace this function, so this example doesn't work with `urLoaderInit`. It should probably be changed to something else that is traceable.

pbalcer (commented):
Is this being targeted for the upcoming release, so 0.7.1 (or something like that)?

source/loader/ur_lib.cpp
```diff
@@ -174,4 +182,14 @@ ur_result_t urLoaderConfigEnableLayer(ur_loader_config_handle_t hLoaderConfig,
     hLoaderConfig->enabledLayers.insert(pLayerName);
     return UR_RESULT_SUCCESS;
 }
+
+ur_result_t urLoaderTearDown(void *pParams) {
+    if (!pParams) {
```
aarongreig, line 187: do we know what the intent of this parameter was? since we don't seem to be using it and now no adapters can use it maybe we should look into getting rid of it or making it optional

aarongreig (approved):

source/loader/ur_lib.cpp
callumfare: I'm not sure what the parameter was meant to be for - @pbalcer do you know?
I suspect it's not needed, the documentation for `ur_tear_down_params_t` mentions parameters for callbacks, but I don't think any teardown callbacks actually exist

examples/collector/collector.cpp
pbalcer: As part of the recent loader changes, there's now one additional env variable to set `UR_ENABLE_LAYERS=UR_LAYER_TRACING`. README needs updating.

This runs in CI, and I've noticed because you had to change the [hello_world output](https://github.com/oneapi-src/unified-runtime/pull/793/files#diff-beee61f5a9e76bfc5266df9381f10343c24bda0775f9d612fc0c2e29d58619d4) ;)

source/loader/ur_lib.cpp
pbalcer: Maybe it's in case we need to add an argument for teardown in future releases?
Generic callbacks were removed sometime back. I vote we remove it.

examples/collector/collector.cpp
callumfare: I didn't even notice that hello_world.output belonged to this test, it's fixed now! README is also updated.

source/loader/ur_lib.cpp
callumfare: I've removed it, we can always re-add it in future if an actual use becomes apparent

scripts/core/CONTRIB.rst
```diff
@@ -137,8 +137,8 @@ defined within them, with the following exceptions:
     enumerations, and structure type enumerations.
 *   `scripts/core/enqueue.yml`_ defines commands which can be enqueued on a
     queue object.
-*   `scripts/core/runtime.yml`_ defines global symbols pertaining to
-    initialization and tear down of the entire runtime.
+*   `scripts/core/loader.yml`_ defines global symbols pertaining to
```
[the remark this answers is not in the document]
callumfare: The line at the top of this says:
> Most of the files are named after the API object who's interface is
> defined within them, with the following exceptions

The adapter entry points all relate to the `ur_adapter_handle_t` object so I don't think it's an exception to that

kbenzie:
> Is this being targeted for the upcoming release, so 0.7.1 (or something like that)?

I don't think so, its too big a change for 2024.0 which 0.7.x is primarily for.

pbalcer (approved):

scripts/core/adapter.yml
```diff
@@ -9,177 +9,9 @@
 #
 --- #--------------------------------------------------------------------------
 type: header
-desc: "Intel $OneApi Unified Runtime APIs for Runtime"
+desc: "Intel $OneApi Unified Runtime APIs for Adapter"
 ordinal: "0"
```
kbenzie, line 13: currently adapter comes before loader in the rendered contents menu, changing the ordinal should fix that.

```suggestion
ordinal: "1"
```

test/loader/platforms/null_platform.match
```diff
@@ -1,3 +1,3 @@
-<TEST>[INFO]: urInit succeeded.
+<TEST>[INFO]: urLoaderInit succeeded.
 <TEST>[INFO]: urPlatformGet found 1 platforms
 <TEST>[INFO]: Found UR_PLATFORM_NULL
```
kbenzie, line 3: looks like this file could use a trailing new line

kbenzie (approved):

callumfare:
Assuming this passes CI this should be ready to merge now - not sure if we're waiting for anything else before that happens

pbalcer (approved):

PR #731, opened 2023-07-19
Test parts of API with libFuzzer.

PatKamin:
Add a fuzz test generating API calls. Related to #566 .

Issue #566, opened 2023-06-05
Perform Fuzz Testing


wlemkows:
Fuzz testing is a common type of automated testing where randomized inputs are injected into the interfaces of a system.
It includes testing the system with invalid, negative, malformed or unexpected inputs.

Any interface can be a target for fuzzing. Fuzzing can quickly identify areas where unsanitized input is improperly handled and hence is extremely effective in finding critical security issues without needing to fully understand the mechanics of the underlying software.
It is worth using a fuzzing tool to implement such tests, for example LibFuzzer for C/C++.

pbalcer:
Partially done in #795. Since UR is a stateful API with *many* entry points, fuzz testing is a bit awkward. We followed libfuzzer [docs](https://github.com/google/fuzzing/blob/master/docs/structure-aware-fuzzing.md#fuzzing-stateful-apis) recommendations for stateful APIs, and @PatKamin implemented a fuzzing solution similar to that of [gRPC](https://github.com/google/fuzzing/blob/master/docs/structure-aware-fuzzing.md#example-grpc-api-fuzzer).
The fuzz tests will run with known-working fuzz corpus for every change, and there is a nightly job for the fuzzer to run for longer. This is to prevent PR checks from being flaky due to the non-deterministic nature of fuzzing.

The current implementation covers the core APIs, but some gaps still need to be addressed. Not all arguments in the existing functions are also subject to fuzzing for now.

I'm closing this issue, and we will create new ones as needed for additional fuzzer funcionality.

PR #795, opened 2023-08-10
Test parts of API with libFuzzer.

PatKamin:
Add a fuzz test generating API calls. Related to #566 . 
Replaces #731 (changed the target branch to adapters).


test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+    UR_PROGRAM_CREATE_WITH_IL,
+    UR_MAX_FUZZER_API_CALL,
+};
+
+typedef struct FuzzerInput {
+    uint8_t *data_ptr;
+    size_t data_size;
+} FuzzerInput;
+
+typedef struct TestState {
```
pbalcer, line 41: do you need to use typedef like in C?

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+        return -1;
+    }
+    *out_data = *input->data_ptr;
+    input->data_ptr += out_data_size;
+    input->data_size -= out_data_size;
+
+    return 0;
+}
+
+int init_random_data(TestState *state) {
```
pbalcer, line 80: it's not really random, right? it's just reading N values from the input to create initial state.

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+    if (state->input.data_size < 5) {
```
pbalcer, line 81: hm, wouldn't it be better to check result of `get_next_input_data`? otherwise, if the size of these fields changes, this will not be correct.

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+        return -1;
+    }
+    get_next_input_data(&state->input, &state->platform_num);
+    get_next_input_data(&state->input, &state->device_num);
+    get_next_input_data(&state->input, &state->context_num);
+    get_next_input_data(&state->input, &state->device_type_fuzz);
+    if (state->device_type_fuzz < 1 || state->device_type_fuzz > 7) {
```
pbalcer, line 88: `UR_DEVICE_TYPE_MAX` or something?

test/fuzz/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+    add_ur_executable(${TEST_TARGET_NAME}
+        ${ARGN})
+    target_link_libraries(${TEST_TARGET_NAME}
+        PRIVATE
+        ${PROJECT_NAME}::loader
+        ${PROJECT_NAME}::headers
+        ${PROJECT_NAME}::common
+        -fsanitize=fuzzer -fprofile-instr-generate -fcoverage-mapping)
+    add_test(NAME ${TEST_TARGET_NAME}
+        COMMAND ${TEST_TARGET_NAME} -max_total_time=600 -seed=1 -shrink=1 -verbosity=1
```
pbalcer, line 17: We should have a set of predefined input data files we know pass for CI, and a separate setup where we run the fuzzer for longer periods of time. Even with a fixed seed, running this on every build might make the CI flaky.

test/fuzz/urFuzz.cpp
```diff
@@ -0,0 +1,377 @@
...
+        return -1;
+    }
+
+    ur_usm_pool_handle_t pool;
+    ur_usm_pool_desc_t pool_desc{UR_STRUCTURE_TYPE_USM_POOL_DESC, nullptr,
+                                 UR_USM_POOL_FLAG_ZERO_INITIALIZE_BLOCK};
+    ur_result_t res =
+        urUSMPoolCreate(state.contexts[state.context_num], &pool_desc, &pool);
+    if (res == UR_RESULT_SUCCESS) {
+        pool_allocs[pool] = {};
```
pbalcer, line 101: pool_allocs.insert(pool) ?

test/fuzz/utils.hpp
pbalcer, line 41: wouldn't it be simpler to make the util functions part of this struct?

test/fuzz/utils.hpp
PatKamin: It's random in a sense that the inupt is generated by a fuzzer. Of course, the fuzzer will probably stick to some values more than the others. Should I replace it with generating random values with the help of a random header?

test/fuzz/utils.hpp
PatKamin: I replaced multiple checks after each and every `get_next_input_data` with this look-ahead check for all values at once. I will change it though, as one have to remember to change this value every time a "random" var is added/removed.

test/fuzz/CMakeLists.txt
PatKamin: OK, and in the predefined input the fuzzer should provide just functions parameters and not generate API calls on the fly, right?

test/fuzz/utils.hpp
pbalcer: No, my point is that the name isn't really accurate. I'd just call it init_state or something.

test/fuzz/CMakeLists.txt
pbalcer: yes, it should be a file that contains all the bytes you use in your program to create some sequence of operations.

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+    uint8_t platform_num;
+    uint8_t device_num;
+    uint8_t context_num;
+    uint8_t device_type_fuzz;
+} TestState;
+
+//////////////////////////////////////////////////////////////////////////////
+template <typename T> int get_next_input_data(FuzzerInput *input, T *out_data) {
+    size_t out_data_size = sizeof(out_data);
```
igchor, line 69: shoudln't it be `sizeof(*out_data)`?

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+    uint32_t num_devices;
+
+    uint8_t platform_num;
+    uint8_t device_num;
+    uint8_t context_num;
+    uint8_t device_type_fuzz;
+} TestState;
+
+//////////////////////////////////////////////////////////////////////////////
+template <typename T> int get_next_input_data(FuzzerInput *input, T *out_data) {
```
igchor, line 68: I'm not sure if you need to write this by yourself. There are already some untils in llvm to split the fuzzing input into multiple 'sub-inputs': https://github.com/google/fuzzing/blob/master/docs/split-inputs.md For me it looks you can just use FuzzedDataProvider  (https://raw.githubusercontent.com/llvm/llvm-project/main/compiler-rt/include/fuzzer/FuzzedDataProvider.h) by calling ConsumeIntegral()/ConsumeEnum, etc.

igchor: And even if you need to write something by yourself I would recommend just wrapping FuzzedDataProvider.ConsumeBytes()

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+    return 0;
+}
+
+int get_next_api_call(FuzzerInput *input) {
+    uint8_t next_api_call;
+    if (get_next_input_data(input, &next_api_call) != 0) {
+        return -1;
+    }
+    return next_api_call % UR_MAX_FUZZER_API_CALL;
```
igchor, line 101: See my comment above, you can probably just use FuzzedDataProvider::ConsumeEnum<ApiCall> or something like this. (where ApiCall is an enum describing the call)

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+        state->devices[0] == nullptr) {
+        return false;
+    }
+
+    return true;
+}
+
+bool check_context_exists(const TestState *state) {
+    if (state->contexts.empty() ||
+        state->context_num >= state->contexts.size() ||
```
igchor, line 115: instead of checking `state->context_num >= state->contexts.size()` I think it would be better to just generate such value that is < size(). Unless, we want to fuzz cases where this number is bigger but perhaps that should be just a special-test case.

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+        state->contexts[0] == nullptr) {
+        return false;
+    }
+
+    return true;
+}
+
+bool check_pool_exists(
```
igchor, line 123: Same comment here, can't we just generate a pool_num value that is < map.size()?

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+int get_next_api_call(FuzzerInput *input) {
+    uint8_t next_api_call;
+    if (get_next_input_data(input, &next_api_call) != 0) {
+        return -1;
+    }
+    return next_api_call % UR_MAX_FUZZER_API_CALL;
+}
+
+bool check_device_exists(const TestState *state) {
```
igchor, line 104: Same comment here (as below).

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+int load_kernel_source(std::vector<char> &binary_out) {
+    std::string source_path = KERNEL_IL_PATH;
+
+    std::ifstream source_file;
+    source_file.open(source_path,
+                     std::ios::binary | std::ios::in | std::ios::ate);
+    if (!source_file.is_open()) {
+        std::cerr << "Failed to open a kernel source file: " << source_path
+                  << std::endl;
+        return -1;
```
igchor, line 170: why not just throw an exception and return a vector? there are probably already some unhandled exceptions in this code (perhaps from open())

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+    source_file.close();
+
+    binary_out = std::vector<char>(std::move(device_binary));
+
+    return 0;
+}
+
+void cleanup(TestState *state) {
+    urLoaderConfigRelease(state->config);
```
igchor, line 192: nit: Can we just wrap all the things that need to be releaed/freed/closed in unique_ptrs instead of implementing this function?

PatKamin: Done

test/CMakeLists.txt
```diff
@@ -24,3 +24,6 @@ add_subdirectory(unit)
 if(UR_BUILD_TOOLS)
   add_subdirectory(tools)
 endif()
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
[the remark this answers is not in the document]
PatKamin: Thanks, done

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+    uint8_t platform_num;
+    uint8_t device_num;
+    uint8_t context_num;
+
+    TestState(std::unique_ptr<FuzzedDataProvider> data_provider)
+        : data_provider(std::move(data_provider)) {}
+
+    template <typename IntType> int get_next_input_data(IntType *data) {
+        if (data_provider->remaining_bytes() < sizeof(IntType)) {
```
igchor, line 101: I'm wondering if checking for remaining_bytes here and returning -1 makes sense. Why not just always call `consume*`, even if there are no bytes left? From the documentation, it will just return the same value every time (e.g. 0 for int). 

I don't know which approach is better for coverage but it certainly seems simpler as we don't need to deal with `-1`.

But feel free to ignore this comment. We can also experiment later with what gives better results.

test/fuzz/utils.hpp
```diff
@@ -0,0 +1,218 @@
...
+struct Context {
+    ur_context_handle_t handle;
+    Pools host_pools;
+    Pools device_pools;
+    Allocs no_pool_host_allocs;
+    Allocs no_pool_device_allocs;
+
+    Context(ur_context_handle_t handle) : handle(handle) {}
+    ~Context() {
+        host_pools.clear();
```
igchor, line 69: nit: no need to call `.clear()`

PatKamin: It's for calling Alloc dtors before removing Context, avoiding use-after-free.

test/fuzz/CMakeLists.txt
PatKamin: Done

test/fuzz/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    target_compile_definitions(${TEST_TARGET_NAME} PRIVATE -DKERNEL_IL_PATH="${UR_CONFORMANCE_DEVICE_BINARIES_DIR}/bar/sycl_spir641.spv")
+    target_include_directories(${TEST_TARGET_NAME} PRIVATE ${UR_CONFORMANCE_DEVICE_BINARIES_DIR})
+
+    add_dependencies(${TEST_TARGET_NAME} generate_device_binaries)
+endfunction()
+
+add_fuzz_test(base fuzz-long
+    -max_total_time=1 -seed=1 -verbosity=1)
+
+set(CORPUS_DIR ${CMAKE_CURRENT_SOURCE_DIR}/corpus)
```
pbalcer, line 34: maybe not for now, but might be useful to have something like:
```
foreach(file with some extension in corpus_dir) {
   add_test(...);
}
```

pbalcer (approved):
I've noticed the python test is failing for some reason. Maybe you need to rebase?

PatKamin:
> I've noticed the python test is failing for some reason. Maybe you need to rebase?

Thanks, rebased.

test/fuzz/corpus/README.md
```diff
@@ -0,0 +1,87 @@
+# Corpra for fuzz tests with fixed API calls scenarios
+These corpra contain UR API calls in a predefined order described below.
+All scenarios begin with single calls to urInit() and urAdapterGet().
+
```
lukaszstolarczuk, line 4: could you please add some info how to create new/update existing corpras...? perhaps some link to a guide or something

test/fuzz/urFuzz.cpp
```diff
@@ -0,0 +1,415 @@
+// Copyright (C) 2023 Intel Corporation
+// Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+// See LICENSE.TXT
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
```
lukaszstolarczuk, line 5: could you please add a short description, what is this file's purpose :wink:

.github/workflows/daily.yml
```diff
@@ -0,0 +1,70 @@
+name: Build and test
```
lukaszstolarczuk, line 1: perhaps call the file "nigthly" and here also a different name than "Build and test" (to not duplicate existing workflow name)

.github/workflows/cmake.yml
```diff
@@ -63,6 +65,10 @@ jobs:
         wget -O ${{github.workspace}}/dpcpp_compiler.tar.gz https://github.com/intel/llvm/releases/download/sycl-nightly%2F20230626/dpcpp-compiler.tar.gz
         tar -xvf ${{github.workspace}}/dpcpp_compiler.tar.gz
 
+    - name: Setup DPC++
```
lukaszstolarczuk, line 68: why do we need this, actually? if it's only needed for fuzzing, please add an `if` here (and perhaps add a comment on the purpose of this)

PatKamin: It's needed for building kernel binaries with sycl. Removed from the path not running fuzztests.

.github/workflows/daily.yml
PatKamin: Done

test/fuzz/urFuzz.cpp
PatKamin: Done

test/fuzz/utils.hpp
PatKamin: This way we would eventually get stuck with multiple calls to urPlatformGet(), as it's the first item in enum representing 0. I prefer to have it this way and terminate the iteration on insufficient data.

.github/workflows/cmake.yml
```diff
@@ -88,6 +88,60 @@ jobs:
...
+        -DUR_USE_UBSAN=ON
+        -DUR_DPCXX=${{github.workspace}}/dpcpp_compiler/bin/clang++
+
+    - name: Generate source from spec, check for uncommitted diff
+      run: cmake --build ${{github.workspace}}/build --target check-generated
+
+    - name: Build
+      run: cmake --build ${{github.workspace}}/build -j $(nproc)
+
+    - name: Fuzz long test
```
lukaszstolarczuk, line 141: fuzz-long? 😉

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,60 @@
+name: Nightly
+
+on:
+  schedule:
+    - cron: '0 23 * * *'
```
lukaszstolarczuk, line 5: For debugging purpose, you can enable it temporarily for PR trigger (to see if it compiles properly)

PatKamin: Sure

test/fuzz/utils.hpp
PatKamin: Done

test/fuzz/utils.hpp
PatKamin: Added TODO

.github/workflows/cmake.yml
PatKamin: Done

test/fuzz/utils.hpp
PatKamin: Added try-catch

lukaszstolarczuk (commented):
👍

test/fuzz/utils.hpp
igchor: Ok, got it. Btw, you can also wrap `ur_context_handle_t` in unique_ptr and then you wouldn't need to implement the dtor at all. But both ways work

igchor (approved):

test/fuzz/corpus/README.md
pbalcer: https://github.com/google/fuzzing/blob/master/tutorial/libFuzzerTutorial.md#seed-corpus

There's still some work to be done to make a good, minimized, seed corpus. But that's work for later. At that point, we should definitely expand this documentation.

pbalcer (approved):

PR #796, opened 2023-08-10
Umf critnib

igchor:


source/common/unified_malloc_framework/src/memory_tracker.c
```diff
@@ -9,112 +9,117 @@
...
-        }
+umf_memory_tracker_handle_t umfMemoryTrackerGet(void) {
+    return (umf_memory_tracker_handle_t)TRACKER;
+}
 
-        auto ret =
-            map.try_emplace(reinterpret_cast<uintptr_t>(ptr), size, pool);
-        return ret.second ? UMF_RESULT_SUCCESS : UMF_RESULT_ERROR_UNKNOWN;
-    }
+struct tracker_value {
```
vinser52, line 55: should we use `_t` suffix for the type? I mean `tracker_value_t`

source/common/unified_malloc_framework/src/memory_tracker.c
```diff
@@ -9,112 +9,117 @@
...
 
-        map.erase(reinterpret_cast<uintptr_t>(ptr));
+    struct tracker_value *value =
+        (struct tracker_value *)malloc(sizeof(struct tracker_value));
+    value->pool = pool;
+    value->size = size;
 
-        // TODO: handle removing part of the range
-        (void)size;
+    int ret = critnib_insert((critnib *)hTracker, (uintptr_t)ptr, value, 0);
```
vinser52, line 70: Why can we not use `void*` as a key and should use `uintptr_t` instead?

test/unified_malloc_framework/memoryProviderAPI.cpp
```diff
@@ -23,7 +23,8 @@ TEST_F(test, memoryProviderTrace) {
 
     size_t call_count = 0;
 
-    auto ret = umfMemoryProviderAlloc(tracingProvider.get(), 0, 0, nullptr);
+    void *ptr;
```
kswiecicki, line 26: Is this and `*ptr = NULL;` in `provider.c` needed?

source/common/unified_malloc_framework/src/memory_tracker.c
```diff
@@ -9,112 +9,117 @@
...
+    }
+    return TRUE;
+}
+#elif defined(_WIN32)
+struct tracker_t {
+    tracker_t() { map = critnib_new(); }
+    ~tracker_t() { critnib_remove(map); }
+    critnib *map;
+};
+tracker_t TRACKER_INSTANCE;
```
kswiecicki, line 41: It seems you need an explicit `struct` keyword before the type.

source/common/unified_malloc_framework/src/memory_tracker.c
```diff
@@ -9,112 +9,117 @@
...
+umf_memory_tracker_handle_t umfMemoryTrackerGet(void) {
+    return (umf_memory_tracker_handle_t)TRACKER;
+}
 
-        auto ret =
-            map.try_emplace(reinterpret_cast<uintptr_t>(ptr), size, pool);
-        return ret.second ? UMF_RESULT_SUCCESS : UMF_RESULT_ERROR_UNKNOWN;
-    }
+struct tracker_value {
+    void *pool;
```
vinser52, line 56: why are we using `void*` to store pool handle instead of `umf_memory_pool_handle_t`?

test/unified_malloc_framework/memoryProviderAPI.cpp
igchor: Yes, this is because the implementation handles NULL pointers differently now (so we need to special case it in the tracker for both alloc and free).

source/common/unified_malloc_framework/src/memory_tracker.c
igchor: Done

source/common/unified_malloc_framework/src/memory_tracker.c
igchor: Done

source/common/unified_malloc_framework/src/memory_tracker.c
vinser52: Sorry, I forgot that it is not a C++ where we can use the power of metaprogramming. :)

pbalcer:
Sorry, I updated your PR incorrectly (through a merge, not a rebase). Can you rebase manually please?

source/common/unified_malloc_framework/src/utils/utils.h
```diff
@@ -0,0 +1,111 @@
...
+    return (unsigned char)ret;
+}
+static __inline unsigned char util_mssb_index(long long value) {
+    unsigned long ret;
+    _BitScanReverse64(&ret, value);
+    return (unsigned char)ret;
+}
+
+// There is no good way to do atomic_load on windows...
+#define util_atomic_load_acquire(object, dest)                                 \
```
pbalcer, line 44: C++ has atomic loads implemented directly, so we could change these to be declarations (should be fine for performance if we enable lto) and implement it using the C++ apis in the .cpp file.

source/common/unified_malloc_framework/src/utils/utils.h
```diff
@@ -0,0 +1,111 @@
...
+#define util_mssb_index(x) ((unsigned char)(63 - __builtin_clzll(x)))
+#define util_atomic_load_acquire(object, dest)                                 \
+    __atomic_load(object, dest, memory_order_acquire)
+#define util_atomic_store_release(object, desired)                             \
+    __atomic_store_n(object, desired, memory_order_release)
+#define util_atomic_increment(object)                                          \
+    __atomic_add_fetch(object, 1, __ATOMIC_ACQ_REL)
+#endif
+
+#define Malloc malloc
```
pbalcer, line 64: I think it's fine for now, but, since we are moving the entire critnib code here, we can just modify it to use ur/umf "native" methods for allocating/logging/asserting.

source/common/unified_malloc_framework/src/memory_tracker.c
```diff
@@ -9,112 +9,97 @@
...
-        auto size = it->second.first;
-        auto pool = it->second.second;
-
-        if (intptr >= address && intptr < address + size) {
-            return pool;
-        }
+    free(value);
 
-        return nullptr;
+    if (ret == ENOMEM) {
```
kswiecicki, line 58: There's also a possible `EEXIST` return value. Maybe we could return `UMF_RESULT_ERROR_INVALID_ARGUMENT` and log that the provided `ptr` argument already exists.

source/common/unified_malloc_framework/src/memory_tracker_windows.cpp
kswiecicki, line 1: You have changed `memory_tracker.cpp` to `memory_tracker.c`, shouldn't this file also have `.c` extension?

source/common/unified_malloc_framework/src/utils/utils.h
igchor: In C++, you need to declare the variable as std::atomic (similarly as with _Atomic in C, and by the way, _Atomic is not supported by MSVC) to use the atomic operations. Only in C++ 20 you have https://en.cppreference.com/w/cpp/atomic/atomic_ref which can be used on any variable.

source/common/unified_malloc_framework/src/memory_tracker_windows.cpp
igchor: Right now I need it to be a C++ file because I use C++ global variables in case we build UMF as static library to do cleanup. We probably could use some windows-specific API here as well but we plan to only support shared lib in the future so this soon won't be necessary.

source/common/unified_malloc_framework/src/memory_tracker.c
igchor: Yeah, I left a comment that we should add some logging there. The problem is that right now we don't have a UMF logging framework and we shouldn't use the UR-one since those files will need to be moved to a different repo.

igchor:
> Sorry, I updated your PR incorrectly (through a merge, not a rebase). Can you rebase manually please?

Done.

PR #797, opened 2023-08-11
[loader] Add use of threads so -lpthread works when building

hdelan:
This patch makes `libur_loader.so` force link against `pthreads` by adding some trivial use of threads. Linking against pthreads is necessary for `std::call_once` to work for gcc 9.4, but it doesn't happen automatically, as even when specifying `-lpthread`, the linker doesn't know that pthreads are necessary for `call_once` (pthreads isn't necessary for call_once for other versions of gcc) so it may elide the linking.

### DIFF
--- a/source/loader/ur_libapi.cpp
+++ b/source/loader/ur_libapi.cpp
@@ -7743,4 +7743,8 @@ ur_result_t UR_APICALL urUsmP2PPeerAccessGetInfoExp(
     return exceptionToResult(std::current_exception());
 }
 
+void uselessFunc() {
+    std::thread NullThread([](){});
+}
+
 } // extern "C"
--- a/source/loader/ur_lib.hpp
+++ b/source/loader/ur_lib.hpp
@@ -26,6 +26,7 @@
 #include <atomic>
 #include <mutex>
 #include <set>
+#include <thread>
 #include <vector>
 
 struct ur_loader_config_handle_t_ {


hdelan:
I am using gcc-9.4.0 and in this version the definition of `std::call_once` requires the binary to be linked with pthreads. If it isn't linked with pthreads it will still compile OK and error out when it is called:

```
 $ g++ call_once.cpp
$ ./a.out
terminate called after throwing an instance of 'std::system_error'
  what():  Unknown error -1
[1]    184552 abort (core dumped)  ./a.out
```
This bug is documented here https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55394

So in the cmake for ur_adapter, Threads are required. But this translates to -lpthread when linking. The linker then seems to do some scanning of the module, and if it only sees call_once and no other mentions of threads, it will not link to pthreads. (presumably because other versions of gcc don't require linking with pthreads in order for call_once to work)(Aside: a heavier version of -lpthread is -pthread which will force link with pthreads as well as define some macros etc and make it work)

```
$ g++ call_once_no_threads.cpp -lpthread
$ ldd a.out
    linux-vdso.so.1 (0x00007fff2b147000)
    libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f44e9a1a000)
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f44e9828000)
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f44e96d9000)
    /lib64/ld-linux-x86-64.so.2 (0x00007f44e9c24000)
    libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f44e96be000)
$ ./a.out
terminate called after throwing an instance of 'std::system_error'
  what():  Unknown error -1
[1]    184552 abort (core dumped)  ./a.out
```

Whereas if we have some use of threads we get:

```
$ g++ call_once_with_a_thread.cpp -lpthread
$ ldd a.out
    linux-vdso.so.1 (0x00007fffcb9fc000)
    libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007ff8f71ef000)
    libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007ff8f700d000)
    libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007ff8f6ff2000)
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff8f6e00000)
    /lib64/ld-linux-x86-64.so.2 (0x00007ff8f723d000)
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff8f6caf000)
$ ./a.out
Success!
```
So in my case the libur_adapter.so was not getting linked with pthreads which was causing an error out on call_once
```
$ ldd lib/libur_loader.so
    linux-vdso.so.1 (0x00007ffd0615f000)
    libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fdc9169c000)
    libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fdc914ba000)
    libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fdc9149f000)
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fdc912ad000)
    /lib64/ld-linux-x86-64.so.2 (0x00007fdc91720000)
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fdc9115c000)
```

With this change the new output is:
```
$ ldd lib/libur_loader.so
	linux-vdso.so.1 (0x00007ffc031af000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fc922aa6000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fc922aa0000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fc9228be000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fc9228a3000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc9226af000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc922b48000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc922560000)
```

hdelan:
Ping @ldrumm

source/loader/ur_libapi.cpp
```diff
@@ -7743,4 +7743,8 @@ ur_result_t UR_APICALL urUsmP2PPeerAccessGetInfoExp(
     return exceptionToResult(std::current_exception());
 }
 
+void uselessFunc() {
```
veselypeta, line 7746: Since this file is autogenerated you need to add it to the mako file https://github.com/oneapi-src/unified-runtime/blob/main/scripts/templates/libapi.cpp.mako. You will then need to run the generator with `ninja generate` and commit all the changes.

callumfare:
`ur_libapi.cpp` is auto-generated from a template (with the `generate` target), so the function would have to live in a different file like `ur_api.cpp`.

One other solution to this would be conditionally adding `-pthread` to the loader in cmake if using a known bad version of gcc. I don't know if that's portable across everything that we target. Looking at the comment on the gcc issue, it was fixed by the call_once implementation changing in `GLIBCXX_3.4.29` which corresponds to GCC 11.1.

source/loader/ur_libapi.cpp
veselypeta: More info https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#generating-source if you need

hdelan:
Aha silly me thanks @veselypeta @callumfare

source/loader/ur_libapi.cpp
hdelan: Is this OK now? As far as I can tell the `ur_lib.hpp` isn't autogenerated as well?

source/loader/ur_libapi.cpp
```diff
@@ -7743,4 +7743,10 @@ ur_result_t UR_APICALL urUsmP2PPeerAccessGetInfoExp(
     return exceptionToResult(std::current_exception());
 }
 
+// This uselessFunc is needed to ensure that pthread is linked correctly against
+// ur_loader.so for gcc-9
+void uselessFunc() {
```
veselypeta, line 7748: Perhaps put this in an annonymous namespace so that it wont conflict with anything?

veselypeta (approved):
LGMT, just one suggestion 👍

source/loader/ur_libapi.cpp
hdelan: Yep good shout. Change made

kbenzie:
> One other solution to this would be conditionally adding `-pthread` to the loader in cmake if using a known bad version of gcc. I don't know if that's portable across everything that we target. Looking at the comment on the gcc issue, it was fixed by the call_once implementation changing in `GLIBCXX_3.4.29` which corresponds to GCC 11.1.

I think this would be a preferable workaround to me.

hdelan:
> I think this would be a preferable workaround to me.

Change made

ldrumm:
Looks good. The cmake approach is the right one, and it makes it clear that this the extra flag is a workaround for a specific compiler, so it can be removed later

Definitely a fun bug to track down!

source/loader/CMakeLists.txt
```diff
@@ -48,6 +48,13 @@ if (UNIX)
     set(THREADS_PREFER_PTHREAD_FLAG TRUE)
     find_package(Threads REQUIRED)
     target_link_libraries(ur_loader PRIVATE Threads::Threads)
+    # Older gcc versions need -pthread, not just -lpthread
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.1)
+            find_library(PTHREAD_LIBRARY pthread)
```
kbenzie, line 54: The result of this `find_library` is not used, if you wanted to assert that `pthread` is available on the system then `PTHREAD_LIBRARY`'s value should be checked to ensure it does not contain `NOTFOUND` or add the `REQUIRED` flag to the `find_library` call. Alternatively, we could assume a POSIX system when using GCC and completely remove this `find_library` call.

hdelan: I've changed so that we are checking for `CMAKE_USE_PTHREADS_INIT` which is set by cmake `Threads`.

kbenzie (approved):

hdelan:
Can someone with merge privileges merge this please?

PR #799, opened 2023-08-11
[cts] Generate conformance tests for adapters

wlemkows:


test/conformance/CMakeLists.txt
```diff
@@ -5,6 +5,17 @@
 
 set(UR_CONFORMANCE_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
+function(add_test_force_adapters name adapter)
```
lukaszstolarczuk, line 8: why not just `add_test_adapter`...?

test/conformance/adapters/hip/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
 # See LICENSE.TXT
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-
+set(USE_LOCAL_TEST_CONFIGURATION ON)
```
lukaszstolarczuk, line 5: pls add an empty line above your change

test/conformance/CMakeLists.txt
```diff
@@ -18,10 +29,27 @@ function(add_conformance_test name)
         ${PROJECT_NAME}::testing
         ${PROJECT_NAME}::common
         GTest::gtest_main)
-    add_test(NAME ${name}
-        COMMAND ${TEST_TARGET_NAME}
-        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
-    set_tests_properties(${name} PROPERTIES LABELS "conformance")
+
+    if(${USE_LOCAL_TEST_CONFIGURATION})
```
lukaszstolarczuk, line 33: in `if` you can skip `${}` syntax (applies here and a few lines below)

test/conformance/CMakeLists.txt
```diff
@@ -18,10 +29,27 @@ function(add_conformance_test name)
...
+	add_test(NAME ${name}
```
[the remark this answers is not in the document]
wlemkows: Done.

test/conformance/CMakeLists.txt
```diff
@@ -18,10 +29,27 @@ function(add_conformance_test name)
...
+            COMMAND ${TEST_TARGET_NAME}
+            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+    else()
+        if(UR_BUILD_ADAPTER_CUDA)
+            add_test_force_adapters(${name} adapter_cuda)
```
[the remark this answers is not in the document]
wlemkows: This prefix is used intentionally, in order to have the correct name for linking adapters and the correct label. In the function we have several uses for this variable, so let me leave it as it is for now. Perhaps in the future, I will change this part of the code.

test/conformance/adapters/hip/CMakeLists.txt
wlemkows: Done.

test/conformance/CMakeLists.txt
wlemkows: Done.

test/conformance/adapters/cuda/CMakeLists.txt
```diff
@@ -3,7 +3,8 @@
 # See LICENSE.TXT
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-add_conformance_test_with_devices_environment(adapter-cuda 
+set(USE_LOCAL_TEST_CONFIGURATION ON)
```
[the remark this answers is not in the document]
wlemkows: This flag makes it not only use a specific adapter, but the entire local test configuration.

test/conformance/CMakeLists.txt
wlemkows: Done.

test/conformance/CMakeLists.txt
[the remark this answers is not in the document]
lukaszstolarczuk: Heh, ok, it's not that big of a deal.
But on the other hand, it's just a matter of prepending `name` with `adapter_` within the function.

lukaszstolarczuk (approved):
👍

veselypeta:
I'm not sure I fully understand what you're trying to do here? Is this so that `ctest` will run against each adapter present?

pbalcer:
> I'm not sure I fully understand what you're trying to do here? Is this so that the conformance tests will be run with `ctest` against each adapter present?

That's right. We needed the ability to select which adapter is being tested at runtime (not configure-time or build-time). And, maybe more importantly, filter out the failing ones. This approach will allow us to easily run cts for each adapter. It also prepares ctests for upcoming changes where we filter out gtest output to allow known-failing tests that are specific for each adapter. Something like:
`COMMAND match.py ${TEST_TARGET_NAME} expected-output-${adapter}.match`. This would run the tests and expect a specific output for an adapter, passing even with known failures.

veselypeta (approved):
Thanks for the clarification @pbalcer.

pbalcer (approved):

Issue #800, opened 2023-08-11
build error: implicit conversion between pointer-to-function and pointer-to-object


smaslov-intel:
This line produces a warning that causes problems when warnings are promoted to errors: https://github.com/oneapi-src/unified-runtime/blob/09cb31ae5528affc7bfb4cb73dbeede9ff5041c8/source/common/windows/ur_lib_loader.cpp#L33

> unified-runtime-src/source/common/windows/ur_lib_loader.cpp(33,12): error: implicit conversion between pointer-to-function and pointer-to-object is a Microsoft extension [-Werror,-Wmicrosoft-cast]
    return GetProcAddress(handle, func_name);
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

PR #802, opened 2023-08-11
Address https://github.com/oneapi-src/unified-runtime/issues/800

smaslov-intel:


smaslov-intel:
@kbenzie , @pbalcer : please review/merge. When merged please update tag here: https://github.com/intel/llvm/blob/d778e56245c8ff2758c5ea56d269f7d12713f616/sycl/plugins/unified_runtime/CMakeLists.txt#L7

veselypeta (approved):

smaslov-intel:
Who can merge this asap? It's blocking some internal builds from completing.

smaslov-intel:
> @kbenzie , @pbalcer : please review/merge. When merged please update tag here: https://github.com/intel/llvm/blob/d778e56245c8ff2758c5ea56d269f7d12713f616/sycl/plugins/unified_runtime/CMakeLists.txt#L7

I've created https://github.com/intel/llvm/pull/10814

igchor:
Look like we'll need to create a new tag. @smaslov-intel should we do this asap (I think @kbenzie is back tomorrow) or do you expect we'll need some other changes?

smaslov-intel:
> Look like we'll need to create a new tag. @smaslov-intel should we do this asap (I think @kbenzie is back tomorrow) or do you expect we'll need some other changes?

This is the only thing that is needed asap. Does it not work to refer to the commit directly (without creating a tag) as I did in https://github.com/intel/llvm/pull/10814?

igchor:
This should work fine for now, but for the release, it would probably be better to use a tag. @kbenzie can probably comment if there's any significant difference (we may do some extra validation on the tag but I'm not sure if that's true at the moment).

PR #805, opened 2023-08-15
Correct level of indirection used in KernelSetArgPointer calls.

### LINKED ISSUE #558: Clarify urKernelSetArgPointer spec
We have the urKernelSetArgPointer entry point to pass USM allocations as parameters to kernels. To get an idea of how this is set up we have these function signatures for some of the relevant entry points:
```
ur_result_t urUSMDeviceAlloc(ur_context_handle_t hContext, ur_device_handle_t hDevice,
                             const ur_usm_desc_t *pUSMDesc, ur_usm_pool_handle_t pool,
                             size_t size, void **ppMem)

ur_result_t urKernelSetArgPointer(ur_kernel_handle_t hKernel, uint32_t argIndex,
                                  const void *pArgValue)
```
Based on this, and the spec description for `urKernelSetArgPointer`'s `pArgValue` parameter
> [in][optional] USM pointer to memory location holding the argument value. If null then argument value is considered null.

Correct usage could look like this
```
void* allocation = nullptr;
size_t allocation_size = 32;

urUSMDeviceAlloc(context, device, nullptr, nullptr,
                 allocation_size, &allocation);

urKernelSetArgPointer(kernel, 0, allocation);
```
or this
```
void* allocation = nullptr;
size_t allocation_size = 32;

urUSMDeviceAlloc(context, device, nullptr, nullptr,
                 allocation_size, &allocation);

urKernelSetArgPointer(kernel, 0, &allocation);
```
depending on whether you put emphasis on
> [in][optional] **USM pointer** to memory location holding the argument value. If null then argument value is considered null.

or
> [in][optional] USM **pointer to memory location holding the argument** value. If null then argument value is considered null.

While implementing tests that make direct use of these entry points I've found that existing implementations expect the latter usage, and by extension this is also how the sycl runtime makes use of the analogous PI entry point.

This could do with some clarification. The wording can be changed to be more explicit, and to better match how USM pointers are passed around elsewhere it might make sense to change the function signature to this
```
ur_result_t urKernelSetArgPointer(ur_kernel_handle_t hKernel, uint32_t argIndex,
                                  void **pArgValue)
```

I would also suggest that how this is currently expected to work is... weird. Because of how the allocation entry points work, and how passing USM pointers to any other entry point works (see `urEnqueueUSMMemcpy` etc.), I found this behaviour pretty surprising. It would take some refactor work across all the UR adapters and unported plugins as well as the sycl runtime itself, but I think it's worth considering whether we should make the former usage in the above examples the expected behaviour, in the interest of making the library more intuitive.

aarongreig:
Also attempt to clarify the wording around this a bit.

fixes #558

LLVM testing https://github.com/intel/llvm/pull/12270

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -5023,8 +5023,8 @@ urKernelSetArgPointer(
     ur_kernel_handle_t hKernel,                            ///< [in] handle of the kernel object
     uint32_t argIndex,                                     ///< [in] argument index in range [0, num args - 1]
     const ur_kernel_arg_pointer_properties_t *pProperties, ///< [in][optional] pointer to USM pointer properties.
-    const void *pArgValue                                  ///< [in][optional] USM pointer to memory location holding the argument
-                                                           ///< value. If null then argument value is considered null.
+    const void *pArgValue                                  ///< [in][optional] Pointer obtained by USM allocation or virtual memory
+                                                           ///< mapping operation. If null then argument value is considered null.
 );
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/scripts/core/kernel.yml
+++ b/scripts/core/kernel.yml
@@ -339,7 +339,7 @@ params:
       desc: "[in][optional] pointer to USM pointer properties."
     - type: "const void*"
       name: pArgValue
-      desc: "[in][optional] USM pointer to memory location holding the argument value. If null then argument value is considered null."
+      desc: "[in][optional] Pointer obtained by USM allocation or virtual memory mapping operation. If null then argument value is considered null."
 returns:
     - $X_RESULT_ERROR_INVALID_KERNEL_ARGUMENT_INDEX
     - $X_RESULT_ERROR_INVALID_KERNEL_ARGUMENT_SIZE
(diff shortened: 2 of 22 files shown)


martygrant (approved):

fabiomestre (approved):

kbenzie (approved):

aarongreig:
I was able to properly test this on more than just cuda this morning and I can confirm this will need a small patch for the cuda and opencl adapters when it gets into intel-llvm. Maybe merging this should wait until the adapters move in here and they can be patched simultaneously?

kbenzie:
> Maybe merging this should wait until the adapters move in here and they can be patched simultaneously?

That makes sense, convert this to draft so we don't merge it unintentionally.

nrspruit (approved):
LGTM for level zero

kbenzie (approved):

kbenzie:
@oneapi-src/unified-runtime-native-cpu-write @mmoadeli would be great to get your reviews on this.

PietroGhg (approved):
Native CPU lgtm, thank you

kbenzie:
@oneapi-src/unified-runtime-cuda-write @oneapi-src/unified-runtime-hip-write would appriciate a review on this, the CUDA/HIP changes are very small.

jchlanda (approved):
CUDA/HIP 👍

EwanC (approved):

kbenzie:
I've removed *ready to merge* since the CUDA checks aren't passing, please add it again once they go green.

PR #810, opened 2023-08-21
[UR][CTS] Upstream CTS parser

veselypeta:
Credit: @martygrant 

We had used this parser script to measure the CTS pass rate of each adapter in our internal CI. Since UR holds all the CTS tests it makes sense to upstream it here.

I've modified martin's original parser such that we run conformance test suites individually. Instead of parsing the output from stdout we can use the `GTEST_OUTPUT` environment variable to specify that GTEST should output it's testing summary into a json file. We can then parse this and summarize the results. 

Example output:
image

### DIFF
--- a/scripts/ctest_parser.py
+++ b/scripts/ctest_parser.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python
+"""
+ Copyright (C) 2022 Intel Corporation
+
+ Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+ See LICENSE.TXT
+ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""
+
+from sys import argv
+from subprocess import PIPE, Popen, STDOUT
+import argparse
+import os
+
+
+def run(command, env):
+    process = Popen(command, env=env, stdout=PIPE,
+                    stderr=STDOUT, cwd=command[1])
+    lines = process.communicate()[0].decode('utf-8').splitlines()
+    results = {"Passed": {}, "Skipped": {}, "Failed": {}, 'Total': 0, 'Success': True}
+    for line in lines:
+        result_types = ['[       OK ]', '[  FAILED  ]', '[  SKIPPED ]']
+        if any([x in line for x in result_types]) and line.endswith("ms)"):
+            name, time = line[line.find(']') + 2:].split(' ', maxsplit=1)
+            if 'OK' in line:
+                results['Passed'][name] = {'time': time}
+            elif 'SKIPPED' in line:
+                results['Skipped'][name] = {'time': time}
+            elif 'FAILED' in line:
+                results['Failed'][name] = {'time': time}
+        elif '[==========] Running' in line:
+            # This is the start of a new test suite, get the number of tests in
+            # the first line e.g: '[==========] Running 29 tests from 9 test suites.'
+            total = line[line.find('g') + 2:line.find('t') - 1]
+            results['Total'] += int(total)
+
+    if process.returncode != 0:
+        results['Success'] = False    
+
+    return results
+
+
+def print_results(results, result_type):
+    print('[CTest Parser] {} tests: '.format(result_type))
+    print("\n".join("\t{}\t{}".format(k, v['time'])
+          for k, v in results.items()))
+
+
+def print_summary(total, total_passed, total_failed, total_skipped, total_crashed):
+    pass_rate_incl_skipped = str(round((total_passed / total) * 100, 2))
+    total_excl_skipped = total - total_skipped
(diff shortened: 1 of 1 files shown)


scripts/ctest_parser.py
```diff
@@ -0,0 +1,129 @@
...
+ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""
+
+from sys import argv
+from subprocess import PIPE, Popen, STDOUT
+import argparse
+import os
+
+
+def run(command, env):
```
pbalcer, line 16: gtest can be made to output json, which is likely to be more reliable than parsing text.

martygrant: Didn't realise that, sounds like it could be better. The script works by running `ctest` in `build/test/conformance` rather than `gtest` directly against the the executables in `build/bin` which had the advantage of not needing to maintain a list of tests to run against. Although it was possible to first parse the relevant executables paths contained in each of the `CTestTestfile.cmake` in `build/test/conformance/<component>` so a list of relevant tests is built automatically.

I'll take a look at the json output from gtest, thanks.

kbenzie: Might be worth checking if there's an environment variable for gtest that will enable the json output without modifying the test commands in the cmake?

veselypeta: I was looking into this briefly. `gtest` has a ``--gtest_output=json:<outfile>`` which includes all the information about passed tests etc. I'm not sure however if you can forward this argument through Ctest into gtest or if you have to manually call each gtest executable.

pbalcer: > Although we could modify the CMake to invoke gtest with this argument i.e.
> 
> ```cmake
>     add_test(NAME ${name}
>         COMMAND ${TEST_TARGET_NAME} "--gtest_output=json:${CMAKE_CURRENT_BINARY_DIR}/${name}-output.json"
>         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
> ```

That makes the most sense to me. Summarizing results would then become a post-processing step on `*-output.json` files. This approach also has the added benefit of no longer requiring python to launch ctests directly.

martygrant: A problem with asking for the json output is if a test crashes (e.g. a seg fault is produced) then the suite and executable will also crash, and no output file gets produced.

martygrant: That's correct, you will only get details from the tests prior to the crash, anything after is lost. With my script I first check how many tests are part of the suite, giving some idea of a pass rate but yes it wouldn't be 100% accurate as there may be a passing test blocked behind an earlier crashed test.

I was thinking a way around that is to run `gtest --gtest-list-tests` to get the names of all tests, then run them individually, but many tests rely on objects being created in the suite init functions.

pbalcer: If we expect crashing tests to be a regular occurrence, I think it might be better to wrap all tests in `EXPECT_EXIT` to run each one as a separate process.

martygrant (commented):
Hey Petr this looks good, wasn't able to make changes myself last week. Would this still have the issue around crashed tests? The total number of tests wouldn't be accurate because any failed suites won't produce an output file?

veselypeta:
> Would this still have the issue around crashed tests? The total number of tests wouldn't be accurate because any failed suites won't produce an output file?

Yes this is still an issue as you say that the output file is not generate. However I think long-term we shouldn't expect there test suites to fail and obtain an accurate value from all the test suites. See discussion here https://github.com/oneapi-src/unified-runtime/pull/817#issuecomment-1693171030

veselypeta:
I now use `gtest_list_tests` to get the test list. If a test suite crashes we just assume that all the tests fail. We could get more granular with this by running each test individually with `Popen` but I think this is unnecessary.

Example output:
image

scripts/ctest_parser.py
```diff
@@ -0,0 +1,134 @@
...
+ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""
+
+from subprocess import  Popen, DEVNULL
+import argparse
+import os
+import json
+
+TMP_RESULTS_FILE = "tmp-results-file.json"
+CTS_TEST_SUITES = ["context", "device", "enqueue", "event", "kernel", "memory",
```
martygrant, line 16: I was populating a list like this automatically when writing the script to save needing to maintain the list for any future changes. You can first run `ctest --show-only` in ` build/test/conformance` and it'll give you a list of each test suite. You can also pass `-V`and you can get the full path to the executable too.

veselypeta: Thanks - I've added this to the script

martygrant (approved):

pbalcer (approved):

PR #814, opened 2023-08-21
[cmake] Ensure code is compiled with -Werror in dev mode

kbenzie:
The recent hotfix in #800 requiring the v0.7.1 tag was fixing an instance of code not compiling when `-Werror` is specified. We do already specify `-Werror` with `UR_DEVELOPER_MODE=ON` in the GitHub Actions, however the `ur_common` library was not being compiled explicitly with this flag.

As setup prior to this patch `ur_common` was an interface library. This means the source files are not compiled into their own static library but instead included in any targets which "link" it.

This patch changes the `ur_common` into a static library that is always compiled into a static library avoiding recompiling the source files in each target which makes use of it. Adapters will now link with the static library instead of including its source files into their own targets build steps. Additionally, the test suite `test-loader-config` was also updated to use `add_ur_exectuable` and thus respect the flags set when `UR_DEVELOPER_MODE=ON`.

### DIFF
--- a/source/common/CMakeLists.txt
+++ b/source/common/CMakeLists.txt
@@ -3,28 +3,25 @@
 # See LICENSE.TXT
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-add_library(ur_common INTERFACE)
+add_subdirectory(unified_malloc_framework)
+add_subdirectory(umf_pools)
+
+add_ur_library(ur_common STATIC
+    umf_helpers.hpp
+    ur_pool_manager.hpp
+    $<$<PLATFORM_ID:Windows>:windows/ur_lib_loader.cpp>
+    $<$<PLATFORM_ID:Linux,Darwin>:linux/ur_lib_loader.cpp>
+)
 add_library(${PROJECT_NAME}::common ALIAS ur_common)
 
-target_include_directories(ur_common INTERFACE
+target_include_directories(ur_common PUBLIC
     ${CMAKE_CURRENT_SOURCE_DIR}
     ${CMAKE_SOURCE_DIR}/include
 )
 
-add_subdirectory(unified_malloc_framework)
-add_subdirectory(umf_pools)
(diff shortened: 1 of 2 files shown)


aarongreig (approved):

pbalcer:
> Because `ur_common` is only used in adapter implementations, this code was never being compiled in our own GitHub Actions.

That's not true. `ur_common` is used in the loader as well as most tests. In fact, #800 was changing code that is only used by the loader. But I agree that this change makes sense since it allows us to use the `add_ur_library` wrapper. I've noticed that warnings/error aren't always emitted correctly for header files.

I checked a [CI build](https://github.com/oneapi-src/unified-runtime/actions/runs/5812123007/job/15756835654) prior to #800, and there's even a warning about this specific problem. It's either a configuration we don't test (different OS/compiler version), or we are missing some extra flag that intel/llvm has.

kbenzie:
> That's not true. `ur_common` is used in the loader as well as most tests.

Ah my mistake, I was grepping for `ur_common` not `${PROJECT_NAME}::common` so missed those uses. I'll update the description/commit message appropriately.

> I checked a [CI build](https://github.com/oneapi-src/unified-runtime/actions/runs/5812123007/job/15756835654) prior to https://github.com/oneapi-src/unified-runtime/issues/800, and there isn't even a warning about this specific problem.

I just looked at #800 again and its a GCC style warning about a Windows only source file, could this be a MinGW configuration?

Perhaps we could also increase the warning level or enable an equivelant warning that `cl.exe` would emit?

kbenzie:
> I just looked at #800 again and its a GCC style warning about a Windows only source file, could this be a MinGW configuration?

Ah, this is actually a clang warning. Perhaps we need a build using the clang frontend on Windows. I've not used that myself before but I beleive its supposed in Visual Studio.

pbalcer:
> Perhaps we could also increase the warning level or enable an equivelant warning that `cl.exe` would emit?

We could try bumping warning level to `/W4` (from `/W3`). But ultimately it might be best to also compile with clang on Windows in CI.

pbalcer (approved):

PR #817, opened 2023-08-23
[UR] Define new test macro to catch crashes

veselypeta:
Define a specific conformance test macro which will fail each test that exits with a crash intead of failing the entire test suite.

If we decide to use this macro, we will need to ensure that we run the testsuites with the `--gtest_death_test_style=threadsafe`

### DIFF
--- a/test/conformance/testing/include/uur/checks.h
+++ b/test/conformance/testing/include/uur/checks.h
@@ -45,6 +45,53 @@ inline std::ostream &operator<<(std::ostream &out, const Result &result) {
 #ifndef EXPECT_SUCCESS
 #define EXPECT_SUCCESS(ACTUAL) EXPECT_EQ_RESULT(UR_RESULT_SUCCESS, ACTUAL)
 #endif
+#ifndef EXPECT_NO_CRASH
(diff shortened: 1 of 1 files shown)


pbalcer:
That looks clunky, but I guess it's out of the way for the most part...

Maybe it would be simpler to run tests one by one? We are already going to be wrapping test execution in a script (#823), so might as well do a loop inside of it. I'm not sure what's better.

kbenzie:
> Maybe it would be simpler to run tests one by one?

In my experience with [UnitCL](https://github.com/codeplaysoftware/oneapi-construction-kit/tree/main/source/cl/test/UnitCL) this ends up being much slower to run. If we do decide to do that though, there is a ready made tool for it [gtest-parallel](https://github.com/google/gtest-parallel).

kbenzie:
Long term we do expect there to be no crashes in the CTS so I feel like this approach is very heavy handed way of dealing with the problem at hand. This gtest functionality is not intended to be used on every single test case but rather to test specific failure modes around process exits.

veselypeta:
Dropping this solution in favour of #810

PR #818, opened 2023-08-23
[EXP][CMDBUF] Add fill commands to cmd buffer experimental feature

Bensuo:
- Adds USM and Buffer fill append commands
- Update feature spec for new commands
- Aligns Append* naming conventions with core equivalents

### DIFF
--- a/scripts/core/exp-command-buffer.yml
+++ b/scripts/core/exp-command-buffer.yml
@@ -180,7 +180,7 @@ returns:
 type: function
 desc: "Append a USM memcpy command to a command-buffer object"
 class: $xCommandBuffer
-name: AppendMemcpyUSMExp
+name: AppendUSMMemcpyExp
 params:
     - type: $x_exp_command_buffer_handle_t
       name: hCommandBuffer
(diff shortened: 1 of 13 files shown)


scripts/core/exp-command-buffer.yml
```diff
@@ -217,6 +217,51 @@ returns:
...
+desc: "Append a USM fill command to a command-buffer object"
+class: $xCommandBuffer
+name: AppendFillUSMExp
+params:
+    - type: $x_exp_command_buffer_handle_t
+      name: hCommandBuffer
+      desc: "[in] handle of the command-buffer object."
+    - type: "void*"
+      name: pMemory
+      desc: "[in] pointer to USM memory object."
```
EwanC, line 229: Preference for slightly tweaked wording
```suggestion
      desc: "[in] pointer to USM allocated memory to fill."
```

scripts/core/exp-command-buffer.yml
```diff
@@ -217,6 +217,51 @@ returns:
...
+    - type: "const void*"
+      name: pPattern
+      desc: "[in] pointer to the fill pattern."
+    - type: "size_t"
+      name: patternSize
+      desc: "[in] size in bytes of the pattern."
```
EwanC, line 235: Are we expecting any restrictions on `patternSize`? e.g Must be a power-of-two, or <= the largest integer or floating-point vector data type supported by the device.

Same comment for buffer equivalent entry-point

EwanC (commented):
Mostly nitpickings, but one comment about valid inputs for `patternSize` parameter

scripts/core/exp-command-buffer.yml
Bensuo: There are restrictions listed in the `returns:` section that are the same as the existing APIs:

```rst
        - "`patternSize == 0 || size == 0`"
        - "`patternSize > size`"
        - "`(patternSize & (patternSize - 1)) != 0`"
        - "`size % patternSize != 0`"
        - "If `size` is higher than the allocation size of `ptr`"
```

EwanC: apologies, missed that. Looks good.

Because `size` has ",must be a multiple of patternSize." after its description it got me thinking down this path.

scripts/core/exp-command-buffer.yml
```diff
@@ -217,6 +217,51 @@ returns:
     - $X_RESULT_ERROR_OUT_OF_RESOURCES
 --- #--------------------------------------------------------------------------
 type: function
+desc: "Append a USM fill command to a command-buffer object"
+class: $xCommandBuffer
+name: AppendFillUSMExp
```
kbenzie, line 222: Also realised this is internally consistent but not externally consistent with `urEnqueueUSMFill`.

```suggestion
name: AppendUSMFillExp
```

kbenzie (approved):
So my too comments about naming conventions don't need to block this PR but they should be made consistent with enquivelent enqueue entry point naming before command buffers come out of experimental support.

Bensuo:
> So my two comments about naming conventions don't need to block this PR but they should be made consistent with enquivelent enqueue entry point naming before command buffers come out of experimental support.

Good catch! I've just addressed these as part of this PR for simplicity, they should be well aligned now.

veselypeta (approved):

PR #823, opened 2023-08-24
Test test

wlemkows:
- [ ] We still need to set up our runners with the correct HW to run this workflow successfully.


cmake/match.py
```diff
@@ -22,10 +22,17 @@
...
 
-if len(match_lines) != len(input_lines):
-    sys.exit(f"Line count doesn't match (is: {len(input_lines)}, expected: {len(match_lines)})")
+if len(match_lines) < len(input_lines):
+    sys.exit(f"Match length < input length (input: {len(input_lines)}, match: {len(match_lines)})")
 
+input_idx = 0
+optional_line = False
 for i, match_line in enumerate(match_lines):
+    opt = "${OPT}"
```
pbalcer, line 31: hm, `{{OPT}}` to be consistent with how regex is specified?

cmake/match.py
```diff
@@ -22,10 +22,17 @@
     input_lines = input.readlines()
     match_lines = match.readlines()
 
-if len(match_lines) != len(input_lines):
-    sys.exit(f"Line count doesn't match (is: {len(input_lines)}, expected: {len(match_lines)})")
+if len(match_lines) < len(input_lines):
+    sys.exit(f"Match length < input length (input: {len(input_lines)}, match: {len(match_lines)})")
 
+input_idx = 0
+optional_line = False
```
lukaszstolarczuk, line 29: shouldn't this setting be done within `for` loop?

// I know we don't have a case where we mix opt with non-opt lines, but just to be sure 😉

cmake/match.py
```diff
@@ -35,9 +42,26 @@
...
+    input_line = input_lines[input_idx].strip()
     if not re.fullmatch(pattern, input_line):
-        print(f"Line {i+1} does not match".format(i+1))
-        print("is:       " + input_line)
-        print("expected: " + match_line.strip())
-        sys.exit(1)
+        if optional_line is True:
+            continue
+        else:
+            print(f"Line {i+1} does not match".format(i+1))
```
lukaszstolarczuk, line 59: hmm maybe I'm mistaken, but it seems you're mixing F-String with `str.format()` syntax 😉 either one or another...?

test/conformance/CMakeLists.txt
```diff
@@ -8,9 +8,18 @@ set(UR_CONFORMANCE_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 function(add_test_adapter name adapter)
     set(TEST_TARGET_NAME test-${name})
     set(TEST_NAME ${name}-${adapter})
+
     add_test(NAME ${TEST_NAME}
-        COMMAND ${TEST_TARGET_NAME}
-        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+        COMMAND ${CMAKE_COMMAND}
+        -D TEST_FILE=${Python3_EXECUTABLE}
```
lukaszstolarczuk, line 14: Is this solution like a temporary one? Or should we maintain this list of "failing" tests forever, even if all of them are green?

Perhaps we should mention somewhere (at least in a comment here) what's the purpose of this extra matching here...

lukaszstolarczuk (commented):
Can you please add some info in commit message? Explain a little why it's needed 😉

test/conformance/CMakeLists.txt
pbalcer: Ideally all adapters would have 100% pass rate, but that's not the case at the moment unfortunately. And it's going to take some time for everything to get fixed. So yes, this is permanent until such time that we fix everything and agree to never break CTS again, which might be a while.

Definitely should be documented somewhere.

test/conformance/context/context_adapter_cuda.match
```diff
@@ -0,0 +1 @@
+urContextCreateWithNativeHandleTest.Success/NVIDIA_CUDA_BACKEND___NVIDIA_GeForce_RTX_3060_
```
pbalcer, line 1: {{.*}} for device names? Same everywhere else.

lukaszstolarczuk:
@wlemkows, please use my latest workflow commit from: https://github.com/lukaszstolarczuk/unified-runtime/commits/adapters-hw-ci

test/conformance/cts_exe.py
```diff
@@ -0,0 +1,31 @@
...
+from argparse import ArgumentParser
+import subprocess  # nosec B404
+
+if __name__ == '__main__':
+
+    parser = ArgumentParser()
+    parser.add_argument("--test_command", help="Ctest test case")
+
+    args = parser.parse_args()
+    result = subprocess.Popen([args.test_command, '--gtest_brief=1'], stdout = subprocess.PIPE, text = True)  # nosec B603
```
pbalcer, line 21: `subprocess.run` ?

wlemkows: Popen is a better version of `run`, I use it here on purpose because as a result, I get the TextIOWrapper class on which I can perform additional operations that are not possible on the str object in the case of  `run`.

cmake/match.py
wlemkows: Done.

cmake/match.py
wlemkows: Done.

test/conformance/CMakeLists.txt
wlemkows: Done.

wlemkows:
> @wlemkows, please use my latest workflow commit from: https://github.com/lukaszstolarczuk/unified-runtime/commits/adapters-hw-ci


Done.

wlemkows:
> Can you please add some info in commit message? Explain a little why it's needed 😉

Done.

.github/workflows/cmake.yml
```diff
@@ -263,4 +313,4 @@ jobs:
         -DUR_FORMAT_CPP_STYLE=ON
         -DUMF_ENABLE_POOL_TRACKING=ON
     - name: Build
-      run: cmake --build ${{github.workspace}}/build -j $(nproc)
+      run: cmake --build ${{github.workspace}}/build -j $(nproc)
```
lukaszstolarczuk, line 316: suddenly EOF - pls revert this change 😉

cmake/match.py
lukaszstolarczuk: Not quite. I think Piotr suggested without `$` - to be in line with what you have a few lines below, in the code:
`# split into parts at {{ }}`

cmake/match.py
```diff
@@ -22,10 +22,18 @@
...
     match_lines = match.readlines()
 
-if len(match_lines) != len(input_lines):
-    sys.exit(f"Line count doesn't match (is: {len(input_lines)}, expected: {len(match_lines)})")
+if len(match_lines) < len(input_lines):
+    sys.exit(f"Match length < input length (input: {len(input_lines)}, match: {len(match_lines)})")
 
+input_idx = 0
 for i, match_line in enumerate(match_lines):
+    opt = "${{OPT}}"
```
lukaszstolarczuk, line 30: heh, this opt can be moved outside of the loop, actually

test/conformance/README.md
```diff
@@ -0,0 +1,8 @@
+# Conformance tests
+
```
lukaszstolarczuk, line 2: maybe it's an obvious statement, but I'd add info that "when you fix tests you have to update match files" and perhaps mention why we have empty match files...?

test/conformance/CMakeLists.txt
lukaszstolarczuk: Thanks!

test/conformance/cts_exe.py
```diff
@@ -0,0 +1,31 @@
+#! /usr/bin/env python3
```
pbalcer, line 1: For now it's fine, but eventually, I'd like to see this script somehow merged with #810. They fulfill a similar role...

pbalcer (approved):

test/conformance/cts_exe.py
```diff
@@ -0,0 +1,31 @@
...
+
+if __name__ == '__main__':
+
+    parser = ArgumentParser()
+    parser.add_argument("--test_command", help="Ctest test case")
+
+    args = parser.parse_args()
+    result = subprocess.Popen([args.test_command, '--gtest_brief=1'], stdout = subprocess.PIPE, text = True)  # nosec B603
+
+    while True:
```
PatKamin, line 23: This could be simplified to sth like:
```
for line in result.stdout.readlines():
...
```

test/conformance/cts_exe.py
```diff
@@ -0,0 +1,31 @@
...
+        line = result.stdout.readline()
+        if not line: break
+        if "[  FAILED  ]" in line:
+            test_case = line.split(" ")[5]
+            if (test_case[-1] == ","):
+                test_case = test_case[:len(test_case)-1]
+                print(test_case)
+    exit(0)
```
PatKamin, line 31: Possibly this exit is redundant, the script should exit with zero code at this point.

cmake/match.py
```diff
@@ -35,9 +43,26 @@
...
+            print("expected: " + match_line.strip())
+            sys.exit(1)
+
+    input_line = input_lines[input_idx].strip()
     if not re.fullmatch(pattern, input_line):
-        print(f"Line {i+1} does not match".format(i+1))
-        print("is:       " + input_line)
-        print("expected: " + match_line.strip())
-        sys.exit(1)
+        if optional_line is True:
```
PatKamin, line 57: As I understand, this makes optional lines hard to check for - they can contain literally anything and the test can pass. Can we make opt lines required given a certain test scenario or setup?

pbalcer: That would require us to somehow make the lines "conditional", so something like `{{skip if compiler==clang}}`. While that might be useful, it's imho out of scope for this work. And no, optional lines can't be anything - this `continue` here will only move forward the match file, not the input file, so if the line doesn't match either the optional line or the next one, it will fail.

test/conformance/cts_exe.py
wlemkows: Sure, we can plan such changes in the future.

pbalcer (approved):

test/conformance/cts_exe.py
```diff
@@ -0,0 +1,36 @@
...
+
+    args = parser.parse_args()
+    result = subprocess.Popen([args.test_command, '--gtest_brief=1'], stdout = subprocess.PIPE, text = True)  # nosec B603
+
+    pat = re.compile(r'\[( )*FAILED( )*\]')
+    for line in result.stdout.readlines():
+        if pat.search(line):
+            test_case = line.split(" ")[5]
+            if (test_case[-1] == ","):
+                test_case = test_case[:len(test_case)-1]
```
PatKamin, line 30: Try simpler `test_case.rstrip(',')`

test/conformance/cts_exe.py
```diff
@@ -0,0 +1,36 @@
...
+            print(test_case)
+
+    output = result.communicate()
```
PatKamin, line 33: Missing output print?

lukaszstolarczuk (approved):

test/conformance/cts_exe.py
wlemkows: Done.

test/conformance/cts_exe.py
wlemkows: At this point I do not intend to use this variable, I just want to call the communicate function. Removed.

PatKamin (approved):

PR #783, opened 2023-08-07
[umf] add out of memory test

DamianDuy:

PR #829, opened 2023-08-25
[umf] extend out of memory test for disjoint pool

DamianDuy:
It is the continuation of the previous out of memory test. This PR adds check for the disjoint pool if after reaching out of memory error from the provider and then freeing some memory, we are able to allocate again.
Based on #783

### DIFF
--- a/test/unified_malloc_framework/common/pool.hpp
+++ b/test/unified_malloc_framework/common/pool.hpp
@@ -23,12 +23,24 @@
 #include <stdlib.h>
 
 #include "base.hpp"
+#include "provider.hpp"
 #include "umf_helpers.hpp"
 
 namespace umf_test {
 
 auto wrapPoolUnique(umf_memory_pool_handle_t hPool) {
     return umf::pool_unique_handle_t(hPool, &umfPoolDestroy);
+}
+
+template <typename T, typename... Args>
+auto makePoolWithOOMProvider(int allocNum, Args &&...args) {
+    auto [ret, provider] =
+        umf::memoryProviderMakeUnique<provider_mock_out_of_mem>(allocNum);
+    EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
+    auto [retp, pool] = umf::poolMakeUnique<T, 1, Args...>(
+        {std::move(provider)}, std::forward<Args>(args)...);
+    EXPECT_EQ(retp, UMF_RESULT_SUCCESS);
+    return std::move(pool);
 }
 
 bool isReallocSupported(umf_memory_pool_handle_t hPool) {
@@ -128,10 +140,13 @@ struct proxy_pool : public pool_base {
     void *calloc(size_t num, size_t size) noexcept {
         void *ptr;
         auto ret = umfMemoryProviderAlloc(provider, num * size, 0, &ptr);
+        umf::getPoolLastStatusRef<proxy_pool>() = ret;
+
+        if (!ptr) {
+            return ptr;
+        }
 
         memset(ptr, 0, num * size);
-
-        umf::getPoolLastStatusRef<proxy_pool>() = ret;
         return ptr;
     }
     void *realloc([[maybe_unused]] void *ptr,
--- a/test/unified_malloc_framework/memoryPool.hpp
+++ b/test/unified_malloc_framework/memoryPool.hpp
@@ -55,9 +55,27 @@ struct umfMultiPoolTest : umfPoolTest {
     std::vector<umf::pool_unique_handle_t> pools;
 };
 
-struct umfMemTest : umfPoolTest {
-    void SetUp() override { umfPoolTest::SetUp(); }
-    void TearDown() override { umfPoolTest::TearDown(); }
+struct umfMemTest
+    : umf_test::test,
+      ::testing::WithParamInterface<
+          std::tuple<std::function<umf::pool_unique_handle_t(void)>, int>> {
+    umfMemTest() : pool(nullptr, nullptr), expectedRecycledPoolAllocs(0) {}
+    void SetUp() override {
+        test::SetUp();
+        initialize();
+    }
+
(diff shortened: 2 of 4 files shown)


test/unified_malloc_framework/memoryPoolAPI.cpp
```diff
@@ -157,6 +157,18 @@ INSTANTIATE_TEST_SUITE_P(
...
+    proxyPoolOOMTest, umfMemTest,
+    ::testing::Values(std::tuple(
+        [] {
+            return umf::poolMakeUnique<umf_test::proxy_pool, 1>(
+                       {umf::memoryProviderMakeUnique<
+                            umf_test::provider_mock_out_of_mem>(10)
+                            .second})
+                .second;
+        },
+        0)));
```
kswiecicki, line 170: You set the provider to be OOM after 10 allocations but at the same time set the `expectedPooledAllocations` as 0.

test/unified_malloc_framework/umf_pools/disjoint_pool.cpp
```diff
@@ -33,6 +33,16 @@ static auto makePool() {
     return std::move(pool);
 }
 
+static auto makePoolOOMProvider() {
```
kswiecicki, line 36: This function returns a pool, not a provider. Also, you could make this a template function and reuse it in the `proxyPoolOOMTest` test instantiation. It would look similar to `poolMakeUnique` where you accept `Args &&...args` as parameters and forward it to the pool `initialize` function.

test/unified_malloc_framework/common/pool.hpp
```diff
@@ -131,9 +131,7 @@ struct proxy_pool : public pool_base {
 
         memset(ptr, 0, num * size);
```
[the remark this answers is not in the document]
DamianDuy: Done.

test/unified_malloc_framework/memoryPool.hpp
```diff
@@ -251,6 +276,46 @@ TEST_P(umfPoolTest, multiThreadedMallocFreeRandomSizes) {
...
+
+    // next part of the test- freeing some memory to allocate it again (as the memory
+    // should be acquired from the pool itself now, not from the provider),
+    // is done only for the disjoint pool for now
+
+    // remove last nullptr from the allocations vector
+    ASSERT_EQ(allocations.back(), nullptr);
+    allocations.pop_back();
+
+    ASSERT_NE(allocations.back(), nullptr);
```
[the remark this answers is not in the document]
DamianDuy: I'm not exactly sure what you mean. Could you clarify? The size of the `allocations` vector is larger than the value of `expectedPooledAllocations`, since the latter is used to free some memory and then for trying to allocate it again (to check that memory is taken now  from the pool, not from the provider).

test/unified_malloc_framework/memoryPoolAPI.cpp
DamianDuy: Yes, that's because there's no pooling for proxy pool so the part of the test where the `expectedPooledAllocations` is used, for now is only done for disjoint pool.

test/unified_malloc_framework/umf_pools/disjoint_pool.cpp
DamianDuy: Done. I've added an extra template function.

test/unified_malloc_framework/memoryPoolAPI.cpp
kswiecicki: Oh, my bad.

test/unified_malloc_framework/memoryPool.hpp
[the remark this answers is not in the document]
kswiecicki: I mistook `expectedPooledAllocation` for the number of allocations from the pool before it goes OOM. In that case maybe `expectedRecycledPoolAllocs` would fit better.

test/unified_malloc_framework/umf_pools/disjoint_pool.cpp
```diff
@@ -72,6 +73,15 @@ TEST_F(test, freeErrorPropagation) {
 INSTANTIATE_TEST_SUITE_P(disjointPoolTests, umfPoolTest,
                          ::testing::Values(makePool));
 
+INSTANTIATE_TEST_SUITE_P(
+    disjointPoolTests, umfMemTest,
+    ::testing::Values(std::make_tuple(
+        [] {
+            return umf_test::makePoolWithOOMProvider<usm::DisjointPool>(
+                4, poolConfig());
```
igchor, line 81: If this 4 is the same 4 as in `config.Capacity` it would be better to make it a constant or just call poolConfig().Capacity here.

igchor: Similarly for '2'. Is it calculated from the config somehow or is it just an arbitrary number?

test/unified_malloc_framework/memoryPool.hpp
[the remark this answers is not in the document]
DamianDuy: Done.

test/unified_malloc_framework/umf_pools/disjoint_pool.cpp
DamianDuy: Done. `2` was an arbitrary number. I changed it to `poolConfig().Capacity / 2` though because I thought that might be a better idea if for example the `Capacity` in config would change.

igchor (approved):

kswiecicki (approved):

lukaszstolarczuk (commented):
You can squash your commits now (the first commit does not introduce much change now 😉 ).

DamianDuy:
Done.

lukaszstolarczuk (commented):
👍

pbalcer (approved):

PR #834, opened 2023-08-30
Fix device CTS tests

fabiomestre:


kbenzie:
There doesn't appear to be anything CUDA specific in this patch, even if that's what drove these changes I'm not sure its worth calling out in this PR.

scripts/core/device.yml
```diff
@@ -144,6 +144,8 @@ params:
             [out][optional] pointer to the number of devices.
             pNumDevices will be updated with the total number of devices available.
 returns:
+    - $X_RESULT_ERROR_INVALID_SIZE:
+        - "`NumEntries == 0 && phDevices != NULL`"
```
callumfare, line 148: The description for `NumEntries` should be updated too since it says that it will return `UR_RESULT_ERROR_INVALID_VALUE` in this case

fabiomestre: Actually, I went with SIZE because that's what the test was using. But on second thought, maybe INVALID_VALUE is better?

kbenzie: It would align with the OpenCL error code. What do other entry points specify for this type of error case?

kbenzie: Consistency is best, yup.

veselypeta: Having the same fault on HIP so 👍 from me

veselypeta (approved):

scripts/core/device.yml
fabiomestre: Done :+1:

pbalcer (approved):

kbenzie (approved):

PR #836, opened 2023-08-31
[CTS] fix cts device adapter L0 match file

wlemkows:


pbalcer (approved):
add a comment in the commit message that this test if flaky.

kswiecicki (approved):
nit: [...] `it` is flaky.

PR #838, opened 2023-08-31
Run fuzz test with L0 when available

PatKamin:


lukaszstolarczuk (commented):
👍

test/fuzz/CMakeLists.txt
```diff
@@ -16,11 +16,19 @@ function(add_fuzz_test name label)
     add_test(NAME ${TEST_TARGET_NAME}
         COMMAND ${TEST_TARGET_NAME} ${ARGN} -verbosity=1
         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+
+    set(ENV_VARS UR_ENABLE_LAYERS=UR_LAYER_FULL_VALIDATION)
+    if (UR_BUILD_ADAPTER_L0)
+        list(APPEND ENV_VARS
+                UR_ADAPTERS_FORCE_LOAD=\"$<TARGET_FILE:ur_adapter_level_zero>\"
+                NEOReadDebugKeys=1
```
pbalcer, line 24: why?

pbalcer (approved):

test/fuzz/CMakeLists.txt
PatKamin: This combination of these two compute-runtime env vars is for dynamic linking of an Intel driver library without DEEPBIND. The error I get when running the test with sanitizers:
```
You are trying to dlopen a libigc.so.1 shared library with RTLD_DEEPBIND flag which is incompatible with sanitizer runtime (see https://github.com/google/sanitizers/issues/611 for details). If you want to run libigc.so.1 library under sanitizers please remove RTLD_DEEPBIND from dlopen flags.
```

pbalcer: Ah, ok, I misunderstood what "read" here means :D

PR #840, opened 2023-08-31
[ci] fix running cts tests that require device binaries

pbalcer:
Creating device binaries silently fails if SYCL isn't available.


test/conformance/enqueue/enqueue_adapter_cuda.match
```diff
@@ -1 +1,56 @@
...
+{{OPT}}urEnqueueUSMMemcpy2DTestWithParam.SuccessNonBlocking/NVIDIA_CUDA_BACKEND___{{.*}}___pitch__1024__width__256__height__256
+{{OPT}}urEnqueueUSMMemcpy2DTestWithParam.SuccessNonBlocking/NVIDIA_CUDA_BACKEND___{{.*}}___pitch__234__width__233__height__23
+{{OPT}}urEnqueueUSMMemcpy2DTestWithParam.SuccessNonBlocking/NVIDIA_CUDA_BACKEND___{{.*}}___pitch__234__width__233__height__1
+{{OPT}}urEnqueueUSMMemcpy2DNegativeTest.InvalidNullHandleQueue/NVIDIA_CUDA_BACKEND___{{.*}}___pitch__1__width__1__height__1
+{{OPT}}urEnqueueUSMMemcpy2DNegativeTest.InvalidNullPointer/NVIDIA_CUDA_BACKEND___{{.*}}___pitch__1__width__1__height__1
+{{OPT}}urEnqueueUSMMemcpy2DNegativeTest.InvalidSize/NVIDIA_CUDA_BACKEND___{{.*}}___pitch__1__width__1__height__1
+{{OPT}}urEnqueueUSMMemcpy2DNegativeTest.InvalidEventWaitList/NVIDIA_CUDA_BACKEND___{{.*}}___pitch__1__width__1__height__1
+{{OPT}}urEnqueueUSMPrefetchWithParamTest.Success/NVIDIA_CUDA_BACKEND___{{.*}}___UR_USM_MIGRATION_FLAG_DEFAULT
+{{OPT}}urEnqueueUSMPrefetchWithParamTest.CheckWaitEvent/NVIDIA_CUDA_BACKEND___{{.*}}___UR_USM_MIGRATION_FLAG_DEFAULT
+{{OPT}}Aborted
```
wlemkows, line 56: eol

PatKamin:
Could you introduce a failure of the whole UR build when kernels fail to build?

pbalcer:
> Could you introduce a failure of the whole UR build when kernels fail to build?

I thought about that, but it would break UR CTS builds on platforms that do not have the right hw to run the binaries (e.g., regular GH-hosted runners).

PatKamin (approved):

lukaszstolarczuk (commented):
👍

PatKamin:
> 

Hmm, if we stick UR_DPCXX define to hw builds only, then this should be ok.  I don't see any benefit of having this define defined in cases where we don't have the right hw. After all, it's purpose is to trigger kernel builds.
The only issue would be that with HIP device code fail to build.

pbalcer:
> > 
> 
> Hmm, if we stick UR_DPCXX define to hw builds only, then this should be ok. I don't see any benefit of having this define defined in cases where we don't have the right hw. After all, it's purpose is to trigger kernel builds. The only issue would be that with HIP device code fail to build.

Right now this variable is being set on main to build the kernel/enqueue/program tests. The build itself passes, but the device binaries are *not* built correctly. So if we were to change the build (https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/device_code/CMakeLists.txt#L16) to make it fail if it can't build device binaries, then I think we might need a second variable to control whether the gtests are built to preserve how things currently work. So instead of having one big if here for all device-code-related tests: https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/device_code/CMakeLists.txt#L16, we'd use a different variable for these three https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/CMakeLists.txt#L75-L77 tests. I'm not sure what's the best approach tbh. @kbenzie @veselypeta any opinions?

aarongreig:
maybe we could change how `UR_CONFORMANCE_TARGET_TRIPLES` works? instead of defaulting to `spir64` (which was always of dubious usefulness) we could require it to be set if you want to build any binaries, and if it isn't set we just build the tests

pbalcer:
> maybe we could change how `UR_CONFORMANCE_TARGET_TRIPLES` works? instead of defaulting to `spir64` (which was always of dubious usefulness) we could require it to be set if you want to build any binaries, and if it isn't set we just build the tests

That makes sense. I'll do that in a separate PR. Thanks!

PR #846, opened 2023-09-01
[main] Speed up CI jobs

PatKamin:


.github/workflows/coverity.yml
```diff
@@ -31,20 +31,12 @@ jobs:
...
-
       - name: Install pip packages
         run: pip install -r third_party/requirements.txt
 
       - name: Configure CMake
-        run: cmake -B $WORKDIR/build -DUR_ENABLE_TRACING=ON -DUR_DEVELOPER_MODE=ON -DUR_BUILD_TESTS=ON -DUR_FORMAT_CPP_STYLE=ON -DUMF_ENABLE_POOL_TRACKING=ON
+        run: cmake -B $WORKDIR/build -DUR_ENABLE_TRACING=ON -DUR_DEVELOPER_MODE=ON -DUR_BUILD_TESTS=ON -DUMF_ENABLE_POOL_TRACKING=ON
 
-      - name: Generate source from spec, check for uncommitted diff
-        run: |
```
kswiecicki, line 46: Do we no longer need to check whether there were differences between generated spec and committed sources?

kswiecicki (approved):

.github/workflows/coverity.yml
lukaszstolarczuk: no in this workflow 😉

veselypeta (approved):
LGTM. Do you have any stats on the speedup?

PR #847, opened 2023-09-04
Fix device binary extension for nvptx

fabiomestre:


veselypeta (approved):

fabiomestre:
> Is this for when there is more than one triplet specified? The current script does work on CI when there's only one triplet set.



> Is this for when there is more than one triplet specified? The current script does work on CI when there's only one triplet set.

Yeah, it seems that this was working as intented. I just misunderstood how the ptx is being generated. I will close this PR

PR #849, opened 2023-09-05
Add cooperative kernels experimental feature

0x12CC:
Add two new functions for working with cooperative kernels:
- `urKernelSuggestMaxCooperativeGroupCount`
- `urEnqueueCooperativeKernelLaunch`

### DIFF
--- a/scripts/core/EXP-COOPERATIVE-KERNELS.rst
+++ b/scripts/core/EXP-COOPERATIVE-KERNELS.rst
@@ -1,0 +1,68 @@
+<%
+    OneApi=tags['$OneApi']
+    x=tags['$x']
+    X=x.upper()
+%>
+
+.. _experimental-cooperative-kernels:
+
+================================================================================
+Cooperative Kernels
+================================================================================
+
+.. warning::
+
+    Experimental features:
+
+    *   May be replaced, updated, or removed at any time.
+    *   Do not require maintaining API/ABI stability of their own additions over
+        time.
+    *   Do not require conformance testing of their own additions.
+
+
+Motivation
+--------------------------------------------------------------------------------
+Cooperative kernels are kernels that use cross-workgroup synchronization
+features. All enqueued workgroups must run concurrently for cooperative kernels
+to execute without hanging. This experimental feature provides an API for
+querying the maximum number of workgroups and launching cooperative kernels.
+
+Any device can support cooperative kernels by restricting the maximum number of
+workgroups to 1. Devices that support cross-workgroup synchronization can
+specify a larger maximum for a given cooperative kernel.
+
+The functions defined here align with those specified in Level Zero.
+
+API
+--------------------------------------------------------------------------------
+
+Macros
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* ${X}_COOPERATIVE_KERNELS_EXTENSION_STRING_EXP
+
+Functions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* ${x}EnqueueCooperativeKernelLaunchExp
+* ${x}KernelSuggestMaxCooperativeGroupCountExp
+
+Changelog
+--------------------------------------------------------------------------------
++-----------+------------------------+
+| Revision  | Changes                |
++===========+========================+
+| 1.0       | Initial Draft           |
++-----------+------------------------+
+
+Support
+--------------------------------------------------------------------------------
+
+Adapters which support this experimental feature *must* return the valid string 
+defined in ``${X}_COOPERATIVE_KERNELS_EXTENSION_STRING_EXP`` 
+as one of the options from ${x}DeviceGetInfo when querying for 
+${X}_DEVICE_INFO_EXTENSIONS. Conversely, before using any of the 
+functionality defined in this experimental feature the user *must* use the 
+device query to determine if the adapter supports this feature.
+
+Contributors
+--------------------------------------------------------------------------------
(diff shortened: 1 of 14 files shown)


scripts/core/EXP-COOPERATIVE-KERNELS.rst
```diff
@@ -0,0 +1,68 @@
...
+API
+--------------------------------------------------------------------------------
+
+Macros
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* ${X}_COOPERATIVE_KERNELS_EXTENSION_STRING_EXP
+
+Functions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* ${x}EnqueueCooperativeKernelLaunch
```
kbenzie, line 45:
```suggestion
* ${x}EnqueueCooperativeKernelLaunchExp
```

scripts/core/EXP-COOPERATIVE-KERNELS.rst
```diff
@@ -0,0 +1,68 @@
...
+* ${x}KernelSuggestMaxCooperativeGroupCount
```
kbenzie, line 46:
```suggestion
* ${x}KernelSuggestMaxCooperativeGroupCountExp
```

scripts/core/exp-cooperative-kernels.yml
```diff
@@ -0,0 +1,85 @@
...
+desc: |
+      The extension string which defines support for cooperative-kernels
+      which is returned when querying device extensions.
+name: $X_COOPERATIVE_KERNELS_EXTENSION_STRING_EXP
+value: "\"$x_exp_cooperative_kernels\""
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Enqueue a command to execute a cooperative kernel"
+class: $xEnqueue
+name: CooperativeKernelLaunch
```
kbenzie, line 25:
```suggestion
name: CooperativeKernelLaunchExp
```

scripts/core/exp-cooperative-kernels.yml
```diff
@@ -0,0 +1,85 @@
...
+    - $X_RESULT_ERROR_INVALID_WORK_DIMENSION
+    - $X_RESULT_ERROR_INVALID_WORK_GROUP_SIZE
+    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Query the maximum number of work groups for a cooperative kernel"
+class: $xKernel
+name: SuggestMaxCooperativeGroupCount
```
kbenzie, line 76:
```suggestion
name: SuggestMaxCooperativeGroupCountExp
```

kbenzie (commented):
Mostly looks good, just need to add the `Exp` suffix to the experimental function names and run the `generate` target again.

kbenzie (approved):

callumfare (approved):

scripts/core/exp-cooperative-kernels.yml
```diff
@@ -0,0 +1,85 @@
...
+    - $X_RESULT_ERROR_INVALID_WORK_DIMENSION
+    - $X_RESULT_ERROR_INVALID_WORK_GROUP_SIZE
+    - $X_RESULT_ERROR_INVALID_VALUE
+    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Query the maximum number of work groups for a cooperative kernel"
+class: $xKernel
+name: SuggestMaxCooperativeGroupCountExp
```
alycm, line 76: This is called `Suggest` when other queries are called `Get`.

Also, why not just make this a property queried via `urKernelGetInfo`?

scripts/core/exp-cooperative-kernels.yml
```diff
@@ -0,0 +1,85 @@
...
+desc: |
+      The extension string which defines support for cooperative-kernels
+      which is returned when querying device extensions.
+name: $X_COOPERATIVE_KERNELS_EXTENSION_STRING_EXP
+value: "\"$x_exp_cooperative_kernels\""
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Enqueue a command to execute a cooperative kernel"
+class: $xEnqueue
+name: CooperativeKernelLaunchExp
```
alycm, line 25: This function is identical to `urKernelLaunch`.  The actual difference is in the kernel itself?  And given the `urSuggestMaxCooperativeGroupCountExp` entry point we can presumably extract that info from the kernel.

Why not have the use of co-operative groups be a property of the kernel and just reuse `urKernelLaunch`?

veselypeta (approved):

scripts/core/exp-cooperative-kernels.yml
alycm: I see, I missed the distinction that device code may select a different option based on how it was launched.  I'll resolve this convo.

scripts/core/exp-cooperative-kernels.yml
alycm: > I named it Suggest to match Level Zero's zeKernelSuggestMaxCooperativeGroupCount. I'm not opposed to renaming it if we agree that another name would be more clear.

I suppose the difference in naming is that the result is a suggestion, not a firm property or limit.  But given this information:

> The main reason is that it's not intended to be a kernel property, but rather a limit for a kernel-device pair. I think urKernelGetGroupInfo is the most similar since it takes a kernel and a device.

It seems there may still be things to learn about implementing this in future iterations of this experimental feature.

Thank you for the extra context, I'm just going to mark this comment as resolved as well.

PR #852, opened 2023-09-08
[Bindless][Exp] Create layered image properties struct

isaacault:
To enable support for layered images, pass this struct via `pNext` when using the experimental feature to create an image. The struct is used to determine number of layers in the layered image.


veselypeta (approved):
LGTM

aarongreig (approved):

PR #853, opened 2023-09-11
[umf] Fix macOS build

kbenzie:
The AppleClang compiler emits warnings in C functions with zero arguments that are not specified as `f(void)`.


pbalcer (approved):
this should fail on any compiler with `-Wall` enabled. There's still something wrong with our cmakes... :/

kbenzie:
> this should fail on any compiler with `-Wall` enabled. There's still something wrong with our cmakes... :/

this was also picked up in the github actions, not sure how it creeped in to main

PR #854, opened 2023-09-11
Clarify the behaviour of urUSMMemAdvise and urUSMPrefetch

fabiomestre:
Clarify the behaviour of urUSMMemAdvise and urUSMPrefetch when the hint is unsupported

### DIFF
--- a/scripts/core/enqueue.yml
+++ b/scripts/core/enqueue.yml
@@ -1088,6 +1088,9 @@ desc: "Enqueue a command to prefetch USM memory"
 class: $xEnqueue
 name: USMPrefetch
 ordinal: "0"
+details:
+    - "Prefetching may not be supported for all devices or allocation types. If memory prefetching 
+    is not supported, the prefetch hint will be ignored."
 params:
     - type: $x_queue_handle_t
       name: hQueue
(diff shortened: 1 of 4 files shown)


veselypeta (approved):

scripts/core/enqueue.yml
```diff
@@ -1088,6 +1088,9 @@ desc: "Enqueue a command to prefetch USM memory"
 class: $xEnqueue
 name: USMPrefetch
 ordinal: "0"
+details:
+    - "Prefetching may not be supported for all devices or allocation types. If memory prefetching 
+    is not supported, the prefetch hint will be ignored."
```
pbalcer, line 1093: That's not entirely true though. The hint will be ignored, but the function will also return `UR_RESULT_ERROR_ADAPTER_SPECIFIC`. At least that's what the current implementations do. Generally, as a user, I wouldn't expect something documented as an ignored hint to cause an error. The CPU prefetch instruction doesn't rise an exception whenever it's ignored.
If it's valuable for software to know whenever prefetch is ignored, I think that scenario should be instead documented in the `returns` section.

fabiomestre: The reason for this PR is to try to make the error codes and behaviours consistent between adapters. Unless I'm mistaken it's also only Cuda that returns UR_RESULT_ERROR_ADAPTER_SPECIFIC while level zero always return UR_RESULT_SUCCESS. If this gets merged, then the Cuda adapter will have to be updated to always return success.

In addition, as far as I understand, UR_RESULT_ERROR_ADAPTER_SPECIFIC should be used only when the error is coming directly from Cuda, Level Zero, etc. An alternative would be to use a different error code.

I haven't merged this yet because I'm still unsure of what is the right way to fix it. When discussing this internally, it seemed that a few people were in favour of just returning UR_RESULT_SUCCESS when the hint is ignored. But I'm open to other suggestions!

pbalcer: HIP also returns an error. Anyway, good - just wanted to make sure I understand the intention behind the change. Thanks.

My preference would be to always return `UR_RESULT_SUCCESS` unless the API is used incorrectly. This would match how CPU behaves when prefetch is called on, e.g., uncacheable memory.

pbalcer (approved):

PR #855, opened 2023-09-12
[UMF] Fix -Wempty-body warning

veselypeta:
When updating the tag of UR in SYCL I got the -Wempty-body warning emitted from UMF. This PR just places braces around the do statement to prevent emitting this warning.

### DIFF
--- a/cmake/helpers.cmake
+++ b/cmake/helpers.cmake
@@ -63,6 +63,7 @@ function(add_ur_target_compile_options name)
             -fPIC
             -Wall
             -Wpedantic
+            -Wempty-body
             $<$<CXX_COMPILER_ID:GNU>:-fdiagnostics-color=always>
             $<$<CXX_COMPILER_ID:Clang,AppleClang>:-fcolor-diagnostics>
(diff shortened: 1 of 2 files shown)


kbenzie:
> When updating the tag of UR in SYCL I got the -Wempty-body warning emitted from UMF. This PR just places braces around the do statement to prevent emitting this warning.

I feel like this patch should also enable the `-Wempty-body` warning.

igchor (approved):

cmake/helpers.cmake
```diff
@@ -63,6 +63,8 @@ function(add_ur_target_compile_options name)
             -fPIC
             -Wall
             -Wpedantic
+            -Wempty-body
+            -Wunused-parameter
```
kbenzie, line 67: While I think adding `-Wunused-paramter` is a good idea, perhaps that should be a separate PR since its exposed additional warnings.

kbenzie (approved):

pbalcer (approved):

PR #857, opened 2023-09-13
Refactor and add checks for optional parameters

### LINKED ISSUE #856: Extend validation layers to check for INVALID_ENUMERATION when the flags are inside a properties struct
At the moment we can check for invalid enumeration if the flag is a parameter (e.g. `UR_MAP_FLAGS_MASK & mapFlags` in urEnqueueMemBufferMap). However, when the flags are hidden behind a properties variable (e.g. urQueueCreate) this feature doesn't work.

The aim of this issue is to extend validation layers support for the second case.

fabiomestre:
Refactor code generation for validation layers and add checks for optional parameters
    
Closes #856


kbenzie (approved):
LGTM

martygrant (approved):

PR #861, opened 2023-09-15
[UR] Check for unused parameters

veselypeta:
Since UR is imported as a library to different project, we should make sure we don't cause warnings with their compiler. A common warning is for unused parameters so we should ensure we don't raise this warning when compiling the code.

### DIFF
--- a/test/loader/platforms/platforms.cpp
+++ b/test/loader/platforms/platforms.cpp
@@ -18,7 +18,7 @@
 using namespace logger;
 
 //////////////////////////////////////////////////////////////////////////
-int main(int argc, char *argv[]) {
+int main([[maybe_unused]] int argc, [[maybe_unused]] char *argv[]) {
     logger::init("TEST");
 
     ur_result_t status;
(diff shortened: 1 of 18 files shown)


martygrant (approved):

test/loader/platforms/platforms.cpp
```diff
@@ -18,7 +18,7 @@
 using namespace logger;
 
 //////////////////////////////////////////////////////////////////////////
-int main(int argc, char *argv[]) {
+int main([[maybe_unused]] int argc, [[maybe_unused]] char *argv[]) {
```
kbenzie, line 21: I think in this case we should just remove these arguments instead.

kbenzie (approved):

pbalcer (commented):
There's no need to use `[[maybe_unused]]`. You can just omit the parameter name.

veselypeta:
> There's no need to use `[[maybe_unused]]`. You can just omit the parameter name.

Do you mean in every occurrence or just the one Benie mentioned above? I thinking it `ur_params.hpp` it is necessary to use `[[maybe_unused]]` since the parameter only sometimes not used. I also think that keeping the parameter names keeps the context of what the function is doing, removing the names might make it more difficult later on to decipher what the parameter is used for?

pbalcer (approved):

PR #863, opened 2023-09-15
[Bindless] Struct for unique addressing modes per dimension

Seanst98:
This MR adds a struct allowing for unique addressing modes per dimension

### DIFF
--- a/scripts/core/exp-bindless-images.yml
+++ b/scripts/core/exp-bindless-images.yml
@@ -110,6 +110,9 @@ etors:
     - name: EXP_LAYERED_IMAGE_PROPERTIES
       desc: $x_exp_layered_image_properties_t
       value: "0x2005"
+    - name: EXP_SAMPLER_ADDR_MODES
+      desc: $x_exp_sampler_addr_modes_t
+      value: "0x2006"
 --- #--------------------------------------------------------------------------
 type: enum
 extend: true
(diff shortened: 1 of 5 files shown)


veselypeta (approved):

kbenzie (approved):

scripts/core/exp-bindless-images.yml
```diff
@@ -176,6 +179,19 @@ members:
...
+desc: "Describes unique sampler addressing mode per dimension"
+details:
+    - Specify these properties in $xSamplerCreate via $x_sampler_desc_t as part
+      of a `pNext` chain.
+class: $xBindlessImages
+name: $x_exp_sampler_addr_modes_t
+base: $x_base_properties_t
+members:
+    - type: $x_sampler_addressing_mode_t
+      name: addrModes[3]
```
veselypeta, line 191: Unfortunatelly this won't work with the current generator. I think the simplest solution would be to just have three named fields for the addressing mode values.

scripts/core/exp-bindless-images.yml
```diff
@@ -176,6 +179,19 @@ members:
...
+      desc: "[in] Specify the address mode of the sampler per dimension"
```
veselypeta, line 192:
```suggestion
    - type: $x_sampler_addressing_mode_t
      name: addModeX
      desc: "[in] Specify the addressing mode of the x-dimension."
    - type: $x_sampler_addressing_mode_t
      name: addModeY
      desc: "[in] Specify the addressing mode of the y-dimension."
    - type: $x_sampler_addressing_mode_t
      name: addModeX
      desc: "[in] Specify the addressing mode of the x-dimension."
```

Seanst98: I've made the change.

scripts/core/exp-bindless-images.yml
veselypeta: I've created https://github.com/oneapi-src/unified-runtime/issues/869 to add this support.

Seanst98: Great, thanks! We'll keep an eye on this but will use the separate variables as a temporary solution.

Issue #828, opened 2023-08-24
Consider exposing serialization in public headers


aarongreig:
Being able to serialize API objects and functions is something many users of UR are going to want to do, and as it stands they have to do it themselves. We do have pretty much all the enum serialization etc you'd ever need in ur_params.hpp but it isn't a public interface. We should consider moving it out of `common` into `include` to allow deduplication of this kind of code

pbalcer:
We need to consider a few things before we make this a public interface:
1. Revise names. This was initially called `ur_params` because it was meant for printing out `_params_t` structures, which is really only useful in tracing. Now that we want to make it more broadly applicable, it might be a good idea to come up with a more descriptive name. `stringify`/`print`/`display`? 
We are also calling some functions in the header `serialize...`. This can be misleading, since a) we don't have a matching deserialize and b) for the most part, it wouldn't even be possible to deserialize what this is outputting since we don't adhere to any specific format (and the process is lossy anyway).
2. This header is huge (~16k lines at the moment). We should consider splitting it into smaller but still usable pieces.
3. The `operator<<` implementations are outside the namespace. This is for convenience because the types they are implemented for are also outside the namespace, so the compiler will not find the operator implementations automatically. If this is meant to be public, I think I'd rather have the operators inside of the namespace and require bringing it into the scope or just fully qualifying the calls.
4. Should we have a C API for this?

Luckily, the current implementation is header-only (I was wrong earlier), so that's one less thing to worry about.

pbalcer:
> I came here to suggest a `std::to_string` type interface

That would imply a memory allocation per displayed type. When implementing the logger, we wanted to implement a poor's man `std::format` using that same code, allowing you to do `auto ur_type_str = format("{}", some_ur_type);`. I'm not sure what happened to that. @PatKamin ?

kbenzie:
I'm not tied to the `std::to_string` interface but it would potentially be fewer allocations than interacting with the `std::stringstream` then converting to `std::string` before then manipuating that string.

> That would imply a memory allocation per displayed type.

If it were a C API would there ever be a need to return anything other than a `const char*` pointing to constant data in the executable?

PatKamin:
Part of this header could be still in common, unexposed to the user. All `serialize...` and `is_handle<>` functions are kind of helper functions for stream operators and as such they could be still internal to the project. This is leaving operators overload part to be exposed in `include` dir and in the API.

[...]

PR #875, opened 2023-09-20
Move serialization functions to include

PatKamin:
Related to issue #828 .

C API will be introduced in another PR.


scripts/templates/print.hpp.mako
```diff
@@ -106,7 +108,9 @@ def findMemberType(_item):
     %endif
 </%def>
 
-namespace ${x}_params {
+namespace ${x}_print {
+## API functions declarations #################################################
+// RFC: Should is_handle stay in the common dir?
```
pbalcer, line 113: as long as it's in a namespace, I don't see why not. Maybe we can move it to `details` (see below).

scripts/templates/print.hpp.mako
```diff
@@ -117,49 +121,49 @@ template <> struct is_handle<${th.make_type_name(n, tags, obj)}> : std::true_typ
 %endfor
 template <typename T>
 inline constexpr bool is_handle_v = is_handle<T>::value;
-template <typename T> inline void serializePtr(std::ostream &os, const T *ptr);
-template <typename T> inline void serializeFlag(std::ostream &os, uint32_t flag);
-template <typename T> inline void serializeTagged(std::ostream &os, const void *ptr, T value, size_t size);
+template <typename T> UR_DLLEXPORT inline ${x}_result_t UR_APICALL ${x}PtrPrint(std::ostream &os, const T *ptr);
```
pbalcer, line 124: Since this is a `print` header, my preference would be to name all these `urPrintSomething`. Makes it easier to find all print functions.

scripts/templates/print.hpp.mako
```diff
@@ -381,14 +389,14 @@ inline std::ostream &operator<<(std::ostream &os, [[maybe_unused]] const struct
 %endfor
 %endfor
 
-namespace ${x}_params {
-
-template <typename T> inline void serializePtr(std::ostream &os, const T *ptr) {
+///////////////////////////////////////////////////////////////////////////////
+// @brief Print pointer value
+template <typename T> inline ${x}_result_t ${x}PtrPrint(std::ostream &os, const T *ptr) {
```
pbalcer, line 394: This should not be part of the public API. Maybe we should have a `ur_print::details` namespace?

scripts/templates/print.hpp.mako
```diff
@@ -401,9 +409,13 @@ template <typename T> inline void serializePtr(std::ostream &os, const T *ptr) {
...
         os << ")";
     }
+
+    return ${X}_RESULT_SUCCESS;
 }
 
-inline int serializeFunctionParams(std::ostream &os, uint32_t function, const void *params) {
+///////////////////////////////////////////////////////////////////////////////
+// @brief Print function parameters
+UR_DLLEXPORT inline ${x}_result_t UR_APICALL ${x}FunctionParamsPrint(std::ostream &os, uint32_t function, const void *params) {
```
pbalcer, line 418: I'm not sure if this is useful outside of urtrace (but I guess someone might want to implement a similar functionality themselves).

```suggestion
UR_DLLEXPORT inline ${x}_result_t UR_APICALL ${x}FunctionParamsPrint(std::ostream &os, enum ur_function_t function, const void *params) {
```

scripts/templates/print.hpp.mako
PatKamin: I would prefer to call these functions this way, too. I just adhered to the function naming convention of UR API (https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/CONTRIB.rst#naming-convention). Can I ignore this convention in the case of a printing API then?

PatKamin: Renamed

scripts/templates/print.hpp.mako
PatKamin: Done

pbalcer (approved):

scripts/Doxyfile
```diff
@@ -848,7 +848,8 @@ RECURSIVE              = YES
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                = README.md
+EXCLUDE                = README.md \
+                         ../include/ur_print.hpp
```
pbalcer, line 852: Just occurred to me that these APIs won't have doxygen docs. We should eventually address this.

scripts/templates/print.hpp.mako
```diff
@@ -117,49 +121,50 @@ template <> struct is_handle<${th.make_type_name(n, tags, obj)}> : std::true_typ
 %endfor
 template <typename T>
 inline constexpr bool is_handle_v = is_handle<T>::value;
-template <typename T> inline void serializePtr(std::ostream &os, const T *ptr);
-template <typename T> inline void serializeFlag(std::ostream &os, uint32_t flag);
-template <typename T> inline void serializeTagged(std::ostream &os, const void *ptr, T value, size_t size);
+template <typename T> UR_DLLEXPORT inline ${x}_result_t UR_APICALL ${x}PrintPtr(std::ostream &os, const T *ptr);
```
igchor, line 124: Hm, do we actually need UR_DLLEXPORT? All those functions are defined in the header, right?

scripts/templates/print.hpp.mako
igchor: But we now have something like `ur_print::urFunctionParamsPrint`, which seems a bit excessive. Why not just `ur::printFunctionParams()` (i.e. rename namespace to `ur`, remove `ur` prefix from the function name and possibly start the function name with `print`)

pbalcer: `ur::print::functionParams()` or maybe `ur::extras::printFunctionParams()`?

I think I'd prefer this whole thing be in its own namespace. Hypothetically, we might want to add some more C++ functionality or bindings, and imho it should be possible to use that functionality without having to import all these operator overloads.

igchor: Well, you don't have to include ur_print.hpp header if you don't want to. But I like the `ur::extras::printFunctionParams()` or perhaps (`ur::info::printFunctionParams()`)

scripts/Doxyfile
PatKamin: Added generation of docs for all operators and a function from ::extras.

PatKamin:
> Hm, do we actually need UR_DLLEXPORT? All those functions are defined in the header, right? 

Replaced with `UR_APIEXPORT` for functions visibility on Windows. Also removed these macros from functions of the `::details` namespace.

PatKamin:
> Just occurred to me that these APIs won't have doxygen docs. We should eventually address this. 

Added generation of docs for all operators and a function from ::extras.

PatKamin:
> Well, you don't have to include ur_print.hpp header if you don't want to. But I like the `ur::extras::printFunctionParams()` or perhaps (`ur::info::printFunctionParams()`) 

Rearanged namespaces in the header.

kbenzie (approved):
Sorry I missed the review request last week. These changes are looking good 👍

pbalcer (approved):

Issue #859, opened 2023-09-14
Followup review feedback on e2e codegen example


kbenzie:
This issue is to track the review feedback [here](https://github.com/oneapi-src/unified-runtime/pull/610#pullrequestreview-1611102999) on the PR #610 which is noted as non-blocking but should be addressed post merge.

Issue #824, opened 2023-08-24
Platform conformance test hangs on L0 adapter


lukaszstolarczuk:
Platform fonromance test on L0 adapter (`platform-adapter_level_zero`) hangs for a long time, causing a timeout. For now, it's disabled on GHA runners (branch adapters).

Hang lasts at least 1,5 hours. Happened on a PC with Intel Arc A750.
Caught when running: https://github.com/oneapi-src/unified-runtime/pull/823

kbenzie:
Which test did it hang on exactly?

kbenzie:
looks like a deadlock

```console
Thread 1 "test-platform" received signal SIGINT, Interrupt.
0x00007ffff7908cab in sched_yield () at ../sysdeps/unix/syscall-template.S:120
120     ../sysdeps/unix/syscall-template.S: No such file or directory.
(gdb) bt
#0  0x00007ffff7908cab in sched_yield () at ../sysdeps/unix/syscall-template.S:120
#1  0x00007ffff7bb299d in urPlatformGet () from /home/benie/Projects/oneapi-src/unified-runtime/build-adapters-r/lib/libur_adapter_level_zero.so.0
#2  0x00007ffff7f841de in urPlatformGet () from /home/benie/Projects/oneapi-src/unified-runtime/build-adapters-r/lib/libur_loader.so.0
#3  0x00005555555776e8 in uur::platform::urPlatformsTest::SetUp() ()
#4  0x0000555555577a34 in uur::platform::urPlatformTest::SetUp() ()
#5  0x00005555555e14a1 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#6  0x00005555555cc6ed in testing::Test::Run() [clone .part.0] ()
#7  0x00005555555cd3e2 in testing::TestInfo::Run() ()
#8  0x00005555555cdc73 in testing::TestSuite::Run() [clone .part.0] ()
#9  0x00005555555d6094 in testing::internal::UnitTestImpl::RunAllTests() ()
#10 0x00005555555cd5c5 in testing::UnitTest::Run() ()
#11 0x0000555555570283 in main ()
(gdb) info threads
  Id   Target Id                                         Frame
* 1    Thread 0x7ffff7e41740 (LWP 36382) "test-platform" 0x00007ffff7908cab in sched_yield () at ../sysdeps/unix/syscall-template.S:120
[...]

PR #876, opened 2023-09-20
Followup review feedback on e2e codegen example

omarahmed1111:
This PR is a follow up PR to address the rest of review [feedback](https://github.com/oneapi-src/unified-runtime/pull/610#pullrequestreview-1611102999) on [610](https://github.com/oneapi-src/unified-runtime/pull/610). It contains:
* readme file to instruct how to run the codegen example.
* filter out adapters that doesn't support spirv kernels from codegen example.
* print the input array to the spv kernel in the codegen example.

Fixes: [859](https://github.com/oneapi-src/unified-runtime/issues/859)

### DIFF
--- a/examples/codegen/README.md
+++ b/examples/codegen/README.md
@@ -1,0 +1,35 @@
+## End-to-end code generation example
+
+This example demonstrates an end-to-end pipeline of code generation and execution that typically occurs in a JIT engine.
+It creates a simple function (adding +1 to an array and storing the result in a second one) via llvm API,
+converts it to spirv, and submits it to the runtime.
+
+## Running the example
+
+The following commands are executed from the project root directory.
+
+### Using conda environment
+
+```
+$ conda env create -f third_party/deps.yml
+$ conda activate examples
+$ mkdir build && cd build
+# configure with:
+$ cmake .. -DUR_BUILD_ADAPTER_L0=ON -DUR_BUILD_EXAMPLE_CODEGEN=ON
+$ make
+$ ./bin/codegen
+```
+
+### Without using conda
+
+To run the example with llvm 13, you will need to make sure that `llvm-13` and `libllvmspirvlib-13-dev` are installed.  
+
+**NOTE**: The example could be working with other llvm versions but it was tested with version 13.
+
+```
+$ mkdir build && cd build
+# configure with:
+$ cmake .. -DUR_BUILD_ADAPTER_L0=ON -DUR_BUILD_EXAMPLE_CODEGEN=ON -DLLVM_DIR=/usr/lib/llvm-13/cmake
+$ make
+$ ./bin/codegen
+```
--- a/examples/codegen/codegen.cpp
+++ b/examples/codegen/codegen.cpp
@@ -20,6 +20,8 @@
 #include "helpers.h"
 #include "ur_api.h"
 
+constexpr unsigned PAGE_SIZE = 4096;
+
 void ur_check(const ur_result_t r) {
     if (r != UR_RESULT_SUCCESS) {
         urTearDown(nullptr);
@@ -38,6 +40,23 @@ std::vector<ur_adapter_handle_t> get_adapters() {
     std::vector<ur_adapter_handle_t> adapters(adapterCount);
     ur_check(urAdapterGet(adapterCount, adapters.data(), nullptr));
     return adapters;
+}
+
+std::vector<ur_adapter_handle_t>
+get_supported_adapters(std::vector<ur_adapter_handle_t> &adapters) {
+    std::vector<ur_adapter_handle_t> supported_adapters;
+    for (auto adapter : adapters) {
+        ur_adapter_backend_t backend;
+        ur_check(urAdapterGetInfo(adapter, UR_ADAPTER_INFO_BACKEND,
+                                  sizeof(ur_adapter_backend_t), &backend,
+                                  nullptr));
+
+        if (backend == UR_ADAPTER_BACKEND_LEVEL_ZERO) {
+            supported_adapters.push_back(adapter);
+        }
+    }
+
+    return supported_adapters;
 }
 
 std::vector<ur_platform_handle_t>
@@ -70,7 +89,7 @@ std::vector<ur_device_handle_t> get_gpus(ur_platform_handle_t p) {
     return devices;
 }
 
-template <typename T, size_t N> struct alignas(4096) AlignedArray {
+template <typename T, size_t N> struct alignas(PAGE_SIZE) AlignedArray {
     T data[N];
 };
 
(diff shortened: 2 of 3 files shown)


omarahmed1111:
> is this for page alignment?
> getpagesize() ?

It would make since that this is related to gpus page size, but not sure how to make sure of that. I think it will need to be generalized if it will run on anything other than L0.

omarahmed1111:
> Now that we have GPU-equipped runners, it should be possible to create a workflow where this example is built and run as part of tests. Otherwise, it will be easy to accidentally break it.

I would think that would still be blocked by this [issue](https://github.com/oneapi-src/unified-runtime/issues/824)

examples/codegen/codegen.cpp
```diff
@@ -40,6 +40,24 @@ std::vector<ur_adapter_handle_t> get_adapters() {
...
+get_supported_adapters(std::vector<ur_adapter_handle_t> &adapters) {
+    std::vector<ur_adapter_handle_t> supported_adapters;
+    size_t size = 0;
+    for (auto adapter : adapters) {
+        ur_adapter_backend_t backend;
+        ur_check(urAdapterGetInfo(adapter, UR_ADAPTER_INFO_BACKEND,
+                                  sizeof(ur_adapter_backend_t), &backend,
+                                  nullptr));
+
+        if (backend == UR_ADAPTER_BACKEND_LEVEL_ZERO) {
```
omarahmed1111, line 53: I filtered that on only l0 as I think the other adapter that would accept spirv is opencl and that is still not merged in intel/llvm so we could add it later.

examples/codegen/codegen.cpp
```diff
@@ -79,7 +97,8 @@ int main() {
     ur_check(urInit(UR_DEVICE_INIT_FLAG_GPU, loader_config));
 
     auto adapters = get_adapters();
-    auto platforms = get_platforms(adapters);
+    auto supported_adapters = get_supported_adapters(adapters);
+    auto platforms = get_platforms(supported_adapters);
     auto gpus = get_gpus(platforms.front());
```
pbalcer, line 102: You need to check if the platforms aren't empty. [Calling `front()` on an empty container is UB](https://en.cppreference.com/w/cpp/container/vector/front).

examples/codegen/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Using conda envrionment
+
+```
+$ conda env create -f third_party/deps.yml
+$ conda activate examples
+# configure with:
+$ cmake .. -DUR_BUILD_ADAPTER_L0=ON -DUR_BUILD_EXAMPLE_CODEGEN=ON
+$ make
+$ UR_ADAPTERS_FORCE_LOAD=./lib/libur_adapter_level_zero.so ./bin/codegen
```
pbalcer, line 17: force loading the adapter shouldn't be needed.

pbalcer:
> > is this for page alignment?
> > getpagesize() ?
> 
> It would make since that this is related to gpus page size, but not sure how to make sure of that. I think it will need to be generalized if it will run on anything other than L0.

I guess 4k is a safe enough value... I'd still do a define/const variable instead of having a magic number though.

pbalcer:
> > Now that we have GPU-equipped runners, it should be possible to create a workflow where this example is built and run as part of tests. Otherwise, it will be easy to accidentally break it.
> 
> I would think that would still be blocked by this [issue](https://github.com/oneapi-src/unified-runtime/issues/824), so it will be hard to test that while failure, I think that could be added as a seperate PR with a seperate issue when the L0 pipeline is fixed.

The L0 workflow is not blocked, that test is simply not ran in CI (plus there's a timeout). But I agree that it might make sense to do this in another PR.

examples/codegen/README.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+### Without using conda
+
+To run the example with llvm 13, you will need to make sure that `llvm-13` and `libllvmspirvlib-13-dev` are installed.  
+
+**NOTE**: The example could be working with other llvm versions but it was tested with version 13.
+
+```
```
lukaszstolarczuk, line 26: Should these commands be executed in the current working dir or in the top-level/root repo dir?

The same question to the above commands - I can see you're using local dir `third_party/deps.yml`, but I believe CMake shouldn't be executed in this example's directory

examples/codegen/codegen.cpp
kbenzie: If there are no platforms `get_platforms()` will `throw std::runtime_error("No platforms available.");` so I think its okay?

pbalcer: right, I haven't noticed that.

examples/codegen/README.md
omarahmed1111: fixed that

examples/codegen/README.md
omarahmed1111: `third_party/deps.yml` should be only to initialize the conda environment and not being part of cmake.  I have added a clarification to note where this should be executed. let me know if it still have any ambiguity.

lukaszstolarczuk: ok, thanks for the clarification - no ambiguity :)

Nonetheless, I thought this `deps.yml` was a local file (in `examples` dir), but it's not.
Is it only prepared/tested with examples? Or should it work with all kinds of development/testing with Unified Runtime? // I'm asking because I'm not sure if we should add some kind of comment there, perhaps...?

omarahmed1111:
> > > Now that we have GPU-equipped runners, it should be possible to create a workflow where this example is built and run as part of tests. Otherwise, it will be easy to accidentally break it.
> > 
> > 
> > I would think that would still be blocked by this [issue](https://github.com/oneapi-src/unified-runtime/issues/824), so it will be hard to test that while failure, I think that could be added as a seperate PR with a seperate issue when the L0 pipeline is fixed.
> 
> The L0 workflow is not blocked, that test is simply not ran in CI (plus there's a timeout). But I agree that it might make sense to do this in another PR.

I created this [issue](https://github.com/oneapi-src/unified-runtime/issues/882) to track this. I was trying to attach it to this pr but it seems that it will not work directly and might be controversial in some parts. so, will submit it to a new pr.

examples/codegen/README.md
omarahmed1111: atm, it is used only to run the codegen example, but theoretically, I guess it should work if we use it to initialize any conda environment to run unified runtime, however i think it have a lot of extra packages that only codegen uses but UR doesn't need. I think it is better to add this comment for now and if the need differ for this file in the future we could change that comment.

omarahmed1111: I added a comment at the beginning of the file to pinpoint that.

pbalcer (approved):

examples/codegen/codegen.cpp
```diff
@@ -20,6 +20,8 @@
 #include "helpers.h"
 #include "ur_api.h"
 
+const unsigned PAGE_SIZE = 4096;
```
veselypeta, line 23: `constexpr`?

veselypeta (approved):

examples/codegen/codegen.cpp
omarahmed1111: Done

PR #877, opened 2023-09-20
CTS fixes for ProgramLink and ProgramCreateWithBinary

aarongreig:
We now no longer attempt to link two programs with duplicate definitions together (the SPIR-V dpc++ generates for us always contains a wrapper function with the same name and function signature)

We also now correctly query program binaries

### DIFF
--- a/test/conformance/program/urProgramCreateWithBinary.cpp
+++ b/test/conformance/program/urProgramCreateWithBinary.cpp
@@ -9,6 +9,11 @@ struct urProgramCreateWithBinaryTest : uur::urProgramTest {
     void SetUp() override {
         UUR_RETURN_ON_FATAL_FAILURE(urProgramTest::SetUp());
         ASSERT_SUCCESS(urProgramBuild(context, program, nullptr));
+        size_t binary_sizes_len = 0;
+        ASSERT_SUCCESS(urProgramGetInfo(program, UR_PROGRAM_INFO_BINARY_SIZES,
+                                        0, nullptr, &binary_sizes_len));
+        // We're expecting one binary
+        ASSERT_EQ(binary_sizes_len / sizeof(size_t), 1);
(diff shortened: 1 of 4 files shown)


test/conformance/program/urProgramCreateWithBinary.cpp
```diff
@@ -10,11 +10,14 @@ struct urProgramCreateWithBinaryTest : uur::urProgramTest {
         UUR_RETURN_ON_FATAL_FAILURE(urProgramTest::SetUp());
         ASSERT_SUCCESS(urProgramBuild(context, program, nullptr));
         size_t binary_size = 0;
-        ASSERT_SUCCESS(urProgramGetInfo(program, UR_PROGRAM_INFO_BINARIES, 0,
-                                        nullptr, &binary_size));
+        ASSERT_SUCCESS(urProgramGetInfo(program, UR_PROGRAM_INFO_BINARY_SIZES,
```
omarahmed1111, line 13: I guess this should be returning a a `size_t[]` so we maybe could expand the test by making sure the returned array size is only one, but not sure if that is adding value tbh, so feel free to ignore that.

callumfare (approved):
LGTM, although it would be good to add a linking test where `count > 1` at some point

omarahmed1111 (approved):
LGTM

aarongreig:
> LGTM, although it would be good to add a linking test where `count > 1` at some point

agreed but I think it might require a fairly drastic change to the device binary infrastructure, I'll make an issue for it

test/conformance/program/urProgramCreateWithBinary.cpp
aarongreig: good shout, added a check before this that just queries the size and makes sure it's in line with the `size_t[1]` we're expecting

PR #879, opened 2023-09-21
Move L0 Sources from intel/llvm to UR.

veselypeta:
We are moving the L0 adapter code from upstream intel/llvm into unified runtime to allow the adapters to be easily built independent of llvm.


pbalcer:
Is the commit history correct? Do these commits actually show correct changes if you do `git diff`? If not I'd squash them all. In the description we can include everyone in `Co-authored-by:` and also have a link to intel/llvm with the valid commit history.

veselypeta:
> Is the commit history correct? Do these commits actually show correct changes if you do `git diff`? If not I'd squash them all. In the description we can include everyone in `Co-authored-by:` and also have a link to intel/llvm with the valid commit history.

We want to try and preserve the git history of the L0 plugin as much as possible when moving the code to unified runtime. I got these changes by creating a subtree from the intel/llvm repository and pulling the changes into unified runtime.

kbenzie:
GitHub doesn't render this but `git show <commit>` shows the patches as they were introduced into intel/llvm. In general I'm really not a fan of throwing away history with squashing.

pbalcer:
> GitHub doesn't render this but `git show <commit>` shows the patches as they were introduced into intel/llvm. In general I'm really not a fan of throwing away history with squashing.

Typically, yes - sure. But this is merging two unrelated histories, with patches that might have touched things outside of this tree. In other words, ur will have history from another tree with semi-invalid commits that might break tooling (does this break bisect?) or just be misleading if you are looking at it manually (if the commits function with `git show` then I guess it's not that bad). Having a single squashed commit with a link to the place with the actual good commit history is simpler and imho a bit more useful.

jandres742:
Adding the whole history is good if we want to track down guilty changes in automated way or with something like git blame. A squash would make that difficult. However, I dont want to put too much burden on @veselypeta trying to do that, so if it is possible, great, if not (because of too many conflicts or changes needed to make it work), then maybe having a squash is more practical

source/adapters/level_zero/CMakeLists.txt
```diff
@@ -73,36 +72,36 @@ target_include_directories(LevelZeroLoader-Headers
 
 add_ur_adapter(${TARGET_NAME}
     SHARED
-    ${L0_DIR}/ur_interface_loader.cpp
```
jandres742, line 76: @veselypeta , @kbenzie : in intel/llvm, we fetch the L0 loader with a specific tag for the headers:

sycl/plugins/level_zero/CMakeLists.txt
```cpp
  set(LEVEL_ZERO_LOADER_REPO "https://github.com/oneapi-src/level-zero.git")
  set(LEVEL_ZERO_LOADER_TAG v1.11.0)
```

I dont see it here. If we added it here then now we need to be careful that both unified-runtime and intel/llvm keep the same L0 loader tag always. Or maybe https://github.com/intel/llvm/pull/11252 can have a change where the L0 loader tag used by the unified-runtime is grabbed in a CMake file in intel/llvm

@smaslov-intel : please see above.

pbalcer: Fetching L0 loader already exists in the same cmake a few lines above: https://github.com/oneapi-src/unified-runtime/blob/adapters/source/adapters/level_zero/CMakeLists.txt#L11

jandres742: ah ok, thanks @pbalcer

source/adapters/level_zero/ur_interface_loader.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+UR_DLLEXPORT ur_result_t UR_APICALL urGetUSMExpProcAddrTable(
+    ur_api_version_t version, ur_usm_exp_dditable_t *pDdiTable) {
+  auto result = validateProcInputs(version, pDdiTable);
+  if (UR_RESULT_SUCCESS != result) {
+    return result;
+  }
+  pDdiTable->pfnPitchedAllocExp = urUSMPitchedAllocExp;
+  return UR_RESULT_SUCCESS;
+}
```
callumfare, line 392: This file is missing definitions for `urGetPhysicalMemProcAddrTable` and `urGetVirtualMemProcAddrTable`, which I saw causing problems with the build on intel/llvm because the adapter .def file requires them. I can follow up with a PR if you don't want to add them at the same time

callumfare (approved):

source/adapters/level_zero/ur_interface_loader.cpp
veselypeta: Yeah I think a follow up PR would be better since this one should just move the code from intel/llvm. How did we manage to build before when we were pulling in the files here?

callumfare: Makes sense :+1: I'm not sure how it worked before, it might only be a problem with intel/llvm because of the way the adapter library gets pulled in

kbenzie (approved):

smaslov-intel:
Are you following up with removal of https://github.com/intel/llvm/blob/sycl/sycl/plugins/unified_runtime/ur/adapters/level_zero/ so the 2 versions do not exist?

veselypeta:
> Are you following up with removal of https://github.com/intel/llvm/blob/sycl/sycl/plugins/unified_runtime/ur/adapters/level_zero/ so the 2 versions do not exist?

See here https://github.com/intel/llvm/pull/11252

smaslov-intel:
> > Are you following up with removal of https://github.com/intel/llvm/blob/sycl/sycl/plugins/unified_runtime/ur/adapters/level_zero/ so the 2 versions do not exist?
> 
> See here [intel/llvm#11252](https://github.com/intel/llvm/pull/11252)

Cool, thanks!

Issue #882, opened 2023-09-21
Add a test workflow for codegen example


omarahmed1111:
follow up on this [comment](https://github.com/oneapi-src/unified-runtime/pull/876#issuecomment-1729126110), we should add a workflow to prevent codegen example from being broken silently.

PR #884, opened 2023-09-21
Add test workflow for codegen example

omarahmed1111:
fixes [882](https://github.com/oneapi-src/unified-runtime/issues/882)


.github/workflows/cmake.yml
```diff
@@ -211,6 +211,51 @@ jobs:
       working-directory: ${{github.workspace}}/build
       run: ctest -C ${{matrix.build_type}} --output-on-failure -L "conformance" --timeout 180
 
+  examples-build-hw:
+    name: Build - examples on HW
+    if: github.repository == 'oneapi-src/unified-runtime'  # run only on upstream; forks won't have the HW
+    strategy:
+      matrix:
+        adapter: [
+          {name: L0, triplet: spir64}
```
lukaszstolarczuk, line 220: `triplet` is not used

.github/workflows/cmake.yml
```diff
@@ -211,6 +211,51 @@ jobs:
...
+    - name: Configure CMake
+      shell: bash -el {0}
+      run: >
+        cmake
+        -B${{github.workspace}}/build
+        -DCMAKE_C_COMPILER=${{matrix.compiler.c}}
+        -DCMAKE_CXX_COMPILER=${{matrix.compiler.cxx}}
+        -DCMAKE_BUILD_TYPE=${{matrix.build_type}}
+        -DUR_BUILD_ADAPTER_${{matrix.adapter.name}}=ON
+        -DUR_BUILD_EXAMPLE_CODEGEN=ON
```
lukaszstolarczuk, line 250: this seems like the only place that build this example, so I'd turn `UR_DEVELOPER_MODE` on.

.github/workflows/cmake.yml
omarahmed1111: fixed

.github/workflows/cmake.yml
omarahmed1111: added that.

lukaszstolarczuk (approved):

callumfare (approved):

PR #892, opened 2023-09-26
[L0] Add missing loader interfaces to L0 adapter

callumfare:
These symbols are required to be exported by the .def file on Windows

### DIFF
--- a/source/adapters/level_zero/ur_interface_loader.cpp
+++ b/source/adapters/level_zero/ur_interface_loader.cpp
@@ -390,3 +390,41 @@ UR_DLLEXPORT ur_result_t UR_APICALL urGetUSMExpProcAddrTable(
   pDdiTable->pfnPitchedAllocExp = urUSMPitchedAllocExp;
   return UR_RESULT_SUCCESS;
 }
+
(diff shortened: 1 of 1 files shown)


pbalcer (approved):
This file looks like it could be autogenerated, and it also seems identical in all the adapters.

kbenzie (approved):

kbenzie:
> This file looks like it could be autogenerated, and it also seems identical in all the adapters.

I think if it were autogenerated and the entry-points did not exist we'd get linker errors. Being able to assign `nullptr` to the function pointers at least allows each adapter to opt in to defining entry-points.

pbalcer:
> > This file looks like it could be autogenerated, and it also seems identical in all the adapters.
> 
> I think if it were autogenerated and the entry-points did not exist we'd get linker errors. Being able to assign `nullptr` to the function pointers at least allows each adapter to opt in to defining entry-points.

Good point. We could probably make the script look for function definitions (shouldn't be hard to grep for) to decide whether to assign null, but that's probably not worth it.

veselypeta (approved):

PR #896, opened 2023-09-26
Move CUDA sources from intel/llvm to UR

fabiomestre:
This PR moves the CUDA adapter for the intel/llvm repository to the unified runtime repository. This was done using git subtrees to preserve the commit history. 

A follow-up PR on intel/llvm will remove the sources from that repo.


source/adapters/cuda/CMakeLists.txt
```diff
@@ -3,43 +3,43 @@
...
-    ${CUDA_DIR}/queue.hpp
-    ${CUDA_DIR}/queue.cpp
-    ${CUDA_DIR}/sampler.hpp
-    ${CUDA_DIR}/sampler.cpp
-    ${CUDA_DIR}/tracing.cpp
-    ${CUDA_DIR}/usm.cpp
-    ${CUDA_DIR}/usm_p2p.cpp
-    ${CUDA_DIR}/../../ur.cpp
-    ${CUDA_DIR}/../../ur.hpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/ur_interface_loader.cpp
```
callumfare, line 12: `image.cpp`/`image.hpp` are missing from the CMake

source/adapters/cuda/ur_interface_loader.cpp
```diff
@@ -0,0 +1,355 @@
...
+UR_DLLEXPORT ur_result_t UR_APICALL urGetUSMExpProcAddrTable(
+    ur_api_version_t version, ur_usm_exp_dditable_t *pDdiTable) {
+  auto result = validateProcInputs(version, pDdiTable);
+  if (UR_RESULT_SUCCESS != result) {
+    return result;
+  }
+  pDdiTable->pfnPitchedAllocExp = urUSMPitchedAllocExp;
+  return UR_RESULT_SUCCESS;
+}
+
```
callumfare, line 352: This file is missing `urGetPhysicalMemProcAddrTable` and `urGetVirtualMemProcAddrTable`, this caused problems with the L0 adapter when it got included in intel/llvm - see #892. It's fine if all the actual entries in the table are `nullptr`.

They could be added now or in a follow up PR

callumfare (approved):

source/adapters/cuda/CMakeLists.txt
```diff
@@ -67,6 +67,12 @@ else()
   )
 endif()
 
+if(UMF_ENABLE_POOL_TRACKING)
+  target_compile_definitions("ur_adapter_cuda" PRIVATE UMF_ENABLE_POOL_TRACKING)
```
veselypeta, line 71: Should we add this option in the readme?

source/adapters/cuda/adapter.cpp
```diff
@@ -0,0 +1,89 @@
+//===--------- adapter.cpp - CUDA Adapter ---------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
```
veselypeta, line 3: These licenses will need updating too, but a follow up PR is fine.

veselypeta (approved):

source/adapters/cuda/CMakeLists.txt
aarongreig: as far as I know the UMF team are reworking stuff so that this option isn't needed, hopefully its days are numbered

fabiomestre: This is already an option in UR. It was just not being used in the adapter

source/adapters/cuda/adapter.cpp
fabiomestre: I will add this in a follow up PR

source/adapters/cuda/CMakeLists.txt
fabiomestre: Adding this caused some compile errors (from werror). To avoid changing the adapter on this PR (which makes it hard to review), I will do this change on a follow up PR.

callumfare: Ok, makes sense!

PR #897, opened 2023-09-27
CI: Cancel stale workflows when new commits are added to a PR

callumfare:
We have a limited number of runners for the adapter testing jobs, so this helps avoid congestion

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -2,6 +2,10 @@ name: Build and test
 
 on: [push, pull_request]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   ubuntu-build:
     name: Build - Ubuntu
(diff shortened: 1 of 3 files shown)


omarahmed1111 (approved):

.github/workflows/cmake.yml
```diff
@@ -2,6 +2,10 @@ name: Build and test
 
 on: [push, pull_request]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
kbenzie, line 7: I guess this can't be restricted to only pull requests?

pbalcer: > I guess this can't be restricted to only pull requests?

:+1: We should probably build every merge on push just to make sure we can safely revert. If I understand the syntax, this should work:
`group: ${{ github.workflow }}-${{ github.event.pull_request.number}}`

callumfare: Thanks, I've changed it and checked that it cancels the PR workflows correctly

lukaszstolarczuk (commented):
You could do the same for Bandit. I know it takes only half a minute, but it looks weird when everything else but that one is canceled 😉

pbalcer (approved):

lukaszstolarczuk (approved):

PR #900, opened 2023-09-27
[CUDA] Update license headers and fix compilation issues

fabiomestre:
This changes are necessary as part of moving the Cuda adapter to the Unified Runtime repository:

- Update license headers for the CUDA adapter
- Add virtual memory entrypoints to ur_interface_loader.cpp
- Add missing source files to the CUDA adapter cmake
- Fix compilation error in image.cpp (uninitialized variable)

### DIFF
--- a/source/adapters/cuda/image.cpp
+++ b/source/adapters/cuda/image.cpp
@@ -1,7 +1,9 @@
 //===--------- image.cpp - CUDA Adapter -----------------------------------===//
 //
-// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
-// See https://llvm.org/LICENSE.txt for license information.
+// Copyright (C) 2023 Intel Corporation
+//
+// Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM
+// Exceptions. See LICENSE.TXT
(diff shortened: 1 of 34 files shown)


source/adapters/cuda/image.cpp
```diff
@@ -780,6 +780,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageGetInfoExp(
     case 4:
       ChannelOrder = UR_IMAGE_CHANNEL_ORDER_RGBA;
       break;
+    default:
```
pbalcer, line 783: Does this fail only on SYCL? The previous PR was built without issues. Is UR missing `-Wswitch-default` or something like that?

fabiomestre: The previous PR built without issues because image.cpp was missing from the CmakeLists. The error is a `[-Werror=maybe-uninitialized`  which can be seen failing here:

https://github.com/oneapi-src/unified-runtime/actions/runs/6326508321/job/17180224616

I  think that on sycl side, we were removing -werror for unified runtime.

martygrant (approved):

callumfare (approved):

PR #904, opened 2023-09-28
[UR][Tests] Update match scripts

lukaszstolarczuk:
Input and match files were printed only on one error. Now, it will be printed in both places (where this makes sense).

This is an extension to commit: be85a494bd

### DIFF
--- a/cmake/match.cmake
+++ b/cmake/match.cmake
@@ -75,7 +75,3 @@ if(TEST_RESULT)
 else()
     message("Passed: The output of test command '${TEST_FILE} ${TEST_ARGS}' (stored in ${CMAKE_CURRENT_BINARY_DIR}/${OUT_FILE}) matches '${MATCH_FILE}'")
 endif()
-
-if(EXISTS ${OUT_FILE})
-    file(REMOVE ${OUT_FILE})
-endif()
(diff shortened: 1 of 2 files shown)


cmake/match.cmake
```diff
@@ -71,9 +71,9 @@ execute_process(
 )
 
 if(TEST_RESULT)
-    message(FATAL_ERROR "Failed (${TEST_RESULT}): The output of ${OUT_FILE} does not match ${MATCH_FILE}")
-elseif()
-    message("Passed: The output ${OUT_FILE} matches ${MATCH_FILE}")
+    message(FATAL_ERROR "Failed (${TEST_RESULT}): The output of test command '${TEST_FILE} ${TEST_ARGS}' does not match '${MATCH_FILE}'")
+else()
+    message("Passed: The output of test command '${TEST_FILE} ${TEST_ARGS}' matches '${MATCH_FILE}'")
```
pbalcer, line 76: I actually prefer to have the tmp file here because I can easily navigate to it from the terminal.

lukaszstolarczuk: hmm... it wasn't even the full path:
```
Failed (1): The output of _matchtmpfile does not match
  /home/u105190/actions-runner/_work/unified-runtime/unified-runtime/test/conformance/sampler/sampler_adapter_level_zero.match
```

Perhaps we could have both - (full, if possible) path and the executed test command...

lukaszstolarczuk: Done.

pbalcer (approved):

Issue #862, opened 2023-09-15
[CTS] Add testing for memory migration across devices in same context


hdelan:
The CTS should add some testing to check that `ur_mem_handle_t_`s will automatically migrate memory from device to device within the same context.

Something like:

```
urEnqueueMemBufferWrite(// on Q1 in ctx
urEnqueueMemBufferRead(// on Q2 in ctx
// Wait and check that memory was migrated across queues
```
But with other entry points as well like kernel launches etc. Also for Image entry points.

jandres742:
> The CTS should add some testing to check that ur_mem_handle_t_s will automatically migrate memory from device to device within the same context.

why are we adding that to a CTS? is that specified in the spec as mandatory behavior? Migrating memory between two devices could be implementation specific. For instance, an adapter may choose not to migrate if the underlying implementation offers faster performance for peer-to-peer access, rather than migration.

hdelan:
My original comment is maybe poorly phrased. What the UR CTS ought to test for is whether memory in a UR buffer/image is available for multiple devices in a `ur_context_handle_t`. The method by which this should work is an implementation detail, and may rely on memory migration or by some other means such as P2P access. Nevertheless we should have some test which does something like:
```
// Using multi device ctx associated with queues {Q1, Q2}
urMemBufferCreate // Create buf
urEnqueueMemBufferWrite // Say on Q1
urEnqueueKernelLaunch // Write into buf on Q2
urEnqueueMemBufferRead // Read from buf on Q1 say. Make sure that the kernel on Q2 has had an effect on buffer.
```

jandres742:
> My original comment is maybe poorly phrased. What the UR CTS ought to test for is whether memory in a UR buffer/image is available for multiple devices in a `ur_context_handle_t`. The method by which this should work is an implementation detail, and may rely on memory migration or by some other means such as P2P access. Nevertheless we should have some test which does something like:
> 
> ```
> // Using multi device ctx associated with queues {Q1, Q2}
> urMemBufferCreate // Create buf
> urEnqueueMemBufferWrite // Say on Q1
> urEnqueueKernelLaunch // Modify contents of buf on Q2
> urEnqueueMemBufferRead // Read from buf on Q1 say. Make sure that the kernel on Q2 has had an effect on buffer.
> ```

ah, thanks for the explanation @hdelan .

PR #907, opened 2023-09-29
Add testing for memory management across multi-device contexts

omarahmed1111:
Fixes [862](https://github.com/oneapi-src/unified-runtime/issues/862)

### DIFF
--- a/test/conformance/testing/include/uur/fixtures.h
+++ b/test/conformance/testing/include/uur/fixtures.h
@@ -601,6 +601,128 @@ struct urMemImageQueueTest : urQueueTest {
                               0};                           // num samples
 };
 
+struct urMultiDeviceMemImageTest : urMultiDeviceContextTest {
+    void SetUp() override {
+        UUR_RETURN_ON_FATAL_FAILURE(urMultiDeviceContextTest::SetUp());
+        ASSERT_SUCCESS(urMemImageCreate(context, UR_MEM_FLAG_READ_WRITE,
+                                        &format, &desc1D, nullptr, &image1D));
+
+        ASSERT_SUCCESS(urMemImageCreate(context, UR_MEM_FLAG_READ_WRITE,
+                                        &format, &desc2D, nullptr, &image2D));
+
+        ASSERT_SUCCESS(urMemImageCreate(context, UR_MEM_FLAG_READ_WRITE,
+                                        &format, &desc3D, nullptr, &image3D));
+    }
+
+    void TearDown() override {
+        if (image1D) {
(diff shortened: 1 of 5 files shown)


test/conformance/testing/include/uur/fixtures.h
```diff
@@ -501,41 +500,22 @@ struct urMultiDeviceMemBufferQueueTest : urMultiDeviceMemBufferTest {
...
             EXPECT_SUCCESS(urQueueRelease(queue));
         }
-        UUR_RETURN_ON_FATAL_FAILURE(urMultiDeviceMemBufferTest::TearDown());
+        UUR_RETURN_ON_FATAL_FAILURE(urMemBufferQueueTest::TearDown());
     }
 
     std::vector<ur_queue_handle_t> queues;
 };
 
-struct urMemBufferQueueTest : urQueueTest {
```
omarahmed1111, line 510: I spotted this could be a  redundant code so wanted to address it, there is other redundant code, also added by this PR, that need to refactor the tests a little to eliminate it.

aarongreig (approved):
Looks good, did you get these tested with an adapter?

omarahmed1111:
> Looks good, did you get these tested with an adapter?

Yeah, when tested with multiple adapters passed, all multidevice tests get skipped. when tested with one adapter forced, it segfault.

jandres742 (approved):

kbenzie (approved):

omarahmed1111:
This PR will currently fail on l0 as the feature of migrating between multiple devices not added yet in l0. issue related: [10555](https://github.com/intel/llvm/issues/10555).

jandres742:
> This PR will currently fail on l0 as the feature of migrating between multiple devices not added yet in l0. issue related: [10555](https://github.com/intel/llvm/issues/10555). @jandres742 just wandering, do you think we could merge this now, or should we leave it till the migrating memory feature is added on l0 so not to segfault?

thanks @omarahmed1111 .  Im confused, why are we continue to refer to migration? As we discussed in https://github.com/oneapi-src/unified-runtime/issues/862#issuecomment-1754306441, migration is implementation specific. What this test is (or should be checking) if peer access, not migration.

In that sense, I will suggest to change the name of the commit message.

Now, it could be that this is failing on L0, since we dont have that support yet in https://github.com/intel/llvm/issues/10555#issuecomment-1652294917.

In this case, then my suggestion is to open a new issue here in the repo so we can track it in the UR L0 adapter and know we need to close that gap.

omarahmed1111:
[950](https://github.com/oneapi-src/unified-runtime/issues/950) created to follow up on l0 memory management feature.

PR #915, opened 2023-10-02
[UR][L0] Fix minor code duplication

al42and:
- Second return has no effect
- Checking the same flag twice has no effect

No functional changes expected

### DIFF
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -432,14 +432,12 @@ ur_result_t ur_context_handle_t_::finalize() {
   for (auto &List : ZeComputeCommandListCache) {
     for (auto &Item : List.second) {
       ze_command_list_handle_t ZeCommandList = Item.first;
-      if (ZeCommandList)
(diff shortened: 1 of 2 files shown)


pbalcer (approved):

lukaszstolarczuk (approved):

jandres742:
thanks @al42and 
 
could you modify commit title to have `[UR][L0]` as it is the convention we have been following on?

jandres742 (approved):

al42and:
@jandres742: Done!

Although, I am not sure that convention is helpful. We are in a dedicated UR repo, so isn't prefixing every commit message with `[UR]` a but redundant?

kbenzie:
> Although, I am not sure that convention is helpful. We are in a dedicated UR repo, so isn't prefixing every commit message with `[UR]` a bit redundant?

Up to now I've personally been using `[UR]` as the prefix for spec changes although this isn't well defined. I think we'll need to define the conventions in the contribution guide.

PR #916, opened 2023-10-02
[UR][L0] Check Global Mem Size as Limit for Free Memory

nrspruit:


nrspruit:
pre-commit checks in the draft PR https://github.com/intel/llvm/pull/11379

source/adapters/level_zero/device.cpp
```diff
@@ -658,7 +669,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
         }
       }
     }
-    return ReturnValue(FreeMemory);
+    return ReturnValue(std::min(GlobalMemSize, FreeMemory));
```
jandres742, line 672: why is that we want to return the min?

nrspruit: We want to return the size up to global memsize, not larger than.

The Free memory could exceed the Global Memory size since Global Memory Size is a calculation, but to avoid OOM condition, we want to have the minimum of the two ie GlobalMemSize > Free Memory, then return Free Memory as the Free Memory Available, but if GlobalMemSize < Free Memory, then return the Global Memory Size as the max "free memory" available.

jandres742:
Changes looking good. We will wait for pre-merge results from https://github.com/intel/llvm/pull/11379 before approving.

jandres742 (approved):

jandres742:
waiting for llvm testing before moving to merging.

fabiomestre:

kbenzie:
I'm going to create a combined intel/llvm PR which will include this and a few other PR's so we can get things merged quicker.

PR #917, opened 2023-10-02
[UR][L0][CMake] Disable warning for ICX on Windows

jsji:
On Windows we may build L0 using ICX , this is to revert changes to
check WIN32 instead of MSVC, to fix build failures with ICX on Windows.

### DIFF
--- a/source/adapters/level_zero/CMakeLists.txt
+++ b/source/adapters/level_zero/CMakeLists.txt
@@ -56,7 +56,8 @@ target_link_libraries(LevelZeroLoader
     INTERFACE "${LEVEL_ZERO_LIB_NAME}"
 )
 
-if (NOT MSVC)
+# Windows build might have warnings (both MSVC and ICX), disable Werror etc.
+if (NOT WIN32)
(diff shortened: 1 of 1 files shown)


smaslov-intel (approved):

igchor (approved):

pbalcer:
I'm not sure if I understand why this change fixes the build. Does ICX pretend to be MSVC for the purpose of CMake? I've been trying to approach this more holistically [here](https://github.com/pbalcer/unified-runtime/commit/4de252552f30d5a50c72778758446ea179798997). But I'm not sure if that would have helped.

According to [CMake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html), the identifier for Intel's compiler is `Intel` and `IntelLLVM`. Maybe we should be using these instead?

kbenzie:
> According to [CMake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html), the identifier for Intel's compiler is `Intel` and `IntelLLVM`. Maybe we should be using these instead?

I agree, using `WIN32` is `if` statements in CMake is the old way of doing things.

PR #920, opened 2023-10-03
[UR][L0] Copy prebuilt L0 to avoid leaking shared folder path

jsji:
We allow reusing existing pre-built L0 Libs and Headers.
However, using LEVEL_ZERO_LIBRARY and LEVEL_ZERO_INCLUDE_DIR directly
may leak the path of shared folder.

This PR introduces two L0_LIBRARY and L0_INCLUDE_DIR ,
and copy L0 loader/headers locally to the build to avoid leaking their
location on a shared drive.

Originally authored by @smaslov-intel .

### DIFF
--- a/source/adapters/level_zero/CMakeLists.txt
+++ b/source/adapters/level_zero/CMakeLists.txt
@@ -6,6 +6,19 @@
 
 set(TARGET_NAME ur_adapter_level_zero)
 
+# Copy L0 loader/headers locally to the build to avoid leaking their path.
+set(LEVEL_ZERO_COPY_DIR ${CMAKE_CURRENT_BINARY_DIR}/level_zero_loader)
+if (DEFINED L0_LIBRARY)
+  get_filename_component(LEVEL_ZERO_LIB_NAME "${L0_LIBRARY}" NAME)
+  set(LEVEL_ZERO_LIBRARY ${LEVEL_ZERO_COPY_DIR}/${LEVEL_ZERO_LIB_NAME})
+  message(STATUS "Copying Level Zero loader and headers to local build tree")
+  file(COPY ${L0_LIBRARY} DESTINATION ${LEVEL_ZERO_COPY_DIR} FOLLOW_SYMLINK_CHAIN)
+endif()
+if (DEFINED L0_INCLUDE_DIR)
+  set(LEVEL_ZERO_INCLUDE_DIR ${LEVEL_ZERO_COPY_DIR}/level_zero)
+  file(COPY ${L0_INCLUDE_DIR}/level_zero DESTINATION ${LEVEL_ZERO_COPY_DIR})
+endif()
+
 if (NOT DEFINED LEVEL_ZERO_LIBRARY OR NOT DEFINED LEVEL_ZERO_INCLUDE_DIR)
     message(STATUS "Download Level Zero loader and headers from github.com")
 


source/adapters/level_zero/CMakeLists.txt
```diff
@@ -6,6 +6,19 @@
 
 set(TARGET_NAME ur_adapter_level_zero)
 
+# Copy L0 loader/headers locally to the build to avoid leaking their path.
```
pbalcer, line 9: This looks to me like we are externalizing details of a specific CI pipeline. Given that you can already specify a path to level zero, would it be possible for this copying step to happen outside of the CMake scripts as one of the CI steps? And then you can point to this location using the `LEVEL_ZERO_LIBRARY` and `LEVEL_ZERO_INCLUDE_DIR` variables.

source/adapters/level_zero/CMakeLists.txt
```diff
@@ -57,7 +70,7 @@ target_link_libraries(LevelZeroLoader
 )
 
 # Windows build might have warnings (both MSVC and ICX), disable Werror etc.
-if (NOT WIN32)
+if (NOT WIN32 AND NOT DEFINED L0_LIBRARY)
```
pbalcer, line 73: Once #923 is merged, this will not be necessary.

source/adapters/level_zero/CMakeLists.txt
jsji: Thanks @pbalcer . Yes, doing it in CI is OK, however, we would still leak path if we build without CI.

kbenzie: Where exactly is the path being leak?

jsji: eg: we are using nfs path to share the pre built L0 library/headers among different projects. If we don't copy them locally, we would leak the full path of nfs path .

pbalcer: Right, but does the copying have to happen during the CMake configuration step? There must be some script/configuration file that triggers the build, which could look like this:
```
cp $L0_LIBRARY /tmp/level-zero/ # copy from NFS to a local directory
cmake ... -DLEVEL_ZERO_LIBRARY=/tmp/level-zero/ # configure unified runtime
```

Or am I misunderstanding something?

kbenzie: Sorry, I should have been clearer. What's the mechanism by which someone might find this information about internal infrastructure?

Do these paths get embedded into the binary objects output by the build?

Are the paths in CI job logs which become public?

Or something else.

jsji: > Do these paths get embedded into the binary objects output by the build?

Yes, the path might be leak in binaries, especially in PDB.

pbalcer: So this is for compatibility with existing tooling that already uses these specific variable names. Sounds reasonable. I'd prefer to avoid special-casing such things in CMake scripts, but I guess merging this is the least disruptive approach.

jsji: @kbenzie Any further questions? Thanks.

kbenzie: None from me.

kbenzie (approved):

pbalcer (approved):

jsji:
Thank you @kbenzie @pbalcer !

PR #925, opened 2023-10-03
[UR][L0] Enables in-order flags for ImmdCmdList & CmdList

abagusetty:


veselypeta:
Can you also create a draft PR in intel/llvm  and link it in this PR. Check [this](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#adapter-change-process) out for additional info 👍

jandres742 (changes requested):
thanks @abagusetty . I was planning on adding in-order support for in-order lists (sorry, I have forgotten to open an issue in the repo to track that work).

Adding support for in-order lists would require additional refactoring to remove code that is not needed with in-order, as well as performance evaluation to confirm in-order driver implementation is better than what we currently have in the UR L0 adapter.

I will be working on those changes as part of  https://github.com/oneapi-src/unified-runtime/issues/941

PR #929, opened 2023-10-05
Add validation layer support for nullptr events

fabiomestre:
Adds checks in the validation layer for the case where the event wait list parameter in urEnqueue entrypoints contains a nullptr event.

- Also adds more tests because this behaviour was only being checked for urEnqueueDeviceGetGlobalVariableRead and urEnqueueDeviceGetGlobalVariableWrite

### DIFF
--- a/scripts/templates/valddi.cpp.mako
+++ b/scripts/templates/valddi.cpp.mako
@@ -63,7 +63,7 @@ namespace ur_validation_layer
             %if func_name in th.get_event_wait_list_functions(specs, n, tags):
             if (phEventWaitList != NULL && numEventsInWaitList > 0) {
                 for (uint32_t i = 0; i < numEventsInWaitList; ++i) {
-                    if (*(phEventWaitList + i) == NULL) {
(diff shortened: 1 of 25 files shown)


scripts/templates/valddi.cpp.mako
```diff
@@ -60,6 +60,16 @@ namespace ur_validation_layer
 
             %endfor
             %endfor
+            %if func_name in th.get_event_wait_list_functions(specs, n, tags):
+            if (phEventWaitList != NULL && numEventsInWaitList > 0) {
+                for (uint32_t i = 0; i < numEventsInWaitList; ++i) {
+                    if (*(phEventWaitList + i) == NULL) {
```
aarongreig, line 66: would `phEventWaitList[i]` work here instead?

aarongreig (approved):

kbenzie (approved):

scripts/templates/valddi.cpp.mako
fabiomestre: Changed it in the latest commit :+1:

PR #931, opened 2023-10-06
[UR] [CUDA] Changed the output of querying localMemSize

konradkusiak97:
With this patch when the user asks for the amount of `localMemSize`, they receive the maximum amount available on the SM. Setting the maximum limit with the env variable still remains an option.

### DIFF
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -284,12 +284,18 @@ setKernelParams(const ur_context_handle_t Context,
                                    CudaImplicitOffset);
     }
 
-    if (Context->getDevice()->maxLocalMemSizeChosen()) {
+    auto Device = Context->getDevice();
+    if (LocalSize > static_cast<uint32_t>(Device->getMaxCapacityLocalMem())) {
+      setErrorMessage("Too much local memory allocated for device",
+                      UR_RESULT_ERROR_ADAPTER_SPECIFIC);
+      return UR_RESULT_ERROR_ADAPTER_SPECIFIC;
+    }
+
+    if (Device->maxLocalMemSizeChosen()) {
       // Set up local memory requirements for kernel.
-      auto Device = Context->getDevice();
       if (Device->getMaxChosenLocalMem() < 0) {
         bool EnvVarHasURPrefix =
-            std::getenv("UR_CUDA_MAX_LOCAL_MEM_SIZE") != nullptr;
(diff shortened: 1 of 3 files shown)


konradkusiak97:
Corresponding PR in intel/llvm: https://github.com/intel/llvm/pull/11454

hdelan:
LGTM

hdelan (approved):

JackAKirk (approved):
LGTM. I guess this should be documented somewhere though.

konradkusiak97:
Yes, I'll adjust the documentation once this is in

kbenzie (changes requested):
@konradkusiak97 please fix the conflict in `source/adapters/cuda/enqueue.cpp`

source/adapters/cuda/enqueue.cpp
```diff
@@ -284,9 +284,15 @@ setKernelParams(const ur_context_handle_t Context,
                                    CudaImplicitOffset);
     }
 
-    if (Context->getDevice()->maxLocalMemSizeChosen()) {
+    auto Device = Context->getDevice();
+    if (LocalSize > static_cast<uint32_t>(Device->getMaxCapacityLocalMem())) {
+      setErrorMessage("Too much local memory allocated for device",
```
mmoadeli, line 289: I'd change the message with something like, `Excessive allocation of local memory on the device`

konradkusiak97: Ok, that does sound more suitable

source/adapters/cuda/enqueue.cpp
```diff
@@ -284,9 +284,15 @@ setKernelParams(const ur_context_handle_t Context,
                                    CudaImplicitOffset);
     }
 
-    if (Context->getDevice()->maxLocalMemSizeChosen()) {
+    auto Device = Context->getDevice();
+    if (LocalSize > static_cast<uint32_t>(Device->getMaxCapacityLocalMem())) {
```
mmoadeli, line 288: do we need this `static_cast`?

konradkusiak97: Without the `static_cast`, we would have here implicit conversion (`Device->getMaxCapacityLocalMem()` is of type `int`) which is also okay...but I think this makes it clearer and more readable what we're comparing. 

Also, using `static_cast` like this is uniform across several parts of this code so I aimed to match with that.

kbenzie (approved):
LGTM will wait for the checks to pass before merging.

PR #932, opened 2023-10-06
[CUDA][EXP] CUDA adapter support for command buffers

Bensuo:
- Implement initial subset of command buffer functions for CUDA
- Report command buffer extension as supported from device query
- Some additional methods and refactoring to reuse existing adapter code


Bensuo:
Linked DPC++ PR: https://github.com/intel/llvm/pull/11133

source/adapters/cuda/command_buffer.cpp
```diff
@@ -9,48 +9,154 @@
...
+static ur_result_t getNodesFromSyncPoints(
+    const ur_exp_command_buffer_handle_t &CommandBuffer,
+    size_t NumSyncPointsInWaitList,
+    const ur_exp_command_buffer_sync_point_t *SyncPointWaitList,
+    std::vector<CUgraphNode> &CuNodesList) {
+  // Map of ur_exp_command_buffer_sync_point_t to ur_event_handle_t defining
+  // the event associated with each sync-point
+  auto SyncPoints = CommandBuffer->SyncPoints;
+
+  // For each sync-point add associated L0 event to the return list.
```
jchlanda, line 65: s/L0/Cuda ?

source/adapters/cuda/command_buffer.cpp
```diff
@@ -9,48 +9,154 @@
...
+  }
+  return UR_RESULT_SUCCESS;
+}
+
+/// Set parameter for General 1D memory copy.
+/// If the source and/or destination is on the device, SrcPtr and/or DstPtr
+/// must be a pointer to a CUdeviceptr
+static void setCopyParams(const void *SrcPtr, const CUmemorytype_enum SrcType,
+                          void *DstPtr, const CUmemorytype_enum DstType,
+                          size_t Size, CUDA_MEMCPY3D &Params) {
```
jchlanda, line 82: I'd be tempted to 0-initialize `Params` (maybe with a comment for clarity) and then punch in the values that are non-0.

source/adapters/cuda/command_buffer.cpp
```diff
@@ -60,37 +166,108 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferAppendKernelLaunchExp(
...
+
+  // Set the number of threads per block to the number of threads per warp
+  // by default unless user has provided a better number
+  size_t ThreadsPerBlock[3] = {32u, 1u, 1u};
+  size_t BlocksPerGrid[3] = {1u, 1u, 1u};
+
+  uint32_t LocalSize = hKernel->getLocalSize();
+  CUfunction CuFunc = hKernel->get();
+
+  if ((Result = setKernelParams(
```
jchlanda, line 205: One set of parenthesis could be dropped. Not sure if it would read better if the `if` statement was rewritten not to contain the `setKernelParams` call?

source/adapters/cuda/command_buffer.cpp
```diff
@@ -99,19 +276,31 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferAppendMembufferCopyExp(
...
-  (void)pSyncPointWaitList;
-  (void)pSyncPoint;
+  ur_result_t Result = UR_RESULT_SUCCESS;
+  CUgraphNode GraphNode;
+  std::vector<CUgraphNode> DepsList;
+  UR_CALL(getNodesFromSyncPoints(hCommandBuffer, numSyncPointsInWaitList,
+                                 pSyncPointWaitList, DepsList));
+
+  try {
+    auto Src = hSrcMem->Mem.BufferMem.get() + srcOffset;
```
jchlanda, line 286: Are we sure that the offset will not push the pointer out of bounds?

source/adapters/cuda/command_buffer.hpp
```diff
@@ -9,7 +9,233 @@
...
+  case UR_RESULT_ERROR_OBJECT_ALLOCATION_FAILURE:
+    return "UR_RESULT_ERROR_OBJECT_ALLOCATION_FAILURE";
+  case UR_RESULT_ERROR_ADAPTER_SPECIFIC:
+    return "UR_RESULT_ERROR_ADAPTER_SPECIFIC";
+  default:
+    return "UR_RESULT_ERROR_UNKNOWN";
+  }
+};
+
+// Trace an internal PI call; returns in case of an error.
```
jchlanda, line 161:
```suggestion
// Trace an internal UR call; returns in case of an error.
```

jchlanda (commented):
I don't have an in depth view of this functionality, so my review is mostly style/nit, apologies.

source/adapters/cuda/command_buffer.hpp
```diff
@@ -9,7 +9,233 @@
...
+#define UR_CALL(Call)                                                          \
+  {                                                                            \
+    if (PrintTrace)                                                            \
+      fprintf(stderr, "UR ---> %s\n", #Call);                                  \
+    ur_result_t Result = (Call);                                               \
+    if (PrintTrace)                                                            \
+      fprintf(stderr, "UR <--- %s(%s)\n", #Call, getUrResultString(Result));   \
+    if (Result != UR_RESULT_SUCCESS)                                           \
+      return Result;                                                           \
```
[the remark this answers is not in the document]
ldrumm: It also appears to not be used in this patch, so perhaps it's leftover?

source/adapters/cuda/enqueue.cpp
```diff
@@ -187,6 +187,146 @@ bool hasExceededMaxRegistersPerBlock(ur_device_handle_t Device,
...
+setKernelParams(const ur_context_handle_t Context,
+                const ur_device_handle_t Device, const uint32_t WorkDim,
+                const size_t *GlobalWorkOffset, const size_t *GlobalWorkSize,
+                const size_t *LocalWorkSize, ur_kernel_handle_t &Kernel,
+                CUfunction &CuFunc, size_t (&ThreadsPerBlock)[3],
+                size_t (&BlocksPerGrid)[3]) {
+  ur_result_t Result = UR_RESULT_SUCCESS;
+  size_t MaxWorkGroupSize = 0u;
+  size_t MaxThreadsPerBlock[3] = {};
+  bool ProvidedLocalWorkGroupSize = (LocalWorkSize != nullptr);
```
ldrumm, line 211:
```suggestion
  bool ProvidedLocalWorkGroupSize = LocalWorkSize != nullptr;
```

source/adapters/cuda/enqueue.cpp
```diff
@@ -187,6 +187,146 @@ bool hasExceededMaxRegistersPerBlock(ur_device_handle_t Device,
...
+      Kernel->setImplicitOffsetArg(sizeof(CudaImplicitOffset),
+                                   CudaImplicitOffset);
+    }
+
+    if (Context->getDevice()->maxLocalMemSizeChosen()) {
+      // Set up local memory requirements for kernel.
+      auto Device = Context->getDevice();
+      if (Device->getMaxChosenLocalMem() < 0) {
+        bool EnvVarHasURPrefix =
+            (std::getenv("UR_CUDA_MAX_LOCAL_MEM_SIZE") != nullptr);
```
ldrumm, line 292:
```suggestion
            std::getenv("UR_CUDA_MAX_LOCAL_MEM_SIZE") != nullptr;
```

source/adapters/cuda/enqueue.cpp
```diff
@@ -187,6 +187,146 @@ bool hasExceededMaxRegistersPerBlock(ur_device_handle_t Device,
...
+        return UR_RESULT_ERROR_ADAPTER_SPECIFIC;
+      }
+      if (LocalSize > static_cast<uint32_t>(Device->getMaxCapacityLocalMem())) {
+        setErrorMessage("Too much local memory allocated for device",
+                        UR_RESULT_ERROR_ADAPTER_SPECIFIC);
+        return UR_RESULT_ERROR_ADAPTER_SPECIFIC;
+      }
+      if (LocalSize > static_cast<uint32_t>(Device->getMaxChosenLocalMem())) {
+        bool EnvVarHasURPrefix =
+            (std::getenv("UR_CUDA_MAX_LOCAL_MEM_SIZE") != nullptr);
```
ldrumm, line 307:
```suggestion
            std::getenv("UR_CUDA_MAX_LOCAL_MEM_SIZE") != nullptr;
```

source/adapters/cuda/enqueue.cpp
```diff
@@ -187,6 +187,146 @@ bool hasExceededMaxRegistersPerBlock(ur_device_handle_t Device,
   return BlockSize * Kernel->getRegsPerThread() > Device->getMaxRegsPerBlock();
 }
 
+// Helper to compute and put in good shape kernel parameters from workload
```
ldrumm, line 190: What do you mean by "put in good shape"? That give the reader far too much latitude for imagination

source/adapters/cuda/command_buffer.cpp
```diff
@@ -9,48 +9,154 @@
...
-    ur_exp_command_buffer_handle_t *phCommandBuffer) {
-  (void)hContext;
-  (void)hDevice;
-  (void)pCommandBufferDesc;
-  (void)phCommandBuffer;
-  detail::ur::die("Experimental Command-buffer feature is not "
-                  "implemented for CUDA adapter.");
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+    const ur_exp_command_buffer_desc_t *hCommandBufferDesc,
+    ur_exp_command_buffer_handle_t *hCommandBuffer) {
```
ldrumm, line 115: why are we taking hCommandBuffer as an argument only to immediately assign it a new value?

source/adapters/cuda/command_buffer.cpp
ldrumm: Dropping the parens in this case will lead to an assigment in conditional type warning. 

Normally I like assignment in conditionals, but when there are this many arguments it becomes hard to read, so I suggest splitting up the assignment and the check

ldrumm (changes requested):
I don't know anything about command buffers, so my review is in the small. I'll leave someone else to think about the higher level flow

source/adapters/cuda/command_buffer.cpp
```diff
@@ -9,48 +9,154 @@
...
+
+  // Release the memory allocated to the CudaGraph
+  cuGraphDestroy(CudaGraph);
+
+  // Release the memory allocated to the CudaGraphExec
+  cuGraphExecDestroy(CudaGraphExec);
+}
 
-/// Stub implementations of UR experimental feature command-buffers
+/// Helper function for finding the Cuda Nodes associated with the
```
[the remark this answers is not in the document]
Bensuo: I am running `clang-format` on the files but it seems to be enforcing the couple of lines that are slightly over 80 columns, which seems odd.

source/adapters/cuda/command_buffer.hpp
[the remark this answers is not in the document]
Bensuo: This is lifted from the Level 0 adapter. It is used in a couple of places for calls to `getNodesFromSyncPoints()`, but there are other places that it should probably be used (calls to retain/release context etc.).

The CUDA adapter doesn't seem to do any tracing of calls inside `ur*` functions to other `ur*` functions. Maybe it would be more useful and appropriate if this lived in the common header for the CUDA adapter?

source/adapters/cuda/command_buffer.hpp
```diff
@@ -9,7 +9,233 @@
...
+#include "context.hpp"
+#include <cuda.h>
+#include <memory>
 
-struct ur_exp_command_buffer_handle_t_ {};
+static auto getUrResultString = [](ur_result_t Result) {
+  switch (Result) {
+  case UR_RESULT_SUCCESS:
+    return "UR_RESULT_SUCCESS";
+  case UR_RESULT_ERROR_INVALID_OPERATION:
```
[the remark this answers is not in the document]
Bensuo: I'm not too familiar with X macros but from a little research it seems like it wouldn't reduce much verbosity here, and that to do so it would probably need to be defined along with the `ur_result_t` enum itself in common headers, which would reduce the need for this here. That sounds to me like a potentially useful change, though probably out of scope for this PR. 

Please let me know if I'm misunderstanding though.

source/adapters/cuda/command_buffer.hpp
```diff
@@ -9,7 +9,233 @@
...
+  ur_exp_command_buffer_sync_point_t NextSyncPoint;
+
+  // Used when retaining an object.
+  uint32_t incrementReferenceCount() noexcept { return ++RefCount; }
+  // Used when releasing an object.
+  uint32_t decrementReferenceCount() noexcept { return --RefCount; }
+  uint32_t getReferenceCount() const noexcept { return RefCount; }
+  // This method allows to guard a code which needs to be executed when object's
+  // ref count becomes zero after release. It is important to notice that only a
+  // single thread can pass through this check. This is true because of several
```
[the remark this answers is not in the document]
Bensuo: I agree this comment is a little confusing, I think this was also lifted from the Level 0 adapter.

As you say I think this is serving the same purpose as `decrementRefCount` (just without the user having to do the comparison), so I will remove it to be more in line with existing adapter code!

source/adapters/cuda/command_buffer.cpp
Bensuo: This is `urCreateCommandBufferExp` so the command buffer handle here is what will be returned from the function. I guess it is strongly expected that the handle will be `nullptr` in this case, and I don't think other UR functions check that is actually the case either.

kbenzie: So its an out parameter. I think it should be `phCommandBuffer` since its a pointer to the handle.

source/adapters/cuda/command_buffer.cpp
Bensuo: These have been separated now.

source/adapters/cuda/command_buffer.cpp
Bensuo: I've added missing checks that were present in the regular `memcpy` enqueue that check the bounds.

source/adapters/cuda/enqueue.cpp
Bensuo: I've just removed this wording from the comment as it didn't really add anything except potential confusion in this case.

source/adapters/cuda/command_buffer.cpp
Bensuo: > So its an out parameter. I think it should be `phCommandBuffer` since its a pointer to the handle.

Ah yes good catch, I've fixed that and also `hCommandBufferDesc` which should be `pCommandBufferDesc` as it is in the specification.

source/adapters/cuda/command_buffer.hpp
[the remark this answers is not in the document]
Bensuo: This is now removed.

source/adapters/cuda/enqueue.cpp
```diff
@@ -187,6 +187,146 @@ bool hasExceededMaxRegistersPerBlock(ur_device_handle_t Device,
...
+// @param [in]  GlobalWorkOffset pointer workload global offsets
+// @param [in]  LocalWorkOffset pointer workload local offsets
+// @param [inout] Kernel handler to the kernel
+// @param [inout] CuFunc handler to the cuda function attached to the kernel
+// @param [out] ThreadsPerBlock Number of threads per block we should run
+// @param [out] BlocksPerGrid Number of blocks per grid we should run
+ur_result_t
+setKernelParams(const ur_context_handle_t Context,
+                const ur_device_handle_t Device, const uint32_t WorkDim,
+                const size_t *GlobalWorkOffset, const size_t *GlobalWorkSize,
```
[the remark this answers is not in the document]
Bensuo: They are named this way in the API, in `urEnqueueKernelLaunch()` so it seems a bit unnecessary to me in this case to diverge the names when this is refactoring parts of that function.

source/adapters/cuda/command_buffer.hpp
```diff
@@ -9,7 +9,233 @@
...
 #include <ur/ur.hpp>
+#include <ur_api.h>
 
-/// Stub implementation of command-buffers for CUDA
+#include "context.hpp"
+#include <cuda.h>
+#include <memory>
 
-struct ur_exp_command_buffer_handle_t_ {};
+static auto getUrResultString = [](ur_result_t Result) {
```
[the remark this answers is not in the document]
Bensuo: I've changed this to an inline function now.

source/adapters/cuda/command_buffer.hpp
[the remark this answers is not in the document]
ldrumm: I'm suggesting it shouldn't live *at all*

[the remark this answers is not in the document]
Bensuo: That is fair, and I suppose my comment doesn't really address your primary concern here. This internal call tracing could maybe be achieved without needing to return from the macro itself. Do you think there isn't a value in tracing these internal calls at all though?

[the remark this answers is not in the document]
ldrumm: Tracing isn't just tracing if it changes the observable behaviour of the program. I'd separate validation from tracing.
In this case, I think you'll get more useful information if you only *log* return and result strings, and not change the control flow. People will be more able to reason about the control flow too.

With all that said, maybe just remove the the
```
 if (Result != UR_RESULT_SUCCESS)                                           \
      return Result;      
```

source/adapters/cuda/command_buffer.cpp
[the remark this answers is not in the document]
Bensuo: Since the CI will fail if the formatting isn't correct I think I'll need to leave this as is for now, since `clang-format` is considering this to be right.

Bensuo:
@ldrumm @jchlanda If you're happy with the changes could you please approve the PR?

jchlanda (approved):

ldrumm (approved):

mfrancepillois (approved):

EwanC:
Based on the [CI fail](https://github.com/oneapi-src/unified-runtime/actions/runs/6689260833/job/18172519554?pr=932) after merging in tip `adapters` branch, think we need to update our PR with the equivalent changes from https://github.com/oneapi-src/unified-runtime/pull/961/ to `cuda/command_buffer.cpp`

PR #924, opened 2023-10-03
[UR][L0] Add the multi-device-compile experimental feature

kbenzie:
Instead of relying on the list of devices used to create a context, provide interfaces which instead take a list of devices. This more closely aligns with PI and OpenCL. Introduced to workaround a regression. May be superseded in future.

Pull's in the changes in #919 to the main branch.

PR #919, opened 2023-10-02
[UR][L0] Add support for passing device list to urProgramBuild

igchor:
piProgramBuild receives a list of devices, while urProgramBuild does not. This produces a series of issues when a UR program needs to be created for a specific device.

So define a new API, called urProgramBuildExp to pass this list.

Authored-by: jaime.a.arteaga.molina@intel.com

PR #934, opened 2023-10-06
[UR][L0] Add support for passing device list to urProgramBuild/Link/Compile

nrspruit:
piProgramBuild receives a list of devices, while urProgramBuild does not. This produces a series of issues when a UR program needs to be created for a specific device.

So define a new API, called urProgramBuildExp to pass this list.


source/adapters/cuda/program.cpp
```diff
@@ -188,6 +188,26 @@ urProgramCompile(ur_context_handle_t hContext, ur_program_handle_t hProgram,
   return urProgramBuild(hContext, hProgram, pOptions);
 }
 
+UR_APIEXPORT ur_result_t UR_APICALL urProgramCompileExp(
+    ur_context_handle_t hContext, ur_program_handle_t hProgram,
+    uint32_t numDevices, ur_device_handle_t *phDevices, const char *pOptions) {
+  std::ignore = numDevices;
+  std::ignore = phDevices;
+  return urProgramBuild(hContext, hProgram, pOptions);
```
jandres742, line 196: should be `urProgramCompile`, right?

source/adapters/hip/program.cpp
```diff
@@ -92,6 +92,26 @@ urProgramCompile(ur_context_handle_t hContext, ur_program_handle_t hProgram,
   return urProgramBuild(hContext, hProgram, pOptions);
 }
 
+UR_APIEXPORT ur_result_t UR_APICALL urProgramCompileExp(
+    ur_context_handle_t hContext, ur_program_handle_t hProgram,
+    uint32_t numDevices, ur_device_handle_t *phDevices, const char *pOptions) {
+  std::ignore = numDevices;
+  std::ignore = phDevices;
+  return urProgramBuild(hContext, hProgram, pOptions);
```
jandres742, line 100: same

source/adapters/cuda/program.cpp
nrspruit: hip/cuda plugins call urProgramBuild in urProgramCompile, this was calling directly to build since compile already does that.

To be cleaner I have just setup the function definitions for Exp functions in hip/cuda to be unsupported since the pi2ur does a fallback anyway.

source/adapters/hip/program.cpp
nrspruit: same as above, compile calls build for hip/cuda plugins so this was trying to prevent one more function jump.

However, I have gone ahead and made these functions unsupported for hip/cuda so the calls just fallback to the standard calls in pi2ur.

jandres742:
thanks @nrspruit . Changes look good. Please check failures.

jandres742:
we have now 3 similar PRs for this

https://github.com/oneapi-src/unified-runtime/pull/919
https://github.com/oneapi-src/unified-runtime/pull/924
https://github.com/oneapi-src/unified-runtime/pull/934

which one we will be using?

kbenzie:
> which one we will be using?

see https://github.com/oneapi-src/unified-runtime/pull/924#issuecomment-1755187884

jchlanda (approved):
CUDA/HIP 👍

kbenzie (commented):
We'll need to bring in the spec changes for this first, then add these entry points to the loader dispatch tables.

nrspruit:
Hello @kbenzie ,

With https://github.com/oneapi-src/unified-runtime/pull/924#issuecomment-1755187884 924 merged, we need to merge this change, then update the PR https://github.com/intel/llvm/pull/11464 with the updated unified runtime version to merge the fix to the latest version.

source/adapters/level_zero/program.cpp
```diff
@@ -142,8 +153,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramBuild(
   ZeModuleDesc.pBuildFlags = Options;
   ZeModuleDesc.pConstants = Shim.ze();
 
-  ze_device_handle_t ZeDevice = Context->Devices[0]->ZeDevice;
+  ze_device_handle_t ZeDevice = phDevices[0]->ZeDevice;
   ze_context_handle_t ZeContext = Program->Context->ZeContext;
+  std::ignore = Context;
+  std::ignore = numDevices;
```
mmoadeli, line 159: wouldn't it be better to remove `Context` and `numDevices` in function definition? i.e. to have argument type but not name.

mmoadeli: > thanks @mmoadeli . I would prefer to leave them. That way, `urProgramBuild` and `urProgramBuildExp` are as compatible as possible, allowing us to see if in the future we can merge the two into a single API, or find other use cases.
> 
> For instance, if in `urProgramBuildExp` numDevices is 0, then the program could be built for all devices in the Context instead, which is kind of falling back again to urProgramBuild.
> 
> what do you think?
I did not mean to change the API, only changing it to below, to have the type for consistency and not to have the name to avoid using std::ignore.
```
UR_APIEXPORT ur_result_t UR_APICALL urProgramBuildExp(
    ur_context_handle_t, ///< [in] handle of the context instance.
    ur_program_handle_t Program, ///< [in] Handle of the program to build.
    uint32_t, ur_device_handle_t *phDevices,
    const char *Options ///< [in][optional] pointer to build options
                        ///< null-terminated string.
)```

**Comment by kbenzie** [comment]
> Hello @kbenzie ,
> 
> With [#924 (comment)](https://github.com/oneapi-src/unified-runtime/pull/924#issuecomment-1755187884) 924 merged, we need to merge this change, then update the PR [intel/llvm#11464](https://github.com/intel/llvm/pull/11464) with the updated unified runtime version to merge the fix to the latest version.

The reason this can't be merged yet is because the spec changes to add these entry point don't exist in the adapters branch. We'd like to get those changes into the adapters branch with #1072 but it depends on #1014 and #1016. Both of these are pulling in lots of extra stuff which is prooving to be a real pain to integrate, especially the Windows E2E test failures in https://github.com/intel/llvm/pull/11771 (to pull in the changes in #1014). Any help you can provide getting these to merge faster will mean this can be merged sooner.

**Comment by kbenzie** [comment]
> > Hello @kbenzie ,
> > With [#924 (comment)](https://github.com/oneapi-src/unified-runtime/pull/924#issuecomment-1755187884) 924 merged, we need to merge this change, then update the PR [intel/llvm#11464](https://github.com/intel/llvm/pull/11464) with the updated unified runtime version to merge the fix to the latest version.
> 
> The reason this can't be merged yet is because the spec changes to add these entry point don't exist in the adapters branch. We'd like to get those changes into the adapters branch with #1072 but it depends on #1014 and #1016. Both of these are pulling in lots of extra stuff which is prooving to be a real pain to integrate, especially the Windows E2E test failures in [intel/llvm#11771](https://github.com/intel/llvm/pull/11771) (to pull in the changes in #1014). Any help you can provide getting these to merge faster will mean this can be merged sooner.

Since we've had trouble with getting the spec changes from main into the adapters branch we are going with plan B which is to cherry-pick the spec changes needed for this PR to the adapters branch in #1083 instead.

**Comment by kbenzie** [comment]
I'm going to pull #1083 into this branch and also add another commit to include these entry points in the UR loader DDI table. Having these in one place will make it easier to pull into intel/llvm once we get a chance to do so (currently there are issues in CI).

FYI this will involve force pushing to your branch.

**Comment by kbenzie** [comment]
So I'm learning that these implementations have different arguments to the entry-points defined in the spec.

**Comment by jandres742** [comment]
> So I'm learning that these implementations have different arguments to the entry-points defined in the spec.

@nrspruit : could you check?

**Comment by nrspruit** [comment]
> > So I'm learning that these implementations have different arguments to the entry-points defined in the spec.
> 
> @nrspruit : could you check?

Apologies, yes the implementation was for the earlier version of the API, I have fixed it now and the build is passing. Please confirm that I did not miss anything.

File: include/ur.py [hunk 4]
```diff
@@ -196,6 +196,9 @@ class ur_function_v(IntEnum):
     ADAPTER_RETAIN = 179                            ## Enumerator for ::urAdapterRetain
     ADAPTER_GET_LAST_ERROR = 180                    ## Enumerator for ::urAdapterGetLastError
     ADAPTER_GET_INFO = 181                          ## Enumerator for ::urAdapterGetInfo
+    PROGRAM_BUILD_EXP = 197                         ## Enumerator for ::urProgramBuildExp
```
jandres742, line 199: dont we have all the changes in spec already merged in https://github.com/oneapi-src/unified-runtime/pull/924? I would expect this PR to only need to modify code in the L0 adapter, as the spec part has been already taken care of. Or is that this is not rebased?

kbenzie: #924 was merged to main not adapters, we created #1083 to cherry-pick the spec changes into adapters since we were having trouble syncing adapters and main. I then made the decision to take the cherry-pick from #1083 into this PR to reduce the number of intel/llvm PR's required to get this feature merged in time for the next release.

source/adapters/level_zero/program.cpp
```diff
@@ -180,11 +192,25 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramBuild(
...
+        hProgram,        ///< [in][out] handle of the program to compile.
+    uint32_t numDevices, ///< [in] number of devices
+    ur_device_handle_t *phDevices, ///< [in][range(0, numDevices)] pointer to
+                                   ///< array of device handles
+    const char *pOptions           ///< [in][optional] pointer to build options
+                                   ///< null-terminated string.
+) {
+  std::ignore = numDevices;
+  std::ignore = phDevices;
+  return urProgramCompile(nullptr, hProgram, pOptions);
```
kbenzie, line 211: Does `urProgramCompile` actually work with a `nullptr` context? 

We do still have access to the context here:

```suggestion
  return urProgramCompile(hProgram->Context, hProgram, pOptions);
```

I'm a bit dubiour about this implementation though. I thought the issue these new interfaces are supposed to solve is to compile for one device at a time rather than all devices in a context. `urProgramCompile()` is defined as compiling for all devices in the given context, rather than *only*. the ones in the `phDevices` list.

nrspruit: Hello @kbenzie ,

I fixed the urProgramCompileExp to use the context handle from the program handle for level zero. It does not use that handle so it was not necessary, but it does not hurt to pass it thru.

The urProgramCompile() for Level Zero does not do anything device specific for Level Zero so the changes for the urProgramCompileExp function did not directly impact the Level Zero adapter, but other adapters would need that functionality if urProgramCompileExp were implemented for those Adapters.

kbenzie: Okay, that makes sense. Thanks!

PR #935, opened 2023-10-09
[umf] change default MinBucketSize and ensure buckets are generated from the min size 8

DamianDuy:
Currently, for when the MinBucketSize parameter equals 1, sizes were generated with duplicates (such as 1,1,2,2,4,4...). Since such small sized buckets aren't needed anyway this PR changes the minimal value of MinBucketSize to 8.

### DIFF
--- a/source/common/umf_pools/disjoint_pool.cpp
+++ b/source/common/umf_pools/disjoint_pool.cpp
@@ -293,6 +293,8 @@ class DisjointPool::AllocImpl {
         // Generate buckets sized such as: 64, 96, 128, 192, ..., CutOff.
         // Powers of 2 and the value halfway between the powers of 2.
         auto Size1 = params.MinBucketSize;
+        // Buckets sized smaller than the bucket default size- 8 aren't needed.
+        Size1 = std::max(Size1, MIN_BUCKET_DEFAULT_SIZE);
         auto Size2 = Size1 + Size1 / 2;
         for (; Size2 < CutOff; Size1 *= 2, Size2 *= 2) {
             Buckets.push_back(std::make_unique<Bucket>(Size1, *this));
--- a/source/common/umf_pools/disjoint_pool.hpp
+++ b/source/common/umf_pools/disjoint_pool.hpp
@@ -17,6 +17,8 @@
 
 namespace usm {
 
+inline constexpr size_t MIN_BUCKET_DEFAULT_SIZE = 8;
+
 // Configuration for specific USM allocator instance
 class DisjointPoolConfig {
   public:
@@ -46,7 +48,7 @@ class DisjointPoolConfig {
 
     // Holds the minimum bucket size valid for allocation of a memory type.
     // This value must be a power of 2.
-    size_t MinBucketSize = 1;
+    size_t MinBucketSize = MIN_BUCKET_DEFAULT_SIZE;
 
     // Holds size of the pool managed by the allocator.
     size_t CurPoolSize = 0;


source/common/umf_pools/disjoint_pool.cpp
```diff
@@ -293,6 +293,15 @@ class DisjointPool::AllocImpl {
         // Generate buckets sized such as: 64, 96, 128, 192, ..., CutOff.
         // Powers of 2 and the value halfway between the powers of 2.
         auto Size1 = params.MinBucketSize;
+        // Special case for when MinBucketSize equals 1. In this case buckets
```
lukaszstolarczuk, line 296: but why is this special case here...? (pls describe at least in the commit message)

DamianDuy: Done.

source/common/umf_pools/disjoint_pool.cpp
```diff
@@ -293,6 +293,16 @@ class DisjointPool::AllocImpl {
         // Generate buckets sized such as: 64, 96, 128, 192, ..., CutOff.
         // Powers of 2 and the value halfway between the powers of 2.
         auto Size1 = params.MinBucketSize;
+        // Special case for when MinBucketSize equals 1 because otherwise bucket sizes starting from 1 value
+        // would generate with duplicates (like 1,1,2,2,4,4...) which is undesirable.
+        // In this case buckets are generated sized such as: 1, 2, 4, 8, 16, 32, 64, 96, 128 ..., Cutoff.
+        // Powers of 2 to the size of 64 and then powers of 2 and the values halfway
+        // between the powers of 2.
+        if (Size1 == 1) {
```
vinser52, line 301: What if `Size1` is greater than `1`? I think the `if` statement is excessive because we do not need half the power of `2` until `64`.

source/common/umf_pools/disjoint_pool.cpp
```diff
@@ -293,6 +293,16 @@ class DisjointPool::AllocImpl {
         // Generate buckets sized such as: 64, 96, 128, 192, ..., CutOff.
         // Powers of 2 and the value halfway between the powers of 2.
         auto Size1 = params.MinBucketSize;
+        // Special case for when MinBucketSize equals 1 because otherwise bucket sizes starting from 1 value
+        // would generate with duplicates (like 1,1,2,2,4,4...) which is undesirable.
```
vinser52, line 297: I think this statement is misleading: `(like 1,1,2,2,4,4...)`. Only `1` will be duplicated. If the `Size1` equals `2` the sequence without this patch is `2, 3, 4, 6, 8, 12, 16, 24, etc`

source/common/umf_pools/disjoint_pool.cpp
vinser52: Another question: does it make sense to have such small bucket sizes? Maybe we can just round up the `Size1` to `64` or `32`?

vinser52 (changes requested):

source/common/umf_pools/disjoint_pool.cpp
igchor: Yeah, I think you can just write that we only generate powers of 2 up to a certain value (64) and then powers of 2 + values in between.

source/common/umf_pools/disjoint_pool.cpp
igchor: I don't think we have a use case for such small buckets but since this option is exposed why should we artificially limit it to 64 or 32?

pbalcer: I would agree that having buckets for anything below 8 bytes is excessive, but it shouldn't be much of an issue (unless it causes some alignment issues or false sharing).

However, 1 appears to be the default value, so we probably shouldn't change it, at least not without investigating the impact.

source/common/umf_pools/disjoint_pool.cpp
```diff
@@ -293,6 +293,16 @@ class DisjointPool::AllocImpl {
...
+        // Powers of 2 to the size of 64 and then powers of 2 and the values halfway
+        // between the powers of 2.
+        if (Size1 == 1) {
+            for (; Size1 < 64; Size1 *= 2) {
+                Buckets.push_back(std::make_unique<Bucket>(Size1, *this));
+            }
+        }
         auto Size2 = Size1 + Size1 / 2;
         for (; Size2 < CutOff; Size1 *= 2, Size2 *= 2) {
             Buckets.push_back(std::make_unique<Bucket>(Size1, *this));
```
pbalcer, line 308: I don't like the way this is done. We are special-casing the default path (1) to calculate the bucket sizes differently than for other values. Wouldn't it be simpler to either do `auto Size1 = max(Size1, 2);` or, if we really want to have that bucket with 1 byte objects, something like `if Size1 == 1 { buckets.push_back(...); Size1 = 2; }`?

source/common/umf_pools/disjoint_pool.cpp
DamianDuy: As for the MinBucketSize default value, It was set to 0 before but I've changed it to 1 in the previous PR (since 0 caused an infinite loop in the sizes generation and also the MinBucketSize was ensured to be a power of two in that PR). So I think it should be safe to change to another value if needed.

source/common/umf_pools/disjoint_pool.cpp
DamianDuy: Yeah, I think I like this better as well.

source/common/umf_pools/disjoint_pool.cpp
DamianDuy: I changed the MinBucketSize default value to 8 and also set 8 to minimal value from which sizes for buckets are generated.

source/common/umf_pools/disjoint_pool.cpp
DamianDuy: Done.

source/common/umf_pools/disjoint_pool.hpp
```diff
@@ -46,7 +46,7 @@ class DisjointPoolConfig {
 
     // Holds the minimum bucket size valid for allocation of a memory type.
     // This value must be a power of 2.
-    size_t MinBucketSize = 1;
+    size_t MinBucketSize = 8;
```
pbalcer, line 49: make an a const global variable for this value (and use it here and in the implementation).

DamianDuy: Done.

pbalcer (approved):

igchor (approved):

PR #936, opened 2023-10-09
[HIP] Revert add prefetch for USM hip allocations a6b8fa66b537753415d24076f…

hdelan:
https://github.com/intel/llvm/pull/10430 caused a performance regression. The new way to ensure safety when using xor atomics is by changing the xor impl to use a CAS loop. See here https://github.com/intel/llvm/pull/11876.

Tested in https://github.com/intel/llvm/pull/11893.

### DIFF
--- a/source/adapters/hip/context.hpp
+++ b/source/adapters/hip/context.hpp
@@ -10,7 +10,6 @@
 #pragma once
 
 #include <set>
-#include <unordered_map>
 
 #include "common.hpp"
 #include "device.hpp"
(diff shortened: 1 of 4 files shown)


jchlanda:
Has this been done through `git revert`? If so why do we need 2 commits?

hdelan:
> Has this been done through `git revert`? If so why do we need 2 commits?

Yes it was but I think maybe there was some conflict so added the second commit. Have squashed the second commit so just one commit now

jchlanda (approved):
LGTM

kbenzie:
This needs an intel/llvm testing PR with passing checks before we can merge here.

hdelan:
https://github.com/intel/llvm/pull/11893

PR #940, opened 2023-10-09
[UR][HIP] Enable kernel finalization using comgr

Naghasan:
For kernel fusion support for hip, we need to finalize the kernels using comgr. The patch finalizes tagged binaries during buildProgram before handing it over to the hip runtime.

Initial / related PR https://github.com/intel/llvm/pull/11003

### DIFF
--- a/source/adapters/hip/CMakeLists.txt
+++ b/source/adapters/hip/CMakeLists.txt
@@ -18,7 +18,7 @@ set(UR_HIP_INCLUDE_DIR "${UR_HIP_ROCM_DIR}/include")
 set(UR_HIP_HSA_INCLUDE_DIR "${UR_HIP_ROCM_DIR}/hsa/include")
 
 # Set HIP lib dir
-set(UR_HIP_LIB_DIR "${UR_HIP_ROCM_DIR}/hip/lib")
+set(UR_HIP_LIB_DIR "${UR_HIP_ROCM_DIR}/lib")
 
 # Check if HIP library path exists (AMD platform only)
 if("${UR_HIP_PLATFORM}" STREQUAL "AMD")
@@ -98,6 +98,18 @@ if("${UR_HIP_PLATFORM}" STREQUAL "AMD")
         INTERFACE_INCLUDE_DIRECTORIES        "${HIP_HEADERS}"
         INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${HIP_HEADERS}"
     )
+
+    if(UR_ENABLE_COMGR)
+        add_library(amd_comgr SHARED IMPORTED GLOBAL)
+        set_target_properties(
+        amd_comgr PROPERTIES
+            IMPORTED_LOCATION                    "${UR_HIP_LIB_DIR}/libamd_comgr.so"
+            INTERFACE_INCLUDE_DIRECTORIES        "${HIP_HEADERS}"
+            INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${HIP_HEADERS}"
+        )
+        target_link_libraries(pi_hip PUBLIC amd_comgr)
+        target_compile_definitions(pi_hip PRIVATE SYCL_ENABLE_KERNEL_FUSION)
+    endif(UR_ENABLE_COMGR)
 
     target_link_libraries(${TARGET_NAME} PRIVATE
     ${PROJECT_NAME}::headers
--- a/source/adapters/hip/program.cpp
+++ b/source/adapters/hip/program.cpp
@@ -10,6 +10,70 @@
 
 #include "program.hpp"
 
+#ifdef SYCL_ENABLE_KERNEL_FUSION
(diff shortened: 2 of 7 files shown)


Naghasan:
@JackAKirk could have a wee look please ? (I think you are one of the code owner) This is the UR part of the patch you approved in the main intel/llvm side

JackAKirk:
> @JackAKirk could have a wee look please ? (I think you are one of the code owner) This is the UR part of the patch you approved in the main intel/llvm side

sure

Naghasan:
Thanks

source/adapters/hip/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ set(UR_HIP_INCLUDE_DIR "${UR_HIP_ROCM_DIR}/include")
 set(UR_HIP_HSA_INCLUDE_DIR "${UR_HIP_ROCM_DIR}/hsa/include")
 
 # Set HIP lib dir
-set(UR_HIP_LIB_DIR "${UR_HIP_ROCM_DIR}/hip/lib")
+set(UR_HIP_LIB_DIR "${UR_HIP_ROCM_DIR}/lib")
```
JackAKirk, line 21: CC @kbenzie for awareness. This change implies UR_HIP_ROCM_DIR semantic has changed. I guess this might affect the CMAKE gymnastics with regard to pointing unified-runtime/intel-llvm to the correct rocm.

source/adapters/hip/program.cpp
```diff
@@ -259,6 +412,16 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramCreateWithBinary(
 
   // TODO: Set metadata here and use reqd_work_group_size information.
   // See urProgramCreateWithBinary in CUDA adapter.
+  if (pProperties) {
```
JackAKirk, line 415: @jchlanda I was just wondering whether this affects your work on `reqd_work_group_size` metadata?

jchlanda: Yeap, I've added essentially the same snippet to set the metadata.
I guess we should be OK though, whichever patch gets in first will cause a merge conflict for the other, but that should be easy to resolve.

source/adapters/hip/program.cpp
```diff
@@ -28,7 +108,80 @@ ur_result_t ur_program_handle_t_::setBinary(const char *Source, size_t Length) {
   return UR_RESULT_SUCCESS;
 }
 
+ur_result_t ur_program_handle_t_::finalizeRelocatable() {
+#ifndef SYCL_ENABLE_KERNEL_FUSION
+  assert(false && "Relocation only available with fusion");
+  return UR_RESULT_ERROR_UNKNOWN;
```
JackAKirk, line 114: This line will never be reached right? because of the above assert being set to false.

Also it would be better to use a backend specific error that then gets automatically passed to a sycl extension, using `setErrorMessage`, like here: https://github.com/intel/llvm/pull/10953/files

source/adapters/hip/program.cpp
jchlanda: Agreed, git will likely resolve it for us. Not to worry.

source/adapters/hip/CMakeLists.txt
kbenzie: Thanks, this looks more logical to me but I'm not well versed in _finding rocm_ CMake game.

source/adapters/hip/program.cpp
Naghasan: > This line will never be reached right? 

In a normal setup, correct. Hence the `false` here.

In order to be reached, you would need to either enable fusion in intel/llvm and actively disable it in the adapter or use the finalize property in some other patch. Either way, you need to actively change something in the build (cmake changes) or add a patch (which if tested should fail CI) to trigger this.

I can add a `setErrorMessage` if you care about a non sycl focus use of the adapter, otherwise I don't think it is that useful.

source/adapters/hip/CMakeLists.txt
Naghasan: libs found in `hip/lib` are also found in `lib` via a symlink. But logically, the path is no longer HIP but ROCM (which include ROCM). Hence the simple change here.

source/adapters/hip/program.cpp
JackAKirk: Yeah it's fine to leave it as it is.

JackAKirk (approved):

Naghasan:
Unless there is more comments, can we have this merged please ?

thanks

kbenzie:
> Unless there is more comments, can we have this merged please ?

Updating UR in intel/llvm is currently blocked by https://github.com/intel/llvm/pull/11312 - I'm hoping to get that resolved today but want to do that before merging any more adapter changed to UR, things are already kinda complicated. I'll add it to the list to merge once things are unblocked.

PR #954, opened 2023-10-12
[HIP] Handle required wg size attribute in HIP

jchlanda:


kbenzie:
`[SYCL]` prefix doesn't make much sense in this repo, should be `[HIP]` instead.

jchlanda:
Seems like build error is a blip in CI:
```
error: RPC failed; curl 56 GnuTLS recv error (-9): Error decoding the received TLS packet.
error: 15490 bytes of body are still expected
```
https://github.com/oneapi-src/unified-runtime/actions/runs/6494327570/job/17637060114?pr=954

jchlanda:
> `[SYCL]` prefix doesn't make much sense in this repo, should be `[HIP]` instead.

Ha, ha, that's a leftover from intel/llvm. Will change.

jchlanda:
This PR is blocked on the outcome of the discussion here: https://github.com/llvm/llvm-project/pull/68872

fabiomestre:

jchlanda:
Corresponding intel/llvm PR: https://github.com/intel/llvm/pull/13600

source/adapters/hip/kernel.hpp
```diff
@@ -42,6 +42,10 @@ struct ur_kernel_handle_t_ {
   ur_program_handle_t Program;
   std::atomic_uint32_t RefCount;
 
+  static constexpr uint32_t ReqdThreadsPerBlockDimensions = 3u;
+  std::unordered_map<ur_device_handle_t, std::array<size_t, 3>>
```
hdelan, line 46: Why do we need this? A `ur_kernel_handle_t_` is unique to a program which is unique to a device.

jchlanda: The way it's set up on [Cuda](https://github.com/oneapi-src/unified-runtime/blob/b98584f742226278a79548a0583e715bf890ef1e/source/adapters/cuda/kernel.hpp#L144C28-L144C48) is through a call to `urKernelGetGroupInfo` which takes a `device`. Now that we have multi-device context I though we would need to keep a per-device mapping. If it's safe to always use the first device, we could mimic what Cuda does and have it as a single array of 3 values.

hdelan: It does indeed take a device, but that doesn't mean we need to use it. Every kernel is associated with a single device so we don't need this map. The map adds unnecessary complexity as well as overhead.

jchlanda: So should it always use the first available device in the context?

jchlanda: We've discussed this offline, done in: 3b8d6c5328cf982f4adf17e3f4140e6d1960b9b9

hdelan (approved):
Thanks for making changes. LGTM

PietroGhg (approved):
Native CPU LGTM, thank you (the corresponding llvm PR may have to update the UR tag for native CPU)

PR #961, opened 2023-10-13
[HIP][CUDA] Change unions in ur_mem_handle_t_ to stdvariant

hdelan:
Std::variant is safer than unions in the `ur_mem_handle_t_` class and it will make it easier to incrementally move to multi device ctx, which requires adding `std::vector`s to the MemImpl, which is not possible in a union.

### DIFF
--- a/source/adapters/cuda/memory.cpp
+++ b/source/adapters/cuda/memory.cpp
@@ -44,22 +44,21 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemBufferCreate(
     CUdeviceptr Ptr = 0;
     auto HostPtr = pProperties ? pProperties->pHost : nullptr;
 
-    ur_mem_handle_t_::MemImpl::BufferMem::AllocMode AllocMode =
-        ur_mem_handle_t_::MemImpl::BufferMem::AllocMode::Classic;
+    BufferMem::AllocMode AllocMode = BufferMem::AllocMode::Classic;
 
     if ((flags & UR_MEM_FLAG_USE_HOST_POINTER) && EnableUseHostPtr) {
       UR_CHECK_ERROR(
           cuMemHostRegister(HostPtr, size, CU_MEMHOSTREGISTER_DEVICEMAP));
       UR_CHECK_ERROR(cuMemHostGetDevicePointer(&Ptr, HostPtr, 0));
-      AllocMode = ur_mem_handle_t_::MemImpl::BufferMem::AllocMode::UseHostPtr;
+      AllocMode = BufferMem::AllocMode::UseHostPtr;
     } else if (flags & UR_MEM_FLAG_ALLOC_HOST_POINTER) {
       UR_CHECK_ERROR(cuMemAllocHost(&HostPtr, size));
       UR_CHECK_ERROR(cuMemHostGetDevicePointer(&Ptr, HostPtr, 0));
-      AllocMode = ur_mem_handle_t_::MemImpl::BufferMem::AllocMode::AllocHostPtr;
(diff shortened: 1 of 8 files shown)


source/adapters/cuda/memory.cpp
```diff
@@ -44,22 +44,21 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemBufferCreate(
     CUdeviceptr Ptr = 0;
     auto HostPtr = pProperties ? pProperties->pHost : nullptr;
 
-    ur_mem_handle_t_::MemImpl::BufferMem::AllocMode AllocMode =
-        ur_mem_handle_t_::MemImpl::BufferMem::AllocMode::Classic;
+    BufferMem::AllocMode AllocMode = BufferMem::AllocMode::Classic;
```
jchlanda, line 47: Lack of `ur_mem_handle_t_::MemImpl`, is because of the now gone union, right? Not some namespace mis-match?

source/adapters/cuda/memory.hpp
```diff
@@ -35,136 +163,16 @@ struct ur_mem_handle_t_ {
...
-    Mem.BufferMem.Parent = Parent;
-    Mem.BufferMem.HostPtr = HostPtr;
-    Mem.BufferMem.Size = Size;
-    Mem.BufferMem.MapSize = 0;
-    Mem.BufferMem.MapOffset = 0;
-    Mem.BufferMem.MapPtr = nullptr;
-    Mem.BufferMem.MapFlags = UR_MAP_FLAG_WRITE;
-    Mem.BufferMem.MemAllocMode = Mode;
+      : Context{Context}, RefCount{1}, MemType{Type::Buffer},
+        MemFlags{MemFlags}, Mem{BufferMem{Parent, Mode, Ptr, HostPtr, Size}} {
```
jchlanda, line 173: I think `Mem` was cleaner outside of the initializer list.

source/adapters/cuda/memory.hpp
```diff
@@ -174,43 +182,30 @@ struct ur_mem_handle_t_ {
...
                    ur_mem_type_t ImageType, void *HostPtr)
-      : Context{Context}, RefCount{1}, MemType{Type::Surface}, MemFlags{
-                                                                   MemFlags} {
-    (void)HostPtr;
-
-    Mem.SurfaceMem.Array = Array;
-    Mem.SurfaceMem.SurfObj = Surf;
-    Mem.SurfaceMem.ImageType = ImageType;
+      : Context{Context}, RefCount{1}, MemType{Type::Surface},
+        MemFlags{MemFlags}, Mem{SurfaceMem{Array, Surf, ImageType, HostPtr}} {
```
jchlanda, line 186: Same here, and below.

jchlanda (approved):
LGTM, sorry about the wait.

source/adapters/cuda/memory.cpp
hdelan: Correct

source/adapters/cuda/memory.hpp
hdelan: Unfortunately we need to do this in member initializer list because the default constructor of std::variant is deleted

source/adapters/cuda/memory.hpp
hdelan: Same as above

hdelan:
@kbenzie can we merge this please?

kbenzie:
> @kbenzie can we merge this please?

We already have 2 UR changes that can't merge to intel/llvm because of test failures, I feel like merging this will only add to that problem.

hdelan:
@kbenzie OK sure yes no rush

PR #962, opened 2023-10-13
[UR][L0] Correctly wait on barrier on urEnqueueEventsWaitWithBarrier

jandres742:
When event list is null, a barrier is still needed for all previous commands, so fix it.


jandres742:
@smaslov-intel : please take a look.

source/adapters/level_zero/event.cpp
```diff
@@ -964,8 +966,7 @@ ur_result_t CleanupCompletedEvent(ur_event_handle_t Event, bool QueueLocked,
 ur_result_t EventCreate(ur_context_handle_t Context, ur_queue_handle_t Queue,
                         bool HostVisible, ur_event_handle_t *RetEvent) {
 
-  bool ProfilingEnabled =
-      !Queue || (Queue->Properties & UR_QUEUE_FLAG_PROFILING_ENABLE) != 0;
+  bool ProfilingEnabled = !Queue || Queue->isProfilingEnabled();
```
smaslov-intel, line 969: I know you did not change this, but it looks weird that when Queue == null then we create an event from a profiled pool

source/adapters/level_zero/event.cpp
```diff
@@ -168,7 +168,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueEventsWaitWithBarrier(
         // TODO: this and other special handling of in-order queues to be
         // updated when/if Level Zero adds native support for in-order queues.
         //
-        if (Queue->isInOrderQueue() && InOrderBarrierBySignal) {
+        if (Queue->isInOrderQueue() && InOrderBarrierBySignal &&
+            !Queue->isProfilingEnabled()) {
```
smaslov-intel, line 172: It's not clear to me how profiling affects synchronization.

jandres742: @smaslov-intel : please see here 

https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=appendsignalevent#_CPPv430zeCommandListAppendSignalEvent24ze_command_list_handle_t17ze_event_handle_t

> zeCommandListAppendSignalEvent
>
> The duration of an event created from an event pool that was created using [ZE_EVENT_POOL_FLAG_KERNEL_TIMESTAMP](https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=appendsignalevent#ze__api_8h_1a148be5a8b03e9d401a415d2227cf9f11af1bab0559057f7ec51e2c77a1746894f) or [ZE_EVENT_POOL_FLAG_KERNEL_MAPPED_TIMESTAMP](https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=appendsignalevent#ze__api_8h_1a148be5a8b03e9d401a415d2227cf9f11a853fca9e461e0b8ae1ad7947476c9795) flags is undefined

Therefore,, when profiling is used, we cannot just used signalEvent if EventWaitList.Lenght == 0. We need to fallback directly to barrier to have correct timestamps.

source/adapters/level_zero/event.cpp
jandres742: @smaslov-intel : I know, I saw that and it looks weird. Maybe that was added as a WA, or a better logic could be used. i'm planning on addressing that in another patch.

source/adapters/level_zero/event.cpp
smaslov-intel: OK, but please add source code comment explaining this.

smaslov-intel (approved):

source/adapters/level_zero/event.cpp
jandres742: thanks @smaslov-intel. I have added comment.

jandres742:
@kbenzie : please merge when possible

kbenzie (approved):

fabiomestre:

nrspruit (approved):

PR #965, opened 2023-10-16
[Command-buffer][L0] Reset sync-point events

EwanC:
The L0 events used to implement [UR sync-points](https://oneapi-src.github.io/unified-runtime/core/EXP-COMMAND-BUFFER.html#sync-points) in a command-buffer aren't reset after the first execution. Leaving them in a signaled state before subsequent submissions of the UR command-buffer.

This patch resets the L0 events used as UR command-buffer sync-points to the non-signaled state, by appending event reset commands to the end of the user defined command-list. There is also a reset added for the UR command-buffer wait event that is signaled when the wait list dependencies of `urCommandBufferEnqueueExp` are met.

Additionally this patch removes the host scope flag from being set on L0 events created by the UR command-buffer, as we never signal/reset them from host, only device.

Companion DPC++ PR - https://github.com/intel/llvm/pull/11553

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -38,6 +38,8 @@
 
   There is also a WaitEvent used by the `ur_exp_command_buffer_handle_t` class
   in the prefix to wait on any dependencies passed in the enqueue wait-list.
+  This WaitEvent is reset at the end of the suffix, along with reset commands
+  to reset the L0 events used to implement the UR sync-points.
 
   ┌──────────┬────────────────────────────────────────────────┬─────────┐
   │  Prefix  │ Commands added to UR command-buffer by UR user │ Suffix  │
@@ -47,10 +49,10 @@
   Prefix    │Reset signal event │ Barrier waiting on wait event│
             └───────────────────┴──────────────────────────────┘
 
-            ┌─────────────────────────────────────────┐
-  Suffix    │Signal the UR command-buffer signal event│
-            └─────────────────────────────────────────┘
-
+            ┌─────────────────────────────────────────────┐──────────────┐
+  Suffix    │Barrier waiting on sync-point event,         │ Reset events │
+            │signalling the UR command-buffer signal event│              │
+            └─────────────────────────────────────────────┘──────────────┘
 
   For a call to `urCommandBufferEnqueueExp` with an event_list `EL`,
   command-buffer `CB`, and return event `RE` our implementation has to create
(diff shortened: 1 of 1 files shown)


source/adapters/level_zero/command_buffer.cpp
EwanC, line 1: The comment on line 51 needs updated before this is taken out of draft

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -437,6 +437,21 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t CommandBuffer) {
 
 UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t CommandBuffer) {
+  // Reset the L0 events we use for command-buffer internal sync-points to the
+  // non-signalled state
+  ZE2UR_CALL(zeCommandListAppendBarrier,
+             (CommandBuffer->ZeCommandList, nullptr, 0, nullptr));
```
smaslov-intel, line 443: This will sync the entire queue, right? Is this the intention? Why not just wait for the specific events you are trying to reset here?

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -437,6 +437,21 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t CommandBuffer) {
...
+    ur_event_handle_t Event = SyncPoint.second;
+    ZE2UR_CALL(zeCommandListAppendEventReset,
+               (CommandBuffer->ZeCommandList, Event->ZeEvent));
+  }
+
+  // Reset the wait-event for the UR command-buffer that is signalled when its
+  // submission dependencies have been satisfied.
+  ZE2UR_CALL(zeCommandListAppendEventReset,
+             (CommandBuffer->ZeCommandList, CommandBuffer->WaitEvent->ZeEvent));
+
```
reble, line 454:
```suggestion
  size_t NumEvents = CommandBuffer->SyncPoints.size() + 1;
  std::vector<ze_event_handle_t> TmpEventList{NumEvents};

  // Reset the wait-event for the UR command-buffer that is signalled when its
  // submission dependencies have been satisfied.
  TmpEventList[0] = CommandBuffer->WaitEvent->ZeEvent;

  // Reset the L0 events we use for command-buffer internal sync-points
  for (size_t i = 1; i < NumEvents; i++) {
    TmpEventList[i] = CommandBuffer->SyncPoints[i - 1]->ZeEvent;
  }

  ZE2UR_CALL(zeCommandListAppendBarrier, (CommandBuffer->ZeCommandList, nullptr,
                                          NumEvents, TmpEventList.data()));

  for (size_t i = 0; i < NumEvents; i++) {
    ZE2UR_CALL(zeCommandListAppendEventReset,
               (CommandBuffer->ZeCommandList, TmpEventList[i]));
  }
```

source/adapters/level_zero/command_buffer.cpp
EwanC: Good point, yes this would wait on the whole queue, which is unnecessary. Updated to wait on the specific events. I've also set the signal event parameter of the barrier to signal an event that we we're previously using a standalone `zeCommandListAppendSignalEvent` command for.

source/adapters/level_zero/command_buffer.cpp
EwanC: Thanks for the patch, I used it for inspiration but didn't apply it directly, as I don't think it handles ` CommandBuffer->WaitEvent` correctly. We already have a barrier command to wait on this event at the start of the command-list (added in `urCommandBufferCreateExp`), so don't need to wait on it again. We also need to reset this, which is code that's removed by this patch.

EwanC:
> thanks @EwanC . Please check the failures in the intel/llvm patch

I've had a look through these, and I think they are unrelated. See https://github.com/intel/llvm/pull/11553#issuecomment-1766717083 for rationale

jandres742:
> > thanks @EwanC . Please check the failures in the intel/llvm patch
> 
> I've had a look through these, and I think they are unrelated. See [intel/llvm#11553 (comment)](https://github.com/intel/llvm/pull/11553#issuecomment-1766717083) for rationale. So I'll take this PR out of draft.

thanks @EwanC . Please check the comment in https://github.com/intel/llvm/pull/11553#issuecomment-1772118519 to get the results clean. Once that's done, we can merge, as changes look good.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -407,7 +409,7 @@ urCommandBufferCreateExp(ur_context_handle_t Context, ur_device_handle_t Device,
   // Create signal & wait events to be used in the command-list for sync
   // on command-buffer enqueue.
   auto RetCommandBuffer = *CommandBuffer;
-  UR_CALL(EventCreate(Context, nullptr, true, &RetCommandBuffer->SignalEvent));
+  UR_CALL(EventCreate(Context, nullptr, false, &RetCommandBuffer->SignalEvent));
```
mfrancepillois, line 412: Are you sure there is not use cases where the host needs to access the SignalEvent? E.g. post-process the data on the host?

mfrancepillois (approved):

source/adapters/level_zero/command_buffer.cpp
EwanC: This event is never returned by to the user, so it would only be our runtime code that would want to access `SignalEvent`. We currently don't have any calls to `zeEventHostSignal` or `zeEventHostSynchronize` in our code to need host visibility, but if we wanted to use these for post-processing, then we can set this back to `true` as part of that future PR.

EwanC:
> > > thanks @EwanC . Please check the failures in the intel/llvm patch
> > 
> > 
> > I've had a look through these, and I think they are unrelated. See [intel/llvm#11553 (comment)](https://github.com/intel/llvm/pull/11553#issuecomment-1766717083) for rationale. So I'll take this PR out of draft.
> 
> thanks @EwanC . Please check the comment in [intel/llvm#11553 (comment)](https://github.com/intel/llvm/pull/11553#issuecomment-1772118519) to get the results clean. Once that's done, we can merge, as changes look good.

Testing all green on matching DPC++ PR https://github.com/intel/llvm/pull/11553 now

kbenzie (approved):

PR #966, opened 2023-10-17
[OpenCL] Add Command Buffer extension to OpenCL adapter.

martygrant:
Adds implementations for the [experimental Command Buffer](https://oneapi-src.github.io/unified-runtime/core/api.html#command-buffer-experimental) entry points. Also fixes minor build error with missing namespace in enqueue.cpp.

Added companion intel-llvm PR which also updates many of the E2E tests to run against OpenCL - https://github.com/intel/llvm/pull/11718


source/adapters/opencl/command_buffer.cpp
```diff
@@ -1,71 +1,127 @@
...
+
+  CL_RETURN_ON_FAILURE(F(hCommandBuffer->clCommandBuffer));
+  return UR_RESULT_SUCCESS;
 }
 
-UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferReleaseExp(
-    [[maybe_unused]] ur_exp_command_buffer_handle_t hCommandBuffer) {
+UR_APIEXPORT ur_result_t UR_APICALL
+urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
+  ur_result_t result = urQueueRelease(hCommandBuffer->hInternalQueue);
```
EwanC, line 64: `Result` for consistent variable naming

source/adapters/opencl/command_buffer.cpp
```diff
@@ -1,71 +1,127 @@
...
-    [[maybe_unused]] ur_device_handle_t hDevice,
+    ur_context_handle_t hContext, ur_device_handle_t hDevice,
     [[maybe_unused]] const ur_exp_command_buffer_desc_t *pCommandBufferDesc,
-    [[maybe_unused]] ur_exp_command_buffer_handle_t *phCommandBuffer) {
+    ur_exp_command_buffer_handle_t *phCommandBuffer) {
 
-  cl_adapter::die("Experimental Command-buffer feature is not "
-                  "implemented for OpenCL adapter.");
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+  ur_queue_handle_t queue = nullptr;
```
EwanC, line 17: `Queue` to be PascalCase

source/adapters/opencl/command_buffer.cpp
```diff
@@ -1,71 +1,127 @@
...
+  ur_result_t Result = urQueueCreate(hContext, hDevice, nullptr, &queue);
+  if (Result != UR_RESULT_SUCCESS) {
+    return Result;
+  }
+
+  cl_context clContext = cl_adapter::cast<cl_context>(hContext);
```
EwanC, line 23: `clContext` should start with a capital for PascalCase variable naming, see this called `CLContext ` in other files. Same comment for `clCommandBuffer` and `urCommandBuffer` later on

source/adapters/opencl/command_buffer.cpp
```diff
@@ -1,71 +1,127 @@
...
+  cl_ext::clCreateCommandBufferKHR_fn F = nullptr;
+  cl_int Res = cl_ext::getExtFuncFromContext<decltype(F)>(
+      clContext, cl_ext::ExtFuncPtrCache->clCreateCommandBufferKHRCache,
+      cl_ext::CreateCommandBufferName, &F);
+
+  if (!F || Res != CL_SUCCESS)
+    return UR_RESULT_ERROR_INVALID_OPERATION;
+
+  auto clCommandBuffer =
+      F(1, cl_adapter::cast<cl_command_queue *>(&queue), nullptr, &Res);
```
EwanC, line 33: Should we be checking the return value from  `Res` here?

source/adapters/opencl/command_buffer.hpp
```diff
@@ -8,8 +8,19 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include <CL/cl_ext.h>
 #include <ur/ur.hpp>
 
 /// Stub implementation of command-buffers for OpenCL
```
EwanC, line 14: Can remove this comment

source/adapters/opencl/common.hpp
```diff
@@ -226,6 +236,58 @@ cl_int(CL_API_CALL *)(cl_command_queue queue, cl_program program,
                       cl_uint num_events_in_waitlist,
                       const cl_event *events_waitlist, cl_event *event);
 
+using clCreateCommandBufferKHR_fn = CL_API_ENTRY cl_command_buffer_khr(
```
EwanC, line 239: Rather than defining these ourselves, can we use the symbols from the OpenCL-Headers instead? https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl_ext.h#L102-L110

Not sure what version of the headers you're testing, with but these should have been in for a while

source/adapters/opencl/command_buffer.cpp
```diff
@@ -1,71 +1,127 @@
...
-// Copyright (C) 2023 Intel Corporation
-//
-// Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM
-// Exceptions. See LICENSE.TXT
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-//===----------------------------------------------------------------------===//
+//===-----------------------------------------------------------------===//
```
EwanC, line 7: Why are we touching these license lines? If this is relevant think it should be a standalone PR that updates all the affected files

kbenzie: Yeah, this license change should be reverted. The removed license comments are the correct ones for UR.

source/adapters/opencl/command_buffer.cpp
martygrant: Fixed, thanks.

source/adapters/opencl/command_buffer.cpp
martygrant: Fixed, thanks.

source/adapters/opencl/command_buffer.cpp
martygrant: Fixed, thanks. Added a CL_RETURN_ON_FAILURE_AND_SET_NULL for `Res`.

source/adapters/opencl/command_buffer.hpp
martygrant: Fixed, thanks.

source/adapters/opencl/common.hpp
martygrant: Discussed this offline. Leaving this as is to be safe for the Github CI.

source/adapters/opencl/command_buffer.cpp
martygrant: Fixed thanks think this was an accidental copy paste error.

veselypeta (approved):

source/adapters/opencl/command_buffer.cpp
```diff
@@ -180,19 +277,57 @@ ur_result_t UR_APICALL urCommandBufferAppendMembufferReadRectExp(
...
+  cl_context CLContext = cl_adapter::cast<cl_context>(hCommandBuffer->hContext);
+  cl_ext::clEnqueueCommandBufferKHR_fn F = nullptr;
+  cl_int Res = cl_ext::getExtFuncFromContext<decltype(F)>(
+      CLContext, cl_ext::ExtFuncPtrCache->clEnqueueCommandBufferKHRCache,
+      cl_ext::EnqueueCommandBufferName, &F);
+
+  if (!F || Res != CL_SUCCESS)
+    return UR_RESULT_ERROR_INVALID_OPERATION;
+
+  uint32_t numQueues = (hQueue) ? 1 : 0;
```
EwanC, line 324: Why do we need this check? [urCommandBufferEnqueueExp](https://oneapi-src.github.io/unified-runtime/core/api.html#urcommandbufferenqueueexp) specifies that if `hQueue` is NULL then `UR_RESULT_ERROR_INVALID_NULL_HANDLE` should be returned

martygrant: I added this from reading the below from the [OpenCL spec](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_add_new_section_section_5_x_2_enqueuing_a_command_buffer).
```
queues A pointer to an ordered list of command-queues [compatible](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#compatible) with the command-queues used on recording. queues can be NULL in which case the default command-queues used on command-buffer creation are used and num_queues must be 0.
```

source/adapters/opencl/command_buffer.cpp
```diff
@@ -11,61 +11,120 @@
...
+
+  CL_RETURN_ON_FAILURE(F(hCommandBuffer->CLCommandBuffer));
+  return UR_RESULT_SUCCESS;
 }
 
-UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferReleaseExp(
-    [[maybe_unused]] ur_exp_command_buffer_handle_t hCommandBuffer) {
+UR_APIEXPORT ur_result_t UR_APICALL
+urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
+  ur_result_t Result = urQueueRelease(hCommandBuffer->hInternalQueue);
```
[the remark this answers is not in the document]
martygrant: Yeah that should be called too I'll make sure to add that in, thanks.

source/adapters/opencl/command_buffer.cpp
EwanC: Sure, but that OpenCL behaviour isn't a valid implementation of this UR entry-point. For example, if we had a CTS test that tested calling `urCommandBufferEnqueueExp` with a null `hQueue` parameter, then the expected behaviour is to return an error `UR_RESULT_ERROR_INVALID_NULL_HANDLE `, not enqueue the command-buffer.

source/adapters/opencl/command_buffer.cpp
```diff
@@ -11,61 +11,120 @@
...
     [[maybe_unused]] const ur_exp_command_buffer_desc_t *pCommandBufferDesc,
-    [[maybe_unused]] ur_exp_command_buffer_handle_t *phCommandBuffer) {
+    ur_exp_command_buffer_handle_t *phCommandBuffer) {
 
-  cl_adapter::die("Experimental Command-buffer feature is not "
-                  "implemented for OpenCL adapter.");
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+  ur_queue_handle_t Queue = nullptr;
+  ur_result_t Result = urQueueCreate(hContext, hDevice, nullptr, &Queue);
+  if (Result != UR_RESULT_SUCCESS) {
```
fabiomestre, line 21: Minor: There is a macro in the adapter called UR_RETURN_ON_FAILURE that could be used to simplify this error checks

source/adapters/opencl/command_buffer.cpp
martygrant: Apologies that's my misunderstanding. I wasn't aware of how the validation layer works before but have now found out those checks specified in the spec are done [in that layer](https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/layers/validation/ur_valddi.cpp#L7434). I've removed the `uint32_t numQueues = (hQueue) ? 1 : 0;`. Cheers.

EwanC (changes requested):
This code LGTM, but I think this PR should have a companion change in DPC++. Once this UR commit get's pulled into DPC++ the SYCL-Graph extension will start reporting support for OpenCL with undocumented behaviour. Also without DPC++ testing, i'm not sure how to use this code path, given that I don't think there's UR CTS testing for experimental features.

What i'd imagine the companion DPC++ PR to cover:
* Add a OpenCL section to the design doc for SYCL-Graphs https://github.com/intel/llvm/blob/sycl/sycl/doc/design/CommandGraph.md#backend-implementation documenting the implementation on cl_khr_command_buffer, and explicitly documenting the types of nodes that aren't currently supported.
* Add opencl to the REQUIRES line of `test-e2e/Graph` tests which now pass with this change.
* Add a `device.get_info<exp_ext::info::device::graph_support>();` check to all tests, skipping returning early for those which don't report support. As even for an OpenCL device, support is dependent on the cl_khr_command_buffer extension being available. This will probably be the case for all DPC++ CI testing, which I don't think uses a device with cl_khr_command_buffer support.

martygrant:
Hey @EwanC, thanks for the really thorough review. That all sounds fine for a new PR into DPC++, I'll open up an issue with the details you've provided. Should one PR be merged before the other?

We're temporarily holding off merging any more stuff into the adapters branch here until a L0 issue is reverted but I'll have both of these changes ready to go once that's sorted.

EwanC:
> Hey @EwanC, thanks for the really thorough review. That all sounds fine for a new PR into DPC++, I'll open up an issue with the details you've provided. Should one PR be merged before the other?
> 
> We're temporarily holding off merging any more stuff into the adapters branch here until a L0 issue is reverted but I'll have both of these changes ready to go once that's sorted.

I think following the existing adapter change process is fine https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#adapter-change-process This will need to be merged first, but only after the DPC++ PR has been approved and CI passing.

martygrant:
Have opened up an issue for the work required in DPC++ - https://github.com/intel/llvm/issues/11599

martygrant:
Corrected an error in `urCommandBufferAppendKernelLaunchExp` where the `cl_command_queue command_queue` parameter for `clCommandNDRangeKernelKHR` must be NULL. Given the extension function pointers more meaningful names instead of just `F`.

source/adapters/opencl/command_buffer.cpp
```diff
@@ -74,43 +134,82 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferAppendMemcpyUSMExp(
...
+  cl_int Res = cl_ext::getExtFuncFromContext<decltype(clCommandCopyBufferKHR)>(
+      CLContext, cl_ext::ExtFuncPtrCache->clCommandCopyBufferKHRCache,
+      cl_ext::CommandCopyBufferName, &clCommandCopyBufferKHR);
+
+  if (!clCommandCopyBufferKHR || Res != CL_SUCCESS)
+    return UR_RESULT_ERROR_INVALID_OPERATION;
+
+  CL_RETURN_ON_FAILURE(clCommandCopyBufferKHR(
+      hCommandBuffer->CLCommandBuffer,
+      cl_adapter::cast<cl_command_queue>(hCommandBuffer->hInternalQueue),
```
EwanC, line 173: From the spec wording from `clCommandCopyBufferKHR` for where we are passing the internal queue:

> command_queue Specifies the command-queue the command will be recorded to. Parameter is unused by this extension as only a single command-queue is supported and must be NULL.

Similar wording for other entry-points. We should replace this internal queue parameter passed on command-recording entry-points with NULL

source/adapters/opencl/command_buffer.cpp
```diff
@@ -74,43 +134,82 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferAppendMemcpyUSMExp(
         *pSyncPointWaitList,
     [[maybe_unused]] ur_exp_command_buffer_sync_point_t *pSyncPoint) {
 
-  cl_adapter::die("Experimental Command-buffer feature is not "
+  cl_adapter::die("Experimental Command-buffer entry point is not "
```
EwanC, line 137: Can we remove these die calls? they result in an abort which isn't ideal now that it's highly likely a user could use these entry-points by accident. Returning the error code will be enough for the SYCL runtime to throw an sensible exception that the end user can deal with how they please.

source/adapters/opencl/command_buffer.cpp
martygrant: Fixed: `clCommandCopyBufferKHR`, `clCommandCopyBufferRectKHR` and `clCommandFillBufferKHR`

EwanC (approved):
LGTM 😃

PR #968, opened 2023-10-17
[OpenCL] Add version check for urProgramSetSpecializationConstants 

martygrant:
Adds a version check to only look up the extension function pointer for urProgramSetSpecializationConstants if the platform or any attached device OpenCL version is < 2.2. Also fixes minor build error with missing namespace in enqueue.cpp.

intel-llvm PR - https://github.com/intel/llvm/pull/11720

### DIFF
--- a/source/adapters/opencl/program.cpp
+++ b/source/adapters/opencl/program.cpp
@@ -316,20 +316,58 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramSetSpecializationConstants(
   CL_RETURN_ON_FAILURE(clGetProgramInfo(CLProg, CL_PROGRAM_CONTEXT, sizeof(Ctx),
                                         &Ctx, &RetSize));
 
-  cl_ext::clSetProgramSpecializationConstant_fn F = nullptr;
-  const ur_result_t URResult = cl_ext::getExtFuncFromContext<decltype(F)>(
-      Ctx, cl_ext::ExtFuncPtrCache->clSetProgramSpecializationConstantCache,
-      cl_ext::SetProgramSpecializationConstantName, &F);
+  std::unique_ptr<std::vector<cl_device_id>> DevicesInCtx;
+  cl_adapter::getDevicesFromContext(cl_adapter::cast<ur_context_handle_t>(Ctx),
+                                    DevicesInCtx);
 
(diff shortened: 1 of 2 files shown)


source/adapters/opencl/program.cpp
```diff
@@ -316,20 +316,54 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramSetSpecializationConstants(
...
-    return URResult;
-  }
+  cl_platform_id CurPlatform;
+  clGetDeviceInfo((*DevicesInCtx)[0], CL_DEVICE_PLATFORM,
+                  sizeof(cl_platform_id), &CurPlatform, nullptr);
+
+  oclv::OpenCLVersion PlatVer;
+  cl_adapter::getPlatformVersion(CurPlatform, PlatVer);
+
+  bool useExtensionLookup = false;
```
aarongreig, line 330: `UseExtensionLookup` if we're sticking to the LLVM convention

aarongreig (approved):

source/adapters/opencl/program.cpp
martygrant: Fixed now, cheers.

source/adapters/opencl/program.cpp
```diff
@@ -316,20 +316,54 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramSetSpecializationConstants(
...
   }
 
+  if (UseExtensionLookup == false) {
+    for (uint32_t i = 0; i < count; ++i) {
+      CL_RETURN_ON_FAILURE(clSetProgramSpecializationConstant(
+          CLProg, pSpecConstants[i].id, pSpecConstants[i].size,
+          pSpecConstants[i].pValue));
+    }
+  } else {
+    cl_ext::clSetProgramSpecializationConstant_fn F = nullptr;
```
kbenzie, line 353: Could we can this `SetProgramSpecializationConstant` instead of `F`? Would be much clearer when its called lower down.

kbenzie (approved):

source/adapters/opencl/program.cpp
martygrant: Good idea, I've updated this now. I've done the same in #966 so probably a good idea to make the same change there too.

PR #969, opened 2023-10-17
[OpenCL] Use OpenCL::OpenCL cmake target instead

fabiomestre:
The OpenCL::OpenCL target is defined in cmake since version 3.7 and provides a simpler alternative to import the opencl library without having to use add_library(). This also prevents conflicts caused by adding the OpenCL-ICD library multiple times.


kbenzie (approved):
Much cleaner 🙂

martygrant (approved):

fabiomestre:
> Much cleaner 🙂

Unfortunately, I had to change it a bit because upstream CI is unable to use find_package. You might want to review this again.

source/adapters/opencl/CMakeLists.txt
```diff
@@ -43,41 +43,24 @@ set_target_properties(${TARGET_NAME} PROPERTIES
 
 find_package(Threads REQUIRED)
 
-if (NOT DEFINED OpenCL_LIBRARY OR NOT DEFINED OpenCL_INCLUDE_DIR)
-    message(WARNING "OpenCL_LIBRARY and OpenCL_INCLUDE_DIR are not set. Using find_package() to find an OpenCL installation in the system.")
+# The OpenCL target can be set manually on upstream cmake to avoid using find_package().
+if (NOT OPENCL_TARGET)
```
kbenzie, line 47: I think we should use the `UR_` prefix to avoid name collisions and be more explicit about the variable containing the path to the OpenCL-ICD-Loader library, something like: `UR_OPENCL_ICD_LOADER_LIBRARY`

kbenzie (changes requested):

source/adapters/opencl/CMakeLists.txt
fabiomestre: That makes sense. Updated the name in the latest commit

Issue #842, opened 2023-09-01
Compiling device binaries for HIP in CTS


pbalcer:
The target for building device binaries  currently does not support `amdgcn` triplets:
```
error: missing AMDGPU architecture for SYCL offloading; specify it with '-Xsycl-target-backend --offload-arch'
```

Even if that were made to work, the tests would likely not work anyway because of missing IL postfix:
https://github.com/oneapi-src/unified-runtime/blob/a8e4955a516fdc53a90e7f1da9eb89a1a8133755/test/conformance/source/environment.cpp#L275

This method currently fails anyway because the HIP adapter does not implement `UR_DEVICE_INFO_IL_VERSION`.

veselypeta:
I'm using a amd card with gfx1032 architecture. The command that works for me is 👍 
```
clang++ -fsycl -fsycl-targets=amdgcn-amd-amdhsa -Xsycl-target-backend=amdgcn-amd-amdhsa --offload-ard=gfx1032
```

PR #970, opened 2023-10-18
[HIP] Correctly set HIP Kernel Buffer Arguments

veselypeta:
Device binary code exported from SYCL produces different interfaces depending on the backend selected. On CUDA and L0 buffer arguments are encoded with two parameters 1) A pointer to the buffer 2) a sycl::id struct (the accessor). However, on HIP the accessor argument is encoded as three individual i64 arguments.
This PR correctly sets the kernel arguments for kernels compiled and run with the HIP backend.


ldrumm:
s/Compliation/Compilation/g

test/conformance/kernel/urKernelCreate.cpp
```diff
@@ -28,12 +28,12 @@ struct urKernelCreateTest : uur::urProgramTest {
 UUR_INSTANTIATE_KERNEL_TEST_SUITE_P(urKernelCreateTest);
 
 TEST_P(urKernelCreateTest, Success) {
-    ASSERT_SUCCESS(urKernelCreate(program, kernel_name.data(), &kernel));
+    ASSERT_SUCCESS(urKernelCreate(program, kernel_name.c_str(), &kernel));
```
ldrumm, line 31: These changes seem unrelated

.github/workflows/cmake.yml
```diff
@@ -180,7 +180,7 @@ jobs:
 
     - name: Download DPC++
       run: |
-        wget -O ${{github.workspace}}/dpcpp_compiler.tar.gz https://github.com/intel/llvm/releases/download/nightly-2023-09-21/sycl_linux.tar.gz
+        wget -O ${{github.workspace}}/dpcpp_compiler.tar.gz https://github.com/intel/llvm/releases/download/nightly-2023-10-18/sycl_linux.tar.gz
```
ldrumm, line 183: This change seems unrelated

test/conformance/device_code/CMakeLists.txt
```diff
@@ -9,9 +9,13 @@ macro(add_device_binary SOURCE_FILE)
...
     foreach(TRIPLE ${TARGET_TRIPLES})
         set(EXE_PATH "${DEVICE_BINARY_DIR}/${KERNEL_NAME}_${TRIPLE}")
+        if(${TRIPLE} MATCHES "amd")
+            set(AMD_TARGET_BACKEND -Xsycl-target-backend=${TRIPLE})
+            set(AMD_OFFLOAD_ARCH  --offload-arch=${AMD_ARCH})
+        endif()
         add_custom_command(OUTPUT ${EXE_PATH}
-            COMMAND ${UR_DPCXX} -fsycl -fsycl-targets=${TRIPLE} -fsycl-device-code-split=off
-            ${SOURCE_FILE} -o ${EXE_PATH}
+            COMMAND ${UR_DPCXX} -fsycl -fsycl-targets=${TRIPLE} -fsycl-device-code-split=off ${AMD_TARGET_BACKEND} ${AMD_OFFLOAD_ARCH}
```
ldrumm, line 17: this line's getting a bit long

source/adapters/hip/device.cpp
```diff
@@ -30,6 +30,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   static constexpr uint32_t MaxWorkItemDimensions = 3u;
 
   switch ((uint32_t)propName) {
+  case UR_DEVICE_INFO_VIRTUAL_MEMORY_SUPPORT: {
```
ldrumm, line 33: Is this related?

ldrumm (changes requested):
Some changes don't look like they belong here. Please clarify their purpose, or remove unrelated cleanups

test/conformance/kernel/urKernelCreate.cpp
veselypeta: 👍

test/conformance/device_code/CMakeLists.txt
veselypeta: 👍

source/adapters/hip/device.cpp
veselypeta: Not directly 👍 - will include in a separate patch

test/conformance/source/environment.cpp
```diff
@@ -266,6 +266,17 @@ std::string KernelsEnvironment::getSupportedILPostfix(uint32_t device_index) {
         return {};
     }
 
+    // special case for AMD as it doesn't support IL.
```
pbalcer, line 269: Is this a TODO? should be there an issue to fix IL then?

test/conformance/device_code/CMakeLists.txt
```diff
@@ -9,9 +9,15 @@ macro(add_device_binary SOURCE_FILE)
...
+        if(${TRIPLE} MATCHES "amd")
+            set(AMD_TARGET_BACKEND -Xsycl-target-backend=${TRIPLE})
+            set(AMD_OFFLOAD_ARCH  --offload-arch=${AMD_ARCH})
+        endif()
         add_custom_command(OUTPUT ${EXE_PATH}
-            COMMAND ${UR_DPCXX} -fsycl -fsycl-targets=${TRIPLE} -fsycl-device-code-split=off
-            ${SOURCE_FILE} -o ${EXE_PATH}
+            COMMAND ${UR_DPCXX} -fsycl -fsycl-targets=${TRIPLE} -fsycl-device-code-split=off 
+            ${AMD_TARGET_BACKEND} ${AMD_OFFLOAD_ARCH} ${SOURCE_FILE} 
+            -o ${EXE_PATH} || (exit 0)
```
pbalcer, line 19: Ignoring the error on dumping the device binaries already makes it easy to accidentally misconfigure these tests. Ignoring compiler errors will only make the problem worse. Can you please add a comment why the errors are ignored?

I'm wondering whether it would be possible to check if running `${EXE_PATH}` created a file in cwd. Maybe with inotify?

pbalcer:
Closes #842

test/conformance/source/environment.cpp
veselypeta: Possibly, although from speaking to the team implementing the plugin they are likely to not support this query as the HIP does not accept 'IL' as such.

test/conformance/device_code/CMakeLists.txt
veselypeta: This is a slightly strange issue I've found. AMD does not support images (at least through SYCL) and so the compilation of `image_copy.cpp` fails. I was using Ninja as a generator locally and it's perfectly happy to ignore this error and let the build continue, but using "Unix Makefiles" stops the build and exits. Perhaps we should just skip compiling this code on AMD and avoid having to silence the error. I've already added checks to the image/sampler CTS to ensure that this program is not loaded on AMD. 👍

ldrumm (approved):

pbalcer (approved):

test/conformance/enqueue/enqueue_adapter_hip.match
```diff
@@ -1 +1,87 @@
-Segmentation fault
+{{OPT}}Segmentation Fault
```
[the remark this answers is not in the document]
veselypeta: It seems to be related to what hardware the conformance tests are running on. I've created #985 to track this. I still think this is an improvement as none of these tests were being run on AMD before.

test/conformance/device_code/CMakeLists.txt
```diff
@@ -9,9 +9,18 @@ macro(add_device_binary SOURCE_FILE)
     file(MAKE_DIRECTORY ${DEVICE_BINARY_DIR})
     foreach(TRIPLE ${TARGET_TRIPLES})
         set(EXE_PATH "${DEVICE_BINARY_DIR}/${KERNEL_NAME}_${TRIPLE}")
+        if(${TRIPLE} MATCHES "amd")
+            set(AMD_TARGET_BACKEND -Xsycl-target-backend=${TRIPLE})
+            set(AMD_OFFLOAD_ARCH  --offload-arch=${AMD_ARCH})
+        endif()
+        if(${TRIPLE} MATCHES "amd" AND ${KERNEL_NAME} MATCHES "image_copy")
+            continue()
```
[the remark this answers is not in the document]
veselypeta: I agree. Not sure what the best solution is overall. In intel/llvm the e2e tests that require sampler and image support specifically exclude amd with `UNSUPPORTED:` https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/Sampler/unnormalized-none-nearest.cpp#L2

pbalcer (approved):

jinz2014:
@veselypeta  Do you know the cause of the differences in accessor arguments between HIP and CUDA ?  Thanks.

veselypeta:
Covered in #1030

PR #975, opened 2023-10-20
[OpenCL] Add more mappings from CL error codes to UR error codes.

aarongreig:
Also merge urQueueCreate InvalidValueProperties test into InvalidQueueProperties test.


kbenzie:
Looks like this is now passing in the CUDA adapter. Need to update the match file.

```
urQueueCreateTest.InvalidValueProperties/NVIDIA_CUDA_BACKEND___{{.*}}_
```

kbenzie (approved):

fabiomestre (approved):
LGTM

kbenzie:
@aarongreig there does not appear to be an https://github.com/intel/llvm PR for this as per the [Adapter Change Process](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#adapter-change-process)

aarongreig:
forgot to do that before I went on holiday, is it worth doing now? this is technically a cts fix so it could wait to go in with the rest of them

kbenzie:
> forgot to do that before I went on holiday, is it worth doing now? this is technically a cts fix so it could wait to go in with the rest of them

Okay, we can test it in the combined branch.

PR #976, opened 2023-10-20
[OpenCL] Implement urEnqueueUSMMemcpy2D and allow large fill patterns.

aarongreig:
Normally OpenCL limits fill type operations to a max pattern size of 128, this patch includes a workaround to extend that.

### DIFF
--- a/source/adapters/opencl/usm.cpp
+++ b/source/adapters/opencl/usm.cpp
@@ -257,24 +257,39 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMFill(
       &CopyEvent));
 
   struct DeleteCallbackInfo {
+    DeleteCallbackInfo(clMemBlockingFreeINTEL_fn USMFree, cl_context CLContext,
+                       void *HostBuffer)
+        : USMFree(USMFree), CLContext(CLContext), HostBuffer(HostBuffer) {
+      clRetainContext(CLContext);
+    }
+    ~DeleteCallbackInfo() {
+      USMFree(CLContext, HostBuffer);
+      clReleaseContext(CLContext);
+    }
+    DeleteCallbackInfo(const DeleteCallbackInfo &) = delete;
+    DeleteCallbackInfo &operator=(const DeleteCallbackInfo &) = delete;
+
     clMemBlockingFreeINTEL_fn USMFree;
     cl_context CLContext;
     void *HostBuffer;
-    void execute() {
-      USMFree(CLContext, HostBuffer);
-      delete this;
-    }
   };
 
-  auto Info = new DeleteCallbackInfo{USMFree, CLContext, HostBuffer};
+  auto Info = new DeleteCallbackInfo(USMFree, CLContext, HostBuffer);
 
   auto DeleteCallback = [](cl_event, cl_int, void *pUserData) {
-    static_cast<DeleteCallbackInfo *>(pUserData)->execute();
+    auto Info = static_cast<DeleteCallbackInfo *>(pUserData);
+    delete Info;
   };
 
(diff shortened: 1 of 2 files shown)


source/adapters/opencl/usm.cpp
```diff
@@ -343,18 +404,59 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMFill2D(
...
+  if (blocking) {
+    ClResult = clWaitForEvents(Events.size(), Events.data());
+  }
+  if (phEvent && ClResult == CL_SUCCESS) {
+    ClResult = clEnqueueBarrierWithWaitList(
+        cl_adapter::cast<cl_command_queue>(hQueue), Events.size(),
+        Events.data(), cl_adapter::cast<cl_event *>(phEvent));
+  }
+  for (const auto &E : Events) {
+    clReleaseEvent(E);
```
fabiomestre, line 456: Probably should add CL_RETURN_ON_ERROR to this call as well. Same for line 441

source/adapters/opencl/usm.cpp
```diff
@@ -343,18 +404,59 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMFill2D(
...
+  clEnqueueMemcpyINTEL_fn FuncPtr = nullptr;
+  ur_result_t RetVal = cl_ext::getExtFuncFromContext<clEnqueueMemcpyINTEL_fn>(
+      CLContext, cl_ext::ExtFuncPtrCache->clEnqueueMemcpyINTELCache,
+      cl_ext::EnqueueMemcpyName, &FuncPtr);
+
+  if (!FuncPtr) {
+    return RetVal;
+  }
+
+  std::vector<cl_event> Events;
```
fabiomestre, line 429: Small performance potential improvement could be to reserve the vector memory in advance instead of relying to pull_back to resize it

source/adapters/opencl/usm.cpp
```diff
@@ -207,20 +206,82 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMFill(
...
+      cl_adapter::cast<cl_command_queue>(hQueue), false, ptr, HostBuffer, size,
+      numEventsInWaitList, cl_adapter::cast<const cl_event *>(phEventWaitList),
+      &CopyEvent));
+
+  struct DeleteCallbackInfo {
+    clMemBlockingFreeINTEL_fn USMFree;
+    cl_context CLContext;
+    void *HostBuffer;
+    void execute() {
+      USMFree(CLContext, HostBuffer);
```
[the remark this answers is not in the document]
aarongreig: Since I was adding a destructor anyway I gave the struct a proper constructor with does a retain on the context (with matching Release in the destructor), should prevent the released context scenario. Using a `new` pointer would probably work but it seems sketchier.

source/adapters/opencl/usm.cpp
```diff
@@ -207,20 +206,82 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMFill(
...
+      delete this;
+    }
+  };
+
+  auto Info = new DeleteCallbackInfo{USMFree, CLContext, HostBuffer};
+
+  auto DeleteCallback = [](cl_event, cl_int, void *pUserData) {
+    static_cast<DeleteCallbackInfo *>(pUserData)->execute();
```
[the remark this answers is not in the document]
aarongreig: agreed, this looks a lot nicer

source/adapters/opencl/usm.cpp
aarongreig: Added here, I'm going to leave 441 as it is to prioritise returning the error from the memcpy

PR #979, opened 2023-10-20
[CI] Clean up after conda init in cmake.yml

lukaszstolarczuk:
When initializing conda few extra commands are added to user's profile. These commands make it impossible to log in as the test user to the machine, since conda may be missing.

We can safely remove the whole '.profile' after the job.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -265,6 +265,12 @@ jobs:
       working-directory: ${{github.workspace}}/build
       run: bin/codegen
 
+    # conda init adds content to user's profile making it failing (if conda is gone)
+    - name: Cleanup after conda init
+      run: |
+        cat ${HOME}/.profile || true
+        rm ${HOME}/.profile || true
+
   windows-build:
     name: Build - Windows
     strategy:


kswiecicki (approved):
LGTM, but you have a failing CI job https://github.com/oneapi-src/unified-runtime/actions/runs/6587485286/job/18077584172?pr=979. Probably needs just a re-run.

lukaszstolarczuk:
> LGTM, but you have a failing CI job https://github.com/oneapi-src/unified-runtime/actions/runs/6587485286/job/18077584172?pr=979. Probably needs just a re-run.

It happened quite a few times on this PR, I suspect it may be caused by my changes 😉 I still need to investigate this.

lukaszstolarczuk:
Apparently `remove-profiles` parameter in conda-incubator/setup-miniconda action was doing more than its name suggested.
Also, if these profile files were not removed before `conda init`, then the initialization behavior was different.

Anyway, this fix should work just fine.

kswiecicki (approved):

bratpiorka (approved):

PR #986, opened 2023-10-23
[UR] Discourage using SYCL tc to generate kernels

hdelan:
It was observed that a SYCL kernel lambda can generate significantly different signatures when compiling the same kernel lambda for different arches. This can make generic multi platform UR code complex and difficult to debug since kernel params are different and have different orders. 

This discourages using a SYCL TC to generate kernels to be run with UR for this reason.


hdelan:
ping @kbenzie just a doc change so this should be uncontroversial

scripts/core/CUDA.rst
```diff
@@ -21,6 +21,14 @@ CUDA device images can be generated by a CUDA-capable compiler toolchain. Most
 CUDA compiler toolchains are capable of generating PTX, SASS and/or bundles of
 PTX and SASS.
 
+When generating device code to be launched using Unified Runtime, it is
+recommended to use a programming model with explicit kernel parameters, such as
+openCL or CUDA. This is because kernels generated by a programming model with
```
kbenzie, line 26:
```suggestion
OpenCL or CUDA. This is because kernels generated by a programming model with
```

scripts/core/HIP.rst
```diff
@@ -27,6 +27,14 @@ HIPCC can generate device code for a particular arch using the ``--genco`` flag
 
     $ hipcc --genco hello.cu --amdgpu-target=gfx906 -o hello.hsaco
 
+When generating device code to be launched using Unified Runtime, it is
+recommended to use a programming model with explicit kernel parameters, such as
+openCL or HIP. This is because kernels generated by a programming model with
```
kbenzie, line 32:
```suggestion
OpenCL or HIP. This is because kernels generated by a programming model with
```

scripts/core/CUDA.rst
hdelan: Thanks change made

scripts/core/HIP.rst
hdelan: Thanks change made

PR #988, opened 2023-10-23
Enable fp16 runtime support for hip

MartinWehking:
Needed for https://github.com/intel/llvm/pull/11626


ldrumm:
Does this feature need to be enabled conditionally depending on the arch properties?

MartinWehking:
@ldrumm The only problematic one is Tahiti, which doesn't support fp16, but we decided to ignore Tahiti in this case.
There's unfortunately no possibility to query fp16 support dynamically (for fp64 there's a member inside 'hipDeviceArch_t', rocm/include/hip/hip_runtime_api.h, but not for fp16)

JackAKirk:
> Tahiti

Tahiti is not officially supported by ROCm anyway. This should be fine.

JackAKirk (approved):

fabiomestre:

PR #990, opened 2023-10-23
[CUDA] Add support for binary type query

fabiomestre:
CUDA does not make a distinction between binaryTypes (it treats PTX and binaries using the same entrypoints). 

However, for UR, by definition:
-  `urProgramCompile` should set the binary type to `UR_PROGRAM_BINARY_TYPE_COMPILED_OBJECT` 
-  `urProgramBuild` / `urProgramLink` should set it to `UR_PROGRAM_BINARY_TYPE_EXECUTABLE`. 
- `urProgramCreateWithBinary` should set the binary type to `UR_PROGRAM_BINARY_TYPE_COMPILED_OBJECT`

### DIFF
--- a/source/adapters/cuda/program.cpp
+++ b/source/adapters/cuda/program.cpp
@@ -180,8 +180,6 @@ ur_result_t createProgram(ur_context_handle_t hContext,
             UR_RESULT_ERROR_INVALID_CONTEXT);
   UR_ASSERT(size, UR_RESULT_ERROR_INVALID_SIZE);
 
-  ur_result_t Result = UR_RESULT_SUCCESS;
-
   std::unique_ptr<ur_program_handle_t_> RetProgram{
       new ur_program_handle_t_{hContext}};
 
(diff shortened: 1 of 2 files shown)


fabiomestre:
@oneapi-src/unified-runtime-cuda-write Would be great if someone could have a look a this PR

source/adapters/cuda/program.cpp
```diff
@@ -165,6 +165,47 @@ ur_result_t getKernelNames(ur_program_handle_t) {
   return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
 }
 
+/// Loads images from a list of PTX or CUBIN binaries.
+/// Note: No calls to CUDA driver API in this function, only store binaries
+/// for later.
+///
+/// Note: Only supports one device
```
jchlanda, line 172: @hdelan is that something we need to think about with multi device context (also the comment at the bottom of the diff)?

source/adapters/cuda/program.cpp
```diff
@@ -165,6 +165,47 @@ ur_result_t getKernelNames(ur_program_handle_t) {
...
+ur_result_t createProgram(ur_context_handle_t hContext,
+                          ur_device_handle_t hDevice, size_t size,
+                          const uint8_t *pBinary,
+                          const ur_program_properties_t *pProperties,
+                          ur_program_handle_t *phProgram) {
+  UR_ASSERT(hContext->getDevice()->get() == hDevice->get(),
+            UR_RESULT_ERROR_INVALID_CONTEXT);
+  UR_ASSERT(size, UR_RESULT_ERROR_INVALID_SIZE);
+
+  ur_result_t Result = UR_RESULT_SUCCESS;
```
jchlanda, line 183: `Result` doesn't seem to be needed, maybe wrap the calls directly in `UR_ASSERT` and return `UR_SUCCESS` at the end to simplify?

jchlanda (approved):
LGTM

Does it need any new test, or is it already covered?

source/adapters/cuda/program.cpp
fabiomestre: I updated it to use UR_CHECK_ERROR :+1:

fabiomestre:
> LGTM
> 
> Does it need any new test, or is it already covered?

There is already a test in the UR CTS that checks this flag

source/adapters/cuda/program.cpp
hdelan: This should be OK. The entry point has a Device param, so multi dev ctx just changes constructor of program to take the device as well

PR #991, opened 2023-10-23
[SPEC] Clarify the BinaryType output of urProgramCreateWithBinary

fabiomestre:


scripts/core/program.yml
```diff
@@ -127,6 +127,7 @@ analogue:
     - "**clCreateProgramWithBinary**"
 details:
     - "The application may call this function from simultaneous threads."
+    - "Following a successful call to this entry point, `phProgram` will contain a binary of type $X_PROGRAM_BINARY_TYPE_COMPILED_OBJECT for `hDevice`."
```
aarongreig, line 130: I'm not sure this is necessarily true, COMPILED_OBJECT implies that you still need to pass the program to urProgramLink before you can use it, but you could might have queried the binary that you're passing to CreateWithBinary from a program that has already been linked, in which case it would be a BINARY_TYPE_EXECUTABLE (or possibly a BINARY_TYPE_LIBRARY) and not require further linking. I'm looking at the CL spec to see if it says anything on how this works there but it might be safer to just say that you can query UR_PROGRAM_BUILD_INFO_BINARY_TYPE with ProgramGetBuildInfo to find out the binary type.

aarongreig: ok it's weird, I thought the common thing would be that you use CreateWithBinary passing a binary with TYPE_EXECUTABLE, but CL makes no mention of that: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_PROGRAM_BINARY_TYPE what it does mention is that the type can be LIBRARY

fabiomestre: I remember getting quite confused about this in the OpenCL spec. But in the end, I think my conclusion was that, if `clCreateProgramWithBinary` is called, it cannot be an EXECUTABLE because `clLinkProgram` is required for that.

I might have missed the BINARY_TYPE_LIBRARY. I'm not sure how that would work in UR though. Can we distinguish between libraries and compiled binaries in any other adapter except for OpenCL?

aarongreig: It isn't necessarily useful for other adapters, but we do have the TYPE_LIBRARY enum in UR, and with the CL adapter you could do a series of calls (with the right SPIR-V module) that results in a ProgramCreateWithBinary call producing a program with a TYPE_LIBRARY binary in it. I don't think it's worth writing rules preventing that so we might as well mention it here

fabiomestre: Fair enough. I will update the PR to mention that the type can be either COMPILED_OBJECT or LIBRARY :+1:

aarongreig (approved):

Issue #871, opened 2023-09-19
Optimize bucket selection for allocation sizes


DamianDuy:
The existing algorithm for mapping an allocation request to a bucket currently simply iterates over buckets to find a bucket capable of handling the requested size. In many allocators, this is optimized to be an array lookup.

PR #994, opened 2023-10-25
[umf] optimize bucket selection for allocation sizes

DamianDuy:
This PR changes linear search for buckets to calculating the bucket index based on calculating the power of two (or halfway power of two) larger than Size parameter by using bitwise operations.

Solves: #871


pbalcer:
This affects SYCL, so please follow the process outlined in the [contributing guide](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#adapter-change-process).

pbalcer (commented):
hm, can you please do a micro-benchmark to see which implementation is better? My initial thinking was something like https://github.com/oneapi-src/unified-runtime/commit/0f296e73fc1fdc2f45808ae38d8d634e9f0b4aaf - that's the common approach to this problem.

source/common/umf_pools/disjoint_pool.cpp
```diff
@@ -902,6 +904,10 @@ umf_result_t DisjointPool::initialize(umf_memory_provider_handle_t *providers,
         !((parameters.MinBucketSize & (parameters.MinBucketSize - 1)) == 0)) {
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
+    // MinBucketSize parameter can't be larger than CutOff.
+    if (parameters.MinBucketSize > CutOff) {
```
igchor, line 908: Is CutOff exposed somehow to the user? If not, how will user know what is the max size he can set?
If this is not exposed, I think you should just set MinBucketSize to CutOff if it's bigger.

igchor:
> hm, can you please do a micro-benchmark to see which implementation is better? My initial thinking was something like [0f296e7](https://github.com/oneapi-src/unified-runtime/commit/0f296e73fc1fdc2f45808ae38d8d634e9f0b4aaf) - that's the common approach to this problem.

I'm not sure if this is worth the effort, but you could also try moving the `size` outside of the bucket class (create a second array/vector of the same size as the current `Buckets` that would only keep `size`) and do a binary search on that. This way you would have much fewer cache misses.

source/common/umf_pools/disjoint_pool.cpp
```diff
@@ -821,13 +821,15 @@ void *DisjointPool::AllocImpl::allocate(size_t Size, size_t Alignment,
 Bucket &DisjointPool::AllocImpl::findBucket(size_t Size) {
     assert(Size <= CutOff && "Unexpected size");
 
-    auto It = std::find_if(
-        Buckets.begin(), Buckets.end(),
-        [Size](const auto &BucketPtr) { return BucketPtr->getSize() >= Size; });
+    auto bucketIterator = std::lower_bound(
+        Buckets.begin(), Buckets.end(), Size,
+        [](const std::unique_ptr<Bucket> &bucket, std::size_t size) {
+            return size > bucket->getSize();
```
kswiecicki, line 827: Shouldn't this be `return size >= bucket->getSize();` (bigger or equal)?

PatKamin (approved):

source/common/umf_pools/disjoint_pool.cpp
DamianDuy: Done.

DamianDuy:
After microbenchmarking I changed the implementation from binary search (which was much slower) to the approach with calculating index using bitwise operations.

source/common/umf_pools/disjoint_pool.cpp
```diff
@@ -818,16 +828,39 @@ void *DisjointPool::AllocImpl::allocate(size_t Size, size_t Alignment,
...
+    size_t position = getLeftmostSetBitPos(Size);
 
-    auto It = std::find_if(
-        Buckets.begin(), Buckets.end(),
-        [Size](const auto &BucketPtr) { return BucketPtr->getSize() >= Size; });
+    size_t lower_bound = (size_t)1 << position;
+    size_t diff = (position - MinBucketSizeExp) << 1;
 
-    assert((It != Buckets.end()) && "Bucket should always exist");
+    if (Size == lower_bound) {
```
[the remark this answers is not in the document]
DamianDuy: Unfortunately, it doesn't work correctly with halfway values and personally I find `if` branches more readable. Also wouldn't this code branch anyway because of a usage of the `&&` operator?

DamianDuy:
I am closing this PR because there is a corresponding one in the unified-memory-framework repositorium:
https://github.com/oneapi-src/unified-memory-framework/pull/37

PR #1005, opened 2023-10-26
Bump adapters branch

omarahmed1111:
bump adapters branch to catch up to main branch to this [commit](https://github.com/oneapi-src/unified-runtime/commit/9d82232b9c7e57997325608320e2d89c5fb95187)


pbalcer:
Are you cherry-picking individual commits? That risks duplicating commits if we were to ever merge the two branches again.

omarahmed1111:
> Are you cherry-picking individual commits? That risks duplicating commits if we were to ever merge the two branches again.

Yes, but I have retrieved the list of commits that doesn't exist in adapters branch and was picking from them, and was checking that this changes doesn't exist in the adapters branch before cherry-picking them. but, maybe there is another cleaner way and more safer, but I picked this approach as i thought it might be the safest in terms of not dropping any functionalities.

omarahmed1111:
> Are you cherry-picking individual commits? That risks duplicating commits if we were to ever merge the two branches again.

However, I don't think we could avoid duplication completely as some changes was submitted to the 2 branches on 2 different PRs, so as I understand, that commits will be duplicated when merging the two branches.

pbalcer:
If the intention is to merge adapters into main, it should be possible to remove these duplicate commits at that point.
As for this change, unless the purpose here is to be very selective about what changes are being moved to the adapters branch, wouldn't it be better to merge main into adapters? That would preserve commit hashes and would make it simpler to then merge adapters back into main (no need to manually untangle the commit history).

omarahmed1111:
> Are you cherry-picking individual commits? That risks duplicating commits if we were to ever merge the two branches again.

This should be a merge now rather than a cherry-pick so it should have all the commit hashes as in main branch so should avoid duplication after merging.

callumfare (approved):

pbalcer (approved):

veselypeta (approved):

kbenzie:
> wouldn't it be better to merge main into adapters?

The branches have diverges quite a bit now and its been a bit tricky to actually get everything udpated so we decided to split it into a few steps rather than all at once.

kbenzie (approved):

PR #1013, opened 2023-10-30
[HIP] Fix get mem size segfault

veselypeta:
This PR:
* Fixes segfault when querying memory for it's size as the BasePtr was not defined
* Supports getting the memory size for surface types
* Adds CTS for `urMemGetInfo` with an Image

intel-llvm testing [here](https://github.com/intel/llvm/pull/11709)

### DIFF
--- a/source/adapters/hip/memory.cpp
+++ b/source/adapters/hip/memory.cpp
@@ -29,7 +29,7 @@ size_t GetHipFormatPixelSize(hipArray_Format Format) {
   case HIP_AD_FORMAT_FLOAT:
     return 4;
   default:
-    assert(0 && "Invalid hipFormat specifier.");
+    detail::ur::die("Invalid HIP format specifier");
   }
 }
 
@@ -257,7 +257,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
 
   UR_ASSERT(MemInfoType <= UR_MEM_INFO_CONTEXT,
             UR_RESULT_ERROR_INVALID_ENUMERATION);
-  UR_ASSERT(hMemory->isBuffer(), UR_RESULT_ERROR_INVALID_MEM_OBJECT);
 
   UrReturnHelper ReturnValue(propSize, pMemInfo, pPropSizeRet);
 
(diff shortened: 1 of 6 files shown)


source/common/ur_util.hpp
```diff
@@ -293,4 +293,6 @@ inline ur_result_t exceptionToResult(std::exception_ptr eptr) {
     }
 }
 
+template <class> inline constexpr bool ur_always_false_t = false;
```
veselypeta, line 296: Defer the evaluation of this until it is actually instantiated. using `static_assert(false, "blah");` will always fail.

source/adapters/hip/memory.cpp
```diff
@@ -243,9 +265,31 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
   switch (MemInfoType) {
   case UR_MEM_INFO_SIZE: {
     try {
-      size_t AllocSize = 0;
-      UR_CHECK_ERROR(hipMemGetAddressRange(
-          nullptr, &AllocSize, std::get<BufferMem>(hMemory->Mem).Ptr));
+      const auto MemVisitor = [](auto &&Mem) -> size_t {
+        using T = std::decay_t<decltype(Mem)>;
```
ldrumm, line 269: Why is this templated? Surely we know the type here?

If it's mean

ldrumm: Whoops. Sent that without finishing my sentence.

Just wondering why we need std::visit and a generic lambda.

veselypeta: Not necessarily - It can be a struct with the call operator overloaded for each of the types of the variant.
I've seen before people define a helper struct to which you can add the overloads:

i.e. 
```cpp
template<class... Ts> struct Overload : Ts...{
using Ts::operator()...;
};
template<class... Ts> Overload(Ts...) -> Overload<Ts...>;

// then you can something like this.
std::variant<int, float, long> var;
std::visit(Overload{
[](int x){ /* run on int*/},
[](float x){ /* run on float*/},
[](long x){ /* run on long*/},
}, var);

```

jinz2014:
@veselypeta 
I have a question about the memory types. What are other memory types besides BufferMemory and SurfaceMemory ?

veselypeta:
> @veselypeta I have a question about the memory types. What are other memory types besides BufferMemory and SurfaceMemory ?

Those are the only ones I believe

aarongreig (approved):

veselypeta:
Covered in #1030

PR #870, opened 2023-09-18
[UR] Support C style arrays

veselypeta:
This PR:
* Adds support to type members of structs as c-style arrays
* Changes `ur_exp_sampler_addr_modes_t` to use array as requested in #863 

Closes #869

Issue #136, opened 2022-12-14
A new Python project should use type annotations


keryell:
Make the API clear and type-safe by relying on https://peps.python.org/pep-0484/ and https://peps.python.org/pep-0526/

kbenzie:
Python bindings were removed.

Issue #135, opened 2022-12-14
Add some hierarchy in the Python header


keryell:
The Python code looks flat like C from 1972.
But there are classes and modules in Python. :-)
For example `urProgramCreateWithBinary` could become `oneapi.program.create_with_binary`.
I am unsure that squashing the imported modules like
```Python
from ctypes import *
from enum import *
```
instead of
```Python
from ctypes import *
from enum import *
```
improves the code quality and readability. https://peps.python.org/pep-0008/#imports
I am curious about how the C++ API will look like. :-)

pbalcer:
The `ur.py` code contains automatically generated raw FFI bindings. Eventually, it might make sense to build a rich idiomatic python library on top of it, but there are no immediate plans to do so. That's because the UR interface is still being defined and would be a moving target for any higher-level software layer.

keryell:
Thanks for the clarification.
But since it is automatically generated, this is even better: you could improve the automatic generator to add the right types at the right place and the right hierarchy. Otherwise it will pollute all the Python global namespace.

kbenzie:
Python bindings were removed.

PR #1014, opened 2023-10-30
Second patch of bumping adapters branch

omarahmed1111:
bump adapters branch to catch to main till this [commit](https://github.com/oneapi-src/unified-runtime/commit/b12b606c35bb7469c1dfdf932e2917238635f02d). This is a continuation of this [PR](https://github.com/oneapi-src/unified-runtime/pull/1016)

Intel/LLVM testing and adding some fixes on intel/llvm side associated to this changes: [here](https://github.com/intel/llvm/pull/11771)


omarahmed1111:
Hi @pbalcer , the changes here were failing in python-basic test, the reason of the failure is that `ur.py` file was not able to catch the function pointer for `cooperative kernels` experimental feature from the loader. This is also happening for adding any experimental feature in this branch. I also tried with the current adapters branch and the same issue was there. Do you have any idea what could have caused that? or where does `ur.py` look into the loader (files linked to generate `libur_loader.so`) when it tries to link the ur loader entries with the python part? Thanks!

pbalcer:
> Hi @pbalcer , the changes here were failing in python-basic test, the reason of the failure is that `ur.py` file was not able to catch the function pointer for `cooperative kernels` experimental feature from the loader. This is also happening for adding any experimental feature in this branch. I also tried with the current adapters branch and the same issue was there. Do you have any idea what could have caused that? or where does `ur.py` look into the loader (files linked to generate `libur_loader.so`) when it tries to link the ur loader entries with the python part? Thanks!

I'll take a look tomorrow, I'm guessing experimental features just aren't getting pulled into the python api properly.

But I think a better question is: do we need the python bindings in the first place? they are annoying to keep in a working state, we barely test them, and, as far as I know, no one uses them. They are a leftover from the origin of the project as a L0 loader fork. @kbenzie what do you think?

kbenzie:
> But I think a better question is: do we need the python bindings in the first place? they are annoying to keep in a working state, we barely test them, and, as far as I know, no one uses them. They are a leftover from the origin of the project as a L0 loader fork. @kbenzie what do you think?

This is a good question. We don't have a use case for these right now. I'm not attached to the Python bindings at all, and its been noted in #136 and #135 that they are not idiomatic either.

Practically, removing the Python bindings while synchronizing the main and adapters branches might be a bit painful.

pbalcer:
> not able to catch the function pointer for `cooperative kernels` experimental feature from the loader. 

#1036 

> Practically, removing the Python bindings while synchronizing the main and adapters branches might be a bit painful.

I was more thinking about just suppressing the python test errors for the time being. But let's revisit the topic once all this branch merging is sorted.

kbenzie:
> I was more thinking about just suppressing the python test errors for the time being. But let's revisit the topic once all this branch merging is sorted.

Ah yes, disabling the basic python test would be simpler. Okay @omarahmed1111 lets to that to unblock this, I think commenting out the `add_subdirectory(python)` in `test/CMakeLists.txt` would be the easiest way.

EwanC (approved):
command-buffer changes LGTM, this PR is very helpful in unblocking some other changes we have pending.

isaacault (approved):
Bindless Images LGTM!

source/adapters/cuda/common.cpp
```diff
@@ -111,7 +111,7 @@ void detail::ur::assertion(bool Condition, const char *Message) {
 }
 
 void detail::ur::cuPrint(const char *Message) {
-  std::cerr << "ur_print: " << Message << std::endl;
+  std::fprintf(stderr, "ur_print: %s\n", Message);
```
fabiomestre, line 114: Why was this change necessary?

omarahmed1111: `cerr` was catching a weird seg fault that was avoided with `fprintf`

fabiomestre: It worries me a bit that we never found out what exactly was causing the segfault. It might be a segfault / undefined behavior in some other part of the adapter that for some reason tends to crash inside std::cerr. I think we should investigate this further. Maybe running a sanitizer or valgrind might help.

However, since it seems to pass with the change to fprintf, maybe this PR can still be merged as long as we look into it afterwards.

kbenzie: > However, since it seems to pass with the change to fprintf, maybe this PR can still be merged as long as we look into it afterwards.

Agreed, open an issue to investigate this later.

omarahmed1111: The issue here was that the changes from this [PR](https://github.com/oneapi-src/unified-runtime/pull/1016) was not included due to rebasing on adapters revert commit, so the deepbind loading of the loaders was still there, and that was causing this issue. once I have included the revert of the revert commit (the commit for this [PR](https://github.com/oneapi-src/unified-runtime/pull/1016)) and the deepbind removal was included. This problem have been fixed. So, it should be fixed now.

Seanst98 (changes requested):
My commit adding unique addressing modes per dimension for the bindless images extension found [here](https://github.com/oneapi-src/unified-runtime/pull/1014/commits/5640ea63f5995a0636ff1f264a8c17b73a26cfef) and [PR here](https://github.com/oneapi-src/unified-runtime/pull/863) has been added in this bump. 

At the time, UR did not support C style arrays so the above commit compromised by using 3 different vars to represent each dimension and was merged into main. A new commit was pushed to main and merged [here](https://github.com/oneapi-src/unified-runtime/pull/870/commits/7217fd7a540d22a6eee153f15bffea2beceea725) and [PR here](https://github.com/oneapi-src/unified-runtime/pull/870) which added C style array support and modified the extension to use C style arrays. 

I can't find the commits which update to using C style arrays in this bump. Can we add it?

kbenzie:
> I can't find the commits which update to using C style arrays in this bump. Can we add it?

We've split up the main bumps because that was easier to handle the issues which arose from this process. We will be pulling in the rest of the changes from main in a third PR. Then, I think we'll merge them all back to back so they should land in adapters branch at the same time.

Seanst98 (approved):

fabiomestre (approved):

omarahmed1111:
Due to a windows failure that will be fixed [here](https://github.com/oneapi-src/unified-runtime/pull/1084) and to make it easier, we will combine the second and third bump into one [here](https://github.com/oneapi-src/unified-runtime/pull/1072). So closing this for now.

PR #1017, opened 2023-10-30
[HIP] Implement urMemImageGetInfo

veselypeta:
This PR:
* uses hipArray3DgetDescriptor to implement urMemImageGetInfo in the Hip adapter.

intel/llvm testing [here](https://github.com/intel/llvm/pull/11717)

### DIFF
--- a/source/adapters/hip/memory.cpp
+++ b/source/adapters/hip/memory.cpp
@@ -481,11 +481,91 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemImageCreate(
   return Result;
 }
 
-/// \TODO Not implemented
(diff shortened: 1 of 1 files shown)


jinz2014:
Hip->HIP in 'Invalid Hip format specified' ?

aarongreig:
why doesn't this seem to need a change to the CTS match files? I'd expect it to make the MemImageGetInfo tests start passing

veselypeta:
> why doesn't this seem to need a change to the CTS match files? I'd expect it to make the MemImageGetInfo tests start passing

This will need to be done, but the memory tests are currently segfaulting until #1013 is merged. 👍

veselypeta:
Covered in #1030

PR #1018, opened 2023-10-30
Trim some jobs from PR workflows to reduce runner usage.

aarongreig:
Moving the adapter code into this repo has brought a lot more PR traffic and as a result we are starving other projects in oneapi-src of github runners. This change reduces the number of jobs run per-pr from 47 to 37, which in turn reduces the number of runners used and should improve the runner situation for the other projects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,8 @@ option(UR_BUILD_EXAMPLE_CODEGEN "Build the codegen example." OFF)
 option(VAL_USE_LIBBACKTRACE_BACKTRACE "enable libbacktrace validation backtrace for linux" OFF)
 option(UR_ENABLE_ASSERTIONS "Enable assertions for all build types" OFF)
 
+include(Assertions)
+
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -77,6 +78,32 @@ if(NOT MSVC)
     endif()
 endif()
 
+# This is copied wholesale from the implementation of LLVM_ENABLE_ASSERTIONS
```
aarongreig, line 81: does this need some kind of extra copyright statement or something? I'm not *not* going to steal this, it's a nice known good implementation that does exactly what we need

kbenzie: Probably, it would be easier to put this in a separate file e.g. `cmake/Assertions.cmake` and add a separate license notice in that file, then include it here.

aarongreig: I've done that, hopefully its sufficient

kbenzie: Looks good, could also add a permalink to the line in the file where the code came from in the license comment to clarify of source.

kbenzie (approved):
LGTM

PR #1021, opened 2023-10-31
[CUDA][HIP] Move adapter specific testing out of CTS

kbenzie:
This patch moves the `test-adapter-cuda` and `test-adapter-hip` test suites out of the `test/conformance` directory. The goal of this move is to ensure that all adapters will be tested with the same set and number of conformance tests, making conformance pass rates directly comparable between adapters. Up to now the CUDA and HIP adapters have included
additional unit tests of internals which skews pass rates.


aarongreig (approved):

test/adapters/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+    target_link_libraries(${target} PRIVATE
+        ${PROJECT_NAME}::loader
+        ${PROJECT_NAME}::headers
+        ${PROJECT_NAME}::testing
+        ${PROJECT_NAME}::common
+        GTest::gtest)
+
+    add_test(NAME ${target} COMMAND $<TARGET_FILE:${target}>)
+    set_tests_properties(${target} PROPERTIES
+        LABELS "adapter:${name}"
```
pbalcer, line 36: With this change the adapter tests no longer run in CI. I think this needs to be `LABELS "adapter;${name}"`.
Also, this removes the match file machinery from these tests, but it appears it was never necessary for these particular tests (or I can't find the corresponding match files).

kbenzie: > With this change the adapter tests no longer run in CI. I think this needs to be LABELS "adapter;${name}".

Good spot, I'll fix this.

> Also, this removes the match file machinery from these tests, but it appears it was never necessary for these particular tests (or I can't find the corresponding match files).

These should aways pass and since they are adapter specific by nature we don't need the ability to run them against different adapters with different match files. We can also lean on GTest's facilities for skipping tests if necessary but so far its not.

kbenzie: `ctest -L "adapter"` was matching too many tests so I changed this label to `adapter-specific` and added the *Test adapter specific* step to the workflow.

pbalcer (approved):

PR #1025, opened 2023-11-01
Add UR_ENABLE_ASSERTIONS flag for enabling asserts on all build types.

aarongreig:


kbenzie (approved):

pbalcer (approved):

cmake/Assertions.cmake
```diff
@@ -0,0 +1,30 @@
...
+# https://github.com/llvm/llvm-project/blob/6be0e979896f7dd610abf263f845c532f1be3762/llvm/cmake/modules/HandleLLVMOptions.cmake#L89
+if(UR_ENABLE_ASSERTIONS)
+  # MSVC doesn't like _DEBUG on release builds
+  if( NOT MSVC )
+    add_compile_definitions(_DEBUG)
+  endif()
+  # On non-Debug builds cmake automatically defines NDEBUG, so we
+  # explicitly undefine it:
+  if( NOT uppercase_CMAKE_BUILD_TYPE STREQUAL "DEBUG" )
+    add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-UNDEBUG>)
```
pbalcer, line 15: `target_compile_options?`
I know this is off by default, but could enabling this option affect projects that use FetchContent with UR? We added [wrappers](https://github.com/oneapi-src/unified-runtime/blob/main/cmake/helpers.cmake#L117) to avoid operating on global CMake variables.

kbenzie: > I know this is off by default, but could enabling this option affect projects that use FetchContent with UR? 

Would this be projects that `FetchContent` UR intel/llvm or projects that UR itself calls `FetchContent`?

I think because CMake handles assertions globally the old way in the `CMAKE_{C,CXX}_FLAGS` specifying `-UDEBUG` globally as well feels necessary, espeically on platforms like Windows which really dislike linking objects compiler with/without assertions.

pbalcer: > Would this be projects that `FetchContent` UR intel/llvm or projects that UR itself calls `FetchContent`?

I was thinking mostly about the former.

> I think because CMake handles assertions globally the old way in the CMAKE_{C,CXX}_FLAGS specifying -UDEBUG globally as well feels necessary, espeically on platforms like Windows which really dislike linking objects compiler with/without assertions.

Sure, makes sense. I'm not really that concerned since this is mostly a dev option. Just curious. Thanks.

PR #1026, opened 2023-11-01
[CTS] Add UR_SYCL_LIBRARY_DIR CMake variable

kbenzie:
The CTS is already dependant on the DPC++ compiler to generate programs inputs for the program, kernel, and enqueue test suites specified via the `UR_DPCXX` CMake variable. If the DPC++ compiler is not installed on the system the executables it outputs will likely fail to find the SYCL runtime library when executed, breaking the generation of CTS program inputs. The patch introduces the `UR_SYCL_LIBRARY_DIR` CMake variable enabling the user to specify the path to the SYCL runtime library for use when generating CTS program inputs.


pbalcer (approved):
Might be useful to also update the CI workflow:
https://github.com/oneapi-src/unified-runtime/blob/adapters/.github/workflows/cmake.yml#L203

README.md
```diff
@@ -133,6 +133,8 @@ List of options provided by CMake:
 | UR_BUILD_ADAPTER_HIP    | Fetch and use hip adapter from SYCL                    | ON/OFF     | OFF     |
 | UR_HIP_PLATFORM         | Build hip adapter for AMD or NVIDIA platform           | AMD/NVIDIA | AMD     |
 | UR_ENABLE_COMGR         | Enable comgr lib usage           | AMD/NVIDIA | AMD     |
+| UR_DPCXX | Path of the DPC++ compiler executable | File path | `""` |
+| UR_SYCL_LIBRARY_DIR | Path of the SYCL runtime library directory | Directory path | `""` |
```
callumfare, line 137: Could be worth mentioning that these env vars are used for the conformance tests, someone not familiar with the project  seeing these might assume that UR has a dependency on DPC++

callumfare (approved):

README.md
kbenzie: They're CMake varaibles but yes, worth a mention.

kbenzie: Fixed

PR #1029, opened 2023-11-02
[UR] Add adapter leak-checking tests

kswiecicki:


pbalcer (commented):
can you add a test?

kswiecicki:
> can you add a test?

Done.
Apparently there's a limit to the number of groups in a regex that ctest can compile (it fails as if the match failed) so I've changed `(.*)` to `.*` everywhere.

pbalcer (approved):

pbalcer:
> > 
> 
> Done. Apparently there's a limit to the number of groups in a regex that ctest can compile (it fails as if the match failed) so I've changed `(.*)` to `.*` everywhere.

Use the python match scripts in the future.

kswiecicki:
> > > 
> > 
> > 
> > Done. Apparently there's a limit to the number of groups in a regex that ctest can compile (it fails as if the match failed) so I've changed `(.*)` to `.*` everywhere.
> 
> Use the python match scripts in the future.

I've looked into the `.match.py` but it seems to support only line-by-line matching. In my case I needed to match some unknown number of backtrace lines.

cmake/match.py
```diff
@@ -5,39 +5,136 @@
...
+    with open(input_file, 'r') as input, open(match_file, 'r') as match:
+        input_lines = input.readlines()
+        match_lines = match.readlines()
+
+    ignored_lines = []
+
+    input_idx = 0
+    match_idx = 0
+    tag_in_effect = None
+    while status := check_status(input_lines[input_idx:], match_lines[match_idx:]):
```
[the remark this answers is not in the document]
pbalcer: no, let's not bump required python version unnecessarily.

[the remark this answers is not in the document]
kswiecicki: I've removed the walrus operator.

PatKamin (approved):

cmake/match.py
```diff
@@ -5,39 +5,137 @@
...
+    print("".join(ignored_lines).strip())
+    print("-" * 80)
 
-with open(input_file, 'r') as input, open(match_file, 'r') as match:
-    input_lines = input.readlines()
-    match_lines = match.readlines()
 
-if len(match_lines) != len(input_lines):
-    sys.exit(f"Line count doesn't match (is: {len(input_lines)}, expected: {len(match_lines)})")
+def print_incorrect_match(match_line, present, expected):
```
pbalcer, line 34: brief?:P

cmake/match.py
```diff
@@ -5,39 +5,137 @@
...
+        input_lines = input.readlines()
+        match_lines = match.readlines()
+
+    ignored_lines = []
+
+    input_idx = 0
+    match_idx = 0
+    tags_in_effect = []
+    status = Status.PROCESSING
+    while status == Status.PROCESSING:
```
pbalcer, line 85: this doesn't seem necessary, this condition will never be false as far as I can tell. You can just do a while True: loop and check status at the end (and exit appropriately).

cmake/match.py
```diff
@@ -5,39 +5,137 @@
...
+        elif status == Status.INPUT_END:
+            print_incorrect_match(match_idx + 1, "", match_lines[match_idx].strip());
+            print_content(input_lines, match_lines, ignored_lines)
+            sys.exit(1)
+        else:
+            input_line = input_lines[input_idx].strip()
+
+        match_line = match_lines[match_idx]
+        if match_line.startswith(Tag.OPT.value):
+            tags_in_effect.append(Tag.OPT)
```
pbalcer, line 106: doesn't seem like there can be more than one active tag?

pbalcer:
I don't think we can merge this is as, it will lead to conflicts with the adapters branch. This patch needs to be rebased on adapters or wait until the branches are merged.

cmake/match.py
kswiecicki: Done.

cmake/match.py
kswiecicki: In the case below, there would be a point when ignore tag (active until the next match) is applied, then the optional tag (active until the next line) is applied next.
```
{{IGNORE}}
{{OPT}} optional line
```

cmake/match.py
kswiecicki: I've moved the status variable inside the loop and changed the loop condition.
If I were to move the status check at the end, then I would need to check if either input or match file is empty to avoid index error. Logger has such test where both input and match files are empty.

kswiecicki:
> I don't think we can merge this is as, it will lead to conflicts with the adapters branch. This patch needs to be rebased on adapters or wait until the branches are merged.

Alright, I can wait for the adapters branch to be merged with main.

cmake/match.py
kswiecicki: Also, the match case statement (switch equivalent) is not available in our current supported python version.

pbalcer:
this does not appear to be changing anything but tests?

kbenzie:
I'm getting similar failures locally which are the result of missing device binaries although not exactly the same. I'll continue investigating.

kbenzie:
This appears to have been caused by [this change](https://github.com/oneapi-src/unified-runtime/commit/5dc6e464c83fa41045d4f352a300db9b16400d1d#diff-6bc28da4be765fa3d871baab7e4d458635e40dc2b131a573167f24c7cbc7c9b2R217) where `urProgramCreateWithIL` would formerly pass through to `urProgramCreateWithBinary` but no longer does, instead calling `detail::ur::die()`.

What I don't currently understand is why this isn't affecting all PR's.

kswiecicki:
> This appears to have been caused by [this change](https://github.com/oneapi-src/unified-runtime/commit/5dc6e464c83fa41045d4f352a300db9b16400d1d#diff-6bc28da4be765fa3d871baab7e4d458635e40dc2b131a573167f24c7cbc7c9b2R217) where `urProgramCreateWithIL` would formerly pass through to `urProgramCreateWithBinary` but no longer does, instead calling `detail::ur::die()`.
> 
> What I don't currently understand is why this isn't affecting all PR's.

Thank you for the investigation @kbenzie. With @pbalcer help, we were able to figure out the cause of those test fails.
Previous match script had a bug that allowed a test to pass despite not matching the `Aborted` input if all of the match file lines were marked as `{{OPT}}`. As a temporary solution I've added `Aborted` as an expected input, I've seen other adapter tests do that as well.

pbalcer (approved):

PatKamin (approved):

kbenzie (approved):

PR #1030, opened 2023-11-02
benie/hip cts fixes combined

kbenzie:
- [HIP] Fix Kernel Compilation on AMD
- [HIP] Don't link against device binaries to fix CI
- [HIP] Fix host/device synchronization
- [HIP] Make synchronized time optional
- [HIP][CTS] Fix Device CTS failures
- [UR] Remove passing queue tests from match file
- [HIP] Fix get mem size segfault
- [OO] Add CTS for urMemGetInfo with images
- [UR] Disable Image query on CUDA
- [UR] Disable Image query on L0
- [UR] Windows doesn't like std::array without template arguments
- [HIP] Implement urMemImageGetInfo
- [HIP] Define all UR entry points


pbalcer (approved):

kbenzie:
@oneapi-src/unified-runtime-hip-write would be good to get this reviewed, we're aiming to include this in 2024.1.

jchlanda (approved):
HIP 👍

PR #1031, opened 2023-11-02
[HIP] Add missing symbols hip

hdelan:
The global exported symbols are here https://github.com/oneapi-src/unified-runtime/blob/adapters/source/adapters/adapter.map.in

This file https://github.com/oneapi-src/unified-runtime/blob/adapters/source/adapters/hip/ur_interface_loader.cpp is missing some symbols that are written in the lib map. This caused a linker error for LLD when linking in DPC++. This patch adds them, fixing the linker error.

### DIFF
--- a/source/adapters/hip/ur_interface_loader.cpp
+++ b/source/adapters/hip/ur_interface_loader.cpp
@@ -310,13 +310,17 @@ UR_DLLEXPORT ur_result_t UR_APICALL urGetUsmP2PExpProcAddrTable(
 // TODO: Implement
 UR_DLLEXPORT ur_result_t UR_APICALL urGetBindlessImagesExpProcAddrTable(
     ur_api_version_t, ur_bindless_images_exp_dditable_t *) {
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+  // This needs to return UR_RESULT_SUCCESS or else the platform can't be
+  // initialized
+  return UR_RESULT_SUCCESS;
 }
 
 // TODO: Implement
 UR_DLLEXPORT ur_result_t UR_APICALL
 urGetUSMExpProcAddrTable(ur_api_version_t, ur_usm_exp_dditable_t *) {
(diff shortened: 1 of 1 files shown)


kbenzie:
The testing doesn't like this change, will need to investigagte that before we can merge.

hdelan:
> The testing doesn't like this change, will need to investigagte that before we can merge.

Investigating

pbalcer:
Looking at the loader implementation, it expects the addr table to at least be null-initialized.

pbalcer:
It also just occurred to me that you can't return an error from these functions if you want the adapter to be functional. Once once get table function fails, the rest of them are skipped (https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_ldrddi.cpp#L8626).

I think we have a bug that such platform is then made accessible to the user.

hdelan:
> It also just occurred to me that you can't return an error from these functions if you want the adapter to be functional. Once one get table function fails, the rest of them are skipped (https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_ldrddi.cpp#L8626).
> 
> I think we have a bug that such platform is then made accessible to the user.

Yeah the testing passes if you change retail to `UR_RESULT_SUCCESS`. I'm not sure if this is desirable but have added a comment in the code to say we need this to return `UR_RESULT_SUCCESS` or else platform initialisation will fail.

hdelan:
@kbenzie any word on this? Would be nice to fix lld builds in DPC++

pbalcer (approved):

mmoadeli (commented):
Is there any need for a (warning or error ) message inside the functions annotated with `// TODO: Implement` ?

kbenzie:
> @kbenzie any word on this? Would be nice to fix lld builds in DPC++

We need to prioritise merging PR's for the next release which this isn't targetted for. Will aim to merge this after we've drained that release PR backlog.

PR #1035, opened 2023-11-03
[EXP][CMDBUF] Add support for recovering CommandBuffer profiling information

mfrancepillois:
Copy command-buffer event timestamps into a dedicated USM memory region.
Get the command-buffer start and end timestamps from this memory.
Move events reset from the middle command list to the first to allow the copy of the profiling info in the last command list and relax command list order.

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -45,13 +45,13 @@
   │  Prefix  │ Commands added to UR command-buffer by UR user │ Suffix  │
   └──────────┴────────────────────────────────────────────────┴─────────┘
 
-            ┌───────────────────┬──────────────────────────────┐
-  Prefix    │Reset signal event │ Barrier waiting on wait event│
-            └───────────────────┴──────────────────────────────┘
+            ┌───────────────────┬──────────────┐──────────────────────────────┐
+  Prefix    │Reset signal event │ Reset events │ Barrier waiting on wait event│
+            └───────────────────┴──────────────┘──────────────────────────────┘
 
             ┌─────────────────────────────────────────────┐──────────────┐
-  Suffix    │Barrier waiting on sync-point event,         │ Reset events │
-            │signalling the UR command-buffer signal event│              │
+  Suffix    │Barrier waiting on sync-point event,         │  Query CMD   │
+            │signalling the UR command-buffer signal event│  Timestamps  │
             └─────────────────────────────────────────────┘──────────────┘
 
   For a call to `urCommandBufferEnqueueExp` with an event_list `EL`,
@@ -433,6 +433,10 @@ urCommandBufferCreateExp(ur_context_handle_t Context, ur_device_handle_t Device,
 
   ZeStruct<ze_command_list_desc_t> ZeCommandListDesc;
   ZeCommandListDesc.commandQueueGroupOrdinal = QueueGroupOrdinal;
+  // Dependencies between commands are explicitly enforced by sync points when
+  // enqueuing. Consequently, relax the command ordering in the command list
+  // can enable the backend to further optimize the workload
+  ZeCommandListDesc.flags = ZE_COMMAND_LIST_FLAG_RELAXED_ORDERING;
 
   ze_command_list_handle_t ZeCommandList;
   // TODO We could optimize this by pooling both Level Zero command-lists and UR
@@ -498,13 +502,6 @@ urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t CommandBuffer) {
   // submission dependencies have been satisfied.
   ZE2UR_CALL(zeCommandListAppendEventReset,
              (CommandBuffer->ZeCommandList, CommandBuffer->WaitEvent->ZeEvent));
-
-  // Reset the L0 events we use for command-buffer internal sync-points to the
-  // non-signalled state
-  for (auto Event : WaitEventList) {
-    ZE2UR_CALL(zeCommandListAppendEventReset,
-               (CommandBuffer->ZeCommandList, Event));
-  }
 
   // Close the command list and have it ready for dispatch.
   ZE2UR_CALL(zeCommandListClose, (CommandBuffer->ZeCommandList));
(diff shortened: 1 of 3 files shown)


mfrancepillois:
Linked DPC++ PR: https://github.com/intel/llvm/pull/11324

source/adapters/level_zero/event.cpp
```diff
@@ -463,6 +465,70 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventGetProfilingInfo(
...
 
+  // A Command-buffer consists of three command-lists for which only a single
+  // event is returned to users. The actual profiling information related to the
+  // command-buffer should therefore be extrated from graph events themsleves.
+  // The timestamps of these events are saved in a memory region attached to
+  // event usning CommandData field. The timings must therefore be recovered
+  // from this memory.
+  if (Event->CommandType == UR_COMMAND_COMMAND_BUFFER_ENQUEUE_EXP) {
+    if (Event->CommandData) {
+      struct command_buffer_profiling_t *ProfilingsPtr;
```
EwanC, line 476: Shouldn't need the `struct` prefix, that's a C convention rather than C++ (same for other places)
```suggestion
      command_buffer_profiling_t *ProfilingsPtr;
```

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -764,6 +772,41 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
...
+
+      struct command_buffer_profiling_t *Profiling =
+          new command_buffer_profiling_t();
+
+      Profiling->NumEvents = WaitEventList.size();
+
+      urUSMSharedAlloc(RetEvent->Context, CommandBuffer->Device, &USMDesc,
+                       nullptr, Size, (void **)&Profiling->Timestamps);
+
+      ZE2UR_CALL(zeCommandListAppendQueryKernelTimestamps,
```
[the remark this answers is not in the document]
mfrancepillois: Command-buffer events are all created from an event pool that was created using this flag.
Indeed, if no queue_handler is passed when calling the `EventCreate` funtion (as is the case when the command-buffer enqueues commands), the `ProfilingEnabled` flag is set to ensure that events come from a profiling-enabled pool.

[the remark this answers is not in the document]
EwanC: cool 👍

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -764,6 +772,41 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
...
+
+      size_t Size = WaitEventList.size() * sizeof(ze_kernel_timestamp_result_t);
+
+      struct command_buffer_profiling_t *Profiling =
+          new command_buffer_profiling_t();
+
+      Profiling->NumEvents = WaitEventList.size();
+
+      urUSMSharedAlloc(RetEvent->Context, CommandBuffer->Device, &USMDesc,
+                       nullptr, Size, (void **)&Profiling->Timestamps);
```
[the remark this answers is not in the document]
mfrancepillois: The implementation has been updated to make it use host only memory instead of USMShared since `zeCommandListAppendQueryKernelTimestamps` seems to be able to copy directly to host memory.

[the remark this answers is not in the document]
EwanC: Can we guarantee that in all cases if this is the usage wording?

> The application must ensure the memory pointed to by both dstptr and pOffsets is accessible by the device on which the command list was created.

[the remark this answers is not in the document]
EwanC: Trying to read up on L0 USM memory support https://spec.oneapi.io/level-zero/latest/core/PROG.html#types 
and turns out the minimum capabilities are actually different from SYCL/USM. I incorrectly assumed they were the same, but L0 devices do always support shared USM. So I think we can use your previous approve using the shared USM allocation

## L0 support matrix
image

## SYCL/OpenCL support matrix
image

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -753,17 +756,43 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
...
+      Profiling->Timestamps =
+          new ze_kernel_timestamp_result_t[Profiling->NumEvents];
+
+      ZE2UR_CALL(zeCommandListAppendQueryKernelTimestamps,
+                 (SignalCommandList->first, WaitEventList.size(),
+                  WaitEventList.data(), (void *)Profiling->Timestamps, 0,
+                  RetEvent->ZeEvent, 1,
+                  &(CommandBuffer->SignalEvent->ZeEvent)));
+
+      RetEvent->CommandData = static_cast<void *>(Profiling);
```
jandres742, line 790: so in the before we were adding always a barrier, but now only if PROFILING is enabled. why don't we need the barrier on this branch?

source/adapters/level_zero/event.cpp
```diff
@@ -463,6 +465,70 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventGetProfilingInfo(
...
+      }
+      case UR_PROFILING_INFO_COMMAND_END: {
+        ProfilingsPtr =
+            static_cast<command_buffer_profiling_t *>(Event->CommandData);
+        // Sync-point order does not necessarily match to the order of
+        // execution. We therefore look for the last command executed.
+        uint64_t MaxEnd = ProfilingsPtr->Timestamps[0].global.kernelEnd;
+        uint64_t LastStart = ProfilingsPtr->Timestamps[0].global.kernelStart;
+        for (uint64_t i = 1; i < ProfilingsPtr->NumEvents; i++) {
+          uint64_t Timestamp = ProfilingsPtr->Timestamps[i].global.kernelEnd;
```
jandres742, line 502: I dont follow this logic for MaxEnd and the same for MinStart in line 487, could you elaborate?

source/adapters/level_zero/command_buffer.cpp
mfrancepillois: The barrier was used to signal the end of this specific execution of the CommandBuffer.
The same CommandBuffer can be executed multiple times. A first barrier is added in the main command-list. https://github.com/oneapi-src/unified-runtime/pull/1035/files#diff-ab021f9451ccb6dd66e3df853d0e123b2f898233a1f609b1e698bef4f6011c71R455
This barrier waits for all commands to complete and signals its completion with `SignalEvent`.
So this barrier is the effective one. 
But `SignalEvent` is common to all executions of the CommandBuffer. 
We therefore add a second barrier in the suffix command-list that waits for `SignalEvent` and triggers a other signal, exclusive to this specific execution of the CommandBuffer. And this exclusive event is returned to the runtime.
Now, if profiling is enabled, we directly used `zeCommandListAppendQueryKernelTimestamps` to do the job since this command accepts events to wait for, and allows signaling its completion with an event.

source/adapters/level_zero/event.cpp
mfrancepillois: Since we are relaxing the command execution order in a command list (https://github.com/oneapi-src/unified-runtime/pull/1035/files#diff-ab021f9451ccb6dd66e3df853d0e123b2f898233a1f609b1e698bef4f6011c71R397), there is no guarantee that the order of commands in list (order of `SyncPoints`) is the actual execution order.
We thus loop over all the event timestamps of CommandBuffer commands and search for the minimum start time, which corresponds to the actual start time of the CommandBuffer execution. And the same for the execution end time, we go through the events and look for the last task to complete (max `kernelEnd` timestamp).

EwanC (approved):

EwanC:
@jandres742 Could you take another look at this PR and see if the responses to your questions are satisfactory.

fabiomestre:

EwanC:
@jandres742 reminder that this PR is waiting on your review.

source/adapters/level_zero/event.cpp
jandres742: thanks @mfrancepillois . I understand now. main problem there is if the number of commands is very huge, so the timestamp traversal would be substantial. 

Now, the relaxation of ordering you are referring to is because you are adding `ZE_COMMAND_LIST_FLAG_RELAXED_ORDERING`, right? in that case, I don't think the application (in this case UR) should be doing the traversal of timestamps to ordering them. It should be the driver. However the L0 spec is not clear on that.

I have opened an issue in the L0 spec to clarify that https://github.com/oneapi-src/level-zero-spec/issues/249

In the meantime, let's continue with this patch, and then once the L0 spec is clarified, we know whether this code can stay or needs to be removed.

jandres742 (approved):

PR #1037, opened 2023-11-03
[UR][CUDA] Fix compatibility with CUDA 11.x

al42and:
The code introduced in 74f42f8fdd3b92b355c5769bd13d4d982e839c78 uses the signature of `cuGraphInstantiate` from CUDA 12.x. In CUDA 11.4-11.8, the same function is called `cuGraphInstantiateWithFlags`, so we use it. In older CUDA, there is only the old version of `cuGraphInstantiate` which does not accept flags.

- https://docs.nvidia.com/cuda/archive/11.8.0/cuda-driver-api/group__CUDA__GRAPH.html#group__CUDA__GRAPH_1g433ae118a751c9f2087f53d7add7bc2c
- https://docs.nvidia.com/cuda/archive/12.0.0/cuda-driver-api/group__CUDA__GRAPH.html#group__CUDA__GRAPH_1g433ae118a751c9f2087f53d7add7bc2c

### DIFF
--- a/source/adapters/cuda/command_buffer.cpp
+++ b/source/adapters/cuda/command_buffer.cpp
@@ -140,12 +140,13 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
 UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
   try {
-#if CUDART_VERSION >= 12000
+#if CUDA_VERSION >= 12000
     UR_CHECK_ERROR(cuGraphInstantiate(&hCommandBuffer->CudaGraphExec,
                                       hCommandBuffer->CudaGraph, 0));
 #else
     UR_CHECK_ERROR(cuGraphInstantiate(&hCommandBuffer->CudaGraphExec,
-                                      hCommandBuffer->CudaGraph, nullptr, nullptr, 0));
+                                      hCommandBuffer->CudaGraph, nullptr,
+                                      nullptr, 0));
 #endif
   } catch (...) {
     return UR_RESULT_ERROR_UNKNOWN;
--- a/source/adapters/cuda/command_buffer.cpp
+++ b/source/adapters/cuda/command_buffer.cpp
@@ -140,8 +140,13 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
 UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
   try {
+#if CUDART_VERSION >= 12000
     UR_CHECK_ERROR(cuGraphInstantiate(&hCommandBuffer->CudaGraphExec,
                                       hCommandBuffer->CudaGraph, 0));
+#else
+    UR_CHECK_ERROR(cuGraphInstantiate(&hCommandBuffer->CudaGraphExec,
+                                      hCommandBuffer->CudaGraph, nullptr, nullptr, 0));
+#endif
   } catch (...) {
     return UR_RESULT_ERROR_UNKNOWN;
   }
--- a/source/adapters/cuda/command_buffer.cpp
+++ b/source/adapters/cuda/command_buffer.cpp
@@ -140,10 +140,15 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
 UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
   try {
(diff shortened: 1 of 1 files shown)


source/adapters/cuda/command_buffer.cpp
```diff
@@ -140,8 +140,14 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
 UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
   try {
+#if CUDA_VERSION >= 12000
     UR_CHECK_ERROR(cuGraphInstantiate(&hCommandBuffer->CudaGraphExec,
                                       hCommandBuffer->CudaGraph, 0));
+#else
+    UR_CHECK_ERROR(cuGraphInstantiate(&hCommandBuffer->CudaGraphExec,
```
al42and, line 147: If there are plans to use flags, we can use `cuGraphInstantiateWithFlags` here, but it is only available in CUDA >=11.4,<12, so we would have to handle three cases here.

Bensuo: We don't have any immediate plans for using flags, though we are looking at utilizing the graph owned allocations in the future so this may become relevant then.

So I don't think there's a need to handle this yet, but could you add a comment noting this may be required in the future?

Bensuo (commented):
Thanks for catching this! Changes look good to me just one minor comment.

source/adapters/cuda/command_buffer.cpp
al42and: I added the code, it's not that much. But I don't have CUDA-11.4 at hand to actually test it.

al42and: > We don't have any immediate plans for using flags, though we are looking at utilizing the graph owned allocations in the future so this may become relevant then.

[cudaGraphInstantiateFlagUseNodePriority](https://docs.nvidia.com/cuda/archive/12.2.0/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg638577f2b0f6a98b5e6a0e6a72d0900495adbcaa8c89c323deec4b2070a4bab5) could also be interesting (together with [`sycl_ext_oneapi_queue_priority`](https://github.com/al42and/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_queue_priority.asciidoc#sycl_ext_oneapi_queue_priority)).

Bensuo: > I added the code, it's not that much. But I don't have CUDA-11.4 at hand to actually test it.

I'll see if I can test this middle case locally with an appropriate CUDA version, since this is not tested on CI.

al42and: I formulated that incorrectly. I have CUDA 11.8, so the middle case is tested. It's the third, <11.4, case that I could not test.

Bensuo: Thanks for confirming that, I've been trying to test but had difficulties getting access to an environment with an early enough CUDA version.

Bensuo (approved):

abagusetty:
Can we get this merged.

Fixes: https://github.com/oneapi-src/unified-runtime/issues/1119

kbenzie:
> Can we get this merged.
> 
> Fixes: #1119

About to do so.

> PR for pre-merge testing: https://github.com/intel/llvm/pull/11773

I'm going to close this testing PR. I'll replace it with a new intel/llvm PR that combines a number of UR PR's together with the goal of getting patches merged more quickly.

al42and:
> I'm going to close this testing PR. I'll replace it with a new intel/llvm PR that combines a number of UR PR's together with the goal of getting patches merged more quickly.

Sure thing!

By the way, is the whole "draft MRs in a different repo just to test" shtick a temporary solution, or is it intended to be the long-term way of running the project? I see the benefit when the changes to IntelLLVM and UR are tightly coupled, but, with two separate repos, I got the impression that the general direction was towards untangling them and making the development more independent?

kbenzie:
> By the way, is the whole "draft MRs in a different repo just to test" shtick a temporary solution, or is it intended to be the long-term way of running the project? I see the benefit when the changes to IntelLLVM and UR are tightly coupled, but, with two separate repos, I got the impression that the general direction was towards untangling them and making the development more independent?

I agree this is an unusual setup and it's primarily due to historical reasons. UR was born out of the Plugin Interface (PI) in intel/llvm which we are still in the process of fully migrating in the SYCL runtime.

We are still working towards decoupling UR from intel/llvm further but it's a time consuming process and must be balanced with our other commitments. The main area which needs improved is our level of testing, both API coverage and the OS/hardware support matrix.

Having now completed the move of all the adapter implementations to this repo the maintenance workload has increased quite a bit for me personally as a result of the current process. So I am personally invested in streamlining things for everyone involved in the project.

al42and:
Thanks for the explanation, @kbenzie

PR #1039, opened 2023-11-03
[UR][OPENCL] Eliminate usage of regex in opencl

omarahmed1111:
Usage of regex was downgrading the performance of opencl adapter strongly. Using [Intel PIN](https://www.intel.com/content/www/us/en/developer/articles/tool/pin-a-dynamic-binary-instrumentation-tool.html) tool, querying the total number of instructions using a custom pin tool, it reported that the affective number of instructions that was executed from inside `libpi_opencl.so` to run [simple-sycl-app](https://github.com/intel/llvm/blob/sycl/sycl/doc/GetStartedGuide.md#run-simple-dpc-application) was **269469** which is an order of magnitude higher than any other adapter. Analyzing this, most of these instructions was generated from the regex library. This was also definetly have a strong effect on the runtime performance of sycl-opencl applications.

The changes in this PR will decrease the effective number of instructions executed in opencl adapter to **~12000** which is a reasonable number comparing it to other adapters (all are in the range of **10000 - ~20000**).


omarahmed1111:
Intel/llvm testing [here](https://github.com/intel/llvm/pull/11204)

source/adapters/opencl/kernel.cpp
```diff
@@ -9,6 +9,8 @@
 //===----------------------------------------------------------------------===//
 #include "common.hpp"
 
+#include <memory>
```
aarongreig, line 12: what's this for? did including `<regex>` in common.hpp pull this in before it was removed?

aarongreig (approved):

source/adapters/opencl/kernel.cpp
omarahmed1111: yes, the build was not seeing `shared_ptr` any more, I think that was because of regex was including this.

fabiomestre (approved):
LGTM

PR #1049, opened 2023-11-07
[OpenCL] Add bounds checking to the Enqueue memory operations.

aarongreig:
This allows us to return UR_ERROR_INVALID_SIZE when we should. Extra
checks are only performed on a non-success error code.


aarongreig:
I had to base this on the mega fix branch as there was a non-trivial overlap in EnqueueMemBufferFill. Only the top commit is relevant.

aarongreig:
(and by top commit I mean most recent... which on github is the bottom commit)

kbenzie (approved):
LGTM

source/adapters/opencl/device.cpp
```diff
@@ -345,6 +345,23 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
...
+
+    if (!Supported) {
+      return UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION;
+    }
+
+    cl_device_pci_bus_info_khr PciInfo = {};
+    CL_RETURN_ON_FAILURE(clGetDeviceInfo(
+        cl_adapter::cast<cl_device_id>(hDevice), CL_DEVICE_PCI_BUS_INFO_KHR,
+        sizeof(PciInfo), &PciInfo, nullptr));
+    return ReturnValue(PciInfo.pci_device);
```
omarahmed1111, line 362: I think we could check here for the version of opencl first as I think this query should be only available from version 3.0, I guess we could guard that behind a preprocessor if, not sure if this should work in a preprocessor or not but there is `UR_EXT_DEVICE_INFO_OPENCL_C_VERSION`define in `plugins/unified_runtime/ur/ur.hpp`

omarahmed1111 (approved):
LGTM, just a small nit

source/adapters/opencl/device.cpp
aarongreig: I think regardless of underlying driver version we don't support building with a CL loader/headers that would be old enough not to have this, if it isn't supported by the device that should get caught by the check for cl_khr_pci_bus_info above

omarahmed1111: ah okay, that make sense. I thought we could as I saw cmake searching for cl version 1.0 and 2.0 before moving to fetching headers from upstream.

Issue #1002, opened 2023-10-26
Gather system info in CI workflows run on self-hosted runners


lukaszstolarczuk:
We could gather system info about platform, tests run on.

At least:
- sycl-ls output
- OS
- drivers version

PR #1050, opened 2023-11-08
Add system info in CI workflows run on self-hosted runners

szadam:


.github/workflows/cmake.yml
```diff
@@ -175,6 +175,12 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
+    - name: Information about platform
```
lukaszstolarczuk, line 178: I'm missing a verb in this name 😉 let's add e.g. "Gather"

.github/workflows/cmake.yml
```diff
@@ -175,6 +175,12 @@ jobs:
...
+      run: |
+        sycl-ls
```
lukaszstolarczuk, line 180: sycl-ls is only available after you download dpcpp. I think we might do this step at the very end of the job (but it has to run with `if: always()` then)

.github/workflows/cmake.yml
```diff
@@ -175,6 +175,12 @@ jobs:
...
+        lsb_release -a
+        lspci | grep VGA
+
```
lukaszstolarczuk, line 183: you may consider adding `clinfo` here (but please, check if all our platforms have it installed)

lukaszstolarczuk (commented):
pls link your commit to the issue (https://github.com/oneapi-src/unified-runtime/issues/1002)

.github/workflows/cmake.yml
```diff
@@ -203,6 +203,16 @@ jobs:
       # This is so that device binaries can find the sycl runtime library
       run: cmake --build ${{github.workspace}}/build -j $(nproc)
 
+    - name: Information about platform
+      if: always()
+      run: |
+        export PATH=${workspace}/dpcpp_compiler/bin:$PATH
+        export CPATH=${workspace}/dpcpp_compiler/include:$CPATH
```
lukaszstolarczuk, line 210: are these exports for `sycl-ls` command? you can then just use them inline with the command itself (meaning, not setting exports) 😉

.github/workflows/cmake.yml
```diff
@@ -203,6 +203,16 @@ jobs:
...
+        sycl-ls
+        lsb_release -a
+        lspci | grep VGA
+        clinfo | less
+
```
lukaszstolarczuk, line 215: I just realized we had a similar script in PMDK - you can gather even more info, like info about all installed packages in the OS. Also, you can make it a script in .github/scripts

 https://github.com/pmem/pmdk/blob/master/utils/gha-runners/get-system-info.sh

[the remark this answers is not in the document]
szadam: Done

.github/workflows/cmake.yml
szadam: I add the whole script

.github/workflows/cmake.yml
```diff
@@ -203,6 +203,16 @@ jobs:
...
 
+    - name: Information about platform
+      if: always()
+      run: |
+        export PATH=${workspace}/dpcpp_compiler/bin:$PATH
+        export CPATH=${workspace}/dpcpp_compiler/include:$CPATH
+        sycl-ls
+        lsb_release -a
+        lspci | grep VGA
+        clinfo | less
```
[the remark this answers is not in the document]
szadam: You are right

.github/workflows/cmake.yml
```diff
@@ -203,6 +203,16 @@ jobs:
       # This is so that device binaries can find the sycl runtime library
       run: cmake --build ${{github.workspace}}/build -j $(nproc)
 
+    - name: Information about platform
+      if: always()
+      run: |
+        export PATH=${workspace}/dpcpp_compiler/bin:$PATH
+        export CPATH=${workspace}/dpcpp_compiler/include:$CPATH
+        sycl-ls
+        lsb_release -a
```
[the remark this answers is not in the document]
szadam: You are right

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+#  Copyright (C) 2023 Intel Corporation
+#  Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+#  See LICENSE.TXT
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# get-system-info.sh - Script for printing system info
```
lukaszstolarczuk, line 8: name here is inconsistent with the file name

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,45 @@
...
+
+function system_info {
+	echo "********** system_info **********"
+	cat /etc/os-release | grep -oP "PRETTY_NAME=\K.*"
+	uname -r
+    echo "**********SYCL-LS**********"
```
lukaszstolarczuk, line 14: pls align all lines the same way (these few lines use spaces instead of a tab)

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,45 @@
...
+#  Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+#  See LICENSE.TXT
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# get-system-info.sh - Script for printing system info
+
+function system_info {
+	echo "********** system_info **********"
+	cat /etc/os-release | grep -oP "PRETTY_NAME=\K.*"
+	uname -r
```
lukaszstolarczuk, line 13: it seems this is not needed, since you're printing `/proc/version` below

.github/workflows/cmake.yml
lukaszstolarczuk: not done, and since you call the script "get system info" you can use "get" here as well 😉

.github/workflows/cmake.yml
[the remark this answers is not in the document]
lukaszstolarczuk: not done (only `uname` was added)

lukaszstolarczuk: I think packages are ok, you can trim env down, to only print useful variables, like `(C)PATH`,  `(LD_)LIBRARY_PATH`, etc.

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,45 @@
...
+	echo "**********/proc/cmdline**********"
+	cat /proc/cmdline
+	echo "**********/proc/modules**********"
+	cat /proc/modules
+	echo "**********/proc/cpuinfo**********"
+	cat /proc/cpuinfo
+	echo "**********/proc/meminfo**********"
+	cat /proc/meminfo
+	echo "**********/proc/swaps**********"
+	cat /proc/swaps
```
lukaszstolarczuk, line 33: this may be of no use, I believe

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,45 @@
...
+	echo "**********/proc/version**********"
+	cat /proc/version
+	echo "**********list-enviroment**********"
```
lukaszstolarczuk, line 36: misspell `enviroment`

.github/workflows/cmake.yml
lukaszstolarczuk: and what about `clinfo`? it provides different infos than what it's in that PMDK script

.github/workflows/cmake.yml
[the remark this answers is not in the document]
szadam: Now done

szadam: Okey

.github/workflows/cmake.yml
szadam: now done

.github/workflows/cmake.yml
szadam: okey

.github/scripts/get_system_info.sh
szadam: okey

.github/scripts/get_system_info.sh
szadam: done

.github/scripts/get_system_info.sh
szadam: okey

.github/scripts/get_system_info.sh
szadam: You're right

.github/scripts/get_system_info.sh
szadam: done

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,65 @@
...
+	if command -v rocminfo &> /dev/null; then
+		rocminfo
+	else
+		echo "ROCm not installed"
+	fi
+	echo "**********/proc/cmdline**********"
+	cat /proc/cmdline
+	echo "**********/proc/modules**********"
+	cat /proc/modules
+	echo "**********/proc/cpuinfo**********"
```
lukaszstolarczuk, line 41: instead of this we could actually use `lscpu`, it has much simpler output

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,65 @@
...
+	cat /proc/version
+	echo "**********list-environment**********"
+	echo "PATH=$PATH"
+	echo
+	echo "CPATH=$CPATH"
+	echo
+	echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
+	echo
+	echo "LIBRARY_PATH=$LIBRARY_PATH"
+	echo
```
lukaszstolarczuk, line 55: missing `PKG_CONFIG_PATH`

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,65 @@
...
+	echo "**********/proc/cmdline**********"
+	cat /proc/cmdline
+	echo "**********/proc/modules**********"
+	cat /proc/modules
+	echo "**********/proc/cpuinfo**********"
+	cat /proc/cpuinfo
+	echo "**********/proc/meminfo**********"
+	cat /proc/meminfo
+	echo "**********/proc/version**********"
+	cat /proc/version
```
lukaszstolarczuk, line 46: move this up to `system_info` section

.github/workflows/cmake.yml
```diff
@@ -203,6 +203,10 @@ jobs:
       # This is so that device binaries can find the sycl runtime library
       run: cmake --build ${{github.workspace}}/build -j $(nproc)
 
+    - name: Get information about platform
```
lukaszstolarczuk, line 206: move to the very bottom of the job

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,65 @@
...
+# get_system_info.sh - Script for printing system info
+
+function system_info {
+	echo "********** system_info **********"
+	cat /etc/os-release | grep -oP "PRETTY_NAME=\K.*"
+	echo "**********SYCL-LS**********"
+	source /opt/intel/oneapi/setvars.sh
+	sycl-ls
+	echo "*************** installed-packages ***************"
+	# Instructions below will return some minor errors, as they are dependent on the Linux distribution.
```
lukaszstolarczuk, line 17: you can move this section to the bottom; in general we should have the most significant (for us) info on the top, I think

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,65 @@
...
+	yum list installed 2>/dev/null || true
+	echo "**********VGA**********"
+	lspci | grep VGA
+	echo "**********CUDA Version**********"
+	if command -v nvidia-smi &> /dev/null; then
+		nvidia-smi
+	else
+		echo "CUDA not installed"
+	fi
+	echo "**********L0 Version**********"
```
lukaszstolarczuk, line 29: no `compute-runtime` info

.github/workflows/cmake.yml
```diff
@@ -203,6 +203,10 @@ jobs:
...
+      if: always()
+      run: .github/scripts/get_system_info.sh
```
lukaszstolarczuk, line 208: you could additionally run (from built repo) `(repo)/build/bin/urinfo || true`

lukaszstolarczuk (commented):
in general it looks good now, thanks

.github/scripts/get_system_info.sh
szadam: It is in clinfo ;)

.github/scripts/get_system_info.sh
szadam: done

.github/scripts/get_system_info.sh
szadam: done

.github/workflows/cmake.yml
szadam: done

.github/scripts/get_system_info.sh
szadam: done

.github/workflows/cmake.yml
szadam: done in script

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,71 @@
...
+	dpkg -l | grep level-zero
```
lukaszstolarczuk, line 26: now I think, this will break if we add different OS 😉

lukaszstolarczuk (approved):

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,71 @@
...
+	echo "CPATH=$CPATH"
+	echo
+	echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
+	echo
+	echo "LIBRARY_PATH=$LIBRARY_PATH"
+	echo
+	echo "PKG_CONFIG_PATH=$PKG_CONFIG_PATH"
+	echo
+	echo "******list-build-system-versions*******"
+	gcc --version 2>/dev/null || true
```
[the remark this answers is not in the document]
PatKamin: You're right, it's not worth of future maintenance.

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,71 @@
...
+	echo
+	clang --version 2>/dev/null || true
+	echo
+	make --version 2>/dev/null || true
+	echo "**********/proc/modules**********"
+	cat /proc/modules
+	echo "*************** installed-packages ***************"
```
kswiecicki, line 63: nit: You are alternating between making a space before and after in different `echo` section messages.

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,71 @@
...
+	cat /proc/cmdline
+	echo "**********CPU info**********"
+	lscpu
+	echo "**********/proc/meminfo**********"
+	cat /proc/meminfo
+	echo "**********build/bin/urinfo**********"
+	$(dirname "$(readlink -f "$0")")/../../build/bin/urinfo || true
+	echo "******OpenCL*******"
+	# The driver version of OpenCL Graphics is the compute-runtime version
+	clinfo
```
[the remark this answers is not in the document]
kswiecicki: Since there are quite a few commands that can fail, you could add a bash function that calls given command and prints some message when failed.

.github/scripts/get_system_info.sh
szadam: Done

.github/scripts/get_system_info.sh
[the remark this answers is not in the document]
szadam: I added it. I think it is unnecessary to create a new bash function and complicate this script. It is easier like that.

.github/scripts/get_system_info.sh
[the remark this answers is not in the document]
szadam: :) I agree

.github/scripts/get_system_info.sh
szadam: I've changed it.

lukaszstolarczuk (commented):
👍

PatKamin (approved):

.github/scripts/get_system_info.sh
```diff
@@ -0,0 +1,72 @@
...
+	lspci | grep VGA
+	echo "**********CUDA Version**********"
+	if command -v nvidia-smi &> /dev/null; then
+		nvidia-smi
+	else
+		echo "CUDA not installed"
+	fi
+	echo "**********L0 Version**********"
+	# Instructions below will return some minor errors, as they are dependent on the Linux distribution.
+	dpkg -l | grep level-zero || rpm -qa | grep level-zero || zypper se level-zero || echo "level-zero not installed"
```
pbalcer, line 27: instead of always calling dpkg/rpm/zypper/apt/yum, can you create a wrapper to find which one exists and use that?
Right now you get things like:
```
.github/scripts/get_system_info.sh: line 27: rpm: command not found
.github/scripts/get_system_info.sh: line 27: zypper: command not found
```

At the very least we should suppress these errors.

szadam: Done

PR #1053, opened 2023-11-08
[UR][L0] Add UR_L0_LEAKS_DEBUG key

jandres742:
Use a new environment variable, UR_L0_LEAKS_DEBUG, to check for leaks in the UR L0 adapter, instead of relying on a specific value being set in UR_L0_DEBUG.

LLVM testing: https://github.com/intel/llvm/pull/11820


jandres742:
@smaslov-intel : please take a look.

smaslov-intel (approved):

jandres742:
The llvm testing patch https://github.com/intel/llvm/pull/11820 does more than just testing. it also changes some of the tests to make them pass, so the changes from that patch are needed when upgrading the UR version in intel/llvm.

kbenzie (approved):

Issue #762, opened 2023-08-01
Split `UR_RESULT_ADAPTER_SPECIFIC_ERROR` into two separate `ur_result_t`, for warnings and for errors


hdelan:
`UR_RESULT_ADAPTER_SPECIFIC_ERROR`  can be used to emit adapter specific warnings as well as errors. 

In CUDA adapter warnings are reported if `setErrorMessage` is called with `UR_SUCCESS` followed by a return of `UR_RESULT_ADAPTER_SPECIFIC_ERROR`. Errors are reported  if `setErrorMessage` is called with `UR_RESULT_ADAPTER_SPECIFIC_ERROR` followed by a return of `UR_RESULT_ADAPTER_SPECIFIC_ERROR`. This is unintuitive. It would be nice if we could return either 

- `UR_RESULT_ADAPTER_SPECIFIC_ERROR` for errors
- `UR_RESULT_ADAPTER_SPECIFIC_WARNING` for warnings

And `setErrorMessage` would not need to take an extra `ur_result_t` param.

kbenzie:
Closing in preference of #1330

PR #1055, opened 2023-11-09
[SPEC] Clarify the behaviour of UR warning messages

fabiomestre:
* This commit updates the specification to clarify the behaviour
of UR_RESULT_ERROR_ADAPTER_SPECIFIC when returning warning messages.
* This is a temporary change until specific error codes for warnings
are added.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -856,11 +856,11 @@ urAdapterRetain(
 /// returned by the failed driver entry-point.
 ///
 /// * If `pError` is ::UR_RESULT_SUCCESS, ::UR_RESULT_ERROR_ADAPTER_SPECIFIC
-/// represents a warning instead. This means that the entry-point call did not
-/// fail. However it might not have behaved as expected.
+///   represents a warning instead. This means that the entry-point call did not
+///   fail. However it might not have behaved as expected.
 ///
 /// * Using ::UR_RESULT_ERROR_ADAPTER_SPECIFIC to emit warnings is an optional
-/// feature. Its usage is left at the discretion of adapter maintainers.
+///   feature. Its usage is left at the discretion of adapter maintainers.
 ///
 /// * Implementations *must* store the message and error code in thread-local
 ///   storage prior to returning ::UR_RESULT_ERROR_ADAPTER_SPECIFIC.
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -847,13 +847,17 @@ urAdapterRetain(
 );
 
 ///////////////////////////////////////////////////////////////////////////////
-/// @brief Get the last adapter specific error.
+/// @brief Get the last adapter specific warning or error.
 ///
 /// @details
 /// To be used after another entry-point has returned
 /// ::UR_RESULT_ERROR_ADAPTER_SPECIFIC in order to retrieve a message describing
 /// the circumstances of the underlying driver error and the error code
 /// returned by the failed driver entry-point.
+///
+/// * If `pError` is ::UR_RESULT_SUCCESS, ::UR_RESULT_ERROR_ADAPTER_SPECIFIC
+/// represents a warning instead. This means that the entry-point call did not
+/// fail. However it might not have behaved as expected.
 ///
 /// * Implementations *must* store the message and error code in thread-local
 ///   storage prior to returning ::UR_RESULT_ERROR_ADAPTER_SPECIFIC.
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -859,6 +859,9 @@ urAdapterRetain(
 /// represents a warning instead. This means that the entry-point call did not
 /// fail. However it might not have behaved as expected.
 ///
+/// * Using ::UR_RESULT_ERROR_ADAPTER_SPECIFIC to emit warnings is an optional
+/// feature. Its usage is left at the discretion of adapter maintainers.
+///
 /// * Implementations *must* store the message and error code in thread-local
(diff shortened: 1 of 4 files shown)


include/ur_api.h
```diff
@@ -847,14 +847,18 @@ urAdapterRetain(
...
-/// @brief Get the last adapter specific error.
+/// @brief Get the last adapter specific warning or error.
 ///
 /// @details
 /// To be used after another entry-point has returned
 /// ::UR_RESULT_ERROR_ADAPTER_SPECIFIC in order to retrieve a message describing
 /// the circumstances of the underlying driver error and the error code
 /// returned by the failed driver entry-point.
 ///
+/// * If `pError` is ::UR_RESULT_SUCCESS, ::UR_RESULT_ERROR_ADAPTER_SPECIFIC
```
jandres742, line 858: Although I understand it, I wonder if any adapter would add support to return a warning with SUCESS through `urAdapterGetLastError`. Returning a message through `urAdapterGetLastError` has an overhead (strncpy a mesage, keep track of the last message, etc).

this of course could be optional, but I'm concerned this would open the door for UR customers to start requesting warnings, given that the spec allows it, and then having pushback from underlying drivers or adapters that dont want to implement such warning because of performance considerations.

jandres742:
@fabiomestre

> This is a temporary change until specific error codes for warnings
are added.

could you elaborate on the temporary nature of this change?

fabiomestre:
> @fabiomestre
> 
> > This is a temporary change until specific error codes for warnings
> > are added.
> 
> could you elaborate on the temporary nature of this change?



> @fabiomestre
> 
> > This is a temporary change until specific error codes for warnings
> > are added.
> 
> could you elaborate on the temporary nature of this change?

The solution would temporary because there are plans to split UR_RESULT_ERROR_ADAPTER_SPECIFIC into 2 different errors: https://github.com/oneapi-src/unified-runtime/issues/762

This would still require the adapters to set a message if they want to return a warning though.

include/ur_api.h
fabiomestre: For context, this SPEC change is a follow up to discussion that happened in https://github.com/oneapi-src/unified-runtime/pull/989

At the moment, the UR spec says that UR_RESULT_ERROR_ADAPTER_SPECIFIC can be used to return a warning or error but doesn't specify how. The aim to this PR was to clarify what is the correct way to do it. 

There are already adapters using UR_RESULT_ERROR_ADAPTER_SPECIFIC in this way. For example CUDA is doing it for a few entrypoints. In addition, this is the behaviour that SYCL-RT expects.

I understand that there are performance / design implications to returning a warning. The intent was for it to be an optional feature. 

I guess that we have to decide whether we want to support warnings on UR or not. If the answer is yes, then that should be documented in the spec (which this PR aims to do). If the answer is no, then we need to update the spec, SYCL-RT and all the adapters that are using UR_RESULT_ERROR_ADAPTER_SPECIFIC in this way.

kbenzie: As an interface layer between higher level languages and lower level drivers I don't think UR should stand in the way of either users requesting more information about why a  feature they want to use isn't working or performance is suboptimal.

As @fabiomestre says, the spec already supports this but does so in an unclear way. I you are concearned about the performance implecations in an adatper @jandres742 then perhaps we should think about ways of opting in to their use rather than being on by default, perahps with an environment variable.

jandres742: thanks @kbenzie , @fabiomestre . I wouldn't even want this to be with env var, as that implies having branches on several places, even on critical paths.

Dont know if we need to clarify on this PR or in the spec that the warnings are optional, but if that's the intention, then I'm ok with this.

jandres742 (approved):

include/ur_api.h
kbenzie: I think its reasonable to have spec language stating that emitting warnings is optional at the discretion of the adapter maintainers. @fabiomestre if you could update the spec language to clarify that?

fabiomestre: Added a bullet point to clarify that the warnings are optional in the latest commit.

scripts/core/adapter.yml
```diff
@@ -73,12 +73,19 @@ params:
...
     $X_RESULT_ERROR_ADAPTER_SPECIFIC in order to retrieve a message describing
     the circumstances of the underlying driver error and the error code
     returned by the failed driver entry-point.
+  
+    * If `pError` is $X_RESULT_SUCCESS, $X_RESULT_ERROR_ADAPTER_SPECIFIC
+    represents a warning instead. This means that the entry-point call did not
+    fail. However it might not have behaved as expected.
+  
+    * Using $X_RESULT_ERROR_ADAPTER_SPECIFIC to emit warnings is an optional
+    feature. Its usage is left at the discretion of adapter maintainers.
```
kbenzie, line 88: Indenting the bullet points makes them stand out more to the eye, also matches the surrounding style.

```suggestion
    * If `pError` is $X_RESULT_SUCCESS, $X_RESULT_ERROR_ADAPTER_SPECIFIC
      represents a warning instead. This means that the entry-point call did not
      fail. However it might not have behaved as expected.
  
    * Using $X_RESULT_ERROR_ADAPTER_SPECIFIC to emit warnings is an optional
      feature. Its usage is left at the discretion of adapter maintainers.
```

fabiomestre: Done :+1:

kbenzie (approved):

Issue #926, opened 2023-10-03
Consider whether we should require adapters to bounds check USM operations


aarongreig:
We currently mandate [bounds checks](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/enqueue.yml#L1077) on our EnqueueUSM* type operations. None of the APIs we currently have adapters for have bounds checking on their equivalent operations: [OpenCL](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_unified_shared_memory.html#_filling_and_copying_unified_shared_memory), [Level Zero](https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=kernelsetargumentvalue#zecommandlistappendmemorycopy), [Cuda](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g5f26aaf5582ade791e5688727a178d78) (doesn't explicitly define conditions for each return code but lack of a distinct INVALID_SIZE type code would mean we have to do our own check to return the appropriate error even if cuda does do its own bounds checking). Since this is something we can't relegate to the validation layer implementing these entry points to spec means adding small but measurable, and unavoidable, additional overhead to these operations. The fact that most of the adapters, if not all of them, don't have these checks implemented currently suggests they aren't being missed, so we should consider removing them from the spec.

fabiomestre:
Another similar issue would checking if events inside a wait list are null. This would also require either adding the check in the adapter or modifying the validation layers to iterate through the list of events. 

It currently fails in these tests:

* urEnqueueDeviceGetGlobalVariableReadTest:InvalidEventWaitInvalidEvent
* urEnqueueDeviceGetGlobalVariableWriteTest:InvalidEventWaitInvalidEvent

PR #1060, opened 2023-11-09
Remove the spec requirement for adapters to bounds check USM operations.

aarongreig:
None of the adapter APIs support this kind of bounds checking natively so mandating it in our spec means mandating an apparently unnecessary additional overhead in the adapters.

Addresses #926

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -6619,7 +6619,6 @@ urEnqueueMemUnmap(
 ///         + `patternSize > size`
 ///         + `(patternSize & (patternSize - 1)) != 0`
 ///         + `size % patternSize != 0`
-///         + If `size` is higher than the allocation size of `ptr`
 ///     - ::UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST
 ///         + `phEventWaitList == NULL && numEventsInWaitList > 0`
 ///         + `phEventWaitList != NULL && numEventsInWaitList == 0`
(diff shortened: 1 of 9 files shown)


fabiomestre:
Why are we removing this checks for USM but not for Buffers? Is it that much harder to do the check on USM entrypoints?

I would expect them to be similar in that we need to track the initial allocation of memory and the size of the buffer / usm memory in validation layers.

aarongreig:
the native APIs do generally have bounds checking for buffer operations, the adapters don't need to do anything for that

aarongreig:
...maybe that's only true for OpenCL actually

fabiomestre:
> ...maybe that's only true for OpenCL actually

Yeah. Both Hip and Cuda rely on the same functions for buffer and usm operations. Not entirely sure about level zero.

My feeling about this is that we should either support it for both USM and Buffers using validation layers which avoids adding extra load on the adapters.  Or not support it at all.

include/ur_api.h
```diff
@@ -6619,7 +6619,6 @@ urEnqueueMemUnmap(
 ///         + `patternSize > size`
 ///         + `(patternSize & (patternSize - 1)) != 0`
 ///         + `size % patternSize != 0`
-///         + If `size` is higher than the allocation size of `ptr`
```
jandres742, line 6622: this is not supported by the adapters, but it could be added to the validation layer as common functionality for all. It would be a matter of querying for the allocations' size and compare against the size passed.

aarongreig:
> > ...maybe that's only true for OpenCL actually
> 
> Yeah. Both Hip and Cuda rely on the same functions for buffer and usm operations (and I don't think either of them has this kind of bound check natively). Not entirely sure about level zero.
> 
> My feeling about this is that we should either support it for both USM and Buffers using validation layers which avoids adding extra load on the adapters. Or not support it at all.

yep agreed, in light of every adapter that isn't OpenCL we should just move the checks into the validation layer instead of removing the requirement from the spec

Issue #355, opened 2023-03-14
Provide a way for applications to use a single adapter


pbalcer:
The current UR loader, to support multiple adapters, has an [indirection layer](https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_ldrddi.cpp) that creates and maintains wrappers around UR entities (or function class types, i.e., `platform`, `device`  and so on) that store a pointer to adapter functions. If there's only one adapter, this layer is unused, and the loader calls the adapter functions directly.

This indirection adds an extraneous overhead for applications that use only one adapter but have more available in the system. This issue is to devise a way to allow applications to load and use only the desired adapter implementation, thus avoiding the overhead.

Possible solutions:

* The `UR_ADAPTERS_FORCE_LOAD` environmental variable can be set with a desired adapter, forcing the loader to use it. This is already possible.
* Reintroduce `platform_flags` to `urInit`, with a way of selecting a single adapter.
    * Something like `UR_PLATFORM_USE_FIRST`. But this might be hard to use since the order of adapters is unspecified.
    * Alternatively, we could create a flag per known platform, something like: `UR_PLATFORM_L0`, `UR_PLATFORM_CUDA`, `UR_PLATFORM_HIP`, and then this could be used like this: `urInit(0, UR_PLATFORM_L0 | UR_PLATFORM_CUDA);`. This would only work with predefined platforms.
* Add a way to programmatically filter platforms at `urInit`, for example: `urInit(0, [](struct platform_descriptor d) -> bool { return strcmp(d->name, "ur_adapter_level_zero") == 0; })`. This might be clunky to use from C, but I think is the most universal.
* Add a way of unloading platforms post-initialization. Software would iterate over available platforms and would either call `urPlatformUnload` on the ones it doesn't intend to use or `urPlatformUseOnlyThis` (can't think of a name right now :-)) on the one it does. This fits into the existing API, but might be error-prone and tricky to implement safely.

pbalcer:
ping @jandres742 @smaslov-intel @bmyates @igchor

jandres742:
thanks @pbalcer .

maybe we don't need anything else than UR_ADAPTERS_FORCE_LOAD ? Good thing about it UR_ADAPTERS_FORCE_LOAD  is that users can select one time L0, and the next time CUDA, w/o needing to force a specific change in the code. Having a flag urInit might translate then into another env var in SYCL probably, to know which flag to pass there.

[...]

PR #1067, opened 2023-11-11
[UR][Loader] Fix handling of native handles

alexbatashev:
Native handles are created by adapters and thus are inheritently backend-specific. Loader can not assume anything about these handles, as even nullptr may be a valid value for such a handle.

This patch changes two things about native handles:
1) Native handles are no longer wrapped in UR objects
2) Dispatch table is extracted from any other argument of the API function

The above is true for all interop APIs except for urPlatformCreateWithNativeHandle, which needs a spec change: https://github.com/oneapi-src/unified-runtime/issues/1068


source/loader/ur_ldrddi.cpp
```diff
@@ -2601,14 +2540,6 @@ __urdlllocal ur_result_t UR_APICALL urProgramGetNativeHandle(
         return result;
     }
 
-    try {
-        // convert platform handle to loader handle
-        *phNativeProgram = reinterpret_cast<ur_native_handle_t>(
```
jandres742, line 2606: @bmyates : why is that we need this kind of logic in the L0 loader, of translating driver handles to L0 handles? Do you see any risks on removing this code in the UR loader?

pbalcer: The loader handles are there to find the appropriate adapter-level function to call. The alternative is to either make all functions accept an adapter handle as one of the arguments or create a way to map normal handles to adapter handles.

This mechanism is not used when there's only one adapter, see https://github.com/oneapi-src/unified-runtime/issues/355.

alexbatashev: > The alternative is to either make all functions accept an adapter handle as one of the arguments or create a way to map normal handles to adapter handles.

There's no need to. Most interop calls accept a context, which has the right adapter, and this patch exactly changes the behavior to find the adapter in the context, rather the native handle. The only notable exception is platform: https://github.com/oneapi-src/unified-runtime/issues/1068. That one certainly needs an adapter as an argument.

pbalcer (approved):
Please also modify [`get_adapter_handles`](https://github.com/oneapi-src/unified-runtime/blob/adapters/scripts/templates/helper.py#L579) so that it doesn't return native handles. This will remove the now unused object factories.

source/loader/ur_ldrddi.cpp
pbalcer: > There's no need to.

I know, I agree. I was replying to Jamie's question.

alexbatashev:
> Please also modify [`get_adapter_handles`](https://github.com/oneapi-src/unified-runtime/blob/adapters/scripts/templates/helper.py#L579) so that it doesn't return native handles. This will remove the now unused object factories.

I did try to change it to:

```python
def get_adapter_handles(specs):
    objs = []
    for s in specs:
        for obj in s['objects']:
            if obj_traits.is_handle(obj) and not (obj_traits.is_loader_only(obj) or 'native' in obj['name']):
                objs.append(obj)

    return objs
```

But the compilation would fail with the following error:

```
def get_adapter_handles(specs):
    objs = []
    for s in specs:
        for obj in s['objects']:
            if obj_traits.is_handle(obj) and not (obj_traits.is_loader_only(obj) or 'native' in obj['name']):
                objs.append(obj)

    return objs
```

Am I doing something wrong?

source/loader/ur_ldrddi.cpp
jandres742: thanks @pbalcer , @alexbatashev . Ah, this is only for native handles. That's ok then.

+1

pbalcer:
> Am I doing something wrong?

No, I think it's all correct. The failure is because platform native handles are special-cased for now. This particular change will have to wait until the spec is fixed.

alexbatashev:
> > Am I doing something wrong?
> 
> No, I think it's all correct. The failure is because platform native handles are special-cased for now. This particular change will have to wait until the spec is fixed.

Well, I guess, that is the reason why we can't remove `ur_native_handle_t` factory for now. I don't see a way to leave it only for urProgramCreateWithNative.

PR #1070, opened 2023-11-13
[CUDA] Dynamically load the CUPTI library when tracing

pasaulais:
With these changes, `libcupti.so` is loaded dynamically when CUDA tracing is enabled. This enables XPTI tracing-enabled builds to work on systems that do not have `libcupti.so` or where that library cannot be located on the system.

The `enableCUDATracing` and `disableCUDATracing` functions have been changed to take a context pointer, rather than use global variables for tracing state.

There is a temporary `enableCUDATracing` variant with no parameter for compatibility until the relevant changes have been merged to https://github.com/intel/llvm.

### DIFF
--- a/source/adapters/cuda/CMakeLists.txt
+++ b/source/adapters/cuda/CMakeLists.txt
@@ -76,6 +76,10 @@ else()
   message(WARNING "CUDA adapter USM pools are disabled, set UMF_ENABLE_POOL_TRACKING to enable them")
 endif()
 
+if (CUDA_cupti_LIBRARY)
+  target_compile_definitions("ur_adapter_cuda" PRIVATE CUPTI_LIB_PATH="${CUDA_cupti_LIBRARY}")
+endif()
+
 target_link_libraries(${TARGET_NAME} PRIVATE
     ${PROJECT_NAME}::headers
     ${PROJECT_NAME}::common
--- a/source/adapters/cuda/adapter.cpp
+++ b/source/adapters/cuda/adapter.cpp
@@ -11,13 +11,12 @@
 #include <ur_api.h>
 
 #include "common.hpp"
-
-void enableCUDATracing();
-void disableCUDATracing();
+#include "tracing.hpp"
 
 struct ur_adapter_handle_t_ {
   std::atomic<uint32_t> RefCount = 0;
   std::mutex Mutex;
+  struct cuda_tracing_context_t_ *TracingCtx = nullptr;
 };
 
 ur_adapter_handle_t_ adapter{};
@@ -28,7 +27,8 @@ urAdapterGet(uint32_t NumEntries, ur_adapter_handle_t *phAdapters,
   if (NumEntries > 0 && phAdapters) {
     std::lock_guard<std::mutex> Lock{adapter.Mutex};
     if (adapter.RefCount++ == 0) {
-      enableCUDATracing();
+      adapter.TracingCtx = createCUDATracingContext();
+      enableCUDATracing(adapter.TracingCtx);
     }
 
     *phAdapters = &adapter;
@@ -50,7 +50,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urAdapterRetain(ur_adapter_handle_t) {
 UR_APIEXPORT ur_result_t UR_APICALL urAdapterRelease(ur_adapter_handle_t) {
   std::lock_guard<std::mutex> Lock{adapter.Mutex};
   if (--adapter.RefCount == 0) {
-    disableCUDATracing();
+    disableCUDATracing(adapter.TracingCtx);
+    freeCUDATracingContext(adapter.TracingCtx);
+    adapter.TracingCtx = nullptr;
   }
   return UR_RESULT_SUCCESS;
 }
--- a/source/adapters/cuda/tracing.cpp
+++ b/source/adapters/cuda/tracing.cpp
@@ -43,8 +43,8 @@ CUptiResult (*)(uint32_t enable, CUpti_SubscriberHandle subscriber,
                 CUpti_CallbackDomain domain, CUpti_CallbackId cbid);
 
 #define LOAD_CUPTI_SYM(p, lib, x)                                              \
-  p->x = (cupti##x##_fn)ur_loader::LibLoader::getFunctionPtr(lib.get(),        \
-                                                             "cupti" #x);
+  p.x = (cupti##x##_fn)ur_loader::LibLoader::getFunctionPtr(lib.get(),         \
+                                                            "cupti" #x);
 
 #else
 using tracing_event_t = void *;
@@ -55,15 +55,21 @@ using cuptiEnableDomain_fn = void *;
 using cuptiEnableCallback_fn = void *;
 #endif // XPTI_ENABLE_INSTRUMENTATION
 
+struct cupti_table_t_ {
+  cuptiSubscribe_fn Subscribe = nullptr;
+  cuptiUnsubscribe_fn Unsubscribe = nullptr;
+  cuptiEnableDomain_fn EnableDomain = nullptr;
+  cuptiEnableCallback_fn EnableCallback = nullptr;
+
+  bool isInitialized() const;
+};
+
 struct cuda_tracing_context_t_ {
   tracing_event_t CallEvent = nullptr;
   tracing_event_t DebugEvent = nullptr;
   subscriber_handle_t Subscriber = nullptr;
   ur_loader::LibLoader::Lib Library;
-  cuptiSubscribe_fn Subscribe = nullptr;
-  cuptiUnsubscribe_fn Unsubscribe = nullptr;
-  cuptiEnableDomain_fn EnableDomain = nullptr;
-  cuptiEnableCallback_fn EnableCallback = nullptr;
+  cupti_table_t_ Cupti;
 };
 
 #ifdef XPTI_ENABLE_INSTRUMENTATION
@@ -132,6 +138,10 @@ void freeCUDATracingContext(cuda_tracing_context_t_ *Ctx) {
 #endif // XPTI_ENABLE_INSTRUMENTATION
 }
 
+bool cupti_table_t_::isInitialized() const {
+  return Subscribe && Unsubscribe && EnableDomain && EnableCallback;
+}
+
 bool loadCUDATracingLibrary(cuda_tracing_context_t_ *Ctx) {
 #if defined(XPTI_ENABLE_INSTRUMENTATION) && defined(CUPTI_LIB_PATH)
   if (!Ctx)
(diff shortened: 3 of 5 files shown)


pbalcer (commented):
Does this work/do anything when built out of sycl tree? As far as I can tell, we don't set `XPTI_ENABLE_INSTRUMENTATION` anywhere, and the cuda adapter never links with xpti.

Might be useful to setup a CI job that builds cuda with tracing enabled.

pasaulais:
I haven't built UR out of the SYCL tree, but my understanding is that the CUDA tracing support is #ifdef'd out as you mention. I don't know if xpti should be a required dependency of UR, but this work is about making `XPTI_ENABLE_INSTRUMENTATION` enabled by default for SYCL builds.

That's a good point, I will ask internally about adding CUDA tracing to a CI job.

pasaulais:
> Does this work/do anything when built out of sycl tree? As far as I can tell, we don't set `XPTI_ENABLE_INSTRUMENTATION` anywhere, and the cuda adapter never links with xpti.
> 
> Might be useful to setup a CI job that builds cuda with tracing enabled.

I've asked internally and the consensus is that tracing should be enabled in CI. This is better done in a separate PR due to the dependencies between this repo and https://github.com/intel/llvm (e.g. changing the signature for `enableCUDATracing`).

pasaulais:
I have also added some changes to use `ur_loader::LibLoader` for loading the CUPTI library instead of `dlopen`, which is not available on Windows. This won't affect the CI testing yet as tracing is not enabled, but these changes will be required when it gets enabled by default.

kbenzie:
> I've asked internally and the consensus is that tracing should be enabled in CI. This is better done in a separate PR due to the dependencies between this repo and https://github.com/intel/llvm (e.g. changing the signature for `enableCUDATracing`).

Please create an issue for tracking this https://github.com/oneapi-src/unified-runtime/issues/new

pasaulais:
> > I've asked internally and the consensus is that tracing should be enabled in CI. This is better done in a separate PR due to the dependencies between this repo and https://github.com/intel/llvm (e.g. changing the signature for `enableCUDATracing`).
> 
> Please create an issue for tracking this https://github.com/oneapi-src/unified-runtime/issues/new

I have created https://github.com/oneapi-src/unified-runtime/issues/1098 for this.

fabiomestre:

pasaulais:

Thanks @fabiomestre, I have rebased these changes on top of latest `main`.

source/adapters/cuda/CMakeLists.txt
```diff
@@ -73,6 +73,10 @@ else()
   message(WARNING "CUDA adapter USM pools are disabled, set UMF_ENABLE_POOL_TRACKING to enable them")
 endif()
 
+if (CUDA_cupti_LIBRARY)
+  target_compile_definitions("ur_adapter_cuda" PRIVATE "-DCUPTI_LIB_PATH=\"${CUDA_cupti_LIBRARY}\"")
```
kbenzie, line 77: I don't think need the `-D` or the outer quotes when using `target_compile_definitions`

```suggestion
  target_compile_definitions(ur_adapter_cuda PRIVATE CUPTI_LIB_PATH="${CUDA_cupti_LIBRARY}")
```

alexbatashev:
I wonder, what's the content of `${CUDA_cupti_LIBRARY}`? If it's a full path, then it does not solve the problem, and instead makes it silent and hard to debug for users, who really need tracing and profiling. And if it's just the library name, I'm afraid that's a potential security issue: typically, security guidelines require dynamic library loading with full paths.

source/adapters/cuda/CMakeLists.txt
pasaulais: Good catch, that looks a bit simpler/cleaner now.

pasaulais:
> I wonder, what's the content of `${CUDA_cupti_LIBRARY}`? If it's a full path, then it does not solve the problem, and instead makes it silent and hard to debug for users, who really need tracing and profiling. And if it's just the library name, I'm afraid that's a potential security issue: typically, security guidelines require dynamic library loading with full paths.

On my system, this is the following path: `/usr/local/cuda/lib64/libcupti.so`

$ ls -l /usr/local/cuda/lib64/libcupti.so
lrwxrwxrwx 1 root root 14 Sep  9 04:28 /usr/local/cuda/lib64/libcupti.so -> libcupti.so.12
$ ls -l /usr/local/cuda
lrwxrwxrwx 1 root root 22 Oct 30 15:35 /usr/local/cuda -> /etc/alternatives/cuda
$ ls -l /etc/alternatives/cuda
lrwxrwxrwx 1 root root 20 Oct 30 15:35 /etc/alternatives/cuda -> /usr/local/cuda-12.3

kbenzie:
> The goal of this PR is to solve that particular situation, improving the loading process (e.g. with a user-configurable path to override the build-time path) and feedback (e.g. run-time warnings) is better done in a separate PR.

If you could create a [new issue](https://github.com/oneapi-src/unified-runtime/issues/new/choose) to track this future work @pasaulais that would be much appriciated.

pasaulais:
> > The goal of this PR is to solve that particular situation, improving the loading process (e.g. with a user-configurable path to override the build-time path) and feedback (e.g. run-time warnings) is better done in a separate PR.
> 
> If you could create a [new issue](https://github.com/oneapi-src/unified-runtime/issues/new/choose) to track this future work @pasaulais that would be much appriciated.

Sure, here are issues for both parts: https://github.com/oneapi-src/unified-runtime/issues/1188, https://github.com/oneapi-src/unified-runtime/issues/1189

kbenzie:
Thanks!

pasaulais:
> I wonder, what's the content of `${CUDA_cupti_LIBRARY}`? If it's a full path, then it does not solve the problem, and instead makes it silent and hard to debug for users, who really need tracing and profiling. And if it's just the library name, I'm afraid that's a potential security issue: typically, security guidelines require dynamic library loading with full paths.

Regarding your point about 'just the library name' (relative path), I am curious as to why that would be less secure than explicitly linking against `libcupti.so` when building the CUDA adapter. AFAIU the search path used to locate the library is the same in both cases (at least on Linux). Do you have a specific example where using `dlopen` is less secure?

The reason I am asking is loading `libcupti.so` using a relative path after loading with an absolute path fails could be a way to implement https://github.com/oneapi-src/unified-runtime/issues/1189

alexbatashev:
> > I wonder, what's the content of `${CUDA_cupti_LIBRARY}`? If it's a full path, then it does not solve the problem, and instead makes it silent and hard to debug for users, who really need tracing and profiling. And if it's just the library name, I'm afraid that's a potential security issue: typically, security guidelines require dynamic library loading with full paths.
> 
> Regarding your point about 'just the library name' (relative path), I am curious as to why that would be less secure than explicitly linking against `libcupti.so` when building the CUDA adapter. AFAIU the search path used to locate the library is the same in both cases (at least on Linux). Do you have a specific example where using `dlopen` is less secure?
> 
> The reason I am asking is loading `libcupti.so` using a relative path after loading with an absolute path fails could be a way to implement #1189

I've checked with [ld.so](https://linux.die.net/man/8/ld.so) man page, and it seems the only way to load a library from user-writable path is if that path is part of `LD_LIBRARY_PATH`, which is probably fine (?). But for Windows the first thing in the library search order is either CWD or binary directory, which can be user-writable and can potentially contain a malicious DLL. I remember that SYCL runtime stopped loading libraries with just the name for the same reason. Anyway, I'm not the expert on the topic and you should consult with security champions.

pbalcer:
> I've checked with [ld.so](https://linux.die.net/man/8/ld.so) man page, and it seems the only way to load a library from user-writable path is if that path is part of LD_LIBRARY_PATH, which is probably fine (?). But for Windows the first thing in the library search order is either CWD or binary directory, which can be user-writable and can potentially contain a malicious DLL. I remember that SYCL runtime stopped loading libraries with just the name for the same reason. Anyway, I'm not the expert on the topic and you should consult with security champions.
> 
AFAIK, on Linux it's ok to let the loader find the appropriate library file, but Windows, as you say, we need to provide a fully qualified name for it to be safe. That's what we do in the UR [loader](https://oneapi-src.github.io/unified-runtime/core/INTRO.html#adapter-discovery).

pasaulais:
> > I wonder, what's the content of ${CUDA_cupti_LIBRARY}? If it's a full path, then it does not solve the problem, and instead makes it silent and hard to debug for users, who really need tracing and profiling. And if it's just the library name, I'm afraid that's a potential security issue: typically, security guidelines require dynamic library loading with full paths.
> 
> > Regarding your point about 'just the library name' (relative path), I am curious as to why that would be less secure than explicitly linking against libcupti.so when building the CUDA adapter. AFAIU the search path used to locate the library is the same in both cases (at least on Linux). Do you have a specific example where using dlopen is less secure?
> 
> > The reason I am asking is loading libcupti.so using a relative path after loading with an absolute path fails could be a way to implement #1189
> 
> I've checked with [ld.so](https://linux.die.net/man/8/ld.so) man page, and it seems the only way to load a library from user-writable path is if that path is part of LD_LIBRARY_PATH, which is probably fine (?). But for Windows the first thing in the library search order is either CWD or binary directory, which can be user-writable and can potentially contain a malicious DLL. I remember that SYCL runtime stopped loading libraries with just the name for the same reason. Anyway, I'm not the expert on the topic and you should consult with security champions.
> 
> Two alternative ways would be to either link cupti statically or to make a separate tracing library and to load it from the same location as the adapter library, which would eliminate a case of loading libraries with just the name.

That makes sense, thanks for the clarification about why it's a security issue on Windows. One thing I'm not sure about linking cupti statically is version mismatches. Do you know if there is support for linking against one version of the CUDA toolkit and running on a system with a different toolkit version? With shared libraries, the cupti library will have the same version as the CUDA runtime library (even if a different version was used to build the CUDA adapter).

alexbatashev:
@pasaulais my understanding is that it does not matter, whether you link statically or dynamically. NVIDIA guarantees backwards compatibility inside one major version. That is, if you link against CUDA 12.3, your build will fail on CUDA 12.1 machines no matter what, as the driver version is too low. But other direction should work just fine: 12.1 library should work in 12.3 environment. In fact, `nvcc` links statically by default: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html?highlight=static#cudart-none-shared-static-cudart. And that was done to specifically resolve the issue you describe: make sure the application can work even if CUDA runtime is not installed on the system.

pasaulais:
> @pasaulais my understanding is that it does not matter, whether you link statically or dynamically. NVIDIA guarantees backwards compatibility inside one major version. That is, if you link against CUDA 12.3, your build will fail on CUDA 12.1 machines no matter what, as the driver version is too low. But other direction should work just fine: 12.1 library should work in 12.3 environment. In fact, `nvcc` links statically by default: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html?highlight=static#cudart-none-shared-static-cudart. And that was done to specifically resolve the issue you describe: make sure the application can work even if CUDA runtime is not installed on the system.

This is something that could use revisiting in order to make a good decision for how to link these libraries, so I have created an issue for this: https://github.com/oneapi-src/unified-runtime/issues/1251

PR #1071, opened 2023-11-13
[UR] Enable UMF tracking by default

igchor:
It is needed by adapters's urUSMFree implementations where we rely on umfPoolByPtr (which always returns null if tracking is disabled).

@kbenzie should I also make a PR to the main branch? We plan to move the UMF to a separate repo anyway and I expect this option (UMF_ENABLE_POOL_TRACKING) to not be publicly exposed to UR users anymore.


jandres742 (approved):

pbalcer (approved):

igchor:
I think this is ready to be merged. The failing tests turned out to just be tests that unexpectedly succeeded (most of the CUDA and L0 tests with UsePoolEnabled work after this change).

lukaszstolarczuk:
At this moment, if someone sets `UMF_ENABLE_POOL_TRACKING` manually to OFF, these tests will fail. Do we expect anyone to do that? Perhaps we should remove this option from CMake, after all?

igchor:
@lukaszstolarczuk you are right, but I think I would actually prefer to do this after we merge / cherry-pick this change. Removing the option would also mean that we can remove some of the existing code so the change would be bigger.

jandres742:
Fixes to test-usm in L0 in https://github.com/oneapi-src/unified-runtime/pull/1105 depending on this patch.

jandres742:
@igchor :
is there anything extra to do on this PR, or could we ask @kbenzie  to move it to "ready to merge" ?

kbenzie:
> @kbenzie should I also make a PR to the main branch? We plan to move the UMF to a separate repo anyway and I expect this option (UMF_ENABLE_POOL_TRACKING) to not be publicly exposed to UR users anymore.

I think we can just target adapters for now. I'm hoping we can merged adapters into main branch soon and then stop using the adapters branch so this problem will go away.

kbenzie:
> Done. Do we want to wait until changes from #1084 are cherry-picked? If we merge this first, there might be some extra Windows failures.

If that's a dependency, then yes.

kbenzie:
> > Done. Do we want to wait until changes from #1084 are cherry-picked? If we merge this first, there might be some extra Windows failures.
> 
> If that's a dependency, then yes.

This will be included in #1072 which we'll aiming to merged to adapters next.

kbenzie:
@igchor any idea why the checks failed?

kbenzie:
Thanks @igchor

fabiomestre:

PR #939, opened 2023-10-09
[UR][CUDA][L0][HIP] Add virtual memory adapter implementations

steffenlarsen:
This commit adds the adapter implementations of the virtual memory extension functionality to be used in
https://github.com/intel/llvm/pull/8954.

PR #1074, opened 2023-11-14
[UR][Layer] Add Sanitizer Layer

AllanZyne:
This PR added a new loader layer, named "sanitizer", which implemented the runtime part of device-side sanitizers: AddressSanitizer (`UR_LAYER_ASAN`), MemorySanitizer (`UR_LAYER_MSAN`, planned), and ThreadSanitizer (`UR_LAYER_TSAN`, planned).

Currently, it implemented some basic functions for PVC on the level-zero adapter, which can detect and report out-of-bounds errors on PVC.


pbalcer (commented):
Please add documentation about the new layer here: https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/INTRO.rst#layers.

Also, this patch is adding UR_LOG_SANITIZER. Please add that to the documentation as well.

CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ option(UR_USE_UBSAN "enable UndefinedBehaviorSanitizer" OFF)
 option(UR_USE_MSAN "enable MemorySanitizer" OFF)
 option(UR_USE_TSAN "enable ThreadSanitizer" OFF)
 option(UR_ENABLE_TRACING "enable api tracing through xpti" OFF)
+option(UR_ENABLE_SANITIZER "enable device sanitizer" OFF)
```
pbalcer, line 36: Since this is disabled by default, it will not build in the CI. This patch also doesn't contain any tests.

For CI, I think we should just always have `UR_ENABLE_SANITIZER` set to `ON`. See how tracing does that here:
https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/cmake.yml#L72

For tests, it would be useful to have something that verifies this functionality. From reading the code, it seems this currently works only on PVC? We currently don't have PVC CI runners, but we are hoping to add some soon.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,582 @@
...
+
+        auto AH = &SPIR_DeviceSanitizerReportMem;
+        if (!AH->Flag) {
+            return;
+        }
+
+        const char *File = AH->File[0] ? AH->File : "<unknown file>";
+        const char *Func = AH->Func[0] ? AH->Func : "<unknown func>";
+
+        fprintf(stderr, "\n====ERROR: DeviceSanitizer: %s on %s\n\n",
```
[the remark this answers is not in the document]
AllanZyne: This message is sort of similar to "die()", which needs to be seen by users immediately. 
After printing this message, the program will be aborted.

CMakeLists.txt
AllanZyne: Ok, I will add some tests later.

> From reading the code, it seems this currently works only on PVC? We currently don't have PVC CI runners, but we are hoping to add some soon.

Yes, it only works on PVC now. Support for CPU will be added in an upcoming patch.

source/loader/layers/sanitizer/common.hpp
```diff
@@ -0,0 +1,228 @@
...
+        return "UR_RESULT_ERROR_ADAPTER_SPECIFIC";
+    default:
+        return "UR_RESULT_ERROR_UNKNOWN";
+    }
+};
+
+// Trace an internal PI call; returns in case of an error.
+#define UR_CALL(Call)                                                          \
+    {                                                                          \
+        if (PrintTrace)                                                        \
```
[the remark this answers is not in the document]
AllanZyne: Will do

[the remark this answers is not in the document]
AllanZyne: Done

CMakeLists.txt
AllanZyne: I changed UR_ENABLE_SANITIZER to be enabled by default, because as explained in documentation, this layer should only be enabled by SYCL/OpenMP runtime. There will be no overheads when the user program is not compiled with the flag "-fsanitizer=address".

source/loader/layers/sanitizer/asan_interceptor.hpp
```diff
@@ -0,0 +1,138 @@
...
+                                    ur_event_handle_t *OutEvent);
+
+    ContextInfo &getContextInfo(ur_context_handle_t Context) {
+        std::shared_lock<ur_shared_mutex> Guard(m_ContextMapMutex);
+        assert(m_ContextMap.find(Context) != m_ContextMap.end());
+        return *m_ContextMap[Context].get();
+    }
+
+  private:
+    std::unordered_map<ur_context_handle_t, std::unique_ptr<ContextInfo>>
```
pbalcer, line 133: as far as I can tell, you don't handle removal of context/device/queues. I'm not sure that's safe. It's not impossible for an object handle to have the exact same handle as some earlier deleted one.

source/loader/layers/sanitizer/asan_interceptor.hpp
```diff
@@ -0,0 +1,138 @@
...
+                                    ur_device_handle_t Device,
+                                    ur_queue_handle_t Queue, uptr Addr,
+                                    uptr Size, u8 Value,
+                                    ur_event_handle_t DepEvent,
+                                    ur_event_handle_t *OutEvent);
+
+    ContextInfo &getContextInfo(ur_context_handle_t Context) {
+        std::shared_lock<ur_shared_mutex> Guard(m_ContextMapMutex);
+        assert(m_ContextMap.find(Context) != m_ContextMap.end());
+        return *m_ContextMap[Context].get();
```
pbalcer, line 129: You are dropping the lock and returning a reference to an underlying value of a unique_ptr. While I think this is safe with the current implementation, because these things are only ever removed in the destructor, it might be worth considering using a shared_ptr here.

CMakeLists.txt
pbalcer: > I changed UR_ENABLE_SANITIZER to be enabled by default, because as explained in the documentation, this layer should only be enabled by SYCL/OpenMP runtime. There will be no overheads when the user program is not compiled with the flag "-fsanitizer=address".

Makes sense.

> I found it's hard to add basic tests now, because UR_LAYER_ASAN needs some virtual memory APIs to be active. If not, it will crash on calling a null pointer.
> Hope this PR https://github.com/oneapi-src/unified-runtime/pull/939 can be landed ASAP.

Let's wait with tests until that is merged then. Thanks.

source/loader/layers/sanitizer/common.hpp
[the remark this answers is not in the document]
AllanZyne: Thanks! Will do.

source/loader/layers/sanitizer/ur_sanitizer_layer.hpp
```diff
@@ -0,0 +1,52 @@
...
+    AddressSanitizer,
+    MemorySanitizer,
+    ThreadSanitizer,
+};
+
+///////////////////////////////////////////////////////////////////////////////
+class __urdlllocal context_t : public proxy_layer_context_t {
+  public:
+    ur_dditable_t urDdiTable = {};
+    SanitizerInterceptor *interceptor = {};
```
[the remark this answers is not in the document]
AllanZyne: Thanks. Will do.

source/loader/layers/sanitizer/ur_sanddi.cpp
```diff
@@ -0,0 +1,439 @@
...
+              dditable->VirtualMem.pfnGranularityGetInfo)) {
+            die("Some VirtualMem APIs are needed to enable UR_LAYER_ASAN");
+        }
+
+        if (!dditable->PhysicalMem.pfnCreate) {
+            die("Some PhysicalMem APIs are needed to enable UR_LAYER_ASAN");
+        }
+    }
+
+    // FIXME: Interceptor needs use some of APIs that aren't registered, maybe just copy needed APIs?
```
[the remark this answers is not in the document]
AllanZyne: I found that other layers, like the validation and tracing layer, only save the functions that they intercept into their own DDI table. But for the sanitizer layer, it needs to use some functions that it doesn't intercept, therefore, just for convenience, I copy the whole DDI table here.

README.md
```diff
@@ -126,6 +126,7 @@ List of options provided by CMake:
 | UR_USE_UBSAN | Enable UndefinedBehavior Sanitizer | ON/OFF | OFF |
 | UR_USE_MSAN | Enable MemorySanitizer (clang only) | ON/OFF | OFF |
 | UR_ENABLE_TRACING | Enable XPTI-based tracing layer | ON/OFF | OFF |
+| UR_ENABLE_SANITIZER | Enable device sanitizer layer | ON/OFF | OFF |
```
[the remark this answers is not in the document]
AllanZyne: Will do.

source/loader/layers/sanitizer/asan_interceptor.hpp
AllanZyne: Makes sense. I decided to use shared_ptr, thanks

source/loader/layers/sanitizer/asan_interceptor.hpp
AllanZyne: Thanks for your reminder. I added some interceptions on release functions.

fabiomestre:

AllanZyne:

Thanks for the reminder. I've merged this PR to the main branch.

CMakeLists.txt
pbalcer: So far I've only reviewed the generic/UR-specific parts of this PR, without delving too deeply into the specifics of how the sanitizer works. I was hoping someone familiar with this would review it. I'll take a look today.

As for tests, #939 is ready to merge soon, so it should be possible to add something soon. But unfortunately PVC runners aren't going to be available any time soon. For now, we can merge it without tests since this is fairly self-contained, but please create a github issue to create them.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+ur_result_t SanitizerInterceptor::allocateMemory(
+    ur_context_handle_t Context, ur_device_handle_t Device,
+    const ur_usm_desc_t *Properties, ur_usm_pool_handle_t Pool, size_t Size,
+    void **ResultPtr, USMMemoryType Type) {
+    auto Alignment = Properties->align;
+    assert(Alignment == 0 || IsPowerOfTwo(Alignment));
+
+    auto ContextInfo = getContextInfo(Context);
+    auto DeviceInfo = ContextInfo->getDeviceInfo(Device);
```
pbalcer, line 95: add a comment here saying that `Device` is `nullptr` if `Type == USMMemoryType::HOST`. I know there's a comment like that in the header file, but I think it might be useful here as well.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,594 @@
...
+    UR_CALL(enqueueMemSetShadow(Context, Device, Queue, AllocInfo->AllocBegin,
+                                AllocInfo->AllocSize, 0, LastEvent,
+                                &LastEvent));
+
+    uptr TailBegin = RoundUpTo(AllocInfo->UserEnd, ASAN_SHADOW_GRANULARITY);
+    uptr TailEnd = AllocInfo->AllocBegin + AllocInfo->AllocSize;
+
+    // User tail
+    if (TailBegin != AllocInfo->UserEnd) {
+        auto Value = AllocInfo->UserEnd -
```
pbalcer, line 394: is this specific tail value read somewhere? are is there a layout documentation somewhere?

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,594 @@
...
+                                                std::defer_lock);
+    std::unique_lock<ur_shared_mutex> DeviceGuard(DeviceInfo->Mutex,
+                                                  std::defer_lock);
+    std::scoped_lock<std::shared_lock<ur_shared_mutex>,
+                     std::unique_lock<ur_shared_mutex>, ur_mutex>
+        Guard(HostGuard, DeviceGuard, QueueInfo->Mutex);
+
+    ur_event_handle_t LastEvent = QueueInfo->LastEvent;
+
+    // FIXME: Always update host USM, but it'd be better to update host USM
```
pbalcer, line 453: but this doesn't do anything, no? it will go into `enqueueMemSetShadow`, which will then subsequently die on with unsupported device message.

Can you explain why host memory must be updated every time? won't that overwrite existing data?

source/loader/layers/sanitizer/asan_interceptor.hpp
```diff
@@ -0,0 +1,143 @@
...
+        return DeviceMap[Device];
+    }
+
+    std::shared_ptr<QueueInfo> getQueueInfo(ur_queue_handle_t Queue) {
+        std::shared_lock<ur_shared_mutex> Guard(Mutex);
+        assert(QueueMap.find(Queue) != QueueMap.end());
+        return QueueMap[Queue];
+    }
+
+    std::shared_ptr<USMAllocInfo> getUSMAllocInfo(uptr Address) {
```
pbalcer, line 66: this doesn't appear to be used anywhere.

source/loader/layers/sanitizer/asan_interceptor.hpp
```diff
@@ -0,0 +1,143 @@
...
+    ur_shared_mutex Mutex;
+    // Note: nullptr is host device
+    std::unordered_map<ur_device_handle_t, std::shared_ptr<DeviceInfo>>
+        DeviceMap;
+    std::unordered_map<ur_queue_handle_t, std::shared_ptr<QueueInfo>> QueueMap;
+
+    /// key: USMAllocInfo.AllocBegin
+    /// value: USMAllocInfo
+    /// Use AllocBegin as key can help to detect underflow pointer
+    std::map<uptr, std::shared_ptr<USMAllocInfo>> AllocatedUSMMap;
```
pbalcer, line 81: it might be simpler to use an `unordered_map` indexed by the user pointer.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+    *ResultPtr = reinterpret_cast<void *>(UserBegin);
+
+    auto AllocInfo = std::make_shared<USMAllocInfo>(
+        USMAllocInfo{AllocBegin, UserBegin, UserEnd, NeededSize, Type});
+
+    // For updating shadow memory
+    {
+        std::scoped_lock<ur_shared_mutex> Guard(DeviceInfo->Mutex);
+        DeviceInfo->AllocInfos.emplace_back(AllocInfo);
```
pbalcer, line 140: I don't think `AllocInfos` is ever removed from (apart from the clear on kernel launch). It's not impossible for software to do some sort of allocate/deallocate cycle prior to running any kernel, causing duplicate entries in `AllocInfos`. This might be especially problematic for host allocations where this is never cleared.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,594 @@
...
+                        context.logger.error("zePhysicalMemCreate(): {}", URes);
+                        return URes;
+                    }
+                }
+
+                context.logger.debug("zeVirtualMemMap: {} ~ {}",
+                                     (void *)MappedPtr,
+                                     (void *)(MappedPtr + PageSize - 1));
+
+                // FIXME: No flag to check the failed reason is VA is already mapped
```
pbalcer, line 336: is there a reasonable path towards addressing that? if yes, please add an issue to add the appropriate error code so we don't forget.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,594 @@
...
+            LastEvent = NewEvent;
+            return true;
+        };
+
+        // Device shadow memory offset
+        EnqueueWriteGlobal(kSPIR_AsanShadowMemoryGlobalStart,
+                           &DeviceInfo->ShadowOffset);
+        EnqueueWriteGlobal(kSPIR_AsanShadowMemoryGlobalEnd,
+                           &DeviceInfo->ShadowOffsetEnd);
+    } while (false);
```
pbalcer, line 589: why the single iteration loop?

AllanZyne:
> Please take a look at the build log.
> 
> ```
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In function ‘ur_context_handle_t_* ur_sanitizer_layer::{anonymous}::getContext(ur_queue_handle_t)’:
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:60:10: error: unused variable ‘Result’ [-Werror=unused-variable]
>    60 |     auto Result = context.urDdiTable.Queue.pfnGetInfo(
>       |          ^~~~~~
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In function ‘ur_device_handle_t_* ur_sanitizer_layer::{anonymous}::getDevice(ur_queue_handle_t)’:
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:69:10: error: unused variable ‘Result’ [-Werror=unused-variable]
>    69 |     auto Result = context.urDdiTable.Queue.pfnGetInfo(
>       |          ^~~~~~
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In function ‘ur_program_handle_t_* ur_sanitizer_layer::{anonymous}::getProgram(ur_kernel_handle_t)’:
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:78:10: error: unused variable ‘Result’ [-Werror=unused-variable]
>    78 |     auto Result = context.urDdiTable.Kernel.pfnGetInfo(
>       |          ^~~~~~
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In member function ‘ur_result_t ur_sanitizer_layer::SanitizerInterceptor::allocateMemory(ur_context_handle_t, ur_device_handle_t, const ur_usm_desc_t*, ur_usm_pool_handle_t, size_t, void**, ur_sanitizer_layer::USMMemoryType)’:
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:124:10: error: unused variable ‘AllocEnd’ [-Werror=unused-variable]
>   124 |     uptr AllocEnd = AllocBegin + NeededSize;
>       |          ^~~~~~~~
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In member function ‘std::string ur_sanitizer_layer::SanitizerInterceptor::getKernelName(ur_kernel_handle_t)’:
> /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:248:10: error: variable ‘Res’ set but not used [-Werror=unused-but-set-variable]
>   248 |     auto Res = context.urDdiTable.Kernel.pfnGetInfo(
>       |          ^~~
> cc1plus: all warnings being treated as errors
> ```

Is there any standard method to handle the case that one value is only used in assert?
I'm going to define a macro like the following code, is this a proper way? 
Thanks for your suggestion!

```cpp
#ifndef NDEBUG
#define UR_ASSERT_EQ(Call, Result) assert(Call == Result)
#else
#define UR_ASSERT_EQ(Call, Result) (void)Call
#endif
```

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+#include "asan_interceptor.hpp"
+#include "device_sanitizer_report.hpp"
+#include "ur_sanitizer_layer.hpp"
+
+namespace ur_sanitizer_layer {
+
+namespace {
+
+// These magic values are written to shadow for better error
```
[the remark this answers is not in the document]
AllanZyne: currently no, these values were arbitrarily assigned by me

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,594 @@
...
+    assert(Res == UR_RESULT_SUCCESS);
+    KernelNameBuf[KernelNameSize] = '\0';
+
+    return std::string(KernelNameBuf.data(), KernelNameSize);
+}
+
+ur_result_t SanitizerInterceptor::allocShadowMemory(
+    ur_context_handle_t Context, std::shared_ptr<DeviceInfo> &DeviceInfo) {
+    if (DeviceInfo->Type == DeviceType::CPU) {
+        DeviceInfo->ShadowOffset = 0x00007fff7fffULL;
```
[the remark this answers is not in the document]
AllanZyne: I haven't do enough work on cpu type now, I'll delete these line.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Make sense. Done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Yes, these values will be read in some functions in libdevice.
This idea is from asan in compiler-rt, I think if someone is familiar with asan, then he/she will know what's the layout of shadow memory of asan.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: > but this doesn't do anything, no? it will go into `enqueueMemSetShadow`, which will then subsequently die with unsupported device message.

I don't think so. Here the `Device` is not `nullptr`, as it comes from `Queue`.

> Can you explain why host memory must be updated every time? won't that overwrite existing data?

In fact, the shadow memory of host USM needs to be updated only once per-device, and you can treat these values as read-only data in kernel, and it needn't to be updated until the host USM have been released(I haven't added this logic now).

source/loader/layers/sanitizer/asan_interceptor.hpp
AllanZyne: It will be used to implement the logic of memory releasing.

source/loader/layers/sanitizer/asan_interceptor.hpp
AllanZyne: This is intentional. I need to compare the value of pointer address.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: > is there a reasonable path towards addressing that?

Maybe it not really urgent, but I would be better to have one.

Thanks for your reminder. I added a new issue for it. 
https://github.com/oneapi-src/unified-runtime/issues/1174

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: it's not meaningful now. I'll remove this loop.

AllanZyne:
> > Please take a look at the build log.
> > ```
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In function ‘ur_context_handle_t_* ur_sanitizer_layer::{anonymous}::getContext(ur_queue_handle_t)’:
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:60:10: error: unused variable ‘Result’ [-Werror=unused-variable]
> >    60 |     auto Result = context.urDdiTable.Queue.pfnGetInfo(
> >       |          ^~~~~~
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In function ‘ur_device_handle_t_* ur_sanitizer_layer::{anonymous}::getDevice(ur_queue_handle_t)’:
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:69:10: error: unused variable ‘Result’ [-Werror=unused-variable]
> >    69 |     auto Result = context.urDdiTable.Queue.pfnGetInfo(
> >       |          ^~~~~~
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In function ‘ur_program_handle_t_* ur_sanitizer_layer::{anonymous}::getProgram(ur_kernel_handle_t)’:
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:78:10: error: unused variable ‘Result’ [-Werror=unused-variable]
> >    78 |     auto Result = context.urDdiTable.Kernel.pfnGetInfo(
> >       |          ^~~~~~
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In member function ‘ur_result_t ur_sanitizer_layer::SanitizerInterceptor::allocateMemory(ur_context_handle_t, ur_device_handle_t, const ur_usm_desc_t*, ur_usm_pool_handle_t, size_t, void**, ur_sanitizer_layer::USMMemoryType)’:
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:124:10: error: unused variable ‘AllocEnd’ [-Werror=unused-variable]
> >   124 |     uptr AllocEnd = AllocBegin + NeededSize;
> >       |          ^~~~~~~~
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp: In member function ‘std::string ur_sanitizer_layer::SanitizerInterceptor::getKernelName(ur_kernel_handle_t)’:
> > /home/runner/work/unified-runtime/unified-runtime/source/loader/layers/sanitizer/asan_interceptor.cpp:248:10: error: variable ‘Res’ set but not used [-Werror=unused-but-set-variable]
> >   248 |     auto Res = context.urDdiTable.Kernel.pfnGetInfo(
> >       |          ^~~
> > cc1plus: all warnings being treated as errors
> > ```
> 
> Is there any standard method to handle the case that one value is only used in assert? I'm going to define a macro like the following code, is this a proper way? Thanks for your suggestion!
> 
> ```c++
> #ifndef NDEBUG
> #define UR_ASSERT_EQ(Call, Result) assert(Call == Result)
> #else
> #define UR_ASSERT_EQ(Call, Result) (void)Call
> #endif
> ```

I've found another way. No need to reply. Thanks.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Yes, you're right.
As what I mentioned in the above comments, currently I haven't completed the releasing logic.
In fact, the releasing logic needs much more consideration, so I'll leave it in another patch.

CMakeLists.txt
AllanZyne: > or now, we can merge it without tests since this is fairly self-contained, but please file a github issue to create them.

Ok, I have removed the tests now, and opened a new issue for tracking this.
https://github.com/oneapi-src/unified-runtime/issues/1175

Thanks!

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+        auto AH = &SPIR_DeviceSanitizerReportMem;
+        if (!AH->Flag) {
+            return;
+        }
+
+        const char *File = AH->File[0] ? AH->File : "<unknown file>";
+        const char *Func = AH->Func[0] ? AH->Func : "<unknown func>";
+
+        fprintf(stderr, "\n====ERROR: DeviceSanitizer: %s on %s\n\n",
```
pbalcer, line 231: this is now causing warnings on Windows. Might be simpler to add something like "print always" function:
```patch
index 0575660f..27fcca42 100644
--- a/source/common/logger/ur_logger.hpp
+++ b/source/common/logger/ur_logger.hpp
@@ -43,6 +43,11 @@ inline void error(const char *format, Args &&...args) {
     get_logger().log(logger::Level::ERR, format, std::forward<Args>(args)...);
 }
 
+template <typename... Args>
+inline void always(const char *format, Args &&...args) {
+    get_logger().always(format, std::forward<Args>(args)...);
+}
+
 inline void setLevel(logger::Level level) { get_logger().setLevel(level); }
 
 inline void setFlushLevel(logger::Level level) {
diff --git a/source/common/logger/ur_logger_details.hpp b/source/common/logger/ur_logger_details.hpp
index 4759a2fd..bf63f7ff 100644
--- a/source/common/logger/ur_logger_details.hpp
+++ b/source/common/logger/ur_logger_details.hpp
@@ -51,6 +51,12 @@ class Logger {
         log(logger::Level::ERR, format, std::forward<Args>(args)...);
     }
 
+    template <typename... Args> void always(const char *format, Args &&...args) {
+        if (sink) {
+            sink->log(level, format, std::forward<Args>(args)...);
+        }
+    }
+
     template <typename... Args>
     void log(logger::Level level, const char *format, Args &&...args) {
         if (level < this->level) {
```

source/loader/layers/sanitizer/asan_interceptor.cpp
pbalcer: I understand that, but you are relaying on some (hopefully documented) layout/ABI of the compiler-rt. And if that changes, this code will break. That's why I like having source of truth link near soft dependencies such as this one.

source/loader/layers/sanitizer/asan_interceptor.hpp
pbalcer: For the release logic I'm assuming? The only current use of this map is in `releaseMemory` where you are searching for the base allocation based on a user pointer. That can be done directly with a hashmap.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Makes sense. I've added some comments on memory allocation. Thanks!

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Thanks! Done.

pbalcer (approved):

AllanZyne:
So, do I need more reviewers to approve this PR? Or can someone help to merge this PR?
Thanks!

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: I found that this way is not proper as it will add some unwanted prefixes. @pbalcer

source/common/logger/ur_logger_details.hpp
```diff
@@ -51,6 +51,13 @@ class Logger {
         log(logger::Level::ERR, format, std::forward<Args>(args)...);
     }
 
+    template <typename... Args>
+    void always(const char *format, Args &&...args) {
+        if (sink) {
+            sink->log(level, format, std::forward<Args>(args)...);
```
pbalcer, line 57:
```suggestion
            sink->log(Level::QUIET, format, std::forward<Args>(args)...);
```

source/loader/layers/sanitizer/asan_interceptor.cpp
pbalcer: I didn't test my suggestion :/ Sorry about that. I think I know the problem, see above.

If you don't want the prefix, you can set `skip_prefix` when creating the logger:
`logger::create_logger("sanitizer", true)`

pbalcer:
> So, do I need more reviewers to approve this PR? Or can someone help to merge this PR? Thanks!

This patch is still waiting for #939 to be merged, right? In any case, we typically require two reviewers. I've asked in the team for someone else to take a look, but it might be best if you can find someone on your team with knowledge of how the sanitizers work to also do a review.

kbenzie:
> This patch is still waiting for #939 to be merged, right?

This is correct. I have already sent an email with the subject *Virtual Memory Sanitizer Layer dependency* to @AllanZyne and @cdai2 containing the details of why #939 is currently blocked.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: > I didn't test my suggestion :/ Sorry about that. I think I know the problem, see above.
> 
> If you don't want the prefix, you can set `skip_prefix` when creating the logger: `logger::create_logger("sanitizer", true)`

Thanks! Done.

AllanZyne:
> > This patch is still waiting for #939 to be merged, right?
> 
> This is correct. I have already sent an email with the subject _Virtual Memory Sanitizer Layer dependency_ to @AllanZyne and @cdai2 containing the details of why #939 is currently blocked.

Hi @kbenzie, since https://github.com/oneapi-src/unified-runtime/pull/939 has been merged, can you help review and merge this PR? 

Thank you very much!

source/loader/layers/sanitizer/ur_sanddi.cpp
```diff
@@ -0,0 +1,481 @@
...
```
aarongreig, line 456: based on the discussion I don't think the todo is needed:
```suggestion
```

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,613 @@
...
+    }
+}
+
+std::string SanitizerInterceptor::getKernelName(ur_kernel_handle_t Kernel) {
+    size_t KernelNameSize = 0;
+    [[maybe_unused]] auto Res = context.urDdiTable.Kernel.pfnGetInfo(
+        Kernel, UR_KERNEL_INFO_FUNCTION_NAME, 0, nullptr, &KernelNameSize);
+    assert(Res == UR_RESULT_SUCCESS);
+
+    std::vector<char> KernelNameBuf(KernelNameSize + 1);
```
aarongreig, line 260: I'm 99% sure you can just do `KernelNameBuf(KernelNameSize)`, the spec for `UR_KERNEL_INFO_FUNCTION_NAME` states "Return null-terminated kernel function name.". Unless the drivers you're testing with aren't adding the null character?

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,613 @@
...
+            Context, nullptr, SHADOW_SIZE, (void **)&DeviceInfo->ShadowOffset);
+        if (Result != UR_RESULT_SUCCESS) {
+            context.logger.error(
+                "Shadow memory is allocated failed on PVC ({})", Result);
+            return Result;
+        }
+
+        DeviceInfo->ShadowOffsetEnd = DeviceInfo->ShadowOffset + SHADOW_SIZE;
+    } else {
+        die("Unsupport device type");
```
aarongreig, line 292: instead of `die` here it would be better to output an error with `logger` and return an error code

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,613 @@
...
+    ur_event_handle_t DepEvent, ur_event_handle_t *OutEvent) {
+
+    uint32_t NumEventsInWaitList = DepEvent ? 1 : 0;
+    const ur_event_handle_t *EventsWaitList = DepEvent ? &DepEvent : nullptr;
+
+    auto ContextInfo = getContextInfo(Context);
+    auto DeviceInfo = ContextInfo->getDeviceInfo(Device);
+
+    if (DeviceInfo->Type == DeviceType::CPU) {
+        die("Unsupport device type");
```
aarongreig, line 312: same as my other `die` comment (also applies to other instances of `die` in functions that return `ur_result_t` and have a `logger` available to output a message

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,613 @@
...
+            static ur_physical_mem_handle_t PhysicalMem{};
+
+            // Make sure [Ptr, Ptr + Size] is mapped to physical memory
+            for (auto MappedPtr = RoundDownTo(ShadowBegin, PageSize);
+                 MappedPtr <= ShadowEnd; MappedPtr += PageSize) {
+                if (!PhysicalMem) {
+                    auto URes = context.urDdiTable.PhysicalMem.pfnCreate(
+                        Context, Device, PageSize, &Desc, &PhysicalMem);
+                    if (URes != UR_RESULT_SUCCESS) {
+                        context.logger.error("zePhysicalMemCreate(): {}", URes);
```
aarongreig, line 339:
```suggestion
                        context.logger.error("urPhysicalMemCreate(): {}", URes);
```

source/loader/layers/sanitizer/common.hpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+inline constexpr bool IsAligned(uptr a, uptr alignment) {
+    return (a & (alignment - 1)) == 0;
+}
+
+// Valid redzone sizes are 16, 32, 64, ... 2048, so we encode them in 3 bits.
+// We use adaptive redzones: for larger allocation larger redzones are used.
+inline constexpr u32 RZLog2Size(u32 rz_log) {
+    // CHECK_LT(rz_log, 8);
```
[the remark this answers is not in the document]
AllanZyne: Sure, done

source/loader/layers/sanitizer/common.hpp
```diff
@@ -0,0 +1,90 @@
...
+                 : user_requested_size <= (1 << 14) - 256  ? 4
+                 : user_requested_size <= (1 << 15) - 512  ? 5
+                 : user_requested_size <= (1 << 16) - 1024 ? 6
+                                                           : 7;
+    return rz_log;
+}
+
+// ================================================================
+
+// Trace an internal PI call; returns in case of an error.
```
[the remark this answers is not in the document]
AllanZyne: Done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Makes sense. I tested it and you're right. Thanks!

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: > If you don't want the prefix, you can set `skip_prefix` when creating the logger:
> `logger::create_logger("sanitizer", true)`

Hi @pbalcer, I changed the logic of the logger so that, when the logger level is QUIET, the prefix will not be printed. 

This way is suitable for my needs, and I guess this is also consistent with what QUIET means?

AllanZyne:
> > So, do I need more reviewers to approve this PR? Or can someone help to merge this PR? Thanks!
> 
> This patch is still waiting for #939 to be merged, right? In any case, we typically require two reviewers. I've asked in the team for someone else to take a look, but it might be best if you can find someone on your team with knowledge of how the sanitizers work to also do a review.

Thanks, I've invited @wenju-he to help review this PR.

source/loader/layers/sanitizer/ur_sanddi.cpp
AllanZyne: Done

source/loader/layers/sanitizer/device_sanitizer_report.hpp
```diff
@@ -0,0 +1,95 @@
...
+const char *DeviceSanitizerFormat(DeviceSanitizerErrorType ErrorType) {
+    switch (ErrorType) {
+    case DeviceSanitizerErrorType::OUT_OF_BOUND:
+        return "out-of-bound-access";
+    case DeviceSanitizerErrorType::MISALIGNED:
+        return "misaligned-access";
+    case DeviceSanitizerErrorType::USE_AFTER_FREE:
+        return "use-after-free";
+    case DeviceSanitizerErrorType::OUT_OF_SHADOW_BOUND:
+        return "out-of-shadow-bound-access";
```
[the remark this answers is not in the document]
AllanZyne: Done

source/loader/layers/sanitizer/device_sanitizer_report.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include <cinttypes>
+
+namespace ur_sanitizer_layer {
+
+enum class DeviceSanitizerErrorType : int32_t {
+    UNKNOWN,
+    OUT_OF_BOUND,
+    MISALIGNED,
+    USE_AFTER_FREE,
+    OUT_OF_SHADOW_BOUND,
```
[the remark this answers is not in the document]
AllanZyne: Done

source/loader/layers/sanitizer/device_sanitizer_report.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+#pragma once
+
+#include <cinttypes>
+
+namespace ur_sanitizer_layer {
+
+enum class DeviceSanitizerErrorType : int32_t {
+    UNKNOWN,
+    OUT_OF_BOUND,
```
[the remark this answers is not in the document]
AllanZyne: Done

source/loader/layers/sanitizer/asan_interceptor.hpp
```diff
@@ -0,0 +1,143 @@
...
+};
+
+class SanitizerInterceptor {
+  public:
+    SanitizerInterceptor() {}
```
wenju-he, line 86: can we delete this line so that default constructor is used?

source/loader/layers/sanitizer/common.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+#include <cassert>
+#include <cstdint>
+
+namespace ur_sanitizer_layer {
+
+// ================================================================
+// Copy from LLVM compiler-rt/lib/asan
+
+typedef uintptr_t uptr;
```
wenju-he, line 26: using uptr = uintptr_t;

source/loader/layers/sanitizer/common.hpp
```diff
@@ -0,0 +1,90 @@
...
+typedef unsigned char u8;
+typedef unsigned int u32;
+
+#define ASAN_SHADOW_SCALE 3
```
wenju-he, line 30: constexpr unsigned ASAN_SHADOW_SCALE = 3;

source/loader/layers/sanitizer/ur_sanitizer_layer.hpp
```diff
@@ -0,0 +1,55 @@
...
+///////////////////////////////////////////////////////////////////////////////
+class __urdlllocal context_t : public proxy_layer_context_t {
+  public:
+    ur_dditable_t urDdiTable = {};
+    std::unique_ptr<SanitizerInterceptor> interceptor;
+    logger::Logger logger;
+    SanitizerType enabledType = SanitizerType::None;
+
+    context_t();
+    ~context_t();
```
wenju-he, line 40:
```suggestion
    ~context_t() = default;
```

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+#include "device_sanitizer_report.hpp"
+#include "ur_sanitizer_layer.hpp"
+
+namespace ur_sanitizer_layer {
+
+namespace {
+
+// These magic values are written to shadow for better error
+// reporting.
+const int kUsmDeviceRedzoneMagic = (char)0x81;
```
wenju-he, line 24: cast isn't needed.
```suggestion
constexpr int kUsmDeviceRedzoneMagic = 0x81;
```

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+const int kUsmHostRedzoneMagic = (char)0x82;
+const int kUsmSharedRedzoneMagic = (char)0x83;
+const int kMemBufferRedzoneMagic = (char)0x84;
+const int kUnkownRedzoneMagic = (char)0x8F;
+
+const auto kSPIR_AsanShadowMemoryGlobalStart = "__AsanShadowMemoryGlobalStart";
```
wenju-he, line 30:
```suggestion
constexpr auto kSPIR_AsanShadowMemoryGlobalStart = "__AsanShadowMemoryGlobalStart";
```

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+const auto kSPIR_AsanShadowMemoryGlobalEnd = "__AsanShadowMemoryGlobalEnd";
+
+const auto kSPIR_DeviceSanitizerReportMem = "__DeviceSanitizerReportMem";
+
+DeviceSanitizerReport SPIR_DeviceSanitizerReportMem;
```
wenju-he, line 35:
```suggestion
static DeviceSanitizerReport SPIR_DeviceSanitizerReportMem;
```

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+    return UR_RESULT_SUCCESS;
+}
+
+ur_result_t SanitizerInterceptor::releaseMemory(ur_context_handle_t Context,
+                                                void *Ptr) {
+    auto ContextInfo = getContextInfo(Context);
+
+    std::shared_lock<ur_shared_mutex> Guard(ContextInfo->Mutex);
+
+    auto Addr = (uptr)Ptr;
```
wenju-he, line 172:
```suggestion
    auto Addr = reinterpret_cast<uptr>(Ptr);
```

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+                context.logger.warning("Device Global[{}] Write Failed: {}",
+                                       Name, Result);
+                return false;
+            }
+            LastEvent = NewEvent;
+            return true;
+        };
+
+        // Device shadow memory offset
+        EnqueueWriteGlobal(kSPIR_AsanShadowMemoryGlobalStart,
```
wenju-he, line 608: return value of EnqueueWriteGlobal isn't used? Should prepareLaunch return false in case EnqueueWriteGlobal fails?

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+    // Query device type
+    ur_device_type_t DeviceType;
+    UR_CALL(context.urDdiTable.Device.pfnGetInfo(
+        Device, UR_DEVICE_INFO_TYPE, sizeof(DeviceType), &DeviceType, nullptr));
+    switch (DeviceType) {
+    case UR_DEVICE_TYPE_CPU:
+        DeviceInfo->Type = DeviceType::CPU;
+        break;
+    case UR_DEVICE_TYPE_GPU:
+        DeviceInfo->Type = DeviceType::GPU_PVC;
```
wenju-he, line 532: GPU_PVC is better to be a Name rather than a Type. Perhaps a device name entry is needed in DeviceInfo?

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+        UR_CALL(enqueueAllocInfo(Context, Device, Queue, AllocInfo, LastEvent));
+    }
+    DeviceInfo->AllocInfos.clear();
+
+    QueueInfo->LastEvent = LastEvent;
+
+    return UR_RESULT_SUCCESS;
+}
+
+ur_result_t SanitizerInterceptor::addContext(ur_context_handle_t Context) {
```
wenju-he, line 492: since unordered_map is a map, can we rename addContext to insertContext and rename removeContext to eraseContext to align with the method name in std unordered_map?

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+    DeviceInfo->Alignment = ASAN_SHADOW_GRANULARITY;
+
+    // TODO: Check if host asan is enabled
+    DeviceInfo->ShadowOffset = 0;
+    DeviceInfo->ShadowOffsetEnd = 0;
+
+    ContextInfo->DeviceMap.emplace(nullptr, std::move(DeviceInfo));
+
+    std::scoped_lock<ur_shared_mutex> Guard(m_ContextMapMutex);
+    m_ContextMap.emplace(Context, std::move(ContextInfo));
```
wenju-he, line 507: what if Context is emplaced twice? Is an assert needed here?

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+    }
+
+    // TODO: Update shadow memory
+    return context.urDdiTable.USM.pfnFree(Context,
+                                          (void *)AllocInfo->AllocBegin);
+}
+
+bool SanitizerInterceptor::preLaunchKernel(ur_kernel_handle_t Kernel,
+                                           ur_queue_handle_t Queue,
+                                           ur_event_handle_t &Event) {
```
wenju-he, line 199: Event's type in postLaunchKernel is a pointer. Make the two functions consistent?

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+}
+
+void SanitizerInterceptor::postLaunchKernel(ur_kernel_handle_t Kernel,
+                                            ur_queue_handle_t Queue,
+                                            ur_event_handle_t *Event) {
+    auto Program = getProgram(Kernel);
+    ur_event_handle_t ReadEvent{};
+
+    // If kernel has defined SPIR_DeviceSanitizerReportMem, then we try to read it
+    // to host, but it's okay that it isn't defined
```
wenju-he, line 223: In which case is SPIR_DeviceSanitizerReportMem not defined?

The below code isn't handling the case when DeviceGlobalVariableRead fails even if SPIR_DeviceSanitizerReportMem exists.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+    } else if (DeviceInfo->Type == DeviceType::GPU_PVC) {
+        ur_event_handle_t InternalEvent{};
+        ur_event_handle_t *Event = OutEvent ? OutEvent : &InternalEvent;
+
+        uptr ShadowBegin = MemToShadow_PVC(DeviceInfo->ShadowOffset, Ptr);
+        uptr ShadowEnd =
+            MemToShadow_PVC(DeviceInfo->ShadowOffset, Ptr + Size - 1);
+
+        // Maybe in future, we needn't to map physical memory manually
+        const bool IsNeedMapPhysicalMem = true;
```
wenju-he, line 325: I suggest to remove this variable since at the moment it isn't clear if the `future` case exists.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+
+        if (IsNeedMapPhysicalMem) {
+            // We use fixed GPU PageSize: 64KB
+            const size_t PageSize = 64 * 1024u;
```
wenju-he, line 329: shouldn't the page size be queried from device?

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,617 @@
...
+
+    QueueInfo->LastEvent = LastEvent;
+
+    return UR_RESULT_SUCCESS;
+}
+
+ur_result_t SanitizerInterceptor::addContext(ur_context_handle_t Context) {
+    auto ContextInfo = std::make_shared<ur_sanitizer_layer::ContextInfo>();
+
+    // Host Device
```
wenju-he, line 495: Is this comment accurate? SYCL doesn't have a host device.
I suppose you meant just Host?
However, the code below `DeviceInfo->Type = DeviceType::CPU` is showing a CPU device is initialized.

If the intent is to initialize information for the host, do we currently have logic to handle the coordination between host and a device? I suggest to to remove below code since it is for the future. When the coordination is in place, we can add below code back.

.gitignore
```diff
@@ -86,3 +86,7 @@ out/
 
 # External content
 */**/external
+
+# VS clangd
+/.cache
+/compile_commands.json
```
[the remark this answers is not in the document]
AllanZyne: Yep, these ignores are used for VS clangd plugin

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: > In which case is SPIR_DeviceSanitizerReportMem not defined?

When these are no memory read/write in the kernel, for example, an empty kernel.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: > I suggest to remove this variable since at the moment it isn't clear if the `future` case exists.

I've synced with Michal, system USM will be supported in GPU in the future. At that time, I think we won't need to map physical memory manually.

.gitignore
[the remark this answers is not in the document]
wenju-he: IIUC this file should be generated in build folder, is the root folder of this repo also the build folder?

source/loader/layers/sanitizer/asan_interceptor.cpp
wenju-he: I think a better approach is to query if SPIR_DeviceSanitizerReportMem exists and then read. This would cover all fail cases.

source/loader/layers/sanitizer/asan_interceptor.cpp
wenju-he: I don't think system USM is a viable solution for shadow memory because it is too slow.

.gitignore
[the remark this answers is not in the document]
AllanZyne: > IIUC this file should be generated in build folder, is the root folder of this repo also the build folder?

Yes, this file is generated in the build folder, then I'll link this file to the root folder, which will automatically be detected by the VS plugin and apply it to the whole project. This is very convenient.

You can find the same configuration in intel/llvm project as well.

CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ option(UR_USE_UBSAN "enable UndefinedBehaviorSanitizer" OFF)
 option(UR_USE_MSAN "enable MemorySanitizer" OFF)
 option(UR_USE_TSAN "enable ThreadSanitizer" OFF)
 option(UR_ENABLE_TRACING "enable api tracing through xpti" OFF)
+option(UR_ENABLE_SANITIZER "enable device sanitizer" ON)
```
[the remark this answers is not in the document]
AllanZyne: I follow the same naming style as the tracing layer.

source/loader/layers/sanitizer/asan_interceptor.hpp
AllanZyne: done

source/loader/layers/sanitizer/common.hpp
AllanZyne: done

source/loader/layers/sanitizer/common.hpp
AllanZyne: done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Yep, I'll add an assert here.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: There's no API to query if the device global variables exist.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: I'm not sure. Maybe it will help make Asan more powerful.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: It seems like there's no api do this?

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: > I suppose you meant just Host?

Yes. I don't think host device is a wrong term. Host code runs on the host device.

> If the intent is to initialize information for the host, do we currently have logic to handle the coordination between host and a device?

Yes, we have. I have saved all host USM allocation information in the host device.

CMakeLists.txt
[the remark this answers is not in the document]
wenju-he: > I follow the same naming style as the tracing layer.

I think the confusing still exists even if it the same naming style.

source/loader/layers/sanitizer/ur_sanitizer_layer.hpp
AllanZyne: old code

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: I need the sign bit expansion here, so the cast is needed.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: It's defined in an anonymous namespace.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: I ignore its returning value intentionally. prepareLaunch shouldn't false.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Save it to a string will make the code a little slow, and I don't think using enum value as device type is a big problem.
I'll use the device name to check the device type in the future.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: done

source/loader/layers/sanitizer/asan_interceptor.cpp
wenju-he: > I suggest to remove this variable since at the moment it isn't clear if the `future` case exists.

I still suggest removing the code. Even if you insist that system USM can be used for shadow memory, which I don't agree, you can always add it back when IsNeedMapPhysicalMem becomes used in the future.

source/loader/layers/sanitizer/asan_interceptor.cpp
wenju-he: Can zeVirtualMemQueryPageSize used?

If there is no such API, what is the rational of choosing 64K size here?

source/loader/layers/sanitizer/asan_interceptor.cpp
wenju-he: Being a name doesn't need to be a string. You can define a new enum for the device name.
It is just that the above line doesn't make much sense because the left is device type and the right is device name.

source/loader/layers/sanitizer/asan_interceptor.cpp
wenju-he: I don't understand why prepareLaunch shouldn't check if a call inside is successful. If something goes wrong in the function, you just simply ignore it?

source/loader/layers/sanitizer/asan_interceptor.cpp
wenju-he: both OpenCL and LevelZero have API to get a pointer to a global variable. Do you mean unified runtime doesn't have this API? Is this a gap in unified runtime?
I don't think it is elegant to write to a global variable without querying if it exists.

source/loader/layers/sanitizer/asan_interceptor.cpp
wenju-he: You can still use constexpr

source/loader/layers/sanitizer/asan_interceptor.cpp
wenju-he: SYCL doesn't have host device. What you meant is just "host". Please don't use "host device".

> Yes, we have. I have saved all host USM allocation information in the host device.

Could you please point me to the code where you saved the info?

.gitignore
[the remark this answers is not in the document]
wenju-he: Ok
I don't permission to mark this as resolved.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: I filed a issue to track this.

https://github.com/oneapi-src/unified-runtime/issues/1214

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Here is the same reason as SPIR_DeviceSanitizerReportMem. 
If the kernel is empty, then there's no kSPIR_AsanShadowMemoryGlobalStart, and so on.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: Makes sense. I'll change it.

CMakeLists.txt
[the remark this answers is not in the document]
AllanZyne: I agree. I think the flag to enable layers needs a new naming style.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: > You can define a new enum for the device name.

It seems it's hard to define the enum values of device name.

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: done

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: https://github.com/oneapi-src/unified-runtime/blob/79998b965232ff40e3a0b3829dd57f5466eb23b3/source/loader/layers/sanitizer/asan_interceptor.cpp#L91-L92

wenju-he: according to offline sync, @AllanZyne would you remove this part of code?

AllanZyne: Done

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -88,11 +88,14 @@ ur_result_t SanitizerInterceptor::allocateMemory(
     assert(Alignment == 0 || IsPowerOfTwo(Alignment));
 
     auto ContextInfo = getContextInfo(Context);
-    // Device is nullptr if Type == USMMemoryType::HOST
-    auto DeviceInfo = ContextInfo->getDeviceInfo(Device);
+    std::shared_ptr<DeviceInfo> DeviceInfo;
+    if (Device) {
+        DeviceInfo = ContextInfo->getDeviceInfo(Device);
```
wenju-he, line 93: nit: brace not needed for single line code according to llvm coding style. What's the coding style using in UR? Please also see other changes in this commit.

Is the 4-space indent align with UR coding style?

AllanZyne: > What's the coding style using in UR? Please also see other changes in this commit.

I searched some of UR codes, they add braces for single line code as well.

> Is the 4-space indent aligning with UR coding style?

Yes.

wenju-he (approved):
LGTM.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -0,0 +1,605 @@
...
+        ur_event_handle_t *Event = OutEvent ? OutEvent : &InternalEvent;
+
+        uptr ShadowBegin = MemToShadow_PVC(DeviceInfo->ShadowOffset, Ptr);
+        uptr ShadowEnd =
+            MemToShadow_PVC(DeviceInfo->ShadowOffset, Ptr + Size - 1);
+
+        {
+            static const size_t PageSize = [Context, Device]() {
+                size_t Size;
+                auto Result =
```
kbenzie, line 325: `Result` is only used in non Release builds so is causing our CI to fail.

Since this is a immediately invoked lamba function, I think `Result` should be hoisted outside of here so that when an error occurs it can be returned from `Sanitizer::Interceptor::enqueueMemSetShadow`.

Either that or mark this as `[[maybe_unused]]` to silence the warning (treated as error) which is causing the CI to fail.

AllanZyne: Thanks, I've added `[[maybe_unused]]`

AllanZyne:
Hi, I think I've got 2 approves, is this PR ready to merge?
Thanks!

kbenzie:
> Hi, I think I've got 2 approves, is this PR ready to merge?

Not until it passes all of the GitHub Actions.

AllanZyne:
> > Hi, I think I've got 2 approves, is this PR ready to merge?
> 
> Not until it passes all of the GitHub Actions.

Okay, I've noticed them, and I submitted a fix for them.

pbalcer (approved):
Your PR is now failing due to the use of getenv on Windows in a third-party component (xpti). This is a known issue that, #1206 should suppress the problem.

CMakeLists.txt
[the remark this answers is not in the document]
pbalcer: hmm, `UR_ENABLE_LAYER_XYZ` ?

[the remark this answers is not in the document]
AllanZyne: > hmm, `UR_ENABLE_LAYER_XYZ` ?

Yes, this style is very similar to "UR_BUILD_ADAPTER_XYZ"

cdai2:
> Your PR is now failing due to the use of getenv on Windows in a third-party component (xpti). This is a known issue, #1206 should suppress the problem.

Is there a scheduling to fix this #1206 issue?   
As this PR will not introduce other known issues,  Could you please land this PR first?  It blocks our following work.

kbenzie:
> > Your PR is now failing due to the use of getenv on Windows in a third-party component (xpti). This is a known issue, #1206 should suppress the problem.
> 
> Is there a scheduling to fix this #1206 issue? As this PR will not introduce other known issues, Could you please land this PR first? It blocks our following work.

#1206 has not been merged, please update this branch with the latest changes on the main branch.

AllanZyne:
> > > Your PR is now failing due to the use of getenv on Windows in a third-party component (xpti). This is a known issue, #1206 should suppress the problem.
> > 
> > 
> > Is there a scheduling to fix this #1206 issue? As this PR will not introduce other known issues, Could you please land this PR first? It blocks our following work.
> 
> #1206 has not been merged, please update this branch with the latest changes on the main branch.

Thanks! Done.

AllanZyne:
When will this PR be ready to merge? I'm willing to follow the process to make it land ASAP.

pbalcer:
Once the built passes it should be ready to merge.
Windows CI failed again:
```
  ur_testing.vcxproj -> D:\a\unified-runtime\unified-runtime\build\lib\Debug\ur_testing.lib
D:\a\unified-runtime\unified-runtime\source\loader\layers\sanitizer\asan_interceptor.cpp(102,30): error C2220: the following warning is treated as an error [D:\a\unified-runtime\unified-runtime\build\source\loader\ur_loader.vcxproj]
  test-params.vcxproj -> D:\a\unified-runtime\unified-runtime\build\bin\Debug\test-params.exe
D:\a\unified-runtime\unified-runtime\source\loader\layers\sanitizer\asan_interceptor.cpp(102,30): warning C4244: 'argument': conversion from 'ur_sanitizer_layer::uptr' to 'ur_sanitizer_layer::u32', possible loss of data [D:\a\unified-runtime\unified-runtime\build\source\loader\ur_loader.vcxproj]
D:\a\unified-runtime\unified-runtime\source\loader\layers\sanitizer\asan_interceptor.cpp(422,9): warning C4244: 'argument': conversion from 'ur_sanitizer_layer::uptr' to 'ur_sanitizer_layer::u8', possible loss of data [D:\a\unified-runtime\unified-runtime\build\source\loader\ur_loader.vcxproj]
```

PR #1075, opened 2023-11-14
Remove Python API bindings

callumfare:
Remove the Python bindings for UR. They are unused and just a holdover from the initial port of the scripts/templates used to generate UR.


pbalcer (commented):
remove the [template file](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/templates/api.py.mako) as well.

There's also a few python-specific functions/ifs in helper.py.

pbalcer (approved):

omarahmed1111 (approved):
LGTM, just a small note, 
I guess we need to also remove python from CI like [here](https://github.com/callumfare/unified-runtime/blob/22b34eaad5abfde33b5caaae17fce9730938bb7b/.github/workflows/cmake.yml#L113)

kbenzie (approved):

PR #1076, opened 2023-11-14
[common] add codecov

pbalcer:
Preview: https://app.codecov.io/gh/pbalcer/unified-runtime

Codecov still needs to be enabled for the `oneapi-src` organization. Also, this workflow needs to run on the default branch on the repo, so until the adapter and main branches are merged, we won't see the adapter coverage.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,78 @@
+name: Coverage
+
+on: [push, pull_request]
+
+jobs:
+  ubuntu-build:
+    name: Build - Ubuntu
+    strategy:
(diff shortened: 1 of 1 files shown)


lukaszstolarczuk (approved):
I think you could try and run this workflow on the adapters branch. I believe it always worked for us with multiple branches (master + stables)...

kswiecicki (approved):

pbalcer:
> I think you could try and run this workflow on the adapters branch. I believe it always worked for us with multiple branches (master + stables)...

You are right. I'll do that in a second step.

PR #1079, opened 2023-11-15
[CI][L0] adapt e2e workflow to l0 target

wlemkows:
This change adds support for l0 to e2e tests.
PR will not pass until we have a stable runner configuration with Intel GPU and level zero:
https://jira.devtools.intel.com/browse/GSD-7337

### DIFF
--- a/.github/workflows/e2e_level_zero.yml
+++ b/.github/workflows/e2e_level_zero.yml
@@ -1,0 +1,29 @@
+name: E2E Level Zero
+
+on:
+  schedule:
+    # Run every day at 23:00 UTC
+    - cron: '0 23 * * *'
+  issue_comment:
+    types: [created, edited]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  e2e-build-hw:
+    # trigger only if PR comment contains "e2e-l0"
+    if: ${{ (github.event.issue.pull_request && contains(github.event.comment.body, '/e2e-l0')) || (github.event_name == 'schedule') }}
+    name: Start e2e job
+    # use core flow, run it with level zero specific parameters
+    uses: ./.github/workflows/e2e_core.yml
+    # parameters that we pass to the core flow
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e_nightly.yml
```diff
@@ -1,17 +1,19 @@
...
+#    - cron: '0 23 * * *'
 
 jobs:
   e2e-build-hw:
     name: Build SYCL, UR, run E2E
     strategy:
       matrix:
         adapter: [
-          {name: CUDA}
+          {name: CUDA, str_name: cuda, config: "--cuda- --hip"},
```
lukaszstolarczuk, line 15: redundant `-` after `--cuda`

.github/workflows/e2e_nightly.yml
```diff
@@ -1,17 +1,19 @@
...
+          {name: L0, str_name: level_zero, config: ""}
```
lukaszstolarczuk, line 16: You should be able to just use one name - runners have multiple labels (both `l0` and `level_zero` are accepted) and they are not case sensitive.

.github/workflows/e2e_nightly.yml
```diff
@@ -60,11 +62,17 @@ jobs:
       run: LD_LIBRARY_PATH=${{github.workspace}}/dpcpp_compiler/lib
         cmake --build ${{github.workspace}}/ur-repo/build -j $(nproc)
 
-    - name: Set env vars & pre setup
+    - name: Set prefer UR
```
lukaszstolarczuk, line 65: can't this step be merged with the "Run pre setup" step?

.github/workflows/e2e_nightly.yml
wlemkows: Usage is case-sensitive.

pbalcer:
Closed in favor of #1612

PR #993, opened 2023-10-24
[CI] Add support for OpenCL CI on CPU

fabiomestre:

PR #1085, opened 2023-11-16
[UR][Tests] Add a CMake option to limit the test devices count in CTS

szadam:


test/conformance/source/environment.cpp
```diff
@@ -199,6 +202,19 @@ DevicesEnvironment::DevicesEnvironment(int argc, char **argv)
         error = "Could not find any devices associated with the platform";
         return;
     }
+    // Get the argument (test_devices_count) to limit test devices count.
+    u_long count_set = 0;
+    if (argc >= 3) {
+        count_set = std::strtoul(argv[2], nullptr, 10);
```
pbalcer, line 208: is this always the second argument?

szadam: It depends on cts_exe.py scripts. In this case, it will always be the second argument.

pbalcer: Right, but the gtest binary may be executed manually, the script can change etc. You shouldn't be hard-coding such assumptions.

test/conformance/README.md
```diff
@@ -8,4 +8,11 @@ In the future, when all bugs are fixed, and the tests pass,
...
 When you fix any test, the match file must be updated
 Empty match files indicate that there are no failing tests
-in a particular group for the corresponding adapter.
\ No newline at end of file
+in a particular group for the corresponding adapter.
+
+## How to limit the test devices count
+
+To limit how many devices you want to run the CTS on,
+use CMake option UR_TEST_DEVICES_COUNT. If you want to run
```
jandres742, line 16: could we make it a runtime option instead of a CMAKE one? imagine the case where users have the binaries for the CTS, and have different types of systems (single CPU, 4xGPU, CPU + 2xGPU, etc). They would have to build the binaries for each configuration.

test/conformance/source/environment.cpp
szadam: I've changed it. What about that?

test/conformance/source/environment.cpp
```diff
@@ -199,6 +201,22 @@ DevicesEnvironment::DevicesEnvironment(int argc, char **argv)
         error = "Could not find any devices associated with the platform";
         return;
     }
+    // Get the argument (test_devices_count) to limit test devices count.
+    u_long count_set = 0;
+    for (int i = 1; i < argc; ++i) {
```
pbalcer, line 206:
```suggestion
    for (int i = 1; i < argc - 1; ++i) {
```
and drop the `&& i + 1 < argc` in the if below?

pbalcer (approved):

test/conformance/source/environment.cpp
szadam: You're right. It will work too.

test/conformance/README.md
jandres742: thanks @szadam . could you share why it couldn't be done as part of this PR? Main reason I'm asking is because I have already run into this problem, where I was running the tests in a multi-device system, and it took me a while to figure out the issue was that the tests were built only for 1 device, so I would expect more people to run into the same problem, so addressing sooner would be great.

pbalcer: if you run the tests directly through their binary, you can use the `--test_devices_count 1`  cli option that @szadam is adding. If you are using ctest, this patch will automatically reduce the number of devices it will run with to 1. You don't need to set the cmake option.

There's another patch that @szadam is working on, #992, that will solve the broader problem of devices/platforms having the same names.

lukaszstolarczuk: I think what @jandres742 is proposing is implemented for OpenCL patch (here: https://github.com/oneapi-src/unified-runtime/pull/993/files#diff-d1f50b1ed696f085259b2a9c7d274ae3414700a1cea60f9945677cdae39e25cfR186) and it looks neat as well. The CMake option, though, has, I think, the added value (as Piotr mentioned) that it sets the default value of 1 and still can be overridden with the CLI param.

BTW, Adam has a new version (still on a private branch; to be polished): https://github.com/szadam/unified-runtime/pull/15/commits/ef742824118957136ecfc434f84feb446a366292#diff-d1f50b1ed696f085259b2a9c7d274ae3414700a1cea60f9945677cdae39e25cf

jandres742: thanks @pbalcer and @lukaszstolarczuk for the details.

jandres742:
+1

test/conformance/README.md
szadam: Sorry for the late response. I am preparing a new solution/version as Lukas said. I will push it today.

pbalcer (approved):

test/conformance/source/environment.cpp
```diff
@@ -33,6 +34,23 @@ std::ostream &operator<<(std::ostream &out,
     return out;
 }
 
+std::ostream &operator<<(std::ostream &out, const ur_device_handle_t &device) {
```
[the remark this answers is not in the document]
szadam: Done

test/conformance/README.md
```diff
@@ -8,4 +8,15 @@ In the future, when all bugs are fixed, and the tests pass,
...
+## How to set test device/platform name or limit the test devices/platforms count
+
+To limit how many devices/platforms you want to run the CTS on,
+use CMake option UR_TEST_DEVICES_COUNT or
+UR_TEST_PLATFORMS_COUNT. If you want to run the tests on
+all available devices/platforms, set 0. The default value is 1.
+If you run binaries for the tests, you can use the parameter
+`--platforms_count=COUNT/--devices_count=COUNT`.
+To set test device/platform name you want to run the CTS on, use
+parameter `--platform=NAME/--device=NAME`.
```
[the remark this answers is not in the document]
szadam: Done

PatKamin (approved):

fabiomestre:

PR #1088, opened 2023-11-16
[HIP] Ifdef entry point introduced in HIP 5.6

hdelan:
`hipArray3DGetDescriptor` was introduced in HIP 5.6

### DIFF
--- a/source/adapters/hip/memory.cpp
+++ b/source/adapters/hip/memory.cpp
@@ -264,6 +264,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
 
   switch (MemInfoType) {
   case UR_MEM_INFO_SIZE: {
+#if HIP_VERSION < 50600000
+    return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+#else
     try {
       const auto MemVisitor = [](auto &&Mem) -> size_t {
         using T = std::decay_t<decltype(Mem)>;
@@ -274,12 +277,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
           return AllocSize;
         } else if constexpr (std::is_same_v<T, SurfaceMem>) {
           HIP_ARRAY3D_DESCRIPTOR ArrayDescriptor;
-#if HIP_VERSION >= 50600000
           UR_CHECK_ERROR(
               hipArray3DGetDescriptor(&ArrayDescriptor, Mem.getArray()));
-#else
-          return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
-#endif
           const auto PixelSizeBytes =
               GetHipFormatPixelSize(ArrayDescriptor.Format) *
               ArrayDescriptor.NumChannels;
@@ -301,6 +300,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
     } catch (...) {
       return UR_RESULT_ERROR_UNKNOWN;
     }
+#endif
   }
   case UR_MEM_INFO_CONTEXT: {
     return ReturnValue(hMemory->getContext());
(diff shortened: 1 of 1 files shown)


source/adapters/hip/memory.cpp
```diff
@@ -274,7 +274,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
           return AllocSize;
         } else if constexpr (std::is_same_v<T, SurfaceMem>) {
           HIP_ARRAY3D_DESCRIPTOR ArrayDescriptor;
-          UR_CHECK_ERROR(hipArray3DGetDescriptor(&ArrayDescriptor, Mem.Array));
+#if HIP_VERSION_MAJOR >= 5 && HIP_VERSION_MINOR >= 6
```
alexbatashev, line 277: If I get it right, with this patch the feature will remain unavailable in 6.0. Isn't it better to use [HIP_VERSION](https://rocmdocs.amd.com/projects/HIP/en/develop/user_guide/faq.html#how-can-i-know-the-version-of-hip) macro instead?

hdelan: Yes you are correct. Have changed to `HIP_VERSION`

hdelan:
https://github.com/intel/llvm/pull/11963

JackAKirk (approved):

kbenzie (approved):

source/adapters/hip/memory.cpp
```diff
@@ -274,7 +274,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
           return AllocSize;
         } else if constexpr (std::is_same_v<T, SurfaceMem>) {
           HIP_ARRAY3D_DESCRIPTOR ArrayDescriptor;
-          UR_CHECK_ERROR(hipArray3DGetDescriptor(&ArrayDescriptor, Mem.Array));
+#if HIP_VERSION >= 50600000
+          UR_CHECK_ERROR(
+              hipArray3DGetDescriptor(&ArrayDescriptor, Mem.getArray()));
+#else
+          return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
```
al42and, line 281: I don't think this `return` is correct. We're in a lambda function that's invoked via `const auto AllocSize = std::visit(MemVisitor, hMemory->Mem);` below. Instead of returning the error code from `urMemGetInfo`, we will assign it as a value to the `AllocSize`.

hdelan: Aha good spot thanks @al42and

hdelan: Change made

source/adapters/hip/memory.cpp
```diff
@@ -264,6 +264,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
 
   switch (MemInfoType) {
   case UR_MEM_INFO_SIZE: {
+#if HIP_VERSION < 50600000
+    return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
```
al42and, line 268: But it's only broken for `SurfaceMem`; `BufferMem` should still work. This code, however, returns the error for both.

al42and: I think we should just `throw UR_RESULT_ERROR_UNSUPPORTED_FEATURE;` from `MemVisitor`, and then it will be caught and returned.

al42and:
Shall we also skip [`urMemGetInfoImageTest.Success` test](https://github.com/oneapi-src/unified-runtime/blob/41228d3010d48fee04483cd5a9252914d208e2a0/test/conformance/memory/urMemGetInfo.cpp#L81-L107), or is it better if it fails?

al42and (approved):

hdelan:
We have a fail list here https://github.com/oneapi-src/unified-runtime/blob/adapters/test/conformance/memory/memory_adapter_hip.match so it should be fine already since tests bail with Segmentation Fault before MemImage tests

kbenzie:
I'm going to include this in https://github.com/intel/llvm/pull/12004 since it addresses a compile error and should be merged ASAP.

PR #1089, opened 2023-11-16
[EXP][Command-Buffer] Add kernel command update

EwanC:
This change introduces a new API that allows the kernel commands of a command-buffer to be updated
with a new configuration. For example, modified arguments or ND-Range.

## API

The new API is defined in the following files and then source generated using scripts, so reviewers should look at:
* `scripts/core/EXP-COMMAND-BUFFER.rst`
* `scripts/core/exp-command-buffer.yml`

See [cl_khr_command_buffer_mutable_dispatch](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_command_buffer_mutable_dispatch) as prior art. The differences between the proposed API and the above are:

* Only the append kernel entry-point returns a command handle. I imagine this will be changed in future to enable other commands to do update.
* USM,  buffer, and scalar arguments can be updated, there is not equivalent update struct for `urKernelSetArgLocal` or `urKernelSetArgSampler`
* There is no granularity of optional support for update, an implementer must either implement all the ways to update a kernel configuration, or none of them.
* Command-handles are reference counted in UR, and extend the lifetime of the parent command-buffer.

## Adapters

The CUDA adapter is the only adapter that currently implements this new feature, other adapters don't report support. This is because CUDA is already an adapter supported by UR command-buffers, and the CUDA API for updating nodes already exists as a non-optional feature.

Reviewers should review the changes in `source/adapters/cuda/` to evaluate this,

## CTS

CTS tests are written to verify implementation, as there is not yet a DPC++ feature with testing to stress the code path (see https://github.com/reble/llvm/pull/340 for how that feature could look).

A new test directory has been created to test the command-buffer experimental feature, `test/conformance/exp_command_buffer`, which contains tests to stress using the feature defined by this extension so that it has code coverage. Reviewers should look at the new tests added here, and new device kernels in `test/conformance/device_code` to evaluate these changes.

### DIFF
--- a/scripts/core/EXP-COMMAND-BUFFER.rst
+++ b/scripts/core/EXP-COMMAND-BUFFER.rst
@@ -57,15 +57,19 @@ returned list of supported extensions.
     ${x}DeviceGetInfo(hDevice, ${X}_DEVICE_INFO_EXTENSIONS, 0, nullptr,
                     &returnedSize);
 
-    // Retrieve extension string 
+    // Retrieve extension string
     std::unique_ptr<char[]> returnedExtensions(new char[returnedSize]);
-    ${x}DeviceGetInfo(hDevice, ${X}_DEVICE_INFO_EXTENSIONS, returnedSize, 
+    ${x}DeviceGetInfo(hDevice, ${X}_DEVICE_INFO_EXTENSIONS, returnedSize,
                       returnedExtensions.get(), nullptr);
-    
+
     std::string_view ExtensionsString(returnedExtensions.get());
-    bool CmdBufferSupport = 
+    bool CmdBufferSupport =
         ExtensionsString.find(${X}_COMMAND_BUFFER_EXTENSION_STRING_EXP)
             != std::string::npos;
+
+.. note::
+    The ${X}_DEVICE_INFO_COMMAND_BUFFER_SUPPORT_EXP device info query exists to
+    serve the same purpose as ${X}_COMMAND_BUFFER_EXTENSION_STRING_EXP.
 
 Command-Buffer Creation
 --------------------------------------------------------------------------------
@@ -73,8 +77,9 @@ Command-Buffer Creation
 Command-Buffers are tied to a specific ${x}_context_handle_t and
 ${x}_device_handle_t. ${x}CommandBufferCreateExp optionally takes a descriptor
 to provide additional properties for how the command-buffer should be
-constructed. There are currently no unique members defined for
-${x}_exp_command_buffer_desc_t, however they may be added in the future.
+constructed. The only unique member defined in ${x}_exp_command_buffer_desc_t
+is ``isUpdatable``, which should be set to ``true`` to support :ref:`updating
+command-buffer commands`.
 
 Command-buffers are reference counted and can be retained and released by
 calling ${x}CommandBufferRetainExp and ${x}CommandBufferReleaseExp respectively.
@@ -88,6 +93,11 @@ enqueue functions in the Core API in terms of their command-specific parameters.
 However, they differ in that they take a command-buffer handle instead of a
 queue handle, and the dependencies and return parameters are sync-points instead
 of event handles.
+
+The entry-point for appending a kernel launch command also returns an optional
+handle to the command being appended. This handle can be used to update the
+command configuration between command-buffer executions, see the section on
+:ref:`updating command-buffer commands`.
 
 Currently only the following commands are supported:
 
@@ -103,9 +113,9 @@ Currently only the following commands are supported:
 * ${x}CommandBufferAppendMemBufferFillExp
 * ${x}CommandBufferAppendUSMPrefetchExp
 * ${x}CommandBufferAppendUSMAdviseExp
-  
+
 It is planned to eventually support any command type from the Core API which can
-actually be appended to the equiavalent adapter native constructs.
+actually be appended to the equivalent adapter native constructs.
 
 Sync-Points
 --------------------------------------------------------------------------------
@@ -122,15 +132,15 @@ were obtained from.
     // Append a memcpy with no sync-point dependencies
     ${x}_exp_command_buffer_sync_point_t syncPoint;
 
-    ${x}CommandBufferAppendUSMMemcpyExp(hCommandBuffer, pDst, pSrc, size, 0, 
+    ${x}CommandBufferAppendUSMMemcpyExp(hCommandBuffer, pDst, pSrc, size, 0,
                                         nullptr, &syncPoint);
-    
+
     // Append a kernel launch with syncPoint as a dependency, ignore returned
     // sync-point
-    ${x}CommandBufferAppendKernelLaunchExp(hCommandBuffer, hKernel, workDim, 
-                                           pGlobalWorkOffset, pGlobalWorkSize, 
-                                           pLocalWorkSize, 1, &syncPoint, 
-                                           nullptr);
+    ${x}CommandBufferAppendKernelLaunchExp(hCommandBuffer, hKernel, workDim,
+                                           pGlobalWorkOffset, pGlobalWorkSize,
+                                           pLocalWorkSize, 1, &syncPoint,
+                                           nullptr, nullptr);
 
 Enqueueing Command-Buffers
 --------------------------------------------------------------------------------
@@ -147,6 +157,82 @@ enqueued or executed simultaneously, and submissions may be serialized.
     ${x}CommandBufferEnqueueExp(hCommandBuffer, hQueue, 0, nullptr,
                               &executionEvent);
 
+Updating Command-Buffer Commands
+--------------------------------------------------------------------------------
+
+An adapter implementing the command-buffer experimental feature can optionally
+support updating the configuration of kernel commands recorded to a
+command-buffer. Support for this is reported by returning true in the
+${X}_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP query.
+
+Updating kernel commands is done by passing the new kernel configuration
+to ${x}CommandBufferUpdateKernelLaunchExp along with the command handle of
+the kernel command to update. Configurations that can be changed are the
+parameters to the kernel and the execution ND-Range.
+
+.. parsed-literal::
+
+    // Create a command-buffer with update enabled.
+    ${x}_exp_command_buffer_desc_t desc {
+      ${X}_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_DESC,
+      nullptr,
+      true // isUpdatable
+    };
+    ${x}_exp_command_buffer_handle_t hCommandBuffer;
+    ${x}CommandBufferCreateExp(hContext, hDevice, &desc, &hCommandBuffer);
+
+    // Append a kernel command which has two buffer parameters, an input
+    // and an output.
+    ${x}_exp_command_buffer_command_handle_t hCommand;
+    ${x}CommandBufferAppendKernelLaunchExp(hCommandBuffer, hKernel, workDim,
+                                           pGlobalWorkOffset, pGlobalWorkSize,
+                                           pLocalWorkSize, 0, nullptr,
+                                           nullptr, &hCommand);
+
+    // Close the command-buffer before updating
+    ${x}CommandBufferFinalizeExp(hCommandBuffer);
+
+    // Define kernel argument at index 0 to be a new input buffer object
+    ${x}_exp_command_buffer_update_memobj_arg_desc_t newInputArg {
+        UR_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_UPDATE_MEMOBJ_ARG_DESC, // stype
+        nullptr, // pNext
+        0, // argIndex
+        nullptr, // pProperties
+        newInputBuffer, // hNewMemObjArg
+    };
+
+    // Define kernel argument at index 1 to be a new output buffer object
+    ${x}_exp_command_buffer_update_memobj_arg_desc_t newOutputArg {
+        UR_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_UPDATE_MEMOBJ_ARG_DESC, // stype
+        nullptr, // pNext
+        1, // argIndex
+        nullptr, // pProperties
+        newOutputBuffer, // hNewMemObjArg
+    };
+
+    // Define the new configuration of the kernel command
+    ${x}_exp_command_buffer_update_memobj_arg_desc_t updatedArgs[2] = {newInputArg, newOutputArg};
+    ${x}_exp_command_buffer_update_kernel_launch_desc_t update {
+        UR_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_UPDATE_KERNEL_LAUNCH_DESC, // stype
+        nullptr, // pNext
+        2, // numNewMemobjArgs
+        0, // numNewPointerArgs
+        0, // numNewValueArgs
+        0, // numNewExecInfos
+        0, // newWorkDim
+        new_args, // pNewMemObjArgList
+        nullptr, // pNewPointerArgList
+        nullptr, // pNewValueArgList
+        nullptr, // pNewExecInfoList
+        nullptr, // pNewGlobalWorkOffset
+        nullptr, // pNewGlobalWorkSize
+        nullptr, // pNewLocalWorkSize
+    };
+
+    // Perform the update
+    ${x}CommandBufferUpdateKernelLaunchExp(hCommand, &update);
+
+
 API
 --------------------------------------------------------------------------------
 
@@ -156,12 +242,21 @@ Macros
 
 Enums
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* ${x}_device_info_t
+    * ${X}_DEVICE_INFO_COMMAND_BUFFER_SUPPORT_EXP
+    * ${X}_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP
 * ${x}_result_t
     * ${X}_RESULT_ERROR_INVALID_COMMAND_BUFFER_EXP
     * ${X}_RESULT_ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_EXP
     * ${X}_RESULT_ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_WAIT_LIST_EXP
+    * ${X}_RESULT_ERROR_INVALID_COMMAND_BUFFER_COMMAND_HANDLE_EXP
 * ${x}_structure_type_t
     * ${X}_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_DESC
+    * ${X}_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_UPDATE_KERNEL_LAUNCH_DESC
+    * ${X}_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_UPDATE_MEMOBJ_ARG_DESC
+    * ${X}_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_UPDATE_POINTER_ARG_DESC
+    * ${X}_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_UPDATE_VALUE_ARG_DESC
+    * ${X}_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_UPDATE_EXEC_INFO_DESC
 * ${x}_command_t
     * ${X}_COMMAND_COMMAND_BUFFER_ENQUEUE_EXP
 * ${x}_function_t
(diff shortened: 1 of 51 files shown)


scripts/parse_specs.py
```diff
@@ -286,8 +286,8 @@ def __validate_members(d, tags):
             if not annotation:
                 raise Exception(prefix+"'desc' must start with {'[in]', '[out]', '[in,out]'}")
 
-            if type_traits.is_handle(item['type']):
-                raise Exception(prefix+"'type' must not be '*_handle_t': %s"%item['type'])
+            #if type_traits.is_handle(item['type']):
+            #    raise Exception(prefix+"'type' must not be '*_handle_t': %s"%item['type'])
```
EwanC, line 290: This is a workaround, need to investigate more why it's needed

scripts/core/EXP-COMMAND-BUFFER.rst
```diff
@@ -143,6 +152,71 @@ enqueued or executed simultaneously, and submissions may be serialized.
     ${x}CommandBufferEnqueueExp(hCommandBuffer, hQueue, 0, nullptr,
                               &executionEvent);
 
+Updating Command-Buffers Commands
+--------------------------------------------------------------------------------
+
+A adapter implementing the command-buffer experimental feature can optionally
```
jandres742, line 158: "An adapter ...'

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -780,3 +781,14 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
...
+
+UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
+    ur_exp_command_buffer_handle_t CommandBuffer, ur_queue_handle_t Queue,
+    uint32_t NumEventsInWaitList, const ur_event_handle_t *EventWaitList,
+    ur_event_handle_t *Event) {
+
+  UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferUpdateKernelLaunchExp(
+      ur_exp_command_buffer_command_handle_t,
+      const ur_exp_command_buffer_update_kernel_launch_desc_t *) {
+    return UR_RESULT_ERROR_INVALID_OPERATION
```
jandres742, line 793: is UNSUPPORTED_FEATURE maybe better?

fabiomestre:

scripts/core/EXP-COMMAND-BUFFER.rst
```diff
@@ -147,6 +157,79 @@ enqueued or executed simultaneously, and submissions may be serialized.
...
 
+Updating Command-Buffers Commands
+--------------------------------------------------------------------------------
+
+An adapter implementing the command-buffer experimental feature can optionally
+support updating the configuration of kernel commands recorded to a
+command-buffer. Support of this is reported by returning true in the
+${X}_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP query.
+
+Updating a kernel commands is done by passing the new kernel configuration
```
reble, line 168:
```suggestion
Updating kernel commands is done by passing the new kernel configuration
```

test/conformance/CMakeLists.txt
```diff
@@ -94,6 +94,7 @@ add_subdirectory(event)
 add_subdirectory(queue)
 add_subdirectory(sampler)
 add_subdirectory(virtual_memory)
+add_subdirectory(exp_command_buffer)
```
aarongreig, line 97: I'd move this down to around [118](https://github.com/oneapi-src/unified-runtime/pull/1089/commits/723c2eaf9fb9a11e8cd933bd922f84e8d016e54e#diff-1595fcdcdd0e6bf22312a2db5cbf98fb89265acbbc4bfa9887845dbe193a291fR118) with the other device code tests, maybe the problem is being caused by the `device_code` subdirectory being added after your test

scripts/parse_specs.py
EwanC: Worked out that because the loader adds a layer of indirection to handles, if we embed a handle in a struct this isn't done. To get this functionality to work I had to hack the following into the loader:
```
diff --git a/source/loader/ur_ldrddi.cpp b/source/loader/ur_ldrddi.cpp
index 4671709a..cb10d094 100644
--- a/source/loader/ur_ldrddi.cpp
+++ b/source/loader/ur_ldrddi.cpp
@@ -6885,6 +6885,19 @@ __urdlllocal ur_result_t UR_APICALL urCommandBufferUpdateKernelLaunchExp(
         reinterpret_cast<ur_exp_command_buffer_command_object_t *>(hCommand)
             ->handle;

+    uint32_t NumMemobjArgs = pUpdateKernelLaunch->numMemobjArgs;
+    if (NumMemobjArgs) {
+        for (uint32_t i = 0; i < NumMemobjArgs; i++) {
+            auto &MemobjArgDesc = pUpdateKernelLaunch->pArgMemobjList[i];
+            ur_mem_handle_t &ArgValue =
+                const_cast<ur_mem_handle_t &>(MemobjArgDesc.hArgValue);
+            ArgValue =
+                (ArgValue)
+                    ? reinterpret_cast<ur_mem_object_t *>(ArgValue)->handle
+                    : nullptr;
+        }
+    }
+
     // forward to device-platform
     result = pfnUpdateKernelLaunchExp(hCommand, pUpdateKernelLaunch);
```

Need to come up with a solution where this is generated by the scripts

scripts/core/exp-command-buffer.yml
```diff
@@ -54,7 +87,120 @@ type: struct
...
+      desc: "[in] Length of pArgPointerList."
+    - type: uint32_t
+      name: numValueArgs
+      desc: "[in] Length of pArgValueList."
+    - type: uint32_t
+      name: numExecInfos
+      desc: "[in] Length of pExecInfoList."
+    - type: uint32_t
+      name: workDim
+      desc: "[in][optional] Number of work dimensions in the kernel ND-range, from 1-3."
```
aarongreig, line 182: I don't think [optional] makes sense for something that's passed by value

scripts/core/exp-command-buffer.yml
```diff
@@ -54,7 +87,120 @@ type: struct
...
+    - type: "const $x_exp_command_buffer_update_memobj_arg_desc_t*"
+      name: pArgMemobjList
+      desc: "[in][optional] An array describing the new kernel mem obj arguments for the command."
```
aarongreig, line 185: can you add range to all of these so the scripts know they're arrays, e.g.
```suggestion
      desc: "[in][optional][range(0, numMemobjArgs)] An array describing the new kernel mem obj arguments for the command."
```

scripts/parse_specs.py
aarongreig: the changes in https://github.com/oneapi-src/unified-runtime/pull/1211 should allow this to be generated

scripts/core/exp-command-buffer.yml
```diff
@@ -163,7 +314,11 @@ params:
       desc: "[in][optional] A list of sync points that this command depends on."
     - type: "$x_exp_command_buffer_sync_point_t*"
       name: pSyncPoint
-      desc: "[out][optional] sync point associated with this command"
+      desc: "[out][optional] Sync point associated with this command."
+    - type: "$x_exp_command_buffer_command_handle_t*"
+      name: phCommand
+      desc: "[out][optional] Handle to this command."
```
Bensuo, line 320: Is there a danger here that users may think they have ownership over this handle? Since we're differing from the normal Create/Retain/Release methods of retrieving other handle types.

Perhaps a regular pointer type like with the sync point would prevent any confusion instead of calling it a handle.

EwanC: That's a fair point, calling a type a handle that isn't being created/retained/release could be confusing. Will remove the handler terminology from this type and it's usage

EwanC: After discussion I've added retain/release entry-points to command-handles. This makes them analogous to events in that they are created by being returned from command adding entry-points and reference counted. The relationship between command handle objects and command-buffer objects is the same as kernel/programs in that creating a command handle increments the reference count of the command-buffer.

I've updated the spec with the above, and done a quick impl but for the CUDA adapter but need to write a CTS tests or two to exercise this.

EwanC:
Marking this ready for review as I would like some code owner feedback and believe the remaining CI fails will be resolved by #1211

source/adapters/hip/command_buffer.cpp
```diff
@@ -162,3 +163,41 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
                   "implemented for HIP adapter.");
   return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
 }
+
+UR_APIEXPORT ur_result_t UR_APICALL
+urCommandBufferRetainCommandExp(ur_exp_command_buffer_command_handle_t) {
+  detail::ur::die("Experimental Command-buffer feature is not "
```
martygrant, line 169: Hi Ewan just confirming as per https://github.com/oneapi-src/unified-runtime/pull/1089 if you could remove the `die` calls please, thank you!

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -986,3 +987,41 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
...
+    ur_exp_command_buffer_handle_t hCommandBuffer,
+    ur_exp_command_buffer_info_t propName, size_t propSize, void *pPropValue,
+    size_t *pPropSizeRet) {
+  UrReturnHelper ReturnValue(propSize, pPropValue, pPropSizeRet);
+
+  switch (propName) {
+  case UR_EXP_COMMAND_BUFFER_INFO_REFERENCE_COUNT:
+    return ReturnValue(uint32_t{hCommandBuffer->RefCount.load()});
+  default:
+    die("Command-buffer info request not implemented");
```
martygrant, line 1017: Another `die` to be be removed in here.

source/adapters/native_cpu/command_buffer.cpp
```diff
@@ -133,3 +134,41 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
                   "implemented for the NativeCPU adapter.");
   return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
 }
+
+UR_APIEXPORT ur_result_t UR_APICALL
+urCommandBufferRetainCommandExp(ur_exp_command_buffer_command_handle_t) {
+  detail::ur::die("Experimental Command-buffer feature is not "
```
martygrant, line 140: `die` calls to be removed here too.

source/adapters/opencl/command_buffer.cpp
```diff
@@ -356,3 +357,75 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
 
   return UR_RESULT_SUCCESS;
 }
+
+UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferRetainCommandExp(
+    [[maybe_unused]] ur_exp_command_buffer_command_handle_t hCommand) {
+  cl_adapter::die("Experimental Command-buffer feature is not "
```
martygrant, line 363: `die` calls to be removed here too.

martygrant (changes requested):
LGTM! Just the `die` calls to remove please, thanks!

scripts/core/exp-command-buffer.yml
```diff
@@ -44,6 +77,33 @@ etors:
...
+class: $xCommandBuffer
+name: $x_exp_command_buffer_command_info_t
+typed_etors: True
+etors:
+    - name: REFERENCE_COUNT
+      desc: |
+            [uint32_t] Reference count of the command-buffer object.
+            The reference count returned should be considered immediately stale.
+            It is unsuitable for general use in applications. This feature is provided for identifying memory leaks.
+
```
aarongreig, line 105: nit: I don't think the newlines between the end of the definition and the separator is common elsewhere in the scripts

test/conformance/exp_command_buffer/buffer_fill_kernel_update.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+// Test updating the global size so that the fill outputs to a larger buffer
+TEST_P(BufferFillCommandTest, UpdateGlobalSize) {
+    ASSERT_SUCCESS(urCommandBufferEnqueueExp(updatable_cmd_buf_handle, queue, 0,
+                                             nullptr, nullptr));
+    ASSERT_SUCCESS(urQueueFinish(queue));
+    ValidateBuffer(buffer, sizeof(val) * global_size, val);
+
+    size_t new_global_size = 64;
+    const size_t buffer_size = sizeof(val) * new_global_size;
```
aarongreig, line 131: can we make this a new variable name? it looks like this is shadowing an existing declaration from the [fixture](https://github.com/oneapi-src/unified-runtime/pull/1089/files#diff-9fd0d900f6d64a39d5d438eec69265eeadd6d5c62e5d5815e4f259c920e6d8e3R49) (also applies to the `SeparateUpdateCalls` test)

source/adapters/cuda/command_buffer.hpp
```diff
@@ -175,6 +175,85 @@ static inline const char *getUrResultString(ur_result_t Result) {
...
+      const size_t ZeroSize = sizeof(size_t) * (3 - WorkDim);
+      std::memset(LocalWorkSize + WorkDim, 0, ZeroSize);
+    }
+  }
+
+  // Used when retaining an object.
+  uint32_t incrementReferenceCount() noexcept { return ++RefCount; }
+  // Used when releasing an object.
+  uint32_t decrementReferenceCount() noexcept { return --RefCount; }
+  uint32_t getReferenceCount() const noexcept { return RefCount; }
```
GeorgeWeb, line 241: Just a nit here but prefer to use consistent naming convention, i.e. all of the class' functions and variable members starting with upper case following PascalCase style.

GeorgeWeb (approved):

source/adapters/cuda/command_buffer.hpp
aarongreig: this is correct for llvm style, variables are Pascal and functions are camel

hdelan: Some of the above functions are Pascal

GeorgeWeb: Yeah that's fine with me. The nit comment was rather me pointing out the inconsistency between the function members of the struct where the setters are _Pascal_ style while those getters are _camel_ style. So long as all the member functions follow the correct naming convention, in this case being that from the llvm style, that'll be grand. Thanks.

aarongreig: yep didn't clock the other functions, always get more context than what you see in the email :facepalm:

aarongreig (approved):
CTS changes LGTM, this'll probably be the most rigorously tested feature in the suite

source/adapters/cuda/command_buffer.hpp
EwanC: Good spot, I've updated the other functions defined in this file to llvm style

scripts/core/EXP-COMMAND-BUFFER.rst
```diff
@@ -89,6 +94,12 @@ However, they differ in that they take a command-buffer handle instead of a
 queue handle, and the dependencies and return parameters are sync-points instead
 of event handles.
 
+The entry-point for appending a kernel launch command also returns an optional
+handle to the command being appended. Returning this handle extends the
+lifetime of the parent command-buffer by incrementing its reference count, and
+the parent command-buffer's reference count is decremented again when the command
+handle is destroyed.
```
Bensuo, line 101: To me this seems a bit counterintuitive, I would've thought that the command-buffer would extend the lifetime of the command handles in case it needed them internally. They can't both extend each other's lifetimes, and I'm not sure this matters that much, but was there a specific reason this way was chosen?

source/adapters/cuda/command_buffer.cpp
```diff
@@ -762,3 +772,174 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
...
+    ur_exp_command_buffer_handle_t hCommandBuffer,
+    ur_exp_command_buffer_info_t propName, size_t propSize, void *pPropValue,
+    size_t *pPropSizeRet) {
+  UrReturnHelper ReturnValue(propSize, pPropValue, pPropSizeRet);
+
+  switch (propName) {
+  case UR_EXP_COMMAND_BUFFER_INFO_REFERENCE_COUNT:
+    return ReturnValue(hCommandBuffer->getReferenceCount());
+  default:
+    detail::ur::die("Command-buffer info request not implemented");
```
martygrant, line 925: Another `die` call here to be removed.

source/adapters/cuda/command_buffer.cpp
```diff
@@ -762,3 +772,174 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
...
+    ur_exp_command_buffer_command_handle_t hCommand,
+    ur_exp_command_buffer_command_info_t propName, size_t propSize,
+    void *pPropValue, size_t *pPropSizeRet) {
+  UrReturnHelper ReturnValue(propSize, pPropValue, pPropSizeRet);
+
+  switch (propName) {
+  case UR_EXP_COMMAND_BUFFER_COMMAND_INFO_REFERENCE_COUNT:
+    return ReturnValue(hCommand->getReferenceCount());
+  default:
+    detail::ur::die("Command-buffer command info request not implemented");
```
martygrant, line 941: Another `die` call here to be removed.

martygrant (approved):
LGTM just two more `die` calls to be removed.

scripts/core/EXP-COMMAND-BUFFER.rst
```diff
@@ -147,6 +158,82 @@ enqueued or executed simultaneously, and submissions may be serialized.
...
+    // Append a kernel command which has two buffer parameters, an input
+    // and an output.
+    ${x}_exp_command_buffer_command_handle_t hCommand;
+    ${x}CommandBufferAppendKernelLaunchExp(hCommandBuffer, hKernel, workDim,
+                                           pGlobalWorkOffset, pGlobalWorkSize,
+                                           pLocalWorkSize, 0, nullptr,
+                                           nullptr, &hCommand);
+
+    // Close the command-buffer before updating
+    ${x}CommandBufferFinalizeExp(hCommandBuffer);
```
Bensuo, line 194: This implies to me that finalizing is required before updating, and if that is the intention (it is not reflected in errors returned in the YAML at least), do we need to mandate that the command buffer be finalized before updating? I think it is possible to update a non-executable CUDA Graph as well as an executable one at least.

scripts/core/EXP-COMMAND-BUFFER.rst
```diff
@@ -227,10 +336,13 @@ Changelog
 | 1.3       | Add function definitions for Prefetch and Advise      |
 |           | commands                                              |
 +-----------+-------------------------------------------------------+
+| 1.4       | A function definitions for kernel command update      |
```
Bensuo, line 339:
```suggestion
| 1.4       | Add function definitions for kernel command update      |
```

scripts/core/exp-command-buffer.yml
```diff
@@ -704,3 +879,123 @@ returns:
...
+        - "If `propSize` is less than the real number of bytes needed to return the info."
+    - $X_RESULT_ERROR_INVALID_NULL_POINTER:
+        - "`propSize != 0 && pPropValue == NULL`"
+        - "`pPropValue == NULL && pPropSizeRet == NULL`"
+    - $X_RESULT_ERROR_INVALID_COMMAND_BUFFER_EXP
+    - $X_RESULT_ERROR_OUT_OF_RESOURCES
+    - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Get command-buffer jbject information"
```
Bensuo, line 971:
```suggestion
desc: "Get command-buffer object information"
```

scripts/core/EXP-COMMAND-BUFFER.rst
EwanC: Good question, in [clUpdateMutableCommandsKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#mutable-commands) for the OpenCL backend there is the error:
> CL_INVALID_OPERATION if command_buffer has not been finalized.

So I think for version 1 of this feature we should go with that restriction, then we could look into relaxing it if we need to. I'll update the docs to reflect this.

scripts/core/exp-command-buffer.yml
```diff
@@ -54,7 +109,118 @@ type: struct
...
+name: $x_exp_command_buffer_update_memobj_arg_desc_t
+members:
+    - type: uint32_t
+      name: argIndex
+      desc: "[in] Argument index."
+    - type: "const ur_kernel_arg_mem_obj_properties_t *"
+      name: pProperties
+      desc: "[in][optinal] Pointer to memory object properties."
+    - type: $x_mem_handle_t
+      name: hArgValue
```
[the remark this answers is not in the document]
EwanC: I've copied the names of the associated parameters in the UR set kernel arg functions https://oneapi-src.github.io/unified-runtime/core/api.html#kernel

But no reason we can't use different ones if that's more helpful. Will update.

scripts/core/EXP-COMMAND-BUFFER.rst
EwanC: There's nothing really specified about reference counting in the UR spec, so this behaviour is based on what adapter implementations are doing for kernel/program & event refcounting. Before I was looking at the CUDA adapter which doesn't do anything fancy with refcounting, but now looking into Level Zero it has internal & external event refcounting implemented (presumably for this reason).

I'll try use that as inspiration to do something similar for command handles, because I agree it makes sense.

EwanC: I've pushed a commit which implements command-buffers and command-handles with an internal reference count so that they can extend each others lifetimes. I did however also remove this spec wording, since this is turning into more of an implementation detail and reference counting isn't really talked about elsewhere.

EwanC:
ping @oneapi-src/unified-runtime-native-cpu-write & @oneapi-src/unified-runtime-level-zero-write for review

PietroGhg (approved):
Native CPU lgtm, thank you

Bensuo (approved):

EwanC:
ping @oneapi-src/unified-runtime-level-zero-write for review, although these entry-points are stubs in the L0 adapter now, they should be implementable by https://github.com/oneapi-src/level-zero-spec/pull/264 in the near future.

EwanC:
> is there an LLVM PR for this?

There wasn't because I added the CTS tests to verify the functionality rather than using the DPC++ tests, but I whipped one up quickly to verify there are no regressions https://github.com/intel/llvm/pull/12645

PR #1093, opened 2023-11-17
Add bounds checking for enqueue operations to the validation layer.

aarongreig:
This is accomplished with the various size queries for buffers, images and USM allocations. Since not all adapters have these queries implemented the bounds checking isn't entirely comprehensive on all platforms just yet.

Aims to address issues discussed in #926

### DIFF
--- a/scripts/templates/helper.py
+++ b/scripts/templates/helper.py
@@ -356,6 +356,7 @@ class param_traits:
     RE_RELEASE  = r".*\[release\].*"
     RE_TYPENAME = r".*\[typename\((.+),\s(.+)\)\].*"
     RE_TAGGED   = r".*\[tagged_by\((.+)\)].*"
+    RE_BOUNDS   = r".*\[bounds\((.+),\s*(.+)\)].*"
 
     @classmethod
     def is_mbz(cls, item):
@@ -412,6 +413,13 @@ class param_traits:
             return True if re.match(cls.RE_TAGGED, item['desc']) else False
         except:
             return False
+
+    @classmethod
+    def is_bounds(cls, item):
+        try:
+            return True if re.match(cls.RE_BOUNDS, item['desc']) else False
+        except:
+            return False
     
     @classmethod
     def tagged_member(cls, item):
@@ -452,6 +460,22 @@ class param_traits:
     @classmethod
     def typename_size(cls, item):
         match = re.match(cls.RE_TYPENAME, item['desc'])
+        if match:
+            return match.group(2)
+        else:
+            return None
+
+    @classmethod
+    def bounds_offset(cls, item):
+        match = re.match(cls.RE_BOUNDS, item['desc'])
+        if match:
+            return match.group(1)
+        else:
+            return None
+
+    @classmethod
+    def bounds_size(cls, item):
+        match = re.match(cls.RE_BOUNDS, item['desc'])
         if match:
             return match.group(2)
         else:
@@ -1041,7 +1065,35 @@ def make_pfncb_param_type(namespace, tags, obj):
 
 """
 Public:
-    returns a dict of auto-generated c++ parameter validation checks
+    returns an appropriate bounds helper function call for an entry point
+    parameter with the [bounds] tag
+"""
+def get_bounds_check(param, bounds_error):
+    # Images need their own helper, since function signature wise they would be
+    # identical to buffer rect
+    bounds_function = 'boundsImage' if 'image' in param['name'].lower() else 'bounds'
+    bounds_check = "auto {0} = {1}({2}, {3}, {4})".format(
+        bounds_error,
+        bounds_function,
+        param["name"],
+        param_traits.bounds_offset(param),
+        param_traits.bounds_size(param),
+    )
+    bounds_check += '; {0} != UR_RESULT_SUCCESS'.format(bounds_error)
+
+    # USM bounds checks need the queue handle parameter to be able to use the
+    # GetMemAllocInfo entry point
+    if type_traits.is_pointer(param['type']):
+        # If no `hQueue` parameter exists that should have been caught at spec
+        # generation.
+        return re.sub(r'bounds\(', 'bounds(hQueue, ', bounds_check)
+
+    return bounds_check
+
+"""
+Public:
+    returns a dict of auto-generated c++ parameter validation checks for the
+    given function (specified by `obj`)
 """
 def make_param_checks(namespace, tags, obj, cpp=False, meta=None):
     checks = {}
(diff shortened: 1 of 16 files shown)


aarongreig:
@pbalcer I'd be interested if you had any feedback about this approach. The implementation does work but it isn't very elegant. I especially hate this https://github.com/oneapi-src/unified-runtime/pull/1093/files#diff-6ccceeaee484ce2a4014958578efbcd7732574d1bf5002a94ed76d992d5367adR1309 but I'm struggling to come up with a better alternative.

aarongreig:
The other major issue with this is signposting it so that if anyone ever adds a new enqueue operation that could utilize bounds checking it's obvious that they should come and update this mechanism to account for it.

aarongreig:
> pbalcer I'd be interested if you had any feedback about this approach. The implementation does work but it isn't very elegant. I especially hate this https://github.com/oneapi-src/unified-runtime/pull/1093/files#diff-6ccceeaee484ce2a4014958578efbcd7732574d1bf5002a94ed76d992d5367adR1309 but I'm struggling to come up with a better alternative.


What's there now is marginally nicer but I'm still not thrilled about it.

pbalcer:
Maybe move the list of checks to the spec? So for memcpy, it could look like this:
```
type: function
desc: "Enqueue a command to copy USM memory" 
...
    - type: void*
      name: pDst
      desc: "[in][bounds_check(size)] pointer to the destination USM memory object"
    - type: "const void*"
      name: pSrc
      desc: "[in][bounds_check(size)] pointer to the source USM memory object"
    - type: size_t
      name: size
      desc: "[in] size in bytes to be copied"
...
```

Based on that, you can figure out in the validation layer that this is a pointer (so needs to use usm bounds check), and iterate over the rest of arguments to retrieve the context/queue or whatever other object might be needed.

Alternatively, it might also make sense to align to how the rest of the validation layer is defined and put this information in the returns section:
```
returns:      
    - $X_RESULT_ERROR_INVALID_EVENT
        - "`bounds_check_usm(pDst, size)`"
        - "`bounds_check_usm(pSrc, size)`"
        - "`size == 0`"
        - "If `size` is higher than the allocation size of `pSrc` or `pDst`"
    - $X_RESULT_ERROR_INVALID_EVENT_WAIT_LIST:
        - "`phEventWaitList == NULL && numEventsInWaitList > 0`"
        - "`phEventWaitList != NULL && numEventsInWaitList == 0`"
        - "If event objects in phEventWaitList are not valid events."
    - $X_RESULT_ERROR_INVALID_MEM_OBJECT
```

This is more explicit, but also moves the definition of constrains on a type further away from the type definition itself. But both solutions have merit.

pbalcer:
If you go with the first approach, another thought I just had is that all the checks have similar arguments: `handle/pointer, offset, size` (with offset/size potentially having more than 1 dimension). Maybe we should enforce this interface, and so the usm check would become:
```
    - type: void*
      name: pDst
      desc: "[in][bounds_check(0, size)] pointer to the destination USM memory object"
```

Since we are checking if the pDst fits inside of the 0<>size bounds. Having a consistent interface would make it easier to use in other parts of the spec.

aarongreig:
I've implemented the parameter tagging approach, it makes it nice and visible for anyone looking to implement a new entry point with bounds checking and it should generalize to stuff outside enqueue without script modification (up to a point). Thanks for the suggestions. Just looking to see if there's a nicer way to propagate the errors from the GetInfo calls we'll be doing and this should be good to go

scripts/templates/helper.py
```diff
@@ -1026,6 +1058,15 @@ def make_param_checks(namespace, tags, obj, cpp=False, meta=None):
                     if key not in checks:
                         checks[key] = []
                     checks[key].append(subt(namespace, tags, code.group(1), False, cpp))
+
+    for p in obj.get('params', []):
+        if 'desc' not in p or 'name' not in p or 'type' not in p:
+            break
+        if '[bounds' in p['desc']:
```
pbalcer, line 1065: add bounds related functions to param_traits class.

scripts/templates/helper.py
```diff
@@ -1013,7 +1013,39 @@ def make_pfncb_param_type(namespace, tags, obj):
...
+    )
+    bounds_call += '; {0} != UR_RESULT_SUCCESS'.format(bounds_error)
+    # Image needs its own helper, since function signature wise it's identical
+    # to buffer rect
+    if'image' in param_dict['name'].lower():
+        return re.sub(r'bounds\(', 'boundsImage(', bounds_call)
+
+    # USM bounds checks need the queue handle parameter to be able to use the
+    # GetMemAllocInfo entry point
+    if'void' in param_dict['type']:
```
pbalcer, line 1034: there's a `is_pointer` type trait.

scripts/templates/helper.py
```diff
@@ -1013,7 +1013,39 @@ def make_pfncb_param_type(namespace, tags, obj):
...
+        # TODO: this could be more complex, we could pull the full function
```
pbalcer, line 1035: makes sense, but, if possible, it would be nice to have a check somewhere (maybe during param validation) with that produces a nice error message if bounds is used where `hQueue` isn't defined.

pbalcer (commented):
please add documentation to `YaML.md`.

source/loader/layers/validation/ur_validation_layer.cpp
```diff
@@ -20,4 +22,125 @@ context_t::context_t() : logger(logger::create_logger("validation")) {}
...
 context_t::~context_t() {}
 
+// Some adapters don't support all the queries yet, we should be lenient and
+// just not attempt to validate in those cases to preserve functionality.
+#define RETURN_ON_FAILURE(result)                                              \
+    if (result == UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION ||                   \
+        result == UR_RESULT_ERROR_UNSUPPORTED_FEATURE)                         \
+        return UR_RESULT_SUCCESS;                                              \
+    if (result != UR_RESULT_SUCCESS) {                                         \
+        assert(result == UR_RESULT_SUCCESS);                                   \
```
aarongreig, line 32: I added this assert because I'm worried that error codes coming out of the GetInfo calls might make their way back to an application and be confusing. This way a debug build will quickly point out what's going wrong.

pbalcer: sounds reasonable. maybe add `logger.error("unexpected return value {} from ...", result)`?

pbalcer (approved):
this is much neater than the first version.

jandres742 (approved):

aarongreig:
@oneapi-src/unified-runtime-hip-write by a stroke of luck this change fixes a CI breakage we've had since friday, could I ask somebody to take a look?

PR #1096, opened 2023-11-20
[docs] Generate Print C++ API html docs

PatKamin:
This PR adds html docs generation for Printing C++ API.

[preview](https://patkamin.github.io/unified-runtime/core/api.html#id628)

### DIFF
--- a/scripts/templates/print_helper.py
+++ b/scripts/templates/print_helper.py
@@ -109,7 +109,19 @@ class Func:
 
     @property
     def cpp_name(self) -> str:
-        return ""
+        return "operator<<"
+
+    @property
+    def cpp_args(self) -> str:
+        attribute = (
+            "[[maybe_unused]]" if re.match("const struct", self.print_arg.type) else ""
+        )
+        return str(
+            [
+                Arg("std::ostream &", "os"),
+                Arg(" ".join([attribute, self.print_arg.type]), self.print_arg.name),
+            ]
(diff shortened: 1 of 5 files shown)


scripts/templates/print_helper.py
```diff
@@ -5,103 +5,294 @@
...
+    split_arg = re.split(regex, arg_str)
+    arg = Arg(split_arg[0].strip(), split_arg[1])
+    return arg
+
+
+def _add_null_buffer_comment(func_comment: List[str]) -> List[str]:
+    comment_added = False
+    for line in func_comment:
+        if re.search("RESULT_ERROR_INVALID_NULL_POINTER", line):
+            func_comment.append("///         - `NULL == buffer`\n")
```
kswiecicki, line 199: nit: If `RESULT_ERROR_INVALID_NULL_POINTER` is not the last error label, you could append the `NULL == buffer` condition to the wrong error label.

PatKamin:
> nit: You could add some briefs similar to `helper.py`. 

Done

PatKamin:
> nit: If `RESULT_ERROR_INVALID_NULL_POINTER` is not the last error label, you could append the `NULL == buffer` condition to the wrong error label. 

Why? I iterate through all the lines in the comment, so it should work regardless where the error label is.

kswiecicki (approved):

pbalcer (approved):

PR #1099, opened 2023-11-21
[UR][L0] Unify use of large allocation in L0 adapter

jandres742:
Intel(R) GPUs have two modes of operation in terms of allocations:
Stateful and stateless mode.

Stateful optimizes memory accesses through pointer arithmetic.
This can be done as long as allocations used by the allocation
are smaller than 4GB.

Stateless disables such pointer-arithmetic optimization to
allow the kernel to use allocations larger than 4GB.

Currently, L0 adapter dynamically and automatically requests
the L0 driver large allocations if it detects an allocation size
is larger than 4GB. This creates a problem if a kernel has been
previously compiled for stateful access. This ultimately means
the adapter mixes stateful and stateless behavior, which is not
a user-friendly experience.

This patch aims at correcting this behavior by defining a default
one. On Intel(R) GPUs previous to Intel(R) Data Center GPU Max,
default behavior is now stateless, meaning all allocations are
only allowed by default. Users can opt-in for stateful mode setting
a new environment variable UR_L0_USE_OPTIMIZED_32BIT_ACCESS=1.

Addresses:
https://stackoverflow.com/questions/75621264/sycl-dot-product-code-gives-wrong-results

intel/llvm testing: https://github.com/intel/llvm/pull/11958

### DIFF
--- a/source/adapters/level_zero/device.cpp
+++ b/source/adapters/level_zero/device.cpp
@@ -88,6 +88,24 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGet(
   return UR_RESULT_SUCCESS;
 }
 
+uint64_t calculateGlobalMemSize(ur_device_handle_t Device) {
+  // Cache GlobalMemSize
+  Device->ZeGlobalMemSize.Compute =
+      [Device](struct ze_global_memsize &GlobalMemSize) {
+        for (const auto &ZeDeviceMemoryExtProperty :
+             Device->ZeDeviceMemoryProperties->second) {
+          GlobalMemSize.value += ZeDeviceMemoryExtProperty.physicalSize;
+        }
+        if (GlobalMemSize.value == 0) {
+          for (const auto &ZeDeviceMemoryProperty :
+               Device->ZeDeviceMemoryProperties->first) {
+            GlobalMemSize.value += ZeDeviceMemoryProperty.totalSize;
+          }
+        }
+      };
+  return Device->ZeGlobalMemSize.operator->()->value;
+}
+
 UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
     ur_device_handle_t Device,  ///< [in] handle of the device instance
     ur_device_info_t ParamName, ///< [in] type of the info to retrieve
@@ -251,20 +269,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
   case UR_DEVICE_INFO_MAX_MEM_ALLOC_SIZE:
     return ReturnValue(uint64_t{Device->ZeDeviceProperties->maxMemAllocSize});
   case UR_DEVICE_INFO_GLOBAL_MEM_SIZE: {
-    uint64_t GlobalMemSize = 0;
     // Support to read physicalSize depends on kernel,
     // so fallback into reading totalSize if physicalSize
     // is not available.
-    for (const auto &ZeDeviceMemoryExtProperty :
-         Device->ZeDeviceMemoryProperties->second) {
-      GlobalMemSize += ZeDeviceMemoryExtProperty.physicalSize;
-    }
-    if (GlobalMemSize == 0) {
-      for (const auto &ZeDeviceMemoryProperty :
-           Device->ZeDeviceMemoryProperties->first) {
-        GlobalMemSize += ZeDeviceMemoryProperty.totalSize;
-      }
-    }
+    uint64_t GlobalMemSize = calculateGlobalMemSize(Device);
     return ReturnValue(uint64_t{GlobalMemSize});
   }
   case UR_DEVICE_INFO_LOCAL_MEM_SIZE:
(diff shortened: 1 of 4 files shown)


jandres742:
@smaslov-intel : please review

source/adapters/level_zero/device.cpp
```diff
@@ -900,6 +910,17 @@ ur_device_handle_t_::useImmediateCommandLists() {
   }
 }
 
+bool ur_device_handle_t_::useLargeAllocations() {
+  static const bool UseLargeAllocations = [this] {
+    const char *UrRet = std::getenv("UR_L0_ALLOW_LARGE_ALLOCATIONS");
+    if (!UrRet)
+      return (this->isPVC() ? true : false);
```
MichalMrozek, line 917: why this isPVC check ? 
It is not required on PVC.
PVC Level Zero driver does it by default.

source/adapters/level_zero/device.hpp
```diff
@@ -141,6 +141,17 @@ struct ur_device_handle_t_ : _ur_object {
   // Returns whether immediate command lists are used on this device.
   ImmCmdlistMode ImmCommandListUsed{};
 
+  // Returns whether large allocations are being used or not.
+  // On some Intel GPUs, this influences how kernels are compiled.
+  // If large allocations (>4GB) are requested, then kernels are
+  // compiled with stateless access.
+  // If small allocations (<4GB) are requested, then kernels are
```
MichalMrozek, line 148: This is not accurate, even if -ze-opt-greater-than-4GB-buffer-required is not specified, kernels may still be compiled in stateless mode.

source/adapters/level_zero/device.hpp
```diff
@@ -141,6 +141,17 @@ struct ur_device_handle_t_ : _ur_object {
...
+  // compiled with stateful access, with potential performance
+  // improvements.
+  // Some GPUs support only one mode, such us Intel(R) Data Center GPU Max,
```
MichalMrozek, line 151: Intel(R) Data Center GPU Max supports both stateful and stateless modes.
Level Zero implementation for Intel(R) Data Center GPU Max allows only stateless mode for this device.

source/adapters/level_zero/device.cpp
jandres742: @MichalMrozek : this is so the rest of the adapter has  "large allocation behavior".

source/adapters/level_zero/device.hpp
jandres742: I will fix comment indicating is the driver that has only that uspport.

source/adapters/level_zero/device.hpp
jandres742: thanks, I will remove the comment

jandres742:
@nrspruit: please review.

source/adapters/level_zero/device.cpp
MichalMrozek: But PVC is already in large allocation behavior by default.
There is no need to do any special for this device.
There is no need to add compiler option and it is quite dangerous to bypass max memory allocation size limits by default.
By using those variables it indicates that PVC requires some special handling for large allocations which in fact is not needed.

jandres742: thanks @MichalMrozek . I have changed code to use defaults on PVC.

source/adapters/level_zero/program.cpp
```diff
@@ -234,8 +249,17 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramCompile(
...
   // can use them later.
-  if (Options)
+  if (Options) {
     Program->BuildFlags = Options;
+
+    // if large allocations are selected, then pass
+    // ze-opt-greater-than-4GB-buffer-required to disable
+    // stateful optimizations and be able to use larger than
+    // 4GB allocations on these kernels.
+    if (Context->Devices[0]->useOptimized32bitAccess() == 0) {
```
nrspruit, line 259: Should this be in the Exp function and have the Exp function have the Compile code now? That way you check the specific device being passed in and not just the Context->Devices[0] since you might be on a non-uniform system.

jandres742: thanks @nrspruit . Good idea. However, `urProgramCompileExp` at this moment is unimplemented, so adding implementation for `urProgramCompileExp` on top of these changes would make this PR too big. I think it is better we merge this patch, then we add the support for `urProgramCompileExp`, including using the functionality from this patch. what do you think?

nrspruit: Sure, that would be fine, a follow-up patch would be good improvement on this.

nrspruit (approved):
+1 from me

jandres742:

fabiomestre:

kbenzie:
I'm going to create a combined intel/llvm PR which will include this and a few other PR's so we can get things merged quicker.

PR #1100, opened 2023-11-21
[CUDA][HIP][OpenCL][NATIVECPU] Fix multi-device compile

kbenzie:
Ensure that all adapters have the correct signatures for the multi-device compile experimental feature entry points and that they entry points exist even when returning `UR_RESULT_ERROR_UNSUPPORTED_FEATURE`.

Fixes the failing checks in https://github.com/intel/llvm/pull/11464.

### DIFF
--- a/source/adapters/native_cpu/program.cpp
+++ b/source/adapters/native_cpu/program.cpp
@@ -87,6 +87,26 @@ urProgramLink(ur_context_handle_t hContext, uint32_t count,
   DIE_NO_IMPLEMENTATION
 }
 
+UR_APIEXPORT ur_result_t UR_APICALL urProgramCompileExp(ur_program_handle_t,
+                                                        uint32_t,
+                                                        ur_device_handle_t *,
+                                                        const char *) {
+  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+}
+
+UR_APIEXPORT ur_result_t UR_APICALL urProgramBuildExp(ur_program_handle_t,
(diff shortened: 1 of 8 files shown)


source/adapters/native_cpu/program.cpp
```diff
@@ -87,6 +87,26 @@ urProgramLink(ur_context_handle_t hContext, uint32_t count,
   DIE_NO_IMPLEMENTATION
 }
 
+UR_APIEXPORT ur_result_t UR_APICALL urProgramCompileExp(ur_program_handle_t,
```
callumfare, line 90: These new entry points should be added to the DDI tables in `ur_interface_loader.cpp`

source/adapters/opencl/program.cpp
```diff
@@ -220,6 +220,26 @@ urProgramLink(ur_context_handle_t hContext, uint32_t count,
   return UR_RESULT_SUCCESS;
 }
 
+UR_APIEXPORT ur_result_t UR_APICALL urProgramCompileExp(ur_program_handle_t,
```
callumfare, line 223: These should also be added to `ur_interface_loader.cpp`

callumfare (approved):
LGTM,

source/adapters/opencl/program.cpp
kbenzie: Done.

PietroGhg (approved):
Native CPU LGTM, thank you

PR #1106, opened 2023-11-22
[NATIVECPU] Fix build errors when -Werror=on

wenju-he:


wenju-he:
@uwedolinsky @PietroGhg could you please review?

source/adapters/native_cpu/kernel.cpp
```diff
@@ -171,6 +172,10 @@ urKernelGetSubGroupInfo(ur_kernel_handle_t hKernel, ur_device_handle_t hDevice,
     // todo: set proper values
     return ReturnValue(0);
   }
+  case UR_KERNEL_SUB_GROUP_INFO_FORCE_UINT32: {
```
pbalcer, line 175: this is not really a real value.
`default: {
} break;`
?

kbenzie: This should be an unreachable case so I agree with the second suggestion.

wenju-he: done

pbalcer: could be useful to have something like unreachable in common, but otherwise lgtm.

kbenzie: Would `std::abort()` suffice?

kbenzie (approved):

PR #1108, opened 2023-11-22
[UR] Fix urAdapterGet ignores NumEntries

AllanZyne:


pbalcer (approved):

source/loader/ur_ldrddi.cpp
```diff
@@ -100,6 +100,9 @@ __urdlllocal ur_result_t UR_APICALL urAdapterGet(
                 break;
             }
             adapterIndex++;
+            if (adapterIndex == NumEntries) {
+                break;
+            }
```
aarongreig, line 105: this should be added to the ldrddi template file here https://github.com/oneapi-src/unified-runtime/blob/main/scripts/templates/ldrddi.cpp.mako#L73 with the change propagated to this file by running the `generate` target (same goes for #1109)

AllanZyne: > this should be added to the ldrddi template file here https://github.com/oneapi-src/unified-runtime/blob/main/scripts/templates/ldrddi.cpp.mako#L73 with the change propagated to this file by running the `generate` target (same goes for #1109)

Hi @aarongreig , I found after running the `generate` target, the format of the whole file had been changed.
So I just pushed what I had modified here.

pbalcer (approved):

source/loader/ur_ldrddi.cpp
pbalcer: That might indicate you have an incorrect version of clang-format. See README.md for instructions on how to fix that.

AllanZyne: > That might indicate you have an incorrect version of clang-format. See README.md for instructions on how to fix that.

Done, I found that the generated files are the same as what I pushed here.

pbalcer (approved):

PR #1109, opened 2023-11-22
[UR] fix ur_loader::urXXXXGetInfo APIs

### LINKED ISSUE #1107: The return value of `urProgramGetInfo` (UR_PROGRAM_INFO_CONTEXT) should be converted to loader handle?
When I was developing the sanitizer layer, I found my code crashed at `urProgramBuild` because of the wrong `hContext` value, which was the native handle but not the loader handle.

BTW, I enabled both level-zero and opencl adapter.

AllanZyne:
fix #1107


pbalcer (commented):
This needs to be holistically solved in the template file. Instead of hardcoding specific functions, it should be possible automatically figure out which properties return a handle and need to be converted. This is going to be a bit more complicated, so feel free to close this PR, and I can fix it.

Thanks for reporting the issue.

PR #1114, opened 2023-11-23
[CI][OPENCL] adapt e2e workflow to opencl target

wlemkows:
- [x] Waits for: https://github.com/oneapi-src/unified-runtime/pull/993 ([source/adapters/opencl/CMakeLists.txt](https://github.com/oneapi-src/unified-runtime/pull/993/files#diff-097e8a00604cf0f1635f4a7c6917bc25e00ad1375221c1a39b115934c91ccfb6))

- [x] build pass: -https://github.com/oneapi-src/unified-runtime/actions/runs/7221868465/job/19677743227

### DIFF
--- a/.github/workflows/e2e_nightly.yml
+++ b/.github/workflows/e2e_nightly.yml
@@ -11,7 +11,8 @@ jobs:
     strategy:
       matrix:
         adapter: [
-          {name: CUDA}
+          {name: CUDA, str_name: cuda, prefix: "ext_oneapi_", config: "--cuda --hip", unit: "gpu"},
+          {name: OPENCL, str_name: opencl, prefix: "", config: "", unit: "cpu"}
         ]
         build_type: [Release]
         compiler: [{c: clang, cxx: clang++}]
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e_nightly.yml
```diff
@@ -111,7 +119,7 @@ jobs:
         -GNinja
         -B ${{github.workspace}}/build-e2e/
         -S ${{github.workspace}}/sycl-repo/sycl/test-e2e/
-        -DSYCL_TEST_E2E_TARGETS="ext_oneapi_cuda:gpu"
+        -DSYCL_TEST_E2E_TARGETS="${{matrix.adapter.prefix}}_${{matrix.adapter.str_name}}:${{matrix.adapter.unit}}"
```
lukaszstolarczuk, line 122: this `_` between prefix and name seems not right 😉

.github/workflows/e2e_nightly.yml
```diff
@@ -1,17 +1,19 @@
...
+#  schedule:
+#    # Run every day at 23:00 UTC
+#    - cron: '0 23 * * *'
 
 jobs:
   e2e-build-hw:
     name: Build SYCL, UR, run E2E
     strategy:
       matrix:
         adapter: [
```
lukaszstolarczuk, line 14: you could add `fail-fast: false` - we may want to run all jobs independently (at least, as long as they run in nightly job)

lukaszstolarczuk (commented):
in general it looks good 👍

.github/workflows/e2e_nightly.yml
wlemkows: This is intentional. It is part of the target name.

lukaszstolarczuk: What I meant is - it won't work in case of an empty prefix...?

wlemkows: Fixed.

.github/workflows/e2e_nightly.yml
wlemkows: I think it should fail to not block other runners if anything is failing.

wlemkows:
Runner set, PR passes, ready to merge.

pbalcer (approved):

lukaszstolarczuk (approved):

Issue #867, opened 2023-09-18
add support for detecting use-after-free in the validation layer (for all APIs)


pbalcer:
The existing validation layer detects handle leaks and incorrect use of get/retain/release APIs for handle lifetime management. It's only capable of detecting use-after-free if it happens to be in one of the lifetime management calls.
Given the existing implementation that already tracks handles, the validation layer should be extended to also detect use-after-free and, in general, use of invalid handles across the entire API.

PR #1121, opened 2023-11-24
[UR] Add lifetime validation to validation layer

kswiecicki:
... and refactor the validation layer mako template.

Relies on: https://github.com/oneapi-src/unified-runtime/pull/1029.
Solves: https://github.com/oneapi-src/unified-runtime/issues/867.


scripts/templates/helper.py
```diff
@@ -1297,45 +1297,79 @@ def get_loader_epilogue(namespace, tags, obj, meta):
 
     return epilogue
 
+
+def get_event_wait_list_functions(specs, namespace, tags):
+    funcs = []
+    for s in specs:
+        for obj in s['objects']:
+            if re.match(r"function", obj['type']):
+                if any(x['name'] == 'phEventWaitList' for x in obj['params']) and any(
```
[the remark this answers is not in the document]
kswiecicki: Well, I'm not sure what's the idea behind this function, I've just refactored and moved the `_get_create_get_retain_release_functions` function.

pbalcer (commented):
right now this checks whether the provided ptr is a valid ptr to something, not necessarily an object of the expected type. Shouldn't be too hard to add?

scripts/core/INTRO.rst
```diff
@@ -256,6 +256,8 @@ Layers currently included with the runtime are as follows:
      - Enables non-adapter-specific parameter validation (e.g. checking for null values).
    * - UR_LAYER_LEAK_CHECKING
      - Performs some leak checking for API calls involving object creation/destruction.
+   * - UR_LAYER_LIFETIME_VALIDATION
+     - Performs lifetime validation on objects used in API calls. Requires UR_LAYER_LEAK_CHECKING.
```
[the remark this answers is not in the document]
kswiecicki: Yes, it's not a problem. I was considering either enabling it automatically or making it a requirement.
I'll change it.

scripts/templates/valddi.cpp.mako
```diff
@@ -72,39 +74,52 @@ namespace ur_validation_layer
 
         }
 
+        if( context.enableLifetimeValidation ) {
```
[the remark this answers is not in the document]
kswiecicki: Done.

scripts/core/INTRO.rst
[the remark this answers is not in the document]
kswiecicki: I've added an explanation, take a look.

kswiecicki:
> right now this checks whether the provided ptr is a valid ptr to something, not necessarily an object of the expected type. Shouldn't be too hard to add?

I've added `std::type_index` info to the `RefRuntimeInfo` and check for the index equality when checking a reference validity.
I was trying to also print handle type names but they were mangled and showed internal type names like `ur_adapter_handle_t_ *` instead of `ur_adapter_handle_t`.

test/layers/validation/fixtures.hpp
```diff
@@ -30,36 +30,57 @@ struct urTest : ::testing::Test {
...
     }
 
     std::vector<ur_adapter_handle_t> adapters;
+};
+
+struct valAdapterTest : valAdaptersTest {
+
+    void SetUp() {
+        valAdaptersTest::SetUp();
+        adapter = adapters[0]; // TODO - which to choose?
```
pbalcer, line 60: the way the tests are run there will only ever be one adapter. so [0] is fine.

pbalcer (approved):

pbalcer:
> I was trying to also print handle type names but they were mangled and showed internal type names like `ur_adapter_handle_t_ *` instead of `ur_adapter_handle_t`.

I think that's fine. If you want to do this, one approach might be to pass the type name as a string from the generated layer to the class with runtime info. Would allow you to do away with typeindex and get correct type names. Not sure if it's worth it.

pbalcer (approved):

scripts/templates/valddi.cpp.mako
```diff
@@ -74,37 +76,35 @@ namespace ur_validation_layer
...
+        %if func_name in tp_handle_funcs['create']:
         if( context.enableLeakChecking && result == UR_RESULT_SUCCESS )
         {
-            refCountContext.incrementRefCount(${object_param}, true);
+            refCountContext.createRefCount(*${tp['name']});
         }
-        %elif func_name == n + "AdapterGet":
-        if( context.enableLeakChecking && phAdapters && result == UR_RESULT_SUCCESS )
+        %elif func_name in tp_handle_funcs['get']:
+        if( context.enableLeakChecking && ${tp['name']} && result == UR_RESULT_SUCCESS )
```
PatKamin, line 102: Can the `name` be an empty field?

PatKamin, line 102: Is it possible for `tp['name']` to be empty?

test/layers/validation/lifetime.cpp
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2023 Intel Corporation
```
PatKamin, line 1: Update dates in all files.

test/layers/validation/lifetime.cpp
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "fixtures.hpp"
+
+TEST_F(urTest, testUrAdapterHandleLifetimeExpectFail) {
+    size_t size = 0;
+    ur_adapter_handle_t adapter = (ur_adapter_handle_t)0xC0FFEE;
+    ur_adapter_info_t info_type = UR_ADAPTER_INFO_BACKEND;
+    ASSERT_EQ(urAdapterGetInfo(adapter, info_type, 0, nullptr, &size),
+              UR_RESULT_ERROR_INVALID_ARGUMENT);
```
PatKamin: You return the actual `size` here. Add an assert for this var or ignore it by passing a `nullptr` and removing this variable. Same below.

cmake/match.py
```diff
@@ -5,77 +5,131 @@
...
+        # match or process tags
+        if re.fullmatch(pattern, input_line):
+            input_idx += 1
+            match_idx += 1
+            tags_in_effect = []
+        elif Tag.OPT in tags_in_effect:
+            match_idx += 1
+            tags_in_effect.remove(Tag.OPT)
+        elif Tag.IGNORE in tags_in_effect:
+            ignored_lines.append(input_line + "\n")
```
lukaszstolarczuk, line 126: `\n` -> `os.linesep`...?

cmake/match.py
```diff
@@ -5,77 +5,131 @@
...
+            input_idx += 1
```
lukaszstolarczuk, line 127: hmm.. when is IGNORE tag removed from `tags_in_effect`...?

cmake/match.py
```diff
@@ -5,77 +5,131 @@
...
 
-# check if all lines match in a file
-# lines in a match file can contain regex inside of double curly braces {{}}
+# Check if all input file content matches match file content.
+# Lines in a match file can contain regex inside of double curly braces {{}}.
+# Regex patterns are limited to single line.
+#
+# List of available special tags:
+# {{OPT}}    - makes content in the same line as the tag optional
+# {{IGNORE}} - ignores all content until next match or end of input file
```
lukaszstolarczuk, line 14: just confirming - by "next match" you mean "next IGNORE tag"? perhaps you could be more specific, please

scripts/templates/valddi.cpp.mako
kswiecicki: It should never be an empty field unless specification was incorrectly formulated.

kswiecicki: Answered above.

test/layers/validation/fixtures.hpp
```diff
@@ -30,36 +30,57 @@ struct urTest : ::testing::Test {
...
         }
         urTest::TearDown();
     }
 
     std::vector<ur_adapter_handle_t> adapters;
+};
+
+struct valAdapterTest : valAdaptersTest {
+
+    void SetUp() {
```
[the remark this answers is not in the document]
kswiecicki: Done.

test/layers/validation/fixtures.hpp
```diff
@@ -30,36 +30,57 @@ struct urTest : ::testing::Test {
...
+
+    void SetUp() override {
+        valAdaptersTest::SetUp();
+
+        uint32_t count;
+        ASSERT_EQ(urPlatformGet(adapters.data(),
+                                static_cast<uint32_t>(adapters.size()), 0,
+                                nullptr, &count),
+                  UR_RESULT_SUCCESS);
+        ASSERT_NE(count, 0);
```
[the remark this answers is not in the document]
kswiecicki: Done.

test/layers/validation/lifetime.cpp
kswiecicki: Done.

test/layers/validation/lifetime.cpp
kswiecicki: I've removed the `size` param.

cmake/match.py
kswiecicki: Done.

cmake/match.py
kswiecicki: By next match I mean the next line that's successfully matched, next `{{IGNORE}}` tag will do nothing since `{{IGNORE}}` tag is already in effect.

lukaszstolarczuk: och, I see, then please write that in the comment 😉 `the next successfully matched line`

btw, I'd write: `or the end of the input ...` (missing "the" x2)

test/layers/validation/lifetime.cpp
kswiecicki: Actually, I need `size` param for this function to fail at the expected staged. I don't check its value because I don't want to  rely on the null adapter generic implementation for `urDeviceGetInfo` which does nothing.

scripts/templates/valddi.cpp.mako
PatKamin: This check `tp['name']` can be removed then.

kswiecicki: This is not a check, I'm retrieving the value at `tp['name']` with `${...}`.

PatKamin (approved):

scripts/core/common.yml
```diff
@@ -102,7 +102,7 @@ name: "$x_kernel_handle_t"
 --- #--------------------------------------------------------------------------
 type: handle
 desc: "Handle of a queue object"
-class: $xPlatform
+class: $xQueue
```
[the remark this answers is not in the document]
kswiecicki: Done.

pbalcer, line 105: this needs to be moved into another PR. Otherwise this one lgtm and we can merge it.

kswiecicki: Done. The queue fix is here https://github.com/oneapi-src/unified-runtime/pull/1285.

PR #1124, opened 2023-11-25
[NFC] Add utility function ur_unreachable to simplify code

wenju-he:


wenju-he:
@kbenzie could you please take a look if this PR makes sense?

pbalcer (approved):

PietroGhg (approved):
Native CPU LGTM, thank you

source/common/ur_util.hpp
```diff
@@ -295,4 +295,12 @@ inline ur_result_t exceptionToResult(std::exception_ptr eptr) {
 
 template <class> inline constexpr bool ur_always_false_t = false;
 
+[[noreturn]] inline void ur_unreachable() {
+#ifdef _MSC_VER
+    __assume(0);
+#else
+    __builtin_unreachable();
+#endif
+}
```
kbenzie, line 304: I think we should put this in a namespace `ur::unreachable()`.

wenju-he: done

kbenzie (approved):

kbenzie:
I think there's a missing include in the CUDA file that's attempting to use `ur::unreachable`.

wenju-he:
> I think there's a missing include in the CUDA file that's attempting to use `ur::unreachable`.

added to both cuda/usm.cpp and hip/usm.cpp

kbenzie:
> > I think there's a missing include in the CUDA file that's attempting to use `ur::unreachable`.
> 
> added to both cuda/usm.cpp and hip/usm.cpp. I don't have access to cuda/hip machines to verify. After searching the headers used in hip/usm.cpp, it looks like ur_util.hpp is needed.

If you could add those includes we can let the CI check if the fix works.

fabiomestre:

JackAKirk (approved):

PR #1126, opened 2023-11-27
CMake Flag for building all adapters

martygrant:
Suggestion for a new CMake flag `UR_BUILD_ADAPTER_ALL` to build all supported adapters rather than specifying multiple individual flags. This is something I found useful for local dev work. 

Issue for this to discuss two different options - https://github.com/oneapi-src/unified-runtime/issues/1132


README.md
```diff
@@ -132,6 +132,7 @@ List of options provided by CMake:
 | UR_BUILD_ADAPTER_CUDA   | Fetch and use cuda adapter from SYCL                   | ON/OFF     | OFF     |
 | UR_BUILD_ADAPTER_HIP    | Fetch and use hip adapter from SYCL                    | ON/OFF     | OFF     |
 | UR_BUILD_ADAPTER_NATIVE_CPU | Fetch and use native-cpu adapter from SYCL         | ON/OFF     | OFF     |
+| UR_BUILD_ADAPTER_ALL    | Fetch and use all supported adapter from SYCL          | ON/OFF     | OFF     |
```
lukaszstolarczuk, line 135: `adapters` (misspell - plural)

lukaszstolarczuk (commented):
👍

README.md
jandres742, line 135: unrelated: 

@kbenzie , @pbalcer : why we have here "fetch and use from SYCL"? We build the adapters in UR, and then update the tag in the SYCL repo, so it is actually the SYCL repo the one fetching the changes from the unified-runtime repo.

jandres742 (commented):
I like the idea, +1

@kbenzie : should we have an issue opened first to discuss the merits of the change?

README.md
lukaszstolarczuk: heh, I was thinking about the same thing yesterday... I believe it's here because in the beginning it was the other way around and we just didn't change the description...

pbalcer: That's right. We should change it.

martygrant:
@jandres742 my apologies I decided to just open a PR as it's a minor change not altering any existing functionality. I'll put this PR into draft for now and create an issue, as I suggested the other alternative of having a single `UR_BUILD_ADAPTER` list variable with multiple values and will provide a diff for what that suggestion looks like.

pbalcer:
I prefer having an option per adapter. There's not that many of them, and doing it this way has them show up in `cmake -L` (and, consequently, IDEs or any other UI that helps with cmake configuration). It's also easier to programmatically set a variable instead of appending to one.

README.md
martygrant: Updated this, thanks.

martygrant: I've changed the wording here to simply `Build the X adapter`

callumfare (approved):

Issue #1078, opened 2023-11-14
Remove usage of "die" termination function


martygrant:
Across the UR adapters the `die` function (which calls `std::terminate()`) is used for not yet implemented entry points or as an exit from a particular error scenario. This should be replaced with returning a sensible error code instead.

PR #1127, opened 2023-11-27
[CUDA][HIP][L0][NATIVECPU][OpenCL] Remove usage of die/terminate

martygrant:
For https://github.com/oneapi-src/unified-runtime/issues/1078. 

This PR attempts to replace all die/terminate calls with simply returning an appropriate error code. There may be a better code, or some kind of assert/abort may be more appropriate in some cases so any suggestions are appreciated.

intel-llvm run https://github.com/intel/llvm/pull/12076

### DIFF
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -855,9 +855,11 @@ static ur_result_t imageElementByteSize(CUDA_ARRAY_DESCRIPTOR ArrayDesc,
     *Size = 4;
     break;
   default:
-    return UR_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR;
     *Size = 0;
+    return UR_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR;
   }
+
+  return UR_RESULT_SUCCESS;
 }
 
 /// General ND memory copy operation for images.
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -837,23 +837,25 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferFill(
   }
 }
 
-static size_t imageElementByteSize(CUDA_ARRAY_DESCRIPTOR ArrayDesc) {
+static ur_result_t imageElementByteSize(CUDA_ARRAY_DESCRIPTOR ArrayDesc,
+                                        int *Size) {
   switch (ArrayDesc.Format) {
   case CU_AD_FORMAT_UNSIGNED_INT8:
   case CU_AD_FORMAT_SIGNED_INT8:
-    return 1;
+    *Size = 1;
   case CU_AD_FORMAT_UNSIGNED_INT16:
   case CU_AD_FORMAT_SIGNED_INT16:
   case CU_AD_FORMAT_HALF:
-    return 2;
+    *Size = 2;
   case CU_AD_FORMAT_UNSIGNED_INT32:
   case CU_AD_FORMAT_SIGNED_INT32:
   case CU_AD_FORMAT_FLOAT:
-    return 4;
+    *Size = 4;
   default:
-    detail::ur::die("Invalid image format.");
-    return 0;
+    return UR_RESULT_ERROR_INVALID_IMAGE_SIZE;
+    *Size = 0;
   }
+  return UR_RESULT_SUCCESS;
 }
 
 /// General ND memory copy operation for images.
@@ -949,7 +951,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemImageRead(
     CUDA_ARRAY_DESCRIPTOR ArrayDesc;
     UR_CHECK_ERROR(cuArrayGetDescriptor(&ArrayDesc, Array));
 
-    int ElementByteSize = imageElementByteSize(ArrayDesc);
+    int ElementByteSize = 0;
+    UR_RETURN_ON_FAILURE(imageElementByteSize(ArrayDesc, &ElementByteSize));
 
     size_t ByteOffsetX = origin.x * ElementByteSize * ArrayDesc.NumChannels;
     size_t BytesToCopy = ElementByteSize * ArrayDesc.NumChannels * region.width;
@@ -1021,7 +1024,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemImageWrite(
     CUDA_ARRAY_DESCRIPTOR ArrayDesc;
     UR_CHECK_ERROR(cuArrayGetDescriptor(&ArrayDesc, Array));
 
-    int ElementByteSize = imageElementByteSize(ArrayDesc);
+    int ElementByteSize = 0;
+    UR_RETURN_ON_FAILURE(imageElementByteSize(ArrayDesc, &ElementByteSize));
 
     size_t ByteOffsetX = origin.x * ElementByteSize * ArrayDesc.NumChannels;
     size_t BytesToCopy = ElementByteSize * ArrayDesc.NumChannels * region.width;
@@ -1100,7 +1104,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemImageCopy(
     UR_ASSERT(SrcArrayDesc.NumChannels == DstArrayDesc.NumChannels,
               UR_RESULT_ERROR_INVALID_MEM_OBJECT);
 
-    int ElementByteSize = imageElementByteSize(SrcArrayDesc);
+    int ElementByteSize = 0;
+    UR_RETURN_ON_FAILURE(imageElementByteSize(SrcArrayDesc, &ElementByteSize));
 
     size_t DstByteOffsetX =
         dstOrigin.x * ElementByteSize * SrcArrayDesc.NumChannels;
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -838,24 +838,26 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferFill(
 }
 
 static ur_result_t imageElementByteSize(CUDA_ARRAY_DESCRIPTOR ArrayDesc,
-                                        int *Size) {
+                                        unsigned int *Size) {
   switch (ArrayDesc.Format) {
   case CU_AD_FORMAT_UNSIGNED_INT8:
   case CU_AD_FORMAT_SIGNED_INT8:
     *Size = 1;
+    break;
   case CU_AD_FORMAT_UNSIGNED_INT16:
   case CU_AD_FORMAT_SIGNED_INT16:
   case CU_AD_FORMAT_HALF:
     *Size = 2;
+    break;
   case CU_AD_FORMAT_UNSIGNED_INT32:
   case CU_AD_FORMAT_SIGNED_INT32:
   case CU_AD_FORMAT_FLOAT:
     *Size = 4;
+    break;
   default:
-    return UR_RESULT_ERROR_INVALID_IMAGE_SIZE;
+    return UR_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR;
     *Size = 0;
   }
-  return UR_RESULT_SUCCESS;
 }
 
 /// General ND memory copy operation for images.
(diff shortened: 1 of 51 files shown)


source/adapters/cuda/event.cpp
```diff
@@ -248,17 +245,19 @@ urEventWait(uint32_t numEvents, const ur_event_handle_t *phEventWaitList) {
 UR_APIEXPORT ur_result_t UR_APICALL urEventRetain(ur_event_handle_t hEvent) {
   const auto RefCount = hEvent->incrementReferenceCount();
 
-  detail::ur::assertion(RefCount != 0,
-                        "Reference count overflow detected in urEventRetain.");
```
JackAKirk, line 252: This removes a specific error message with a generic one that does not relate to the specific error message.
If you want to keep specific errors then use setErrorMessage etc as e.g here: https://github.com/intel/llvm/pull/10626

JackAKirk: There are quite a few examples of this in this pr.

source/adapters/opencl/usm_p2p.cpp
```diff
@@ -13,18 +13,12 @@
...
   return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL
 urUsmP2PDisablePeerAccessExp([[maybe_unused]] ur_device_handle_t commandDevice,
                              [[maybe_unused]] ur_device_handle_t peerDevice) {
-
-  cl_adapter::die(
-      "Experimental P2P feature is not implemented for OpenCL adapter.");
   return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
```
JackAKirk, line 22: See here: https://github.com/intel/llvm/blob/88f1d0a2b64e2c61d208d4dcebf9b512f49f4860/sycl/plugins/unified_runtime/pi2ur.hpp#L108

This means that in dpc++ the error will simply be PI_ERROR_INVALID_VALUE which is useless to the user and is probably going to lead to a lot of confusion.

In the long run when it really could return the semantics of UR_RESULT_ERROR_UNSUPPORTED_FEATURE
it is probably worth considering whether this is a degradation wrt end-user usefulness wrt the more specific "Experimental P2P feature is not implemented for OpenCL adapter."

There are many examples of this.

source/adapters/hip/memory.cpp
```diff
@@ -275,9 +278,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
         } else if constexpr (std::is_same_v<T, SurfaceMem>) {
           HIP_ARRAY3D_DESCRIPTOR ArrayDescriptor;
           UR_CHECK_ERROR(hipArray3DGetDescriptor(&ArrayDescriptor, Mem.Array));
-          const auto PixelSizeBytes =
-              GetHipFormatPixelSize(ArrayDescriptor.Format) *
-              ArrayDescriptor.NumChannels;
+          auto PixelSize = 0;
+          UR_RETURN_ON_FAILURE(
+              GetHipFormatPixelSize(ArrayDescriptor.Format, &PixelSize));
```
GeorgeWeb, line 283: Can we define the `PixelSize` as `int` type explicitly instead of using `auto`. Currently it is `auto`-deduced as an `int` due to the `0` on the right hand-side regardless but I don't see the need for `auto` in this specific instance. Also, I think ideally the `GetHipFormatPixelSize` function would accept `unsigned int*` as an argument for the size parameter.

source/adapters/hip/common.cpp
```diff
@@ -188,3 +178,21 @@ ur_result_t urGetLastResult(ur_platform_handle_t, const char **ppMessage) {
   *ppMessage = &ErrorMessage[0];
   return ErrorMessageCode;
 }
+
+ur_result_t GetHipFormatPixelSize(hipArray_Format Format, int *Size) {
```
GeorgeWeb, line 182: I know this is just refactoring existing code but does the `Size` argument need to be (pointer to) an integer. Most of the calculation using the size are with values of type `unsigned int` or `size_t` since size values are supposed to be greater or equal to 0. If it is a requirement to be an `int` due to something that I missed please let me know, I just thought that `unsigned int *Size` here would've been appropriate.

source/adapters/cuda/event.cpp
jandres742: > This removes a specific error message with a generic one that does not relate to the specific error message.

Agree.

source/adapters/cuda/enqueue.cpp
```diff
@@ -843,23 +843,25 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferFill(
   }
 }
 
-static size_t imageElementByteSize(CUDA_ARRAY_DESCRIPTOR ArrayDesc) {
+static ur_result_t imageElementByteSize(CUDA_ARRAY_DESCRIPTOR ArrayDesc,
+                                        int *Size) {
   switch (ArrayDesc.Format) {
```
jchlanda, line 848: Do you need `break`?
https://godbolt.org/z/P766Pvaf1

source/adapters/hip/common.cpp
jchlanda: Same applies to `imageElementByteSize`.

source/adapters/hip/common.hpp
```diff
@@ -15,6 +15,16 @@
 #include <hip/hip_runtime.h>
 #include <ur/ur.hpp>
 
+/**
+ * Call an UR API and, if the result is not UR_RESULT_SUCCESS, automatically
```
jchlanda, line 19:
```suggestion
 * Call a UR API and, if the result is not UR_RESULT_SUCCESS, automatically
```

source/adapters/hip/device.cpp
```diff
@@ -418,15 +465,19 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
...
+      return UR_RESULT_ERROR_INVALID_SIZE;
+    }
     return ReturnValue(static_cast<uint64_t>(LocalMemSize));
   }
   case UR_DEVICE_INFO_ERROR_CORRECTION_SUPPORT: {
     int EccEnabled = 0;
     UR_CHECK_ERROR(hipDeviceGetAttribute(
         &EccEnabled, hipDeviceAttributeEccEnabled, hDevice->get()));
 
-    detail::ur::assertion((EccEnabled == 0) | (EccEnabled == 1));
```
jchlanda, line 429: Same as in CUDA re: binary/logical or.

source/adapters/hip/device.cpp
```diff
@@ -493,8 +546,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     // name instead, this is also what AMD OpenCL devices return.
     if (strlen(Name) == 0) {
       hipDeviceProp_t Props;
-      detail::ur::assertion(hipGetDeviceProperties(&Props, hDevice->get()) ==
-                            hipSuccess);
+      if (hipGetDeviceProperties(&Props, hDevice->get()) != hipSuccess) {
```
jchlanda, line 549: Should we make some effort to try and translate native error returns to UR specific? There is a bunch of native calls (here and in CUDA: `cuMemGetInfo`, `cuDeviceTotalMem` etc.) that just hard code `UR_RESULT_ERROR_INVALID_OPERATION`/`UR_RESULT_ERROR_INVALID_SIZE` on failure. Previously they were wrapped up in assertion, so the value didn't really matter, but now, they will be served up all the way to the user.

jchlanda (commented):
I like the effort to get rid of asserting in favour of providing error message. I do think however, that greater care is needed to make sure that the error codes are meaningful. Admittedly, this is non-trivial and daunting task, but if we let it slip now, it's likely that it will never get fixed.

Similarly, like Jack pointed out, it would be nice to preserve messages (maybe with exception of those unimplemented API entry).

source/adapters/cuda/enqueue.cpp
```diff
@@ -843,23 +843,25 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferFill(
...
   case CU_AD_FORMAT_UNSIGNED_INT8:
   case CU_AD_FORMAT_SIGNED_INT8:
-    return 1;
+    *Size = 1;
```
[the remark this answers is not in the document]
GeorgeWeb: Oh yeah, I just spotted that too. The control will be transferred to the next statement and so on until it reaches a `break` or end of the `switch`. Maybe there are either no tests or not enabled tests to catch this now but it will be giving incorrect element byte sizes for when 8-bit and 16-bit formats are used.

source/adapters/hip/memory.cpp
martygrant: Absolutely, the original line was `auto`.  I changed the `GetHipFormatPixelSize` signature too. Cheers.

source/adapters/hip/common.cpp
martygrant: Have updated these functions, cheers.

source/adapters/cuda/enqueue.cpp
```diff
@@ -843,23 +843,25 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferFill(
...
+    *Size = 2;
   case CU_AD_FORMAT_UNSIGNED_INT32:
   case CU_AD_FORMAT_SIGNED_INT32:
   case CU_AD_FORMAT_FLOAT:
-    return 4;
+    *Size = 4;
   default:
-    detail::ur::die("Invalid image format.");
-    return 0;
+    return UR_RESULT_ERROR_INVALID_IMAGE_SIZE;
```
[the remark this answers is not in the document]
martygrant: Updated to this, thanks.

source/adapters/cuda/memory.cpp
```diff
@@ -306,8 +306,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemImageCreate(
     PixelTypeSizeBytes = 4;
     break;
   default:
-    detail::ur::die(
-        "urMemImageCreate given unsupported image_channel_data_type");
+    return UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION;
```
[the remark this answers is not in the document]
martygrant: Updated to this, thanks.

source/adapters/hip/common.cpp
```diff
@@ -188,3 +178,21 @@ ur_result_t urGetLastResult(ur_platform_handle_t, const char **ppMessage) {
...
+  switch (Format) {
+  case HIP_AD_FORMAT_UNSIGNED_INT8:
+  case HIP_AD_FORMAT_SIGNED_INT8:
+    *Size = 1;
```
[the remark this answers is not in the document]
martygrant: Fixed this, thanks.

source/adapters/cuda/enqueue.cpp
martygrant: Fixed this, thanks.

source/adapters/hip/common.hpp
martygrant: Fixed this, thanks.

source/adapters/hip/device.cpp
martygrant: Fixed this, thanks.

source/adapters/opencl/usm_p2p.cpp
JackAKirk: Sounds good, thanks.

fabiomestre:

Bensuo (approved):
Changes to command buffers LGTM

isaacault (approved):
Bindless Images LGTM

source/adapters/level_zero/event.cpp
```diff
@@ -641,9 +641,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventWait(
       // This ensures that all signalling commands are submitted below and
       // thus proxy events can be waited without a deadlock.
       //
-      ur_event_handle_t_ *Event = ur_cast<ur_event_handle_t_ *>(e);
-      if (!Event->hasExternalRefs())
-        die("urEventsWait must not be called for an internal event");
+      ur_event_handle_t_ *Event =
```
nrspruit, line 644: why drop the use of the variable e? then it will just be unused in the internals.

martygrant:
## Information

## PR Migration

---

martygrant:

---

PR #1128, opened 2023-11-27
Clarify spec for QUEUE_INFO_SIZE and change testing for it accordingly.

aarongreig:
LLVM testing https://github.com/intel/llvm/pull/12269


fabiomestre:

scripts/core/queue.yml
```diff
@@ -108,6 +115,7 @@ returns:
         - "`propSize != 0 && pPropValue == NULL`"
         - "`pPropValue == NULL && pPropSizeRet == NULL`"
     - $X_RESULT_ERROR_INVALID_QUEUE
```
kbenzie, line 117: Missing `:` is required, currently this is treating the bullet on the line below as part of this string rather than an array item.

```suggestion
    - $X_RESULT_ERROR_INVALID_QUEUE:
```

Ends up rendered as this which doesn't make sense:

```cpp
///     - ::UR_RESULT_ERROR_INVALID_QUEUE - "If `hQueue` isn't a valid queue handle or if `propName` isn't supported by `hQueue`."
```

kbenzie:
@GeorgeWeb would be great if you could give this a review, seems like this predates the automatic reviewer team assignment being enabled.

GeorgeWeb (approved):
Sorry it took so long to notice this one. Cuda and HIP adapters look good.

PR #1131, opened 2023-11-28
[UR] Optimize util functions

PatKamin:
by using std::move where possible. Fixes coverity issues.

### DIFF
--- a/source/common/ur_util.hpp
+++ b/source/common/ur_util.hpp
@@ -268,7 +268,7 @@ inline std::optional<EnvVarMap> getenv_to_map(const char *env_var_name,
             }
             values_vec.push_back(value);
         }
-        map[key] = values_vec;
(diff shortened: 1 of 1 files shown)


source/common/ur_util.hpp
```diff
@@ -268,7 +268,7 @@ inline std::optional<EnvVarMap> getenv_to_map(const char *env_var_name,
             }
             values_vec.push_back(value);
         }
-        map[key] = values_vec;
+        map[key] = std::move(values_vec);
```
pbalcer, line 271: I wonder whether we shouldn't just ignore issues like these. Adding extraneous `std::move` in non-performance critical code just adds clutter.

pbalcer:
I'm going to close this. @PatKamin please update the coverity issues.

PR #1135, opened 2023-11-28
[L0] Add support for multi-device kernel compilation

nrspruit:


nrspruit:

Thank you! Much appreciated.

source/adapters/level_zero/kernel.cpp
```diff
@@ -41,6 +41,15 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
         *OutEvent ///< [in,out][optional] return an event object that identifies
                   ///< this particular kernel execution instance.
 ) {
+  auto ZeDevice = Queue->Device->ZeDevice;
+
+  ze_kernel_handle_t ZeKernel{};
+  if (Kernel->ZeKernelMap.empty()) {
+    ZeKernel = Kernel->ZeKernel;
```
raiyanla, line 48: Question: Could you confirm if there would ever be scenario where ZeKernelMap is empty, but ZeKernel is created? I see on Line 393 zeKernelCreate() is called and then added to this ZeKernelMap, is there other place ZeKernel would be created outside of being added to the map?

nrspruit: the kernel map will never be empty, the map is the only place now where kernels can be created.

source/adapters/level_zero/kernel.cpp
```diff
@@ -65,18 +74,22 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
 
   ze_group_count_t ZeThreadGroupDimensions{1, 1, 1};
   uint32_t WG[3]{};
 
   // global_work_size of unused dimensions must be set to 1
-  UR_ASSERT(WorkDim == 3 || GlobalWorkSize[2] == 1,
-            UR_RESULT_ERROR_INVALID_VALUE);
-  UR_ASSERT(WorkDim >= 2 || GlobalWorkSize[1] == 1,
-            UR_RESULT_ERROR_INVALID_VALUE);
+  if (WorkDim >= 2) {
```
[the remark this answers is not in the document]
nrspruit: The || causes invalid memory access when the work dim is not >=2 etc. This is an array so accessing that element when it is not that size will cause an invalid memory access and causes a sporadic segfault. This change resolves those issues.

source/adapters/level_zero/kernel.cpp
raiyanla: Gotcha, so this branch exists for negative testing and should only be taken incase of a failure of memory allocation.

raiyanla (approved):

nrspruit:
Hello @kbenzie , please merge this to main and 0.8.x when you are able.

kbenzie:
> Hello @kbenzie , please merge this to main and 0.8.x when you are able.

I missed this one. I'll go ahead and create v0.8.4 with just this since I don't see any other PR's with the v0.8.x label.

PR #1137, opened 2023-11-29
Update Adapter Change Process to use main branch

kbenzie:
Depends on #1159

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -17,9 +17,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-      # with-ref part to be removed after merging 'adapters' branch with 'main'
-      with:
-        ref: adapters
 
     - name: Install pip packages
       run: pip install -r third_party/requirements.txt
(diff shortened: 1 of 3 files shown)


omarahmed1111 (approved):

omarahmed1111:
I think we need to update the nightly CI also [here](https://github.com/oneapi-src/unified-runtime/blob/9e3bb2f902181c1ea62d8e58315c61174e06a626/.github/workflows/nightly.yml#L22)

kbenzie:
> I think we need to update the nightly CI also [here](https://github.com/oneapi-src/unified-runtime/blob/9e3bb2f902181c1ea62d8e58315c61174e06a626/.github/workflows/nightly.yml#L22)

Good shout! Thanks.

.github/workflows/nightly.yml
```diff
@@ -17,9 +17,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-      # with-ref part to be removed after merging 'adapters' branch with 'main'
```
fabiomestre, line 20: There is something similar in the e2e nightly which probably needs to be removed as well

kbenzie: I'll take a look

kbenzie: Removed.

martygrant (approved):

PR #1139, opened 2023-11-29
Check mako dependency for generate-code target

ldrumm:
This python module is a hard dependency of the header generation scripts. Error out on configure if this target is enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,6 +250,16 @@ install(
 set(API_JSON_FILE ${PROJECT_BINARY_DIR}/unified_runtime.json)
 
 if(UR_FORMAT_CPP_STYLE)
+    execute_process(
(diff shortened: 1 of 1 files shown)


ldrumm:
On second thoughts, there are a bunch of other python dependencies that I just happened to have on my system already. Perhaps we need to find a way to check everything not in the standard library? Or perhaps just forgot about it and continue leaving it until build time?

kbenzie:
I feel like this failure mode has been pointed out before, I agree that we should resolve it to improve usability of the project. I do wonder about setting up a venv and installing all our required dependencies as part of the build but that seems over the top for those with multiple build directories.

ldrumm:
Agreed. Let's decide on this later. Ticket to track it [here](https://github.com/oneapi-src/unified-runtime/issues/1140)

PR #1142, opened 2023-11-30
Update and extend dockers

lukaszstolarczuk:
Add new docker images and update the existing ones with proper dependencies. I've added pip requirements into the docker, hence updated README with proper usage.

These images should be ready to use in CI.

// I've tested these dockers on a sep. branch (`test-dockers`) on my fork, results are all green (except for clang build in Ubuntu 20.04 - but this will be fixed separately):
https://github.com/lukaszstolarczuk/unified-runtime/actions/runs/10073401444
\+ Fedora build: https://github.com/lukaszstolarczuk/unified-runtime/actions/runs/10073471475

### DIFF
--- a/.github/docker/install_dpcpp.sh
+++ b/.github/docker/install_dpcpp.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright (C) 2023 Intel Corporation
+# Copyright (C) 2023-2024 Intel Corporation
 # Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
 # See LICENSE.TXT
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
@@ -15,9 +15,6 @@ if [ "${SKIP_DPCPP_BUILD}" ]; then
 	exit
 fi
 
-apt-get install -y --no-install-recommends \
-	libncurses5
-
-mkdir -p ${DPCPP_PATH}
-wget -O ${DPCPP_PATH}/dpcpp_compiler.tar.gz https://github.com/intel/llvm/releases/download/sycl-nightly%2F20230626/dpcpp-compiler.tar.gz
-tar -xvf ${DPCPP_PATH}/dpcpp_compiler.tar.gz -C ${DPCPP_PATH}/
+mkdir -p ${DPCPP_PATH}/dpcpp_compiler
+wget -O ${DPCPP_PATH}/dpcpp_compiler.tar.gz https://github.com/intel/llvm/releases/download/nightly-2024-01-29/sycl_linux.tar.gz
+tar -xvf ${DPCPP_PATH}/dpcpp_compiler.tar.gz -C ${DPCPP_PATH}/dpcpp_compiler
--- a/.github/docker/rockylinux-8.Dockerfile
+++ b/.github/docker/rockylinux-8.Dockerfile
@@ -4,18 +4,17 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #
-# Dockerfile - a 'recipe' for Docker to build an image of ubuntu-based
+# Dockerfile - a 'recipe' for Docker to build an image of rockylinux-based
 #              environment for building the Unified Runtime project.
 #
 
-# Pull base image ("22.04")
-FROM registry.hub.docker.com/library/ubuntu@sha256:0eb0f877e1c869a300c442c41120e778db7161419244ee5cbc6fa5f134e74736
+# Pull base image ("8.9")
+FROM registry.hub.docker.com/library/rockylinux@sha256:9794037624aaa6212aeada1d28861ef5e0a935adaf93e4ef79837119f2a2d04c
 
 # Set environment variables
-ENV OS ubuntu
-ENV OS_VER 22.04
+ENV OS rockylinux
+ENV OS_VER 8
 ENV NOTTY 1
-ENV DEBIAN_FRONTEND noninteractive
 
 # Additional parameters to build docker without building components.
 # These ARGs can be set in docker building phase and are used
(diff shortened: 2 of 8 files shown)


.github/docker/rockylinux-8.Dockerfile
```diff
@@ -0,0 +1,84 @@
...
+ && dnf --enablerepo devel install -y \
+	${BASE_DEPS} \
+	${UR_DEPS} \
+	${PRE_PYTHON_DEPS} \
+	${MISC_DEPS} \
+ && dnf clean all
+
+# Additional dependencies (installed via pip)
+COPY third_party/requirements.txt requirements.txt
+# Comment out clang-format as it's not working properly on this distro
```
PatKamin, line 65: Do we have to file a new issue for fixing this problem?

lukaszstolarczuk: I guess so. If we decide to merge this PR we should probably track this - either we should fix the weird problems in clang-format installation (via pip) on rockylinux8 or update the version of clang-format...

fabiomestre:

wlemkows:
Please add Hadolint support.

lukaszstolarczuk:
> Please add Hadolint support.

Done in #1253

lukaszstolarczuk:
All Hadolint issues are resolved, except for apt/dnf/zypper pinned dependencies (which is ignored in the workflow). It's a topic to discuss if we want to pin versions instead of using the latest ones available in a distro.

The same applies to Scorecard, but the issue is not ignored 😉

.github/docker/install_dpcpp.sh
```diff
@@ -15,9 +15,6 @@ if [ "${SKIP_DPCPP_BUILD}" ]; then
 	exit
 fi
 
-apt-get install -y --no-install-recommends \
-	libncurses5
-
-mkdir -p ${DPCPP_PATH}
-wget -O ${DPCPP_PATH}/dpcpp_compiler.tar.gz https://github.com/intel/llvm/releases/download/sycl-nightly%2F20230626/dpcpp-compiler.tar.gz
-tar -xvf ${DPCPP_PATH}/dpcpp_compiler.tar.gz -C ${DPCPP_PATH}/
+mkdir -p ${DPCPP_PATH}/dpcpp_compiler
```
PatKamin, line 18: Bump the date in the license.

.github/docker/rockylinux-9.Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+
+# Install libbacktrace
+COPY .github/docker/install_libbacktrace.sh /opt/ur/install_libbacktrace.sh
+RUN /opt/ur/install_libbacktrace.sh
+
+# Add a new (non-root) 'user'
+ENV USER user
+ENV USERPASS pass
+# Change shell to bash with safe pipe usage
+SHELL [ "/bin/bash", "-o", "pipefail", "-c" ]
```
PatKamin, line 75: Why is it needed? Can't the following commands be done with a default shell?

lukaszstolarczuk: it's required (for the security scans) to enable `-o pipefail` and this is the way to do it

.github/docker/install_dpcpp.sh
lukaszstolarczuk: Done.

.github/docker/rockylinux-8.Dockerfile
lukaszstolarczuk: We decided we won't test on regular basis python dependencies in some OSes, due to missing packages in some of these repos (in their pip database to be precise).

Development checks will be run only on selected OS, in our case mostly on Ubuntu 22.04.

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,43 @@
+# Runs linter for Docker files
+name: Trivy
+
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+    paths:
+      - '.github/docker/*Dockerfile'
+      - '.github/workflows/trivy.yml'
```
wlemkows, line 10: Do we scan dockerfiles or dockerfiles and images?

lukaszstolarczuk: in this workflow we only scan for misconfigurations (with command `trivy config`). Docs: https://aquasecurity.github.io/trivy/v0.49/docs/scanner/misconfiguration/

If you want to scan images that's another story 😉 (for now, we don't build them)

.trivyignore
```diff
@@ -0,0 +1,10 @@
+# Docs: https://aquasecurity.github.io/trivy/latest/docs/configuration/filtering/#trivyignore
+
+# In docker files:
+# non-root user is always created within docker, but we switch it only in CI workflows;
+# not enforcing non-root user makes it easier for developers to use their own users in local container
+AVD-DS-0002
+
+# In docker files:
+# HEALTHCHECK is not required for development, nor in CI (failed docker = failed CI)
```
wlemkows, line 9: PSE approval is required for:
Hadolint YAML configuration files that disable any built-in rules. (I assume the same requirement will apply to Trivy)

wlemkows (changes requested):

.github/docker/ubuntu-22.04.Dockerfile
```diff
@@ -34,13 +35,10 @@ ARG UR_DEPS="\
 	python3 \
 	python3-pip"
 
-# Unified Runtime's dependencies (installed via pip)
-ARG UR_PYTHON_DEPS="\
-	clang-format==15.0.7"
-
 # Miscellaneous for our builds/CI (optional)
 ARG MISC_DEPS="\
 	clang \
```
lukaszstolarczuk, line 40: I might be missing `libstdc++-12-dev` for clang build...?

pbalcer (approved):

PR #1143, opened 2023-11-30
[CTS][CUDA] Ignore adapter warnings in conformance tests

fabiomestre:
Adapter warnings are supported in the UR spec for every
entrypoint. This commit updates the CTS to not fail
when a warning is returned. An environment variable was
added to optionally print the warnings to stderr.

A few fixes to the CUDA adapter were also added as part of this
commit:

- Always return a valid event for urEnqueueUSMPrefetch and
urEnqueueUSMAdvise.
- urAdapterGetLastError no longer ignores the error parameter.

E2E run: https://github.com/intel/llvm/pull/12051

### DIFF
--- a/source/adapters/cuda/adapter.cpp
+++ b/source/adapters/cuda/adapter.cpp
@@ -57,7 +57,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urAdapterRelease(ur_adapter_handle_t) {
 
 UR_APIEXPORT ur_result_t UR_APICALL urAdapterGetLastError(
     ur_adapter_handle_t, const char **ppMessage, int32_t *pError) {
-  std::ignore = pError;
+  *pError = ErrorMessageCode;
   *ppMessage = ErrorMessage;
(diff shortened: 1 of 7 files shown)


fabiomestre:

source/adapters/cuda/adapter.cpp
```diff
@@ -57,7 +57,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urAdapterRelease(ur_adapter_handle_t) {
 
 UR_APIEXPORT ur_result_t UR_APICALL urAdapterGetLastError(
     ur_adapter_handle_t, const char **ppMessage, int32_t *pError) {
-  std::ignore = pError;
+  *pError = ErrorMessageCode;
```
kbenzie, line 60: `pError` is supposed to contain the adapter specific error message returned from CUDA/HIP/L0/OpenCL/etc. `ErrorMessageCode` is a `ur_result_t` so it shouldn't be getting used to set `pError`.

PR #1149, opened 2023-12-01
[GHA] Separate HW concurrency groups per apdater

kbenzie:
Currently when any adapter fails testing on hardware all the jobs for all the adapters are cancelled. This can mean issues on adapters which didn't fail first are hidden and require multiple round-trips of bug fixes and GHA runs. This patch changes the HR runner concurrency groupings to be specific to each adapter rather than global.


fabiomestre:

kbenzie:
This has been superceeded by #1530

Issue #1164, opened 2023-12-07
[NATIVE_CPU] Cannot build UR with native cpu support


wlemkows:
commit 861d95d4fa83e14462e96022a995fa2a095084cc
Ubuntu 22.04
kernel 5.19.0
Hardware Class: cpu
Arch: X86-64
Vendor: "GenuineIntel"
Model: 6.94.3 "Intel(R) Core(TM) i5-6600 CPU @ 3.30GHz"

I am building ur with flags:
cmake -B build -DUMF_ENABLE_POOL_TRACKING=ON -DUR_DEVELOPER_MODE=ON  -DUR_BUILD_TESTS=ON -DUR_FORMAT_CPP_STYLE=ON -DCMAKE_BUILD_TYPE=Debug  **-DUR_BUILD_ADAPTER_NATIVE_CPU=ON**

cmake --build ./unified-runtime/build/ -j4

Error:
[ 14%] Building CXX object source/adapters/native_cpu/CMakeFiles/ur_adapter_native_cpu.dir/device.cpp.o
/home/pmdk/Desktop/unified-runtime/source/adapters/native_cpu/device.cpp: In function ‘ur_result_t urDeviceGetInfo(ur_device_handle_t, ur_device_info_t, size_t, void*, size_t*)’:
/home/pmdk/Desktop/unified-runtime/source/adapters/native_cpu/device.cpp:144:3: error: case value ‘4157’ not in enumerated type ‘ur_device_info_t’ [-Werror=switch]
  144 |   case UR_EXT_DEVICE_INFO_OPENCL_C_VERSION:
      |   ^~~~
cc1plus: all warnings being treated as errors
gmake[2]: *** [source/adapters/native_cpu/CMakeFiles/ur_adapter_native_cpu.dir/build.make:132: source/adapters/native_cpu/CMakeFiles/ur_adapter_native_cpu.dir/device.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:1936: source/adapters/native_cpu/CMakeFiles/ur_adapter_native_cpu.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
[ 14%] Building CXX object source/loader/CMakeFiles/ur_loader.dir/ur_libddi.cpp.o
[ 14%] Linking CXX static library ../../../lib/libgmock.a
[ 14%] Built target gmock
[...]

PR #1167, opened 2023-12-07
[NATIVECPU] Use static_cast in switch for ur_device_info_t

PietroGhg:
This should fix https://github.com/oneapi-src/unified-runtime/issues/1164 by using `static_cast` to silence warnings about values that are not yet part of the `ur_device_info_t` enum, similarly to what is done by the other adapters.


source/adapters/native_cpu/device.cpp
```diff
@@ -60,7 +60,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   UR_ASSERT(hDevice, UR_RESULT_ERROR_INVALID_NULL_HANDLE);
   UrReturnHelper ReturnValue(propSize, pPropValue, pPropSizeRet);
 
-  switch (propName) {
+  switch (static_cast<uint32_t>(propName)) {
```
pbalcer, line 63: I understand that this is how the rest of the adapters suppress this problem, but wouldn't it be better to add something like:
```
  default:
     warn("unimplemented query {}", propName);
     break;
```

PR #433, opened 2023-04-10
Expand urContextGetNativeHandle

jandres742:
Add needed members for PI:

NumDevices
Devices
OwnNativeHandle

PR #1176, opened 2023-12-11
[CL] Add UR handles to OpenCL adapter

omarahmed1111:
Redesign OpenCL adapter by adding UR handles and get rid of the casting of the UR handles to OPENCL handles.

https://github.com/intel/llvm/pull/12172

### DIFF
--- a/source/adapters/opencl/context.cpp
+++ b/source/adapters/opencl/context.cpp
@@ -15,24 +15,49 @@
 #include <set>
 #include <unordered_map>
 
-ur_result_t cl_adapter::getDevicesFromContext(
-    ur_context_handle_t hContext,
-    std::unique_ptr<std::vector<cl_device_id>> &DevicesInCtx) {
+ur_result_t
+ur_context_handle_t_::makeWithNative(native_type Ctx, uint32_t DevCount,
+                                     const ur_device_handle_t *phDevices,
+                                     ur_context_handle_t &Context) {
+  try {
+    uint32_t CLDeviceCount;
+    CL_RETURN_ON_FAILURE(clGetContextInfo(Ctx, CL_CONTEXT_NUM_DEVICES,
+                                          sizeof(CLDeviceCount), &CLDeviceCount,
+                                          nullptr));
+    std::vector<cl_device_id> CLDevices(CLDeviceCount);
+    CL_RETURN_ON_FAILURE(clGetContextInfo(
+        Ctx, CL_CONTEXT_DEVICES, sizeof(CLDevices), CLDevices.data(), nullptr));
+    std::vector<ur_device_handle_t> URDevices;
+    if (DevCount) {
+      if (DevCount != CLDeviceCount) {
+        return UR_RESULT_ERROR_INVALID_CONTEXT;
+      }
+      for (uint32_t i = 0; i < DevCount; i++) {
+        if (phDevices[i]->CLDevice != CLDevices[i]) {
+          return UR_RESULT_ERROR_INVALID_CONTEXT;
+        }
+        URDevices.push_back(phDevices[i]);
+      }
+    } else {
+      DevCount = CLDeviceCount;
+      for (uint32_t i = 0; i < CLDeviceCount; i++) {
+        ur_device_handle_t UrDevice = nullptr;
+        ur_native_handle_t hNativeHandle =
+            reinterpret_cast<ur_native_handle_t>(CLDevices[i]);
+        UR_RETURN_ON_FAILURE(urDeviceCreateWithNativeHandle(
+            hNativeHandle, nullptr, nullptr, &UrDevice));
+        URDevices.push_back(UrDevice);
+      }
+    }
 
-  cl_uint DeviceCount;
-  CL_RETURN_ON_FAILURE(clGetContextInfo(cl_adapter::cast<cl_context>(hContext),
-                                        CL_CONTEXT_NUM_DEVICES, sizeof(cl_uint),
-                                        &DeviceCount, nullptr));
-
-  if (DeviceCount < 1) {
-    return UR_RESULT_ERROR_INVALID_CONTEXT;
+    auto URContext =
+        std::make_unique<ur_context_handle_t_>(Ctx, DevCount, URDevices.data());
+    Context = URContext.release();
+  } catch (std::bad_alloc &) {
+    return UR_RESULT_ERROR_OUT_OF_RESOURCES;
+  } catch (...) {
+    return UR_RESULT_ERROR_UNKNOWN;
   }
-
-  DevicesInCtx = std::make_unique<std::vector<cl_device_id>>(DeviceCount);
-
-  CL_RETURN_ON_FAILURE(clGetContextInfo(
-      cl_adapter::cast<cl_context>(hContext), CL_CONTEXT_DEVICES,
-      DeviceCount * sizeof(cl_device_id), (*DevicesInCtx).data(), nullptr));
 
   return UR_RESULT_SUCCESS;
 }
@@ -42,32 +67,26 @@ UR_APIEXPORT ur_result_t UR_APICALL urContextCreate(
     const ur_context_properties_t *, ur_context_handle_t *phContext) {
 
   cl_int Ret;
-  *phContext = cl_adapter::cast<ur_context_handle_t>(
-      clCreateContext(nullptr, cl_adapter::cast<cl_uint>(DeviceCount),
-                      cl_adapter::cast<const cl_device_id *>(phDevices),
-                      nullptr, nullptr, cl_adapter::cast<cl_int *>(&Ret)));
-
-  return mapCLErrorToUR(Ret);
-}
-
-static cl_int mapURContextInfoToCL(ur_context_info_t URPropName) {
-
-  cl_int CLPropName;
-  switch (URPropName) {
-  case UR_CONTEXT_INFO_NUM_DEVICES:
-    CLPropName = CL_CONTEXT_NUM_DEVICES;
-    break;
-  case UR_CONTEXT_INFO_DEVICES:
-    CLPropName = CL_CONTEXT_DEVICES;
-    break;
-  case UR_CONTEXT_INFO_REFERENCE_COUNT:
-    CLPropName = CL_CONTEXT_REFERENCE_COUNT;
-    break;
-  default:
-    CLPropName = -1;
+  std::vector<cl_device_id> CLDevices(DeviceCount);
+  for (size_t i = 0; i < DeviceCount; i++) {
+    CLDevices[i] = phDevices[i]->CLDevice;
   }
 
-  return CLPropName;
+  try {
+    cl_context Ctx = clCreateContext(nullptr, static_cast<cl_uint>(DeviceCount),
+                                     CLDevices.data(), nullptr, nullptr,
+                                     static_cast<cl_int *>(&Ret));
+    CL_RETURN_ON_FAILURE(Ret);
+    auto URContext =
+        std::make_unique<ur_context_handle_t_>(Ctx, DeviceCount, phDevices);
+    *phContext = URContext.release();
+  } catch (std::bad_alloc &) {
+    return UR_RESULT_ERROR_OUT_OF_RESOURCES;
+  } catch (...) {
+    return UR_RESULT_ERROR_UNKNOWN;
+  }
+
+  return mapCLErrorToUR(Ret);
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL
@@ -75,7 +94,6 @@ urContextGetInfo(ur_context_handle_t hContext, ur_context_info_t propName,
                  size_t propSize, void *pPropValue, size_t *pPropSizeRet) {
 
   UrReturnHelper ReturnValue(propSize, pPropValue, pPropSizeRet);
-  const cl_int CLPropName = mapURContextInfoToCL(propName);
 
   switch (static_cast<uint32_t>(propName)) {
   /* 2D USM memops are not supported. */
@@ -91,21 +109,14 @@ urContextGetInfo(ur_context_handle_t hContext, ur_context_info_t propName,
      * queries of each device separately and building the intersection set. */
     return UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION;
   }
-  case UR_CONTEXT_INFO_NUM_DEVICES:
-  case UR_CONTEXT_INFO_DEVICES:
+  case UR_CONTEXT_INFO_NUM_DEVICES: {
+    return ReturnValue(hContext->DeviceCount);
+  }
+  case UR_CONTEXT_INFO_DEVICES: {
+    return ReturnValue(&hContext->Devices[0], hContext->DeviceCount);
+  }
   case UR_CONTEXT_INFO_REFERENCE_COUNT: {
-    size_t CheckPropSize = 0;
-    auto ClResult =
-        clGetContextInfo(cl_adapter::cast<cl_context>(hContext), CLPropName,
-                         propSize, pPropValue, &CheckPropSize);
-    if (pPropValue && CheckPropSize != propSize) {
-      return UR_RESULT_ERROR_INVALID_SIZE;
-    }
-    CL_RETURN_ON_FAILURE(ClResult);
-    if (pPropSizeRet) {
-      *pPropSizeRet = CheckPropSize;
-    }
-    return UR_RESULT_SUCCESS;
+    return ReturnValue(hContext->getReferenceCount());
   }
   default:
     return UR_RESULT_ERROR_INVALID_ENUMERATION;
@@ -119,51 +130,47 @@ urContextRelease(ur_context_handle_t hContext) {
   // should drastically reduce the chances of the pathological case described
   // in the comments in common.hpp.
   static std::mutex contextReleaseMutex;
-  auto clContext = cl_adapter::cast<cl_context>(hContext);
+  auto clContext = hContext->CLContext;
 
-  {
-    std::lock_guard<std::mutex> lock(contextReleaseMutex);
-    size_t refCount = 0;
-    CL_RETURN_ON_FAILURE(clGetContextInfo(clContext, CL_CONTEXT_REFERENCE_COUNT,
-                                          sizeof(size_t), &refCount, nullptr));
-
-    // ExtFuncPtrCache is destroyed in an atexit() callback, so it doesn't
-    // necessarily outlive the adapter (or all the contexts).
-    if (refCount == 1 && cl_ext::ExtFuncPtrCache) {
-      cl_ext::ExtFuncPtrCache->clearCache(clContext);
-    }
+  std::lock_guard<std::mutex> lock(contextReleaseMutex);
+  size_t refCount = hContext->getReferenceCount();
+  // ExtFuncPtrCache is destroyed in an atexit() callback, so it doesn't
+  // necessarily outlive the adapter (or all the contexts).
+  if (refCount == 1 && cl_ext::ExtFuncPtrCache) {
+    cl_ext::ExtFuncPtrCache->clearCache(clContext);
   }
 
-  CL_RETURN_ON_FAILURE(
-      clReleaseContext(cl_adapter::cast<cl_context>(hContext)));
+  if (hContext->decrementReferenceCount() == 0) {
+    delete hContext;
+  }
 
   return UR_RESULT_SUCCESS;
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL
 urContextRetain(ur_context_handle_t hContext) {
-
-  cl_int Ret = clRetainContext(cl_adapter::cast<cl_context>(hContext));
-  return mapCLErrorToUR(Ret);
+  hContext->incrementReferenceCount();
+  return UR_RESULT_SUCCESS;
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL urContextGetNativeHandle(
     ur_context_handle_t hContext, ur_native_handle_t *phNativeContext) {
 
-  *phNativeContext = reinterpret_cast<ur_native_handle_t>(hContext);
+  *phNativeContext = reinterpret_cast<ur_native_handle_t>(hContext->CLContext);
   return UR_RESULT_SUCCESS;
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL urContextCreateWithNativeHandle(
-    ur_native_handle_t hNativeContext, ur_adapter_handle_t, uint32_t,
-    const ur_device_handle_t *,
+    ur_native_handle_t hNativeContext, ur_adapter_handle_t, uint32_t numDevices,
+    const ur_device_handle_t *phDevices,
     const ur_context_native_properties_t *pProperties,
     ur_context_handle_t *phContext) {
 
-  *phContext = reinterpret_cast<ur_context_handle_t>(hNativeContext);
-  if (!pProperties || !pProperties->isNativeHandleOwned) {
-    return urContextRetain(*phContext);
-  }
+  cl_context NativeHandle = reinterpret_cast<cl_context>(hNativeContext);
+  UR_RETURN_ON_FAILURE(ur_context_handle_t_::makeWithNative(
+      NativeHandle, numDevices, phDevices, *phContext));
+  (*phContext)->IsNativeHandleOwned =
+      pProperties ? pProperties->isNativeHandleOwned : false;
   return UR_RESULT_SUCCESS;
 }
 
@@ -221,7 +228,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urContextSetExtendedDeleter(
     C->execute();
   };
   CL_RETURN_ON_FAILURE(ur::cl::getAdapter()->clSetContextDestructorCallback(
-      cl_adapter::cast<cl_context>(hContext), ClCallback, Callback));
+      hContext->CLContext, ClCallback, Callback));
 
   return UR_RESULT_SUCCESS;
 }
--- a/source/adapters/opencl/context.hpp
+++ b/source/adapters/opencl/context.hpp
@@ -10,9 +10,43 @@
 #pragma once
 
 #include "common.hpp"
+#include "device.hpp"
 
-namespace cl_adapter {
-ur_result_t
-getDevicesFromContext(ur_context_handle_t hContext,
-                      std::unique_ptr<std::vector<cl_device_id>> &DevicesInCtx);
-}
+#include <vector>
+
+struct ur_context_handle_t_ {
+  using native_type = cl_context;
+  native_type CLContext;
+  std::vector<ur_device_handle_t> Devices;
+  uint32_t DeviceCount;
+  std::atomic<uint32_t> RefCount = 0;
+  bool IsNativeHandleOwned = true;
+
+  ur_context_handle_t_(native_type Ctx, uint32_t DevCount,
+                       const ur_device_handle_t *phDevices)
+      : CLContext(Ctx), DeviceCount(DevCount) {
+    for (uint32_t i = 0; i < DeviceCount; i++) {
+      Devices.emplace_back(phDevices[i]);
+      urDeviceRetain(phDevices[i]);
+    }
+    RefCount = 1;
+  }
+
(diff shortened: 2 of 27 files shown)


Bensuo (approved):

source/adapters/opencl/memory.cpp
```diff
@@ -224,13 +226,13 @@ cl_map_flags convertURMemFlagsToCL(ur_mem_flags_t URFlags) {
 UR_APIEXPORT ur_result_t UR_APICALL urMemBufferCreate(
     ur_context_handle_t hContext, ur_mem_flags_t flags, size_t size,
     const ur_buffer_properties_t *pProperties, ur_mem_handle_t *phBuffer) {
-
   cl_int RetErr = CL_INVALID_OPERATION;
+  UR_RETURN_ON_FAILURE(urContextRetain(hContext));
```
omarahmed1111, line 230: This is temporary on this PR to pass the intel/llvm tests, but will be fixed in this [PR](https://github.com/oneapi-src/unified-runtime/pull/1230), so after [1230](https://github.com/oneapi-src/unified-runtime/pull/1230) is merged, this PR will be rebased on top of it and be ready to merge.

source/adapters/opencl/context.cpp
```diff
@@ -14,38 +14,22 @@
...
+  std::vector<cl_device_id> CLDevices(DeviceCount);
+  for (size_t i = 0; i < DeviceCount; i++) {
+    CLDevices[i] = phDevices[i]->get();
+  }
 
+  cl_context Ctx = clCreateContext(
+      nullptr, cl_adapter::cast<cl_uint>(DeviceCount), CLDevices.data(),
+      nullptr, nullptr, cl_adapter::cast<cl_int *>(&Ret));
+  auto URContext =
+      std::make_unique<ur_context_handle_t_>(Ctx, DeviceCount, phDevices);
```
kbenzie, line 31: This `std::make_unique` might throw `std::bad_alloc` - we should catch that and return `UR_RESULT_ERROR_OUT_OF_MEMORY`.

source/adapters/opencl/context.cpp
```diff
@@ -14,38 +14,22 @@
...
-                      cl_adapter::cast<const cl_device_id *>(phDevices),
-                      nullptr, nullptr, cl_adapter::cast<cl_int *>(&Ret)));
+  std::vector<cl_device_id> CLDevices(DeviceCount);
+  for (size_t i = 0; i < DeviceCount; i++) {
+    CLDevices[i] = phDevices[i]->get();
+  }
 
+  cl_context Ctx = clCreateContext(
+      nullptr, cl_adapter::cast<cl_uint>(DeviceCount), CLDevices.data(),
+      nullptr, nullptr, cl_adapter::cast<cl_int *>(&Ret));
```
kbenzie, line 29: If `Ret != CL_SUCCESS` then we should return `Ret` not allocate an invalid `ur_context_handle_t`.

source/adapters/opencl/context.cpp
```diff
@@ -90,13 +74,16 @@ urContextGetInfo(ur_context_handle_t hContext, ur_context_info_t propName,
...
+  }
+  case UR_CONTEXT_INFO_DEVICES: {
+    return ReturnValue(&hContext->Devices[0], hContext->DeviceCount);
+  }
   case UR_CONTEXT_INFO_REFERENCE_COUNT: {
     size_t CheckPropSize = 0;
-    auto ClResult =
-        clGetContextInfo(cl_adapter::cast<cl_context>(hContext), CLPropName,
-                         propSize, pPropValue, &CheckPropSize);
+    auto ClResult = clGetContextInfo(hContext->get(), CLPropName, propSize,
```
kbenzie, line 85: Now we have `ur_context_handle_t` I think this should be returning its reference count not the `cl_context` reference count.

source/adapters/opencl/context.hpp
```diff
@@ -10,9 +10,52 @@
...
+            reinterpret_cast<ur_native_handle_t>(CLDevices[i]);
+        UR_RETURN_ON_FAILURE(urDeviceCreateWithNativeHandle(
+            NativeDevice, nullptr, nullptr, &(URContext->Devices[i])));
+      }
+    }
+    Context = URContext.release();
+    return UR_RESULT_SUCCESS;
+  }
+
+  ~ur_context_handle_t_() {}
```
kbenzie, line 58: Should this be calling `clReleaseContext`?

source/adapters/opencl/device.hpp
```diff
@@ -10,11 +10,65 @@
...
+    Supported = true;
+    for (const std::string &Ext : Exts) {
+      if (!(Supported = (ExtStr.find(Ext) != std::string::npos))) {
+        break;
+      }
+    }
+
+    return UR_RESULT_SUCCESS;
+  }
+};
```
kbenzie, line 74: Missing new line at end of file.

kbenzie (commented):
I've not reviewed the whole PR yet. I think we need to run the CTS with the Address Sanitizer enabled to make sure we're not introducing memory leaks.

source/adapters/opencl/context.cpp
omarahmed1111: Fixed

source/adapters/opencl/context.cpp
omarahmed1111: Fixed that and tried to move to other places to make sure everywhere have this check.

source/adapters/opencl/context.cpp
omarahmed1111: I was planning to add the reference counting and resolve the memory leak in a separate PR following this to be more clear. but maybe it is better to extend this PR with that so will do that here.

source/adapters/opencl/device.hpp
omarahmed1111: fixed

source/adapters/opencl/context.cpp
```diff
@@ -114,29 +101,33 @@ urContextGetInfo(ur_context_handle_t hContext, ur_context_info_t propName,
 UR_APIEXPORT ur_result_t UR_APICALL
 urContextRelease(ur_context_handle_t hContext) {
 
-  cl_int Ret = clReleaseContext(cl_adapter::cast<cl_context>(hContext));
+  cl_int Ret = clReleaseContext(hContext->get());
```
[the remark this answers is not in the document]
omarahmed1111: Will extend handles with reference counting and fix this memory leak.

source/adapters/opencl/context.hpp
omarahmed1111: Ah, I think I misunderstood that, that's a good point, I didn't link that with reference counting, it actually would make sense to call `clReleaseContext` with the destructor as the ur retaining will decrease the reference counting and delete the resources.

source/adapters/opencl/context.cpp
kbenzie: I'd like to avoid introducing a known memory leak to main, at any point.

source/adapters/opencl/context.hpp
```diff
@@ -10,9 +10,74 @@
...
+  std::atomic<uint32_t> RefCount = 0;
+
+  ur_context_handle_t_(native_type Ctx, uint32_t DevCount,
+                       const ur_device_handle_t *phDevices)
+      : Context(Ctx), DeviceCount(DevCount) {
+    for (uint32_t i = 0; i < DeviceCount; i++) {
+      Devices.emplace_back(phDevices[i]);
+      urDeviceRetain(phDevices[i]);
+    }
+    urContextRetain(this);
```
aarongreig, line 31: I think we could avoid the function call here and just initialize `RefCount` to 1 in the initializer list, like what the cuda adapter does https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/cuda/context.hpp#L82

source/adapters/opencl/context.hpp
```diff
@@ -10,9 +10,74 @@
...
+                             sizeof(DeviceCount), &DeviceCount, nullptr));
+        std::vector<cl_device_id> CLDevices(DeviceCount);
+        CL_RETURN_ON_FAILURE(clGetContextInfo(NativeContext, CL_CONTEXT_DEVICES,
+                                              sizeof(CLDevices),
+                                              CLDevices.data(), nullptr));
+        URContext->Devices.resize(DeviceCount);
+        for (uint32_t i = 0; i < DeviceCount; i++) {
+          ur_native_handle_t NativeDevice =
+              reinterpret_cast<ur_native_handle_t>(CLDevices[i]);
+          UR_RETURN_ON_FAILURE(urDeviceCreateWithNativeHandle(
```
aarongreig, line 60: do we need to do this here? I thought the application would already have had to do something similar to get the devices for `phDevices`

omarahmed1111: Some e2e tests (especially opencl interop ones) where creating ctx from native context but passing devices as `nullptr`, then when it queries the context handle for the devices, it will create a problem as it will expect UR devices handles related to the context, that will also have the native handle for devices and it didn't pass them in first place, so I need to create ones for it. (In OpenCL Interop functionality, it could create any handle at any point without having to pass the dependant handles so thats why I need to create ones, and that is mostly as also SYCl runtime was built to cast the OPENCL handles directly, so i also had to make some changes there to align with that, the changes are in the testing PR for this PR).

aarongreig: passing nullptr for `phDevices` isn't allowed by the spec, so this needs a spec fix or a sycl fix

omarahmed1111: I think it needs a spec fix as I don't find a reference for it in [here](https://oneapi-src.github.io/unified-runtime/core/api.html#urcontextcreatewithnativehandle)

omarahmed1111: I also think, it might need changes in the e2e tests/sycl-rt

omarahmed1111: Ah okay, I didn't notice it in the spec. I think sycl-rt maybe has something to change for that, will try to retest without this and collect more info on this part.

aarongreig: I can't remember why we added the device list as a non-optional parameter but I do remember that it happened relatively recently, maybe we should re-assess that if it doesn't align with the sycl spec. @kbenzie can you remember what the deal with this was?

kbenzie: They were added in #433 to line up with the PI interface which was added in https://github.com/intel/llvm/pull/2772. Both motivated by Level Zero interop.

source/adapters/opencl/context.cpp
omarahmed1111: Added Ref counting to UR layer and fixed that. Checked it also with Address Sanitizer.

source/adapters/opencl/context.cpp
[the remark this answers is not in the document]
omarahmed1111: Added Ref counting to UR layer and fixed that. Checked it also with Address Sanitizer.

omarahmed1111:
> I've not reviewed the whole PR yet. I think we need to run the CTS with the Address Sanitizer enabled to make sure we're not introducing memory leaks.

Used Address sanitizer to check for memory leaks, fixed some in the UR layer here. Some are generated from OpenCL implementation so left that, others are generated from UR CTS tests that will be fixed in this [PR](https://github.com/oneapi-src/unified-runtime/pull/1271)

source/adapters/opencl/context.hpp
omarahmed1111: Fixed that

source/adapters/opencl/kernel.hpp
```diff
@@ -0,0 +1,94 @@
...
+  native_type Kernel;
+  ur_program_handle_t Program;
+  ur_context_handle_t Context;
+  std::atomic<uint32_t> RefCount = 0;
+
+  ur_kernel_handle_t_(native_type Kernel, ur_program_handle_t Program,
+                      ur_context_handle_t Context)
+      : Kernel(Kernel), Program(Program), Context(Context) {
+    RefCount = 1;
+    if (Program) {
```
aarongreig, line 28: I think we can get away without null checking stuff like this in the constructors and destructors. It isn't optional so it'll already get checked by the validation layer, and if you're running without validation then you can't get too mad about hitting a segfault

source/adapters/opencl/kernel.hpp
```diff
@@ -0,0 +1,94 @@
...
+  uint32_t getReferenceCount() const noexcept { return RefCount; }
+
+  static ur_result_t makeWithNative(native_type NativeKernel,
+                                    ur_program_handle_t Program,
+                                    ur_context_handle_t Context,
+                                    ur_kernel_handle_t &Kernel) {
+    try {
+      auto URKernel =
+          std::make_unique<ur_kernel_handle_t_>(NativeKernel, Program, Context);
+      if (!Program) {
```
aarongreig, line 59: the program isn't optional in `KernelCreateWithNativeHandle`, do we fail an interop test if we remove this section the same way we do with `ContextCreateWithNative` and the device list?

aarongreig (approved):
LGTM pending open discussions. This is an important change and I bet it wasn't fun to implement, thanks @omarahmed1111

source/adapters/opencl/kernel.hpp
omarahmed1111: We did, but I moved all this logic to intel/llvm side and will stick with that and see the tests results, if any failure happens i will try to keep that logic with the fix there.

source/adapters/opencl/context.hpp
```diff
@@ -10,9 +10,75 @@
 #pragma once
 
 #include "common.hpp"
+#include "device.hpp"
```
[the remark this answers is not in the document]
omarahmed1111: It fails locally for me when I remove that giving `invalid use of incomplete type ‘struct ur_device_handle_t_’`

source/adapters/opencl/kernel.hpp
omarahmed1111: I changed that, I added it initially as `CreateWithNativeHandle` was implemented to accept it as nullptr, but now that I moved that logic to intel/llvm, this was removed.

source/adapters/opencl/context.hpp
omarahmed1111: I moved that logic to intel/llvm side so that we don't need to change any spec.

source/adapters/opencl/device.cpp
```diff
@@ -1007,63 +957,91 @@ UR_APIEXPORT ur_result_t UR_APICALL urDevicePartition(
...
+    CL_RETURN_ON_FAILURE(clCreateSubDevices(hDevice->get(), CLProperties.data(),
+                                            CLNumDevicesRet,
+                                            CLSubDevices.data(), nullptr));
+    for (uint32_t i = 0; i < std::min(CLNumDevicesRet, NumDevices); i++) {
+      try {
+        auto URSubDevice = std::make_unique<ur_device_handle_t_>(
+            CLSubDevices[i], hDevice->Platform, hDevice);
+        phSubDevices[i] = URSubDevice.release();
+      } catch (std::bad_alloc &) {
+        return UR_RESULT_ERROR_OUT_OF_RESOURCES;
```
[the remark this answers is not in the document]
omarahmed1111: Good point, fixed that

source/adapters/opencl/program.cpp
```diff
@@ -339,33 +377,30 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramSetSpecializationConstants(
...
   oclv::OpenCLVersion PlatVer;
-  cl_adapter::getPlatformVersion(CurPlatform, PlatVer);
+  CurPlatform->getPlatformVersion(PlatVer);
 
   bool UseExtensionLookup = false;
   if (PlatVer < oclv::V2_2) {
     UseExtensionLookup = true;
   } else {
-    for (cl_device_id Dev : *DevicesInCtx) {
+    for (ur_device_handle_t Dev : DevicesInCtx) {
```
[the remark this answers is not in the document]
omarahmed1111: There was 3 places using this, so, I added that. That is more concise.

source/adapters/opencl/context.hpp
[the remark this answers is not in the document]
aarongreig: huh, yeah me too now that I try it again... :shrug:

source/adapters/opencl/kernel.cpp
```diff
@@ -250,13 +272,15 @@ urKernelGetSubGroupInfo(ur_kernel_handle_t hKernel, ur_device_handle_t hDevice,
...
 UR_APIEXPORT ur_result_t UR_APICALL urKernelRetain(ur_kernel_handle_t hKernel) {
-  CL_RETURN_ON_FAILURE(clRetainKernel(cl_adapter::cast<cl_kernel>(hKernel)));
+  hKernel->incrementReferenceCount();
   return UR_RESULT_SUCCESS;
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL
 urKernelRelease(ur_kernel_handle_t hKernel) {
-  CL_RETURN_ON_FAILURE(clReleaseKernel(cl_adapter::cast<cl_kernel>(hKernel)));
+  if (hKernel->decrementReferenceCount() == 0) {
```
fabiomestre, line 281: Shouldn't this code block be inside a mutex? I don't think using atomics is sufficient here. We can get into scenarios where 2 threads decrement the reference count at the same time (and before the check for == 0). That would cause a memory leak because the reference count will be -1.

source/adapters/opencl/memory.cpp
```diff
@@ -288,10 +306,18 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemImageCreate(
...
-      cl_adapter::cast<cl_context>(hContext), MapFlags, &ImageFormat,
-      &ImageDesc, pHost, cl_adapter::cast<cl_int *>(&RetErr)));
-  CL_RETURN_ON_FAILURE(RetErr);
+  try {
+    cl_mem Mem =
+        clCreateImage(hContext->get(), MapFlags, &ImageFormat, &ImageDesc,
+                      pHost, cl_adapter::cast<cl_int *>(&RetErr));
+    CL_RETURN_ON_FAILURE(RetErr);
+    auto URMem = std::make_unique<ur_mem_handle_t_>(Mem, hContext);
+    *phMem = URMem.release();
```
fabiomestre, line 315: I there an advantage to using a unique_ptr here and then call `release()` on it right away? Wouldn't it be simpler and faster to just use `new`?

I know that `new` is frowned upon. But this is a case where I see little advantage in using smart pointers given that the underlying api requires raw pointers. Please correct me if I'm missing something.

source/adapters/opencl/kernel.cpp
omarahmed1111: I think with this scenario one of them only will delete hKernel resources, and the other that will have a `-1` will already have a deleted resource, and both will return as they release successfully and with no memory leak. The atomic will only prevent them from checking at the same time, and prevent them from delete the resources both of them.

source/adapters/opencl/memory.cpp
fabiomestre: I don't understand why make_unique is safer in this scenario. The `new` operator can also throw `std::bad_alloc` and will delete the object if the constructor throws.

My understanding is that the only situation, where make_unique() would be better for exception safety is if we are allocating multiple objects in the same statement which is not the case. 

But it's not a big deal. Personally, I would prefer to keep it simple and fast by just using `new`. But I understand that this is a style preference and since there is other adapters using this style I'm ok with keeping it as is.

fabiomestre (approved):

source/adapters/opencl/context.hpp
```diff
@@ -10,9 +10,74 @@
...
+      }
+      for (uint32_t i = 0; i < DevCount; i++) {
+        if (phDevices[i]->get() != CLDevices[i]) {
+          return UR_RESULT_ERROR_INVALID_CONTEXT;
+        }
+      }
+      auto URContext =
+          std::make_unique<ur_context_handle_t_>(Ctx, DevCount, phDevices);
+      Context = URContext.release();
+      CL_RETURN_ON_FAILURE(clRetainContext(Ctx));
```
aarongreig, line 63: I think for native handles we should mirror l0's behaviour, that's where `isNativeHandleOwned` was first implemented so we can assume it reflects the original intention. The UR handles don't retain the native handle, and `isNativeHandleOwned` decides whether UR will release the native handle when the UR handle is released.

So for example l0 context tracks whether it owns its native handle (the param gets passed from [here](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/context.cpp#L155) to [here](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/context.hpp#L41)) and only uses that information when it comes to releasing the context [here](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/context.cpp#L369)

aarongreig: here's a draft for the new spec around native handles https://github.com/oneapi-src/unified-runtime/pull/2193

source/adapters/opencl/context.hpp
```diff
@@ -10,9 +10,91 @@
...
+  ~ur_context_handle_t_() {
+    for (uint32_t i = 0; i < DeviceCount; i++) {
+      urDeviceRelease(Devices[i]);
+    }
+    if (IsNativeHandleOwned) {
+      clReleaseContext(Context);
+    }
+  }
+
+  native_type get() { return Context; }
```
kbenzie, line 97: `Context` is public, not sure having a `get()`er really adds much.

source/adapters/opencl/context.hpp
```diff
@@ -10,9 +10,91 @@
...
+
+  const std::vector<ur_device_handle_t> &getDevices() { return Devices; }
```
kbenzie, line 99: Likewise for `getDevices()` - we can just access `Devices` directly.

source/adapters/opencl/device.hpp
```diff
@@ -10,11 +10,97 @@
...
+
+  ~ur_device_handle_t_() {
+    if (ParentDevice && IsNativeHandleOwned) {
+      clReleaseDevice(Device);
+    }
+  }
+
+  uint32_t incrementReferenceCount() noexcept { return ++RefCount; }
+
+  uint32_t decrementReferenceCount() noexcept { return --RefCount; }
```
kbenzie, line 44: Reference count and associated member functions is something we should provide a shared abstraction for. That's out of scope for this PR, though. Seeing this just reminded me.

source/adapters/opencl/device.hpp
```diff
@@ -10,11 +10,97 @@
...
+        clGetDeviceInfo(Device, CL_DEVICE_NAME, 0, nullptr, &NameSize));
+    std::string NameStr(NameSize, '\0');
+    CL_RETURN_ON_FAILURE(clGetDeviceInfo(Device, CL_DEVICE_NAME, NameSize,
+                                         NameStr.data(), nullptr));
+
+    return NameStr.find("Intel(R) FPGA Emulation Device") != std::string::npos;
+  }
+
+  ur_result_t checkDeviceExtensions(const std::vector<std::string> &Exts,
+                                    bool &Supported) {
```
kbenzie, line 79: This interface isn't very nice to use. Do we really need to return the `ur_result_t` here? Could it just return `false`?

This is another issue that can be looked at separately.

source/adapters/opencl/enqueue.cpp
```diff
@@ -30,14 +36,26 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
-      cl_adapter::cast<cl_kernel>(hKernel), workDim, pGlobalWorkOffset,
-      pGlobalWorkSize, pLocalWorkSize, numEventsInWaitList,
-      cl_adapter::cast<const cl_event *>(phEventWaitList),
-      cl_adapter::cast<cl_event *>(phEvent)));
-
+  cl_event Event;
+  std::vector<cl_event> CLWaitEvents(numEventsInWaitList);
+  for (uint32_t i = 0; i < numEventsInWaitList; i++) {
+    CLWaitEvents[i] = phEventWaitList[i]->get();
+  }
```
kbenzie, line 43: This might be a nice thing to make an abstraction for since it's used for every enqueue entry point.

source/adapters/opencl/enqueue.cpp
```diff
@@ -30,14 +36,26 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
+    try {
+      auto UREvent =
+          std::make_unique<ur_event_handle_t_>(Event, hQueue->Context, hQueue);
+      *phEvent = UREvent.release();
+    } catch (std::bad_alloc &) {
+      return UR_RESULT_ERROR_OUT_OF_RESOURCES;
+    } catch (...) {
+      return UR_RESULT_ERROR_UNKNOWN;
+    }
+  }
```
kbenzie, line 58: Likewise for this, an abstraction for this would clean up a lot of repetition.

source/adapters/opencl/platform.cpp
```diff
@@ -83,10 +66,38 @@ UR_APIEXPORT ur_result_t UR_APICALL
...
+            URPlatforms.emplace_back(URPlatform.release());
+          }
+        } catch (std::bad_alloc &) {
+          return CL_OUT_OF_RESOURCES;
+        } catch (...) {
+          return CL_INVALID_PLATFORM;
+        }
+        return Res;
+      },
+      Result);
```
kbenzie, line 100: Do we already have a `call_once` elsewhere? Might be better to combine them if we do.

source/adapters/opencl/queue.hpp
```diff
@@ -0,0 +1,89 @@
...
+      auto URQueue =
+          std::make_unique<ur_queue_handle_t_>(NativeQueue, Context, Device);
+      Queue = URQueue.release();
+    } catch (std::bad_alloc &) {
+      return UR_RESULT_ERROR_OUT_OF_RESOURCES;
+    } catch (...) {
+      return UR_RESULT_ERROR_UNKNOWN;
+    }
+    return UR_RESULT_SUCCESS;
+  }
```
kbenzie, line 70: Does this need to be a public interface or should this move to the `.cpp` file?

source/adapters/opencl/platform.cpp
omarahmed1111: I already refactored this inside adapter so no need for `call_once`.

source/adapters/opencl/enqueue.cpp
omarahmed1111: Done, not only for events but for all handles.

source/adapters/opencl/enqueue.cpp
omarahmed1111: Fixed.

source/adapters/opencl/device.hpp
omarahmed1111: Didn't look into this area much but I think this could be addressed in a follow-up that would redesign these extensions helper functions.

source/adapters/opencl/device.hpp
omarahmed1111: Yeah, good point, that would eliminate a lot of repetition.

source/adapters/opencl/context.hpp
omarahmed1111: Fixed

source/adapters/opencl/context.hpp
omarahmed1111: Removed all `get()` interface.

source/adapters/opencl/device.hpp
kbenzie: Something like this already exists in at least the Native CPU adapter. I'm currently figuring out where this should live as the laoder also has reference counted objects.

source/adapters/opencl/device.hpp
kbenzie: Makes sense.

kbenzie (approved):

martygrant:
## Information

## PR Migration

---

martygrant:

---

RossBrunton:
This change is now here: https://github.com/intel/llvm/pull/17118/commits/87fe1179c269a044d13b748c5a28a3916167bb5f

I'll recreate an MR in intel/llvm once prerequisites are merged.

PR #1178, opened 2023-12-12
[HIP] Remove inline from helper

hdelan:
Different compilers can change the linkage of symbols based on whether `inline` is used. This removes the `inline` specifier so it has external linkage and can be called from within another TU.

### DIFF
--- a/source/adapters/hip/memory.cpp
+++ b/source/adapters/hip/memory.cpp
@@ -458,9 +458,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemRetain(ur_mem_handle_t hMem) {
   return UR_RESULT_SUCCESS;
 }
 
-inline ur_result_t
-allocateMemObjOnDeviceIfNeeded(ur_mem_handle_t Mem,
-                               const ur_device_handle_t hDevice) {
+ur_result_t allocateMemObjOnDeviceIfNeeded(ur_mem_handle_t Mem,
+                                           const ur_device_handle_t hDevice) {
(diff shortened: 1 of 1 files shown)


npmiller (approved):

ldrumm (commented):
Now this function is visible, does it need to start with `ur` or some internal-private prefix?

@kbenzie will know

kbenzie:
> Now this function is visible, does it need to start with `ur` or some internal-private prefix?
> 
> @kbenzie will know

If this is only being made to have external linkage for use within the HIP adapter there is no need for it to have a `ur` preifx. We have linker scripts for stripping symbols which we don't want to be part of the official ABI of the adapters.

hdelan:
> Now this function is visible, does it need to start with `ur` or some internal-private prefix?
> 
> @kbenzie will know

This function already had external linkage with the compilers that I tested (gcc 11 and upstream clang) so it shouldn't need any change. Note that this only broke for gcc 12 since the func is called from another TU

ldrumm:
Thanks for clarifying

hdelan:
Ping @oneapi-src/owner-unified-runtime can we merge this?

kbenzie:
> Ping @oneapi-src/owner-unified-runtime can we merge this?

We are working though the [ready to merge](https://github.com/oneapi-src/unified-runtime/pulls?q=is%3Aopen+is%3Apr+label%3A%22ready+to+merge%22) backlog and will get to this as soon as we can.

PR #1181, opened 2023-12-12
Return device total global memory for MaxAllocSize

DuncanMcBain:
The SYCL specifiction inherits an OpenCL query for the maximum single allocaton that can be made, and requires that at least one-quarter of the device's reported global memory size can be allocated.

CUDA doesn't really have such a limitation as far as I can tell, and will happily allocate anything up to the total size of global memory in one go. This means returning that size is a reasonable guess for MaxAllocSize.

### DIFF
--- a/source/adapters/cuda/device.hpp
+++ b/source/adapters/cuda/device.hpp
@@ -68,17 +68,11 @@ public:
     }
 
     // Max size of memory object allocation in bytes.
-    // The minimum value is max(min(1024 × 1024 ×
-    // 1024, 1/4th of CL_DEVICE_GLOBAL_MEM_SIZE),
-    // 32 × 1024 × 1024) for devices that are not of type
-    // CL_DEVICE_TYPE_CUSTOM.
-    size_t Global = 0;
-    UR_CHECK_ERROR(cuDeviceTotalMem(&Global, cuDevice));
-
-    auto QuarterGlobal = static_cast<uint32_t>(Global / 4u);
-
-    MaxAllocSize = std::max(std::min(1024u * 1024u * 1024u, QuarterGlobal),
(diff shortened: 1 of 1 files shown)


source/adapters/cuda/device.hpp
```diff
@@ -64,17 +64,11 @@ struct ur_device_handle_t_ {
...
-    UR_CHECK_ERROR(cuDeviceTotalMem(&Global, cuDevice));
-
-    auto QuarterGlobal = static_cast<uint32_t>(Global / 4u);
-
-    MaxAllocSize = std::max(std::min(1024u * 1024u * 1024u, QuarterGlobal),
-                            32u * 1024u * 1024u);
+    // The minimum value is max (1/4th of info::device::global_mem_size,
+    // 128*1024*1024) if this SYCL device is not device_type::custom.
+    // CUDA doesn't really have this concept, and could allow almost 100% of
+    // global memory in one allocation, but is dependent on device usage.
```
mmoadeli, line 70: just curious for the reason to have a comment about the minimum value, when `MaxAllocSize` is populated.
would it be possible to have the minimum value as an assert, instead of comment?

DuncanMcBain: I think I can't use `UR_ASSERT` here because it's in a constructor.

That said, I've had a quick look at NVIDIA GPU specifications, and I'm struggling to find devices that don't actually have enough memory. Going back ten years to the 700 series (devices that support sm_50 as required by the compiler) shows devices with ~8x the 128 MB minimum from the SYCL spec, so I'm confident that anything the CUDA adapter sees will meet the requirement.

mmoadeli (approved):

PR #1183, opened 2023-12-13
Fix crash in urAdapterGet with non-initialized platforms

al42and:
Encountered in a bothched ASAN build, but makes sense to check platform status before trying to query it.


source/loader/ur_ldrddi.cpp
```diff
@@ -53,6 +53,9 @@ __urdlllocal ur_result_t UR_APICALL urAdapterGet(
     size_t adapterIndex = 0;
     if (nullptr != phAdapters && NumEntries != 0) {
         for (auto &platform : context->platforms) {
+            if (platform.initStatus != UR_RESULT_SUCCESS) {
```
pbalcer, line 56: this should be changed in the template instead: https://github.com/oneapi-src/unified-runtime/blob/main/scripts/templates/ldrddi.cpp.mako#L59

pbalcer (commented):
I don't think the fix is this simple. If there's only one adapter, we might never get into the loader's `urAdapterGet` implementation. But this is a step in the right direction.
 
Eventually, we might want to rewrite the way adapters are initialized, so that it's possible to just not have the broken adapters in the platforms collection.

source/loader/ur_ldrddi.cpp
al42and: Thanks, done!

pbalcer (approved):

PR #1184, opened 2023-12-13
[CTS] Don't use urProgramCreateWithIL on HIP adapter

kbenzie:
Since 5dc6e464c83fa41045d4f352a300db9b16400d1d CTS tests which execute
kernels have been failing to create programs. This has been hidden due
to the match files containing expected failures in combination with an
bug in the match file script found while debugging #1029 check failures.

This patch adds `uur::KernelsEnvironment::CreateProgram()` which
automatically handles the differences between the HIP adapter and
others. On the hip path `urProgramCreateWithIL` is not longer called
because it is not supported by the adatper and
`urProgramCreateWithBinary` is used instead. The non-HIP path continues
to use `urProgramCreateWithIL`.

### DIFF
--- a/test/conformance/program/urProgramCreateWithIL.cpp
+++ b/test/conformance/program/urProgramCreateWithIL.cpp
@@ -8,6 +8,15 @@
 struct urProgramCreateWithILTest : uur::urContextTest {
     void SetUp() override {
         UUR_RETURN_ON_FATAL_FAILURE(urContextTest::SetUp());
+        // TODO: This should use a query for urProgramCreateWithIL support or
+        // rely on UR_RESULT_ERROR_UNSUPPORTED_FEATURE being returned.
+        ur_platform_backend_t backend;
+        ASSERT_SUCCESS(urPlatformGetInfo(platform, UR_PLATFORM_INFO_BACKEND,
+                                         sizeof(ur_platform_backend_t),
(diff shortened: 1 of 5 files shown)


pbalcer (approved):

test/conformance/program/urProgramCreateWithIL.cpp
```diff
@@ -8,6 +8,15 @@
 struct urProgramCreateWithILTest : uur::urContextTest {
     void SetUp() override {
         UUR_RETURN_ON_FATAL_FAILURE(urContextTest::SetUp());
+        // TODO: This should use a query for urProgramCreateWithIL support or
```
fabiomestre, line 11: I guess this falls into the discussion of which entrypoints should be considered optional. Maybe we should create a github issue to update this later?

fabiomestre (approved):

test/conformance/program/urProgramCreateWithIL.cpp
kbenzie: Having an issue for these discussions makes a lot of sense, yes.

kbenzie: https://github.com/oneapi-src/unified-runtime/issues/1267

PR #1185, opened 2023-12-13
Fix coverity issues in OpenCL, cuda and hip adapters.

aarongreig:
LLVM PR https://github.com/intel/llvm/pull/12176

### DIFF
--- a/source/adapters/cuda/device.cpp
+++ b/source/adapters/cuda/device.cpp
@@ -1143,17 +1143,18 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceCreateWithNativeHandle(
   if (Result != UR_RESULT_SUCCESS)
     return Result;
 
-  ur_platform_handle_t *Plat = static_cast<ur_platform_handle_t *>(
-      malloc(NumPlatforms * sizeof(ur_platform_handle_t)));
-  Result = urPlatformGet(&AdapterHandle, 1, NumPlatforms, Plat, nullptr);
+  std::vector<ur_platform_handle_t> Platforms(NumPlatforms);
+
+  Result = urPlatformGet(&AdapterHandle, 1, Platforms.size(), Platforms.data(),
+                         nullptr);
   if (Result != UR_RESULT_SUCCESS)
     return Result;
 
   // Iterate through platforms to find device that matches nativeHandle
-  for (uint32_t j = 0; j < NumPlatforms; ++j) {
-    auto SearchRes =
-        std::find_if(begin(Plat[j]->Devices), end(Plat[j]->Devices), IsDevice);
-    if (SearchRes != end(Plat[j]->Devices)) {
+  for (auto Platform : Platforms) {
+    auto SearchRes = std::find_if(begin(Platform->Devices),
+                                  end(Platform->Devices), IsDevice);
+    if (SearchRes != end(Platform->Devices)) {
       *phDevice = static_cast<ur_device_handle_t>((*SearchRes).get());
       return UR_RESULT_SUCCESS;
     }
--- a/source/adapters/cuda/device.cpp
+++ b/source/adapters/cuda/device.cpp
@@ -1145,15 +1145,15 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceCreateWithNativeHandle(
 
   std::vector<ur_platform_handle_t> Platforms(NumPlatforms);
 
-  Result = urPlatformGet(&AdapterHandle, 1, Platforms.size(), Platforms.data(),
-                         nullptr);
+  Result =
+      urPlatformGet(&AdapterHandle, 1, NumPlatforms, Platforms.data(), nullptr);
   if (Result != UR_RESULT_SUCCESS)
     return Result;
 
   // Iterate through platforms to find device that matches nativeHandle
-  for (auto Platform : Platforms) {
-    auto SearchRes = std::find_if(begin(Platform->Devices),
-                                  end(Platform->Devices), IsDevice);
+  for (const auto Platform : Platforms) {
+    auto SearchRes = std::find_if(std::begin(Platform->Devices),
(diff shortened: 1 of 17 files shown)


source/adapters/hip/enqueue.cpp
```diff
@@ -1060,7 +1060,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemImageRead(
   } catch (...) {
     return UR_RESULT_ERROR_UNKNOWN;
   }
-  return UR_RESULT_SUCCESS;
   return Result;
```
kbenzie, line 1063: I think the wrong `return` has been removed here, this is a behaviour change.

source/adapters/hip/enqueue.cpp
```diff
@@ -1125,8 +1124,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemImageWrite(
     return UR_RESULT_ERROR_UNKNOWN;
   }
 
-  return UR_RESULT_SUCCESS;
-
   return Result;
```
kbenzie, line 1127: I think the wrong `return` has been removed here, this is a behaviour change.

isaacault (approved):
Bindless images changes LGTM

source/adapters/hip/enqueue.cpp
aarongreig: This pattern is used in a lot of places here, `Result` is what gets returned. On closer inspection though I don't like it, I've updated these two entry points and I'm working on a refactor for this whole file because I think the pattern potentially drops error codes.

EwanC (approved):
command-buffer changes LGTM

source/adapters/cuda/device.cpp
```diff
@@ -1143,17 +1143,18 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceCreateWithNativeHandle(
   if (Result != UR_RESULT_SUCCESS)
     return Result;
 
-  ur_platform_handle_t *Plat = static_cast<ur_platform_handle_t *>(
-      malloc(NumPlatforms * sizeof(ur_platform_handle_t)));
-  Result = urPlatformGet(&AdapterHandle, 1, NumPlatforms, Plat, nullptr);
+  std::vector<ur_platform_handle_t> Platforms(NumPlatforms);
+
+  Result = urPlatformGet(&AdapterHandle, 1, Platforms.size(), Platforms.data(),
+                         nullptr);
```
mmoadeli, line 1149: Could `NumPlatforms` be used instead of `Platforms.size()`?

source/adapters/cuda/device.cpp
```diff
@@ -1143,17 +1143,18 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceCreateWithNativeHandle(
...
 
   // Iterate through platforms to find device that matches nativeHandle
-  for (uint32_t j = 0; j < NumPlatforms; ++j) {
-    auto SearchRes =
-        std::find_if(begin(Plat[j]->Devices), end(Plat[j]->Devices), IsDevice);
-    if (SearchRes != end(Plat[j]->Devices)) {
+  for (auto Platform : Platforms) {
+    auto SearchRes = std::find_if(begin(Platform->Devices),
+                                  end(Platform->Devices), IsDevice);
+    if (SearchRes != end(Platform->Devices)) {
```
mmoadeli, line 1157: a minor inconsistency using namespace `std` for `find_if` but not for `begin()` and `end()`

source/adapters/cuda/event.hpp
```diff
@@ -82,8 +83,17 @@ struct ur_event_handle_t_ {
   static ur_event_handle_t
   makeNative(ur_command_t Type, ur_queue_handle_t Queue, CUstream Stream,
              uint32_t StreamToken = std::numeric_limits<uint32_t>::max()) {
-    return new ur_event_handle_t_(Type, Queue->getContext(), Queue, Stream,
-                                  StreamToken);
+    bool ProfilingEnabled = Queue->URFlags & UR_QUEUE_FLAG_PROFILING_ENABLE;
```
mmoadeli, line 86: `const bool`?

source/adapters/cuda/program.cpp
```diff
@@ -137,7 +137,7 @@ ur_result_t ur_program_handle_t_::buildProgram(const char *BuildOptions) {
 
   if (!this->BuildOptions.empty()) {
     unsigned int MaxRegs;
-    bool Valid = getMaxRegistersJitOptionValue(BuildOptions, MaxRegs);
+    bool Valid = getMaxRegistersJitOptionValue(this->BuildOptions, MaxRegs);
```
mmoadeli, line 140: `const bool`?

source/adapters/cuda/device.cpp
```diff
@@ -1143,17 +1143,18 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceCreateWithNativeHandle(
...
+                         nullptr);
   if (Result != UR_RESULT_SUCCESS)
     return Result;
 
   // Iterate through platforms to find device that matches nativeHandle
-  for (uint32_t j = 0; j < NumPlatforms; ++j) {
-    auto SearchRes =
-        std::find_if(begin(Plat[j]->Devices), end(Plat[j]->Devices), IsDevice);
-    if (SearchRes != end(Plat[j]->Devices)) {
+  for (auto Platform : Platforms) {
```
mmoadeli, line 1154: `(const auto &Platform : Platforms)`?

source/adapters/cuda/device.cpp
aarongreig: we just constructed `Platforms` the line before, and yeah reusing `NumPlatforms` is slightly more efficient than getting the size back out of the vector

mmoadeli (approved):
LGTM hip/cuda.

PR #1190, opened 2023-12-14
[HIP] Refactor error handling in enqueue.cpp

aarongreig:
LLVM testing https://github.com/intel/llvm/pull/12481

### DIFF
--- a/source/adapters/hip/enqueue.cpp
+++ b/source/adapters/hip/enqueue.cpp
@@ -625,7 +625,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferCopy(
   std::unique_ptr<ur_event_handle_t_> RetImplEvent{nullptr};
 
   try {
-    ScopedContext Active(hQueue->getDevice());
+    ScopedDevice Active(hQueue->getDevice());
     auto Stream = hQueue->getNextTransferStream();
 
     if (phEventWaitList) {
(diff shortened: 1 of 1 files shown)


hdelan (approved):
LGTM

jchlanda (approved):
Really like this, needless accumulation of error codes in `Result`-like variables, instead of straight out returns, and not using `UR_CHECK_ERROR` are one of my top pet hates. Thank you!

aarongreig:
@npmiller please take a look when you can

npmiller (approved):
LGTM, sorry about the late review!

source/adapters/hip/enqueue.cpp
```diff
@@ -1516,10 +1500,18 @@ urEnqueueUSMAdvise(ur_queue_handle_t hQueue, const void *pMem, size_t size,
...
         releaseEvent();
-        setErrorMessage("mem_advise is ignored as the advice argument is not "
-                        "supported by this device",
-                        UR_RESULT_SUCCESS);
-        return UR_RESULT_ERROR_ADAPTER_SPECIFIC;
+        // UR_RESULT_ERROR_INVALID_ENUMERATION is returned when using a valid
+        // but currently unmapped advice arguments as not supported by this
+        // platform. Therefore, warn the user instead of throwing and aborting
+        // the runtime.
+        if (Result == UR_RESULT_ERROR_INVALID_ENUMERATION) {
```
GeorgeWeb, line 1507: I may be missing something here but is this `if (Result == UR_RESULT_ERROR_INVALID_ENUMERATION)` meant to be checked again inside the above `if (Result == UR_RESULT_ERROR_INVALID_ENUMERATION)`? It is already true in this case. Is it meant to be wrapped in `if (Result != UR_RESULT_SUCCESS)` -> check if invalid else `throw`?

GeorgeWeb (commented):
Generally looks good. Thank you for cleaning that up!
Just a little comment I was confused about.

source/adapters/hip/enqueue.cpp
aarongreig: I think I messed up the rebase here, now it just falls through to `UR_CHECK_ERROR(Result)` if we aren't doing the adapter specific thing

GeorgeWeb: Yeah that makes sense to me now. Thanks, I actually thuoght it must've been a dodgy rebase by the looks of it.

GeorgeWeb (approved):

PR #1192, opened 2023-12-14
Initial support for sycl_ext_oneapi_composite_device

maarquitos14:
Initial implementation to support `sycl_ext_oneapi_composite_device` specified in https://github.com/intel/llvm/pull/11846.


include/ur_api.h
```diff
@@ -1562,6 +1562,10 @@ typedef enum ur_device_info_t {
                                                                     ///< semaphore resources
     UR_DEVICE_INFO_INTEROP_SEMAPHORE_EXPORT_SUPPORT_EXP = 0x200F,   ///< [::ur_bool_t] returns true if the device supports exporting internal
                                                                     ///< event resources
+    UR_DEVICE_INFO_COMPONENT_DEVICES = 0x2010,                      ///< [::ur_device_handle_t[]] returns the set of component devices contained
```
aarongreig, line 1565: Could you add these to the spec yml here https://github.com/oneapi-src/unified-runtime/blob/dcec3feb7912be6e6fb1a8ef268b2d61afe759c5/scripts/core/device.yml#L153 and run the `generate` target to make these changes? It would also be useful to have these in the DeviceGetInfo test https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/device/urDeviceGetInfo.cpp#L114

maarquitos14: @aarongreig Sure, I can do that, but I'm a bit confused because I don't see the immediately preceding one (`UR_DEVICE_INFO_INTEROP_SEMAPHORE_EXPORT_SUPPORT_EXP`) in that file. Also, I just tried `make generate` and it requires `clang-format-15`, is that right? Or is it 15 or newer?

aarongreig: Ah right, yeah the _EXP enums are generated from separate EXP (for experimental) yml files like [this one](https://github.com/oneapi-src/unified-runtime/blob/67e4d1bcf0057eef7910851bfb5020df9e5befd6/scripts/core/exp-bindless-images.yml#L83), whereas your change is going directly into the core spec. The generate scripts will figure out the order and stuff for you.

Yeah we do require clang-format-15 just to prevent noisy diffs due to changes in the tool's opinion on things between versions

maarquitos14: @aarongreig this PR's goal is to support a new extension, it's not a core SYCL feature. Just pointing it out in case it matters  in terms of what file I should update. I'm going for `device.yml` unless you say otherwise. Thank you for your help!

maarquitos14: I just added a commit with the changes made by `make generate` but, as you can see, it changed the whole `/include/ur_print.hpp` file. Did I do it wrong?

aarongreig: it looks (based on the CI) like your `generate` run formatted a lot of `ur_print.hpp` differently, maybe cmake picked up the wrong clang-format version?

maarquitos14: I don't think so, I didn't have `clang-format-15` and cmake complained about it. I installed it, and after that, it worked, and I could run `make generate`. I just double checked and the output from cmake includes this:

`-- Found clang-format: /iusers2/mmaronas/tools/llvm-project/install/bin/clang-format (version: 15.0)`.

pbalcer: the build is failing because of the formatting changes. You need to have exactly the right version - `clang-format-15.0.7`. This is included in python prerequisites, so you should have it if you follow instructions [here](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#build-environment).

maarquitos14: @pbalcer I have installed the python prerequisites as instructed:
`-- Found clang-format: /iusers2/mmaronas/sycl_workspace/llvm-bld/_deps/unified-runtime-src/.local/bin/clang-format (version: 15.0)`.
 Then, ran `make generate` again and pushed the results in https://github.com/oneapi-src/unified-runtime/pull/1192/commits/84a8b392ebdb08e10fe224390769b19eb5d20f35. It didn't change a single line in `ur_print.hpp`.

GeorgeWeb (approved):
HIP and CUDA adapter changes look good to me.

include/ur_api.h
aarongreig: maybe the format step is somehow silently failing to run, could you try formatting the changes that `generate` makes manually to see if what you get is any different?

maarquitos14: @aarongreig I think you might be right. I think the problem is that the target doesn't complete because I don't have `doxygen`, so it never runs clang-format. I'll try and get `doxygen` and see if that fixes it. Thanks both (@pbalcer) for your help!

maarquitos14: @aarongreig @pbalcer I could run `generate` to completion, and now the diff looks reasonable. I also updated the test you mentioned.

source/adapters/level_zero/device.cpp
```diff
@@ -69,8 +69,16 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGet(
       urPrint("Unknown device type");
       break;
     }
-    if (Matched)
+    // Filter out composite devices when ZE_FLAT_DEVICE_HIERARCHY=COMBINED:
+    // component devices (tiles) enable ZE_DEVICE_PROPERTY_FLAG_SUBDEVICE while
+    // composite devices (cards) don't.
+    bool isComposite =
+        (D->ZeDeviceProperties->flags & ZE_DEVICE_PROPERTY_FLAG_SUBDEVICE) == 0;
+    const char *Mode = std::getenv("ZE_FLAT_DEVICE_HIERARCHY");
```
pbalcer, line 77: `ur_getenv`

But a broader question, should UR read environment variables defined by other components? The environment can change between two getenv calls, resulting in potential subtle bugs.

maarquitos14: I understand the concern, but as far as I know, this envvar should never change once the application has started. I'm open to suggestions here, but I don't see a better way to deal with this.

pbalcer: > envvar should never change once the application has started

```
man 3 getenv

The setenv() function adds the variable name to the environment with the value value .... 
If name does exist in the environment, then its value is changed to value if overwrite is nonzero.
```

```
man 3 putenv

The putenv() function adds or changes the value of environment variables.
The argument string is of the form name=value. ...
If name does exist, then the value of name in the environment is changed to value.
```

The environment change could be unintentional. This is a TOCTOU-like issue.

> I'm open to suggestions here

Ideally L0 would expose what is the device hierarchy it was configured with. Alternatively, maybe it would be possible to infer the hierarchy based on the behavior of `zeGetDevice`/`zeDeviceGetSubDevice` as defined [here](https://spec.oneapi.io/level-zero/latest/core/PROG.html#device-hierarchy)?

maarquitos14: @pbalcer I have just pushed a commit that removes all readings of `ZE_FLAT_DEVICE_HIERARCHY`. Let me know if that addresses your concerns.

maarquitos14:
@pbalcer @aarongreig I'm unable to reproduce the CI errors in my local machine. Any tips on how to reproduce?

source/adapters/level_zero/device.cpp
```diff
@@ -69,8 +70,49 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGet(
...
+    //     the card containing a given tile.
+    //
+    // NOTE: We cannot directly filter out the composite devices here because we
+    // might have the composite device appear earlier than we know we are in
+    // combined mode, so we simply try and infer here if we are in combined
+    // mode, and then, if so, remove composite devices from MatchedDevices.
+    if (!isCombinedMode) {
+      ze_device_handle_t RootDev = nullptr;
+      // Query Root Device
+      ZE2UR_CALL(zeDeviceGetRootDevice, (D->ZeDevice, &RootDev));
```
kbenzie, line 91: Is this a new entry point? I'm wondering if the GitHub Actions failures could be due to an older driver being installed or something like that?

kbenzie:
> @pbalcer @aarongreig I'm unable to reproduce the CI errors in my local machine. Any tips on how to reproduce?

I believe we test L0 on a machine with an A750.

maarquitos14:
> > @pbalcer @aarongreig I'm unable to reproduce the CI errors in my local machine. Any tips on how to reproduce?
> 
> I believe we test L0 on a machine with an A750.

Thanks @kbenzie. @pbalcer gave me access to a machine with such hardware. I'm trying to reproduce the issue in that machine now.

source/adapters/level_zero/device.cpp
maarquitos14: @kbenzie It is a new entry point, but as far as I see, the CI uses level zero v1.15.1 which includes this entry. What I'm seeing is that `zeDeviceGetRootDevice` returns `ZE_RESULT_ERROR_UNSUPPORTED_FEATURE` under certain circumstances (https://github.com/oneapi-src/level-zero/blob/1685d01497428ca4d8b99200972b64685424d5c9/source/lib/ze_libapi.cpp#L463). The CI is running into this, and that makes `ZE2UR_CALL` fail in `UR_CHECK`. I'm not sure how to deal with this, honestly. As of now, I think only PVC should care about this, but I don't think we should have code specific to PVC in UR. Otherwise, we can use `ZE_CALL_NOCHECK` rather than `ZE2UR_CALL` and check manually that the error code is either success or unsupported feature. What do you think? Any other suggestion?

maarquitos14: Btw, I just checked locally using `ZE_CALL_NOCHECK` and all the tests pass.

maarquitos14: Done in https://github.com/oneapi-src/unified-runtime/pull/1192/commits/15d2953dd89d60c615b8ab596c6e2761eba83aec

maarquitos14:
@oneapi-src/unified-runtime-level-zero-write @oneapi-src/unified-runtime-native-cpu-write can I have a review please?

PietroGhg (approved):
Native CPU lgtm, sorry for the late review

source/adapters/level_zero/device.cpp
```diff
@@ -40,6 +40,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGet(
   // Filter available devices based on input DeviceType.
   std::vector<ur_device_handle_t> MatchedDevices;
   std::shared_lock<ur_shared_mutex> Lock(Platform->URDevicesCacheMutex);
+  bool isCombinedMode = false;
```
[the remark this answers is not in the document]
maarquitos14: Done in https://github.com/oneapi-src/unified-runtime/pull/1192/commits/a529440de69a236daab31541c02dffaa8bca3009.

source/adapters/level_zero/device.cpp
```diff
@@ -823,6 +874,58 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
...
+    if (SubDeviceCount == 0)
+      return ReturnValue(0);
+
+    std::vector<ze_device_handle_t> SubDevs(SubDeviceCount);
+    // Second call to get the actual list of devices.
+    ZE2UR_CALL(zeDeviceGetSubDevices,
+               (DevHandle, &SubDeviceCount, SubDevs.data()));
+
+    size_t SubDeviceCount_s{SubDeviceCount};
+    auto ResSize = std::min(SubDeviceCount_s, propSize);
```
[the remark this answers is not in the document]
maarquitos14: Good catch, thanks! Fixed in https://github.com/oneapi-src/unified-runtime/pull/1192/commits/a529440de69a236daab31541c02dffaa8bca3009.

source/adapters/level_zero/device.cpp
```diff
@@ -823,6 +874,58 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
...
+      return ReturnValue(Res.data(), ResSize);
+    }
+    return UR_RESULT_SUCCESS;
+  }
+  case UR_DEVICE_INFO_COMPOSITE_DEVICE: {
+    ur_device_handle_t UrRootDev = nullptr;
+    ze_device_handle_t DevHandle = Device->ZeDevice;
+    ze_device_handle_t RootDev;
+    // Query Root Device
+    ZE2UR_CALL(zeDeviceGetRootDevice, (DevHandle, &RootDev));
```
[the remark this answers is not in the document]
maarquitos14: Didn't notice this one! Thanks! Fixed in https://github.com/oneapi-src/unified-runtime/pull/1192/commits/a529440de69a236daab31541c02dffaa8bca3009.

source/adapters/level_zero/device.cpp
```diff
@@ -69,8 +70,58 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGet(
       urPrint("Unknown device type");
       break;
     }
-    if (Matched)
+
+    // We need to filter out composite devices depending on
+    // ZE_FLAT_DEVICE_HIERARCHY value:
+    //   - If COMPOSITE, L0 returns cards as devices. Thus, zeGetRootDevice must
```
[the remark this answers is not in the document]
maarquitos14: What I understand is that zeDeviceGet returns the devices that do not have a root-device. Then, you can get sub-devices with zeDeviceGetSubDevices, and from those sub-devices, you can get their root-device with zeDeviceGetRootDevice. This is also the behaviour I see in my testing.

source/adapters/level_zero/device.cpp
```diff
@@ -40,6 +40,36 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGet(
...
+                    // Query Root Device for root-devices.
+                    // We cannot use ZE2UR_CALL because under some circumstances
+                    // this call may return ZE_RESULT_ERROR_UNSUPPORTED_FEATURE,
+                    // and ZE2UR_CALL will abort because it's not
+                    // UR_RESULT_SUCCESS. Instead, we use ZE_CALL_NOCHECK and we
+                    // check manually that the result is either
+                    // ZE_RESULT_SUCCESS or ZE_RESULT_ERROR_UNSUPPORTED_FEATURE.
+                    auto errc = ZE_CALL_NOCHECK(zeDeviceGetRootDevice,
+                                                (D->ZeDevice, &RootDev));
+                    assert(errc == ZE_RESULT_SUCCESS ||
```
pbalcer, line 69: [Specs](https://spec.oneapi.io/level-zero/latest/core/api.html#zedevicegetrootdevice) says that there are more allowed return values, so we probably shouldn't assert.

Maybe `if (errc != SUCCESS) return false` ?

pbalcer (approved):
lgtm, just one minor nit.

source/adapters/level_zero/device.cpp
maarquitos14: I can remove the assert and simply return false when `errc != SUCCESS`, but I think it may hide errors. However, we have a couple more calls to `zeDeviceGetRootDevice`, so if there's a new (unknown to us) error, we will catch it there, so I think we can go ahead with your suggestion.

maarquitos14: Done in https://github.com/oneapi-src/unified-runtime/pull/1192/commits/a9746c217adbbe2adcd739261a00ea8a60da5807.

nrspruit (approved):
LGTM for level zero

maarquitos14:
I think this is ready to merge @oneapi-src/unified-runtime-maintain.

aarongreig:
could you update https://github.com/intel/llvm/pull/12178/files#diff-1bf2ed05c0065126d4e2150f0eda4dbd561e5c3625452e96356aadafa1312e6fR64 to point at https://github.com/maarquitos14/unified-runtime/tree/maronas/ext_composite_device so the changes here get tested @maarquitos14 (assuming that PR is going to need to pull the changes in this PR in)

maarquitos14:
> could you update https://github.com/intel/llvm/pull/12178/files#diff-1bf2ed05c0065126d4e2150f0eda4dbd561e5c3625452e96356aadafa1312e6fR64 to point at https://github.com/maarquitos14/unified-runtime/tree/maronas/ext_composite_device so the changes here get tested @maarquitos14 (assuming that PR is going to need to pull the changes in this PR in)

Done in https://github.com/intel/llvm/pull/12178/commits/c55903641aa2c7fffd183572a71fb3d156f77e2d.

aarongreig:
thanks, you'll need to make sure the CI on intel/llvm side is green as well before we can merge this

maarquitos14:
> thanks, you'll need to make sure the CI on intel/llvm side is green as well before we can merge this

CI on intel/llvm is finally all green. How should we proceed now?

aarongreig:
The ready to merge label represents a queue, when we get to this PR we'll just merge it (assuming no conflicts or whatever come up between now and then) and then ping you on the llvm PR to update the UR tag. You should try to get approvals on the LLVM PR so it doesn't hold up subsequent merges when this PR goes in

PR #1193, opened 2023-12-14
[HIP] Support managed allocations in USM free

npmiller:
In ROCm 5.7.1 some USM allocations are reported as managed, these are freed in the same way as device allocations.

Testing in https://github.com/intel/llvm/pull/12380

### DIFF
--- a/source/adapters/hip/usm.cpp
+++ b/source/adapters/hip/usm.cpp
@@ -78,9 +78,10 @@ USMFreeImpl([[maybe_unused]] ur_context_handle_t hContext, void *pMem) {
 #else
     const auto Type = hipPointerAttributeType.memoryType;
 #endif
-    UR_ASSERT(Type == hipMemoryTypeDevice || Type == hipMemoryTypeHost,
+    UR_ASSERT(Type == hipMemoryTypeDevice || Type == hipMemoryTypeHost ||
+                  Type == hipMemoryTypeManaged,
               UR_RESULT_ERROR_INVALID_MEM_OBJECT);
-    if (Type == hipMemoryTypeDevice) {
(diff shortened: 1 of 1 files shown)


GeorgeWeb:
@npmiller This PR is fine as is. I just have a question that ties into the now finally used mananged memory type from ROCm v5.7.1 onwards. So I am wondering whether we need to address this also in `urUSMGetMemAllocInfo` where `hipMemoryTypeManaged` also isn't being checked as a valid memory object type.

npmiller:
It does check this:

```
      Value = hipPointerAttributeType.isManaged;
      if (Value) {
        // pointer to managed memory
        return ReturnValue(UR_USM_TYPE_SHARED);
      }
```

I'm not sure if it's exactly the same thing though

GeorgeWeb:
> It does check this:
> 
> ```
>       Value = hipPointerAttributeType.isManaged;
>       if (Value) {
>         // pointer to managed memory
>         return ReturnValue(UR_USM_TYPE_SHARED);
>       }
> ```
> 
> I'm not sure if it's exactly the same thing though

The `isManaged` member is set for pointers to managed memory via the unified memory allocator is HIP so it should be okay yeah. My bad I didn't spot the `if (Value)` was checking `isManaged`.

EwanC:
I bumped into an issue locally with shared USM while trying out HIP support for SYCL-Graphs on ROCM 5.7.1, and this patch fixed it, would be helpful to get this merged (or tagged ready-to-merge)

ldrumm (approved):

kbenzie:
Rebasing should resolve the unrelated check failures.

PR #1194, opened 2023-12-14
[HIP] Explicitly specify copy direction for USM 2D async memory copies

GeorgeWeb:
There is an issue with `hipMemcpy2D*` when `hipMemcpyDefault` is used, which makes the HIP runtime not correctly derive the copy kind (direction) for the copies introduced since **ROCm 5.6.0**. See: https://github.com/ROCm/clr/issues/40.

This PR aims to provide a work-around in the meantime even though having to introduce the overhead of `hipPointerGetAttributes` in the memcpy2D. It patches it to work for the all possible copy kind directions (host / device / os-allocated host pageable memory) explicitly. Ideally, ROCm will fix the API when using just `hipMemcpyDefault` going forward.

**Testing:** [intel/llvm/pull/12200](https://github.com/intel/llvm/pull/12200)

### DIFF
--- a/source/adapters/hip/enqueue.cpp
+++ b/source/adapters/hip/enqueue.cpp
@@ -1666,6 +1666,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
     // There is an issue with hipMemcpy2D* when hipMemcpyDefault is used, which
     // makes the HIP runtime not correctly derive the copy kind (direction) for
     // the copies since ROCm 5.6.0+. See: https://github.com/ROCm/clr/issues/40
+    // TODO: Add maximum HIP_VERSION when bug has been fixed.
 #if HIP_VERSION >= 50600000
     hipPointerAttribute_t srcAttribs{};
     hipPointerAttribute_t dstAttribs{};
--- a/source/adapters/hip/enqueue.cpp
+++ b/source/adapters/hip/enqueue.cpp
@@ -1663,8 +1663,56 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
       UR_CHECK_ERROR(RetImplEvent->start());
     }
 
+    // There is an issue with hipMemcpy2D* when hipMemcpyDefault is used, which
+    // makes the HIP runtime not correctly derive the copy kind (direction) for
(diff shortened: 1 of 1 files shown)


hdelan:
An `#ifdef` for ROCm version might be a good idea, as well as a comment that the `hipMemcpyDefault` is broken in HIP. (Maybe even link this https://github.com/ROCm/clr/issues/40 in the comment)

GeorgeWeb:
> An `#ifdef` for ROCm version might be a good idea, as well as a comment that the `hipMemcpyDefault` is broken in HIP. (Maybe even link this [ROCm/clr#40](https://github.com/ROCm/clr/issues/40) in the comment)

Update this now, changes applicable to using ROCm HIP version 5.6.0 and above only.

mmoadeli (commented):
LGTM. 
A minor issue might be using empty curly braces on declrations, which seems inconsistent with the rest of the file.

hdelan:
Why did you close this @GeorgeWeb ? We could still use this fix

GeorgeWeb:
> Why did you close this @GeorgeWeb ? We could still use this fix

Didn't mean to actually. Re-opening now.

GeorgeWeb:
Pinging @npmiller @hdelan @mmoadeli for final thoughts. Thanks!

Based on discussion in the original issue raised in ROCm/clr it seems this workaround will be needed as the fix is not going to be cherry-picked into older releases. Also it is unclear which of the next releases will have the fix, when that is clear this patch can be ifdef'd for the range of releases starting from ROCm 5.6.0 until ROCm 6.x that has their fix.

hdelan (approved):
LGTM. I might just add a comment to say `TODO: Add HIP max version when bug has been fixed` just at the `ifdef`

GeorgeWeb:
> `TODO: Add HIP max version when bug has been fixed`

Yeah makes sense. Easy to lose track of these.

npmiller:
> Pinging for final thoughts. Thanks!
> 
> Based on discussion in the original issue raised in ROCm/clr it seems this workaround will be needed as the fix is not going to be cherry-picked into older releases. Also it is unclear which of the next releases will have the fix, when that is clear this patch can be ifdef'd for the range of releases starting from ROCm 5.6.0 until ROCm 6.x that has their fix.

That reasoning makes sense to me

npmiller (approved):

PR #1196, opened 2023-12-15
Replace PROGRAM_INFO_SOURCE with PROGRAM_INFO_IL.

### LINKED ISSUE #1138: Consider adding PROGRAM_INFO_IL
Right now to query out the makings of a given program we have PROGRAM_INFO_SOURCE. Cuda and hip interpret this query as meaning "whatever pre-device-binary material you have for the program" and they return their stored binary - that could be a number of things (ptx, fat ELF binary, etc.) of which none are a source language. OpenCL and level zero interpret it literally: level zero will hit an error if you request program source as it doesn't consume anything that could be considered a source language, OpenCL translates it directly into the CL equivalent which behaves as you would expect.

If we added a PROGRAM_INFO_IL query all the adapters would be able to reliably return something sensible for it, and the CL adapter would still retain the ability to return a program's source code in the unlikely scenario that it's been given ownership of a native program handle that actually was created from source. Otherwise we should just rename PROGRAM_INFO_SOURCE to _IL so it can be useful for all adapters and unambiguous.

aarongreig:
Also make returns for this consistent across adapter implementations (i.e. actual IL if that's supported, otherwise return UNSUPPORTED_ENUMERATION).

resolves #1138

LLVM testing https://github.com/intel/llvm/pull/12265

### DIFF
--- a/source/adapters/cuda/program.cpp
+++ b/source/adapters/cuda/program.cpp
@@ -399,8 +399,6 @@ urProgramGetInfo(ur_program_handle_t hProgram, ur_program_info_t propName,
     return ReturnValue(1u);
   case UR_PROGRAM_INFO_DEVICES:
     return ReturnValue(&hProgram->Device, 1);
-  case UR_PROGRAM_INFO_SOURCE:
-    return ReturnValue(hProgram->Binary);
   case UR_PROGRAM_INFO_BINARY_SIZES:
     return ReturnValue(&hProgram->BinarySizeInBytes, 1);
   case UR_PROGRAM_INFO_BINARIES:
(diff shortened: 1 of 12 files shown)


fabiomestre (approved):
LGTM

source/adapters/cuda/program.cpp
```diff
@@ -382,6 +380,7 @@ urProgramGetInfo(ur_program_handle_t hProgram, ur_program_info_t propName,
     UR_ASSERT(getKernelNames(hProgram), UR_RESULT_ERROR_UNSUPPORTED_FEATURE);
     return UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION;
   case UR_PROGRAM_INFO_NUM_KERNELS:
+  case UR_PROGRAM_INFO_IL:
```
fabiomestre, line 383: I'm wondering, if we wanted to implement this for CUDA, we would have to find way to figure out which binaries are PTX right?

aarongreig: yeah, so it's doable but we should probably wait for a use case to come up before going to the trouble

hdelan (approved):

aarongreig:
@oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-level-zero-write please take a look when you can

aarongreig:
ping @oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-level-zero-write

PietroGhg (approved):
Native CPU lgtm, thank you

aarongreig:
ping @oneapi-src/unified-runtime-level-zero-write

pbalcer (approved):

PR #1198, opened 2023-12-16
[HIP] Fix build with ROCm 6.0.0

al42and:
I see some unit test failures, but there were quite a lot of known issues with earlier ROCm versions too.

### DIFF
--- a/source/adapters/hip/common.hpp
+++ b/source/adapters/hip/common.hpp
@@ -15,23 +15,38 @@
 #include <hip/hip_runtime.h>
 #include <ur/ur.hpp>
 
-// Hipify doesn't support cuArrayGetDescriptor, on AMD the hipArray can just be
-// indexed, but on NVidia it is an opaque type and needs to go through
-// cuArrayGetDescriptor so implement a utility function to get the array
-// properties
-inline void getArrayDesc(hipArray *Array, hipArray_Format &Format,
-                         size_t &Channels) {
+// Before ROCm 6, hipify doesn't support cuArrayGetDescriptor, on AMD the
+// hipArray can just be indexed, but on NVidia it is an opaque type and needs to
+// go through cuArrayGetDescriptor so implement a utility function to get the
+// array properties
+inline static hipError_t getArrayDesc(hipArray *Array, hipArray_Format &Format,
+                                      size_t &Channels) {
+#if HIP_VERSION_MAJOR >= 6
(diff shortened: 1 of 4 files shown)


source/adapters/hip/common.hpp
```diff
@@ -15,24 +15,39 @@
...
-                         size_t &Channels) {
+// Before ROCm 6, hipify doesn't support cuArrayGetDescriptor, on AMD the
+// hipArray can just be indexed, but on NVidia it is an opaque type and needs to
+// go through cuArrayGetDescriptor so implement a utility function to get the
+// array properties
+inline static hipError_t getArrayDesc(hipArray *Array, hipArray_Format &Format,
+                                      size_t &Channels) {
+#if HIP_VERSION_MAJOR >= 6
+  HIP_ARRAY_DESCRIPTOR ArrayDesc;
+  hipError_t err = hipArrayGetDescriptor(&ArrayDesc, Array);
```
hdelan, line 26: Do we need to use `hipArrayGetDescriptor` for ROCm 6+ or can we still rely on the previous behaviour?

al42and: In ROCm 6, `HIP_ARRAY_DESCRIPTOR` is a pointer to a forward-declared type, so once cannot access its fields.

source/adapters/hip/usm.cpp
```diff
@@ -72,7 +72,11 @@ USMFreeImpl([[maybe_unused]] ur_context_handle_t hContext, void *pMem) {
   try {
     hipPointerAttribute_t hipPointerAttributeType;
     UR_CHECK_ERROR(hipPointerGetAttributes(&hipPointerAttributeType, pMem));
-    unsigned int Type = hipPointerAttributeType.memoryType;
+#if HIP_VERSION_MAJOR >= 6
```
GeorgeWeb, line 75: Judging by the HIP headers and release docs, the `memoryType` field name has been deprecated since ROCm 5.6.0, so use of the field name `type` is encouraged since. I think it makes sense to guard in a #if HIP_VERSION >= 50600000. Nothing wrong with only using it from version 6.0 but I think the above suggestion will benefit us to avoid nesting ROCm > 5 with ROCm > 6 in some cases in the source code.

al42and: Good point, agree. Amended the commit.

GeorgeWeb: nice

hdelan (approved):
LGTM

PR #1199, opened 2023-12-18
[draft] Add new experimental entry point for get native mem on device

hdelan:
Accessor interop requires `urMemGetNativeHandle` to return a valid mem handle. Multi dev ctx breaks this because a UR mem can be on many different devices now, so we need a new entry point that also takes a device or a queue param. 

Will wait until @kbenzie is back to discuss the best approach

### DIFF
--- a/scripts/core/memory.yml
+++ b/scripts/core/memory.yml
@@ -440,6 +440,34 @@ returns:
     - $X_RESULT_ERROR_UNSUPPORTED_FEATURE:
         - "If the adapter has no underlying equivalent handle."
 --- #--------------------------------------------------------------------------
+type: function
+desc: "Return platform native mem handle."
+class: $xMem
+name: GetNativeHandleExp
+decl: static
+ordinal: "0"
+details:
+    - "Retrieved native handle can be used for direct interaction with the native platform driver."
+    - "Use interoperability platform extensions to convert native handle to native type."
+    - "The application may call this function from simultaneous threads for the same context."
+    - "The implementation of this function should be thread-safe."
+params:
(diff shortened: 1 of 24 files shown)


scripts/core/memory.yml
```diff
@@ -440,6 +440,34 @@ returns:
...
+    - "The implementation of this function should be thread-safe."
+params:
+    - type: $x_mem_handle_t
+      name: hMem
+      desc: |
+            [in] handle of the mem.
+    - type: $x_device_handle_t
+      name: hDevice
+      desc: |
+            [in] handle of the device that native handle will be resident on.
```
kbenzie, line 462: I think we should just modify the existing `GetNativeHandle` to add this argument.

hdelan: Is that not an ABI breaking change that will take some time to get merged?

kbenzie: We don't claim ABI stability in UR at this time, this won't be an issue as long as the necessary SYCL RT changes are made in tandem with this.

hdelan:
Hi @kbenzie this entry point is needed for both accessor interop as well as buffer interop. So do you think it's OK to have the same entry point for both. For accessor interop we can call `urMemGetNativeHandle(Mem, Dev, &Handle)` and for the buffer interop we just call the same but with the Dev arg as `nullptr`? 

Do you think that is acceptable or should we split this into two entry points, as this PR does?

kbenzie:
I'm not really sure what the implications are here, would need to have a bit more info about why the buffer interop path differs I think.

PR #1203, opened 2023-12-19
[cuda][null][common] fix a few coverity issues

pbalcer:
This should fix all non-test related coverity issues except for:
https://scan7.scan.coverity.com/reports.htm#v21214/p15954/fileInstanceId=155217701&defectInstanceId=19136875&mergedDefectId=1574236, which is a race in the HIP adapter - that looks like it might be a false positive, but I'm not sure.

### DIFF
--- a/source/adapters/cuda/event.cpp
+++ b/source/adapters/cuda/event.cpp
@@ -12,6 +12,8 @@
 #include "context.hpp"
 #include "device.hpp"
 #include "queue.hpp"
+#include "ur_api.h"
+#include "ur_util.hpp"
 
 #include <cassert>
 #include <cuda.h>
@@ -65,7 +67,7 @@ ur_result_t ur_event_handle_t_::start() {
   return Result;
 }
 
-bool ur_event_handle_t_::isCompleted() const noexcept {
+bool ur_event_handle_t_::isCompleted() const noexcept try {
   if (!IsRecorded) {
     return false;
   }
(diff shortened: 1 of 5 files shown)


pbalcer:
https://github.com/intel/llvm/pull/12207

source/adapters/cuda/event.cpp
```diff
@@ -80,6 +82,8 @@ bool ur_event_handle_t_::isCompleted() const noexcept {
     }
   }
   return true;
+} catch (...) {
+  return exceptionToResult(std::current_exception()) == UR_RESULT_SUCCESS;
```
aarongreig, line 86: rather than catching we could just remove the UR_CHECK_ERROR and leave it to hit `return false`, we aren't getting any more information out of the function by comparing against the UR error code as well

pbalcer: I was about to do this change, but it seems that the `UR_CHECK_ERROR` macro prints out the error message:
https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/cuda/common.cpp#L38.
Should I remove it regardless?

aarongreig: ah, we should probably leave it in then

aarongreig (approved):

wlemkows (approved):

mmoadeli (approved):

PR #1206, opened 2023-12-21
Werror fixes

ProGTX:
A bunch of fixes to get `-Werror` builds to pass,
both on Linux and on Windows (`/WX`)

* Disable verbose Windows warnings
  * `_CRT_SECURE_NO_WARNINGS` because of `std::getenv`
  * C4267 because of conversions from `size_t` to other integers
* Define `WIN32_LEAN_AND_MEAN` and `NOMINMAX` on Windows
  * Gets rid of some errors and speeds up the build
* Convert integer CUDA objects to `std::uintptr_t`
  before reinterpreting them to a pointer
* Fixed "unused function" warning for `GetHipFormatPixelSize`
  * There was a lot of duplication, now a single function
    called `imageElementByteSize`
* Mark some variables as potentially unused (only used in asserts)
* Other minor fixes


source/adapters/hip/memory.hpp
```diff
@@ -502,3 +502,22 @@ struct ur_mem_handle_t_ {
     }
   }
 };
+
+constexpr size_t imageElementByteSize(hipArray_Format ArrayFormat) {
```
jchlanda, line 506: Does this need to be a fee floating function? Maybe we could hide it in the cpp file behind an anonymous namespace?

jchlanda (approved):
CUDA/HIP 👍

PietroGhg (approved):
Native CPU LGTM, thank you

test/conformance/source/environment.cpp
```diff
@@ -281,7 +281,7 @@ DevicesEnvironment::DevicesEnvironment(int argc, char **argv)
             error = "urDeviceGet() failed to get devices.";
             return;
         }
-        for (u_long i = 0; i < count; i++) {
+        for (unsigned i = 0; i < count; i++) {
```
pbalcer, line 284: any reason just a regular `int` can't be used?

source/adapters/hip/memory.hpp
ProGTX: This function is used by two separate `.cpp` files, and used to be defined 3 times (with different names) in those two files, so I figured it'd be better to have a single definition. If that's not ideal because the function is now available in the entire adapter, what would be the preffered approach? Have two separate definitions of the function in anonymous namespace, or maybe a function that's defined in one `.cpp` file and `extern`ed in another?

CMakeLists.txt
```diff
@@ -113,9 +113,6 @@ if(UR_ENABLE_TRACING)
     )
     if (MSVC)
         set(TARGET_XPTI $<IF:$<CONFIG:Release>,xpti,xptid>)
-
-        # disable warning C4267: The compiler detected a conversion from size_t to a smaller type.
-        target_compile_options(xptifw PRIVATE /wd4267)
```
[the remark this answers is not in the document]
ProGTX: C4267 affects pretty much the entire codebase, not just XPTI. Ideally I'd like to see that resolved as well, though that would require a lot of code modifications and preferably adding a similar warning flag to Clang/GCC builds (not sure which one).

test/conformance/source/environment.cpp
ProGTX: Using a signed `int` triggers warnings C4018 and `-Wsign-compare`.

source/adapters/hip/memory.hpp
jchlanda: My bad, I only checked the diffs and it looked like it was only used in `memory.cpp`. I'd vote on one def + `extern`.

test/conformance/source/environment.cpp
pbalcer: makes sense.

pbalcer (approved):
Our CI workflows use `UR_DEVELOPER_MODE=ON`, enforcing `-Werror`. Are you using a different compiler? Or are these issue related to configurations we don't build (e.g., non-L0 adapters on windows)?

If it's the latter, it might be useful to create an issue to add these missing configurations so that the build doesn't get broken again.

source/adapters/hip/memory.hpp
ProGTX: Single def + extern in latest version

ProGTX:
> Our CI workflows use `UR_DEVELOPER_MODE=ON`, enforcing `-Werror`. Are you using a different compiler? Or are these issue related to configurations we don't build (e.g., non-L0 adapters on windows)?
> 
> If it's the latter, it might be useful to create an issue to add these missing configurations so that the build doesn't get broken again.

On Linux I built this with `UR_BUILD_ADAPTER_ALL=ON`, and on Windows I built CUDA, L0, and Native CPU. Which ones would need to be added to CI then?

isaacault (approved):
Bindless images LGTM

pbalcer:
> Which ones would need to be added to CI then?

On Linux, we are missing a Native CPU build. On Windows, we are missing CUDA, HIP, OpenCL and Native CPU.

kbenzie:
> On Linux, we are missing a Native CPU build. On Windows, we are missing CUDA, HIP, OpenCL and Native CPU.

The HIP adapter is not supported on Windows currently, so doubt we need to add testing for it just now.

ProGTX:
Friendly ping for @oneapi-src/unified-runtime-level-zero-write

nrspruit (approved):
Good for the level zero side.

PR #1208, opened 2023-12-21
[hip] Fix HSA headers lookup

ldrumm:
ROCm installations prior to v6 don't respect the traditional installation layout and install the HSA headers to `$PREFIX/hsa/include/hsa` whereas in rocm6 it looks like they're putting it in the right place at `$PREFIX/include/hsa`.

That cleanup from AMD is good news, but it means that our workarounds now break and we need to check both places.

### DIFF
--- a/source/adapters/hip/CMakeLists.txt
+++ b/source/adapters/hip/CMakeLists.txt
@@ -13,7 +13,8 @@ set(UR_HIP_ROCM_DIR "/opt/rocm" CACHE STRING "ROCm installation dir")
 
 set(UR_HIP_INCLUDE_DIR "${UR_HIP_ROCM_DIR}/include")
 
-set(UR_HIP_HSA_INCLUDE_DIR "${UR_HIP_ROCM_DIR}/hsa/include")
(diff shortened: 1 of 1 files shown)


source/adapters/hip/CMakeLists.txt
```diff
@@ -31,9 +32,17 @@ if("${UR_HIP_PLATFORM}" STREQUAL "AMD")
...
-                        " please check ROCm installation.")
+    foreach(D IN LISTS UR_HIP_HSA_INCLUDE_DIRS)
+        message(STATUS ${D})
+        if(EXISTS "${D}")
+            set(UR_HIP_HSA_INCLUDE_DIR "${D}")
+            break()
+        endif()
+    endforeach()
+    if(NOT UR_HIP_HSA_INCLUDE_DIR)
+        message(FATAL_ERROR "${UR_HIP_HSA_INCLUDE_DIR} Couldn't find the HSA include directory at in any of
```
ldrumm, line 43: Long line

GeorgeWeb (approved):
Looks good.

jchlanda (approved):

PR #1211, opened 2023-12-27
Update scripts and loader template to allow object handles in structs.

aarongreig:
Also document limitations of this implementation in CONTRIB.rst

### DIFF
--- a/scripts/core/CONTRIB.rst
+++ b/scripts/core/CONTRIB.rst
@@ -226,6 +226,8 @@ following:
   ``[optional]`` tag if that struct (or any of its members, recursively) has
   an object handle member in its definition.
 
+* A struct member which is an object handle must not have the ``[out]`` tag.
+
 Forks and Pull Requests
 =======================
 
--- a/scripts/core/CONTRIB.rst
+++ b/scripts/core/CONTRIB.rst
@@ -214,8 +214,9 @@ they are preventing you from implementing a feature please open an issue and we
 will be happy to try and accommodate your use case. Otherwise beware of the
 following:
 
-* A function parameter which is a struct type that has any of the following
-  members in its type definition must not have the ``[range]`` tag:
+* A function parameter or struct member which is a struct type that has any of
+  the following members in its type definition must not have the ``[range]``
+  tag:
 
   * An object handle with the ``[range]`` tag
 
--- a/scripts/core/CONTRIB.rst
+++ b/scripts/core/CONTRIB.rst
@@ -205,6 +205,26 @@ arguments:
 *   Pointer to handle arguments, such as out parameters, are prefixed with
     ``ph`` i.e. ``phQueue``.
 
+Limitations
+-----------
+
+There are some limitations on the patterns our spec generator can handle. These
+limitations are due to convenience of implementation rather than design: if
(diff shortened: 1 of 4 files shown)


fabiomestre (approved):
The loader magic was a bit hard to understand but I think this PR does a pretty good job at documenting what it is trying to do. 

LGTM

aarongreig:
was digging around in old PRs for unrelated reasons and found this https://github.com/oneapi-src/unified-runtime/pull/302#issuecomment-1446986752 not sure if it (or anything else in that thread, there are a few things brought up) is still a concern

kbenzie:
> was digging around in old PRs for unrelated reasons and found this [#302 (comment)](https://github.com/oneapi-src/unified-runtime/pull/302#issuecomment-1446986752) not sure if it (or anything else in that thread, there are a few things brought up) is still a concern

While the handle pointer would not be replayable in a trace it should be possible to keep track of all the handle types created during a trace and refer to those by index rather than data pointed to by that pointer in to enable replaying traces like this. I personally don't see this as a blocker to supporting handles in structs. We don't support trace/replay yet anyway.

pbalcer:
> > was digging around in old PRs for unrelated reasons and found this [#302 (comment)](https://github.com/oneapi-src/unified-runtime/pull/302#issuecomment-1446986752) not sure if it (or anything else in that thread, there are a few things brought up) is still a concern
> 
> While the handle pointer would not be replayable in a trace it should be possible to keep track of all the handle types created during a trace and refer to those by index rather than data pointed to by that pointer in to enable replaying traces like this. I personally don't see this as a blocker to supporting handles in structs. We don't support trace/replay yet anyway.

We sort-of support replay in the fuzz tests ;) And they do exactly what you describe with using an index into an array. Anyway, I agree, not implementing a useful feature because of some dubious future requirements rarely makes sense.

I'll review the patch later today.

kbenzie:
> We sort-of support replay in the fuzz tests ;)

TIL!

pbalcer (approved):
This is hard to follow without an example output, but, overall, the scripts look ok. Would it be possible to rebase #1089 on top of this so we can see how it works?

One undocumented limitation is that this only converts input loader handles to adapter handles. If there's an output handle in a struct, it will not be converted to a loader handle.

aarongreig:
I rebased #1089 onto this branch, ran `generate` and pushed the results here https://github.com/aarongreig/unified-runtime/commit/457c11fc9d6e73f6bcecb7d62a8d0da8e48f17bf

didn't think about output handles, I'll document that limitation and see about adding a check for it to parse_specs

EwanC:
> I rebased #1089 onto this branch, ran `generate` and pushed the results here [aarongreig@457c11f](https://github.com/aarongreig/unified-runtime/commit/457c11fc9d6e73f6bcecb7d62a8d0da8e48f17bf)
> 
> didn't think about output handles, I'll document that limitation and see about adding a check for it to parse_specs

I've used the diff from [aarongreig@457c11f](https://github.com/aarongreig/unified-runtime/commit/457c11fc9d6e73f6bcecb7d62a8d0da8e48f17bf) in https://github.com/oneapi-src/unified-runtime/pull/1089 to make sure that it works.

PR #1213, opened 2023-12-29
[CUDA] Implement urMemImageGetInfo

fabiomestre:
Adds support for the `urMemImageGetInfo` entrypoint. 

To be able to match the new style for aborts and error codes used in https://github.com/oneapi-src/unified-runtime/pull/1127, I also cherry-picked the `UR_RETURN_ON_FAILURE` macro from that PR.

intel/llvm: https://github.com/intel/llvm/pull/12287


source/adapters/cuda/memory.cpp
```diff
@@ -398,11 +398,109 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemImageCreate(
...
+      default:
+        return UR_RESULT_ERROR_UNSUPPORTED_IMAGE_FORMAT;
+      }
+      return UR_RESULT_SUCCESS;
+    };
+
+    switch (propName) {
+    case UR_IMAGE_INFO_FORMAT:
+      ur_image_channel_type_t ChannelType;
+      UR_RETURN_ON_FAILURE(cuda2urFormat(ArrayInfo.Format, &ChannelType));
```
aarongreig, line 478: is there a benefit to switching to this instead of continuing to use UR_CHECK_ERROR?

fabiomestre: UR_CHECK_ERROR uses aborts which we are trying to eliminate. The addition of this macro is a cherry-pick from https://github.com/oneapi-src/unified-runtime/pull/1127. Essentially I'm just trying to make @martygrant life a bit easier so that he doesn't have to redesign this entrypoint on his PR.

aarongreig: sorry I mean like within this entry point, we're already using UR_CHECK_ERROR above and since we're set up to catch the exception I don't think it can abort unless you set an env var specifically telling it to

fabiomestre: You are correct that we are not going to abort() in here. I forgot that an env var needs to be setup for that.

I'm honestly not sure of what is the correct style for errors. Where is it acceptable to throw (UR_CHECK_ERROR) and when should we return an error code (UR_RETURN_ON_FAILURE). 

The style I'm using here is the same as in https://github.com/oneapi-src/unified-runtime/pull/1127 (the hip implementation of this entrypoint is very similar).

Probably there should be only one macro. Any opinions about this @martygrant ?

GeorgeWeb: Same comment as @aarongreig.

Also I am not too sure if there is general style we follow (maybe threre should be) but aim to throw where the error is a legitimate failure, usually on most driver API calls unless it's just a hint where we ignore the error or treat it as a warning. I think in this specific case we really just want to return the error/result code if `cuda2urFormat` is not successful (invalid convertion) which is what you have done, but since we are already in a try-catch block that handles the thrown exception it is fine to throw via `UR_CHECK_ERROR`.

martygrant: Sorry @fabiomestre I didn't see this before. I do know that it's an objective for us to remove all abort/die calls. Unless I'm mistaken `UR_CHECK_ERROR` doesn't actually throw anything? I think we do probably need to have a discussion about correct error handling.

fabiomestre: Sorry for the delay in addressing these comments.

UR_CHECK_ERROR does throw. But I guess the pattern in the CUDA adapter is to do a try/catch in every entrypoint to catch those throws. I'm not a big fan of that pattern but that's not something that should be addressed in this PR.

I have updated the PR to use only UR_CHECK_ERROR. Would appreciate if you could review it again @aarongreig / @GeorgeWeb

aarongreig (approved):

GeorgeWeb (approved):
adapter LGTM!

PR #1212, opened 2023-12-29
Batch adapter changes: spec motivated refactors

aarongreig:
This is a batch PR containing a number of spec-motivated refactors/small changes. Changes include:
* #1118
* #1082
* #1125
* #1145
* #1215 
* #1136

llvm testing https://github.com/intel/llvm/pull/12681

PR #1215, opened 2024-01-02
A few small changes to move towards empty match files for CL

aarongreig:
* Return UNSUPPORTED_ENUMERATION in device info query when FP16
  extension isn't supported
* Add a mapping from CL_INVALID_ARG_SIZE to
  UR_RESULT_ERROR_INVALID_KERNEL_ARGUMENT_SIZE
* Remove a bunch of {{OPT}} entries from the enqueue match file, these
  should all be passing since bounds checking was added to the
  validation layer
* Write in a special case for PROGRAM_INFO_BINARIES test - this query
  works pretty differently to the others

LLVM testing https://github.com/intel/llvm/pull/12558


kbenzie (approved):

callumfare (approved):

test/conformance/program/urProgramGetInfo.cpp
```diff
@@ -25,13 +25,31 @@ UUR_TEST_SUITE_P(
...
+                                        0, nullptr, &binary_sizes_len));
+        // Due to how the fixtures + env are set up we should only have one
+        // device associated with program, so one binary.
+        ASSERT_EQ(binary_sizes_len / sizeof(size_t), 1);
+        size_t binary_sizes[1] = {binary_sizes_len};
+        ASSERT_SUCCESS(urProgramGetInfo(program, UR_PROGRAM_INFO_BINARY_SIZES,
+                                        binary_sizes_len, binary_sizes,
+                                        nullptr));
+        std::vector<char> binary(binary_sizes[0]);
+        char *binaries[1] = {binary.data()};
```
omarahmed1111, line 41: I think we could check that the pointer returned is the same as the binary pointer that we initialized the program with

omarahmed1111 (approved):

test/conformance/program/urProgramGetInfo.cpp
aarongreig: This is for querying out the device binary, I'll add something like that for `GET_IL` in https://github.com/oneapi-src/unified-runtime/pull/1196. I started adding a bit to this test where we feed the binary that was returned back into `urProgramCreateWithBinary` and then query the binary out of the new program and verify it against the binary we passed in, but then it occurred to me that `GET_PROGRAM_BINARIES` is very implementation dependant. The opencl equivalent of the query says the binary you get out is completely implementation defined, so as far as we know a given CL implementation might create a program with a binary that contains slightly different bits if it was created with `CreateWithBinary` and then built with `ProgramBuild`. There's no guarantee of consistency.

omarahmed1111: That make sense. Thanks for clarification.

aarongreig:
merged into https://github.com/oneapi-src/unified-runtime/pull/1212

PR #1216, opened 2024-01-02
[UR] Remove UMF sources and use standalone UMF repo instead

igchor:


igchor:
Right now UMF depends on libnuma (because of OS memory provider). 

@pbalcer @vinser52  Are we OK adding libnuma as a dependency for UR? (I guess it's not ideal). If not, should we make OS provider compilation optional for now?

bratpiorka (approved):

ldorau (approved):

ldrumm:
Do you have a rationale for such a change? Splitting repos can make synchronization quite hard despite some marginal maintenance-adjacent benefits. I'm wary about such splits while it still requires 3 merge requests between UR and intel/llvm before a patch is truly landed. Until that situation is sorted, I fear a change like this will compound this maintenance complexity.

Thoughts?

bratpiorka:
> Do you have a rationale for such a change? Splitting repos can make synchronization quite hard despite some marginal maintenance-adjacent benefits. I'm wary about such splits while it still requires 3 merge requests between UR and intel/llvm before a patch is truly landed. Until that situation is sorted, I fear a change like this will compound this maintenance complexity.
> 
> Thoughts?

UMF has become a separate project with its own repository. It is currently under active development and is intended to be used by various oneAPI components, not just UR. All further enhancements and fixes will only go into the UMF repository - we do not plan to merge such changes here in the UR repository and into the UMF repository separately, so downloading UMF from its own repository makes it easier to manage.

igchor:
@kbenzie could take a look? 

Also, I have one question: **Would it be OK if UMF (and hence UR) had a dependency on hwloc?** 

Hwloc will be needed to enable 'OS memory provider' that is currently disabled and for other features like topology discovery that will be used by other oneAPI components.

ldrumm:
> UMF has become a separate project with its own repository. It is currently under active development and is intended to be used by various oneAPI components, not just UR. All further enhancements and fixes will only go into the UMF repository - we do not plan to merge such changes here in the UR repository and into the UMF repository separately, so downloading UMF from its own repository makes it easier to manage.

Thanks. That should go in the body of the commit message. Otherwise, I defer to kbenzie's expert judgment

igchor:
> > UMF has become a separate project with its own repository. It is currently under active development and is intended to be used by various oneAPI components, not just UR. All further enhancements and fixes will only go into the UMF repository - we do not plan to merge such changes here in the UR repository and into the UMF repository separately, so downloading UMF from its own repository makes it easier to manage.
> 
> Thanks. That should go in the body of the commit message. Otherwise, I defer to kbenzie's expert judgment

Of course, done.

igchor:
> > @kbenzie could you please take a look?
> 
> As for UMF being a separate project, I think this change makes sense. 👍
> 
> > Also, I have one question: **Would it be OK if UMF (and hence UR) had a dependency on hwloc?**
> > Hwloc will be needed to enable 'OS memory provider' that is currently disabled and for other features like topology discovery that will be used by other oneAPI components.
> 
> Would the dependency be dynamic? In principal I don't have an issue with the dependency but I'd like to avoid a situation where a user doesn't have hwloc available and that causes SYCL programs to crash.

Ideally hwloc would be a static dependency. As far as I know hwloc is already distributed inside oneapi packages so there should not be any issues with distribution.

kbenzie:
> Ideally hwloc would be a static dependency. As far as I know hwloc is already distributed inside oneapi packages so there should not be any issues with distribution. (But in any way, I do not add any extra dependency in this PR).

Oh, if its statically linked users shouldn't even notice, right? In that case I don't see any issues adding the dependency.

igchor:
> > Ideally hwloc would be a static dependency. As far as I know hwloc is already distributed inside oneapi packages so there should not be any issues with distribution. (But in any way, I do not add any extra dependency in this PR).
> 
> Oh, if its statically linked users shouldn't even notice, right? In that case I don't see any issues adding the dependency.

Actually, sorry, I'm not yet sure if we'll link statically or dynamically. I will take a look at how other oneAPI components are using it and I'll specify that when I create a PR with adding the dependency. We can discuss it then and change the approach if necessary.

igchor:
@kbenzie are you OK merging this PR as is, or do you need more time to review it?

kbenzie:
> @kbenzie are you OK merging this PR as is, or do you need more time to review it?

We'll need approvals from all the touched 4 adapters (@oneapi-src/unified-runtime-opencl-write, @oneapi-src/unified-runtime-hip-write, @oneapi-src/unified-runtime-cuda-write, @oneapi-src/unified-runtime-level-zero-write) as well as shared code (@oneapi-src/unified-runtime-maintain).

I'll take a look for the OpenCL adapter and shared code but the other adapter teams will need to approve also.

igchor:
@ldrumm could you please take a look at HIP and CUDA changes?
@nrspruit could you please take a look at L0 changes?

Changes in the adapters are pretty minor - I only change the way the pool is being created (so that we now use UMF_DISJOINT_POOL_OPS from UMF repo)

nrspruit (approved):
LGTM for level-zero

source/common/CMakeLists.txt
```diff
@@ -19,9 +17,31 @@ target_include_directories(ur_common PUBLIC
     ${CMAKE_SOURCE_DIR}/include
 )
 
+message(STATUS "Download Unified Memory Framework from github.com")
```
ldrumm, line 20: in oneapi we have a switch to enable building against a local checkout of unified runtime. Should we have the same thing here?

c.f [intel/llvm/139b13172](https://github.com/intel/llvm/commit/139b1317278fc777aa36f71096d6639fe6fd3ff0)

ldrumm (approved):
HIP and CUDA look fine

source/common/CMakeLists.txt
igchor: I don't think there will be many changes required in the UR when bumping UMF version in the future. 

However, we've had some discussion about potentially moving some of the memory provider logic for L0 and CUDA to UMF - if we decide to do that, then adding such a switch might a good idea.

lukaszstolarczuk (commented):
👍

igchor:
FYI, I've create a testing PR to llvm (with this patch rebased on main from a few days ago): https://github.com/intel/llvm/pull/12570

igchor:
@kbenzie do you have an estimate on when you will merge it? is there any extra testing you would like me to do?

kbenzie:
> @kbenzie do you have an estimate on when you will merge it? is there any extra testing you would like me to do?

I was on holiday last week. No additionally testing necessary, we'll aim to merge it this week.

igchor:
> @igchor I rebased this to test against the latest main changes before merging, seems like a fix in required for the native cpu cmake, hopefull its a straight forward fix?
> 
> ```
> CMake Error at cmake/helpers.cmake:131 (add_library):
>   Target "ur_adapter_native_cpu" links to target
>   "unified-runtime::unified_malloc_framework" but the target was not found.
>   Perhaps a find_package() call is missing for an IMPORTED target, or an
>   ALIAS target is missing?
> Call Stack (most recent call first):
>   source/adapters/CMakeLists.txt:7 (add_ur_library)
>   source/adapters/native_cpu/CMakeLists.txt:10 (add_ur_adapter)
> ```

Yes, I just had to remove reference to `unified-runtime::unified_malloc_framework` from native CPU CMakeLists.txt - it's enough to just link with `ur_common` now.

PR #1217, opened 2024-01-03
[HIP] Cache some of the HIP driver calls from kernel enqueue

konradkusiak97:
Those changes ought to mirror the CUDA ones from [[SYCL][CUDA] Cached info from CUDA API calls to the kernel and device classes](https://github.com/intel/llvm/pull/10104). It turns out, there are fewer HIP driver calls to cache than in the previous case


source/adapters/hip/device.hpp
```diff
@@ -35,6 +37,13 @@ struct ur_device_handle_t_ {
 
   ~ur_device_handle_t_() noexcept(false) {
```
jchlanda, line 38: Are you sure you want to obtain those values at destruction time, calling it from the constructor would appear to be more appropriate?

konradkusiak97: Indeed!

jchlanda (approved):

PR #1220, opened 2024-01-03
[CUDA] Add support for multiple active mappings

fabiomestre:
The CUDA adapter currently does not support multiple active mappings on the same buffer when using urEnqueueMemBufferMap. This change aims to add support for that feature.

However, the unsupported scenario of overlapping maps that were both created with the `UR_MAP_FLAG_WRITE` is not currently being checked. That's better done using validation layers for which I have created an issue:  https://github.com/oneapi-src/unified-runtime/issues/1221

intel/llvm: https://github.com/intel/llvm/pull/12285


jchlanda (approved):
As a general comment, a wee explanation of what the PR does wouldn't go amiss.

fabiomestre:
@oneapi-src/unified-runtime-maintain would be great if someone could review these changes

aarongreig (approved):

PR #1226, opened 2024-01-04
[UR] Add extra param to urMemGetNativeHandle

hdelan:
Add an extra param to `urMemGetNativeHandle` as suggested here https://github.com/oneapi-src/unified-runtime/pull/1199#discussion_r1440464125

### DIFF
--- a/scripts/core/memory.yml
+++ b/scripts/core/memory.yml
@@ -432,6 +432,10 @@ params:
       name: hMem
       desc: |
             [in] handle of the mem.
+    - type: $x_device_handle_t
+      name: hDevice
+      desc: |
+            [in] handle of the device that the native handle will be resident on.
     - type: $x_native_handle_t*
       name: phNativeMem
       desc: |
--- a/source/adapters/hip/memory.cpp
+++ b/source/adapters/hip/memory.cpp
@@ -279,15 +279,34 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
 /// \param[out] phNativeMem Set to the native handle of the UR mem object.
 ///
 /// \return UR_RESULT_SUCCESS
-UR_APIEXPORT ur_result_t UR_APICALL urMemGetNativeHandle(ur_mem_handle_t,
-                                                         ur_native_handle_t *) {
-  // FIXME: there is no good way of doing this with a multi device context.
-  // If we return a single pointer, how would we know which device's allocation
-  // it should be?
-  // If we return a vector of pointers, this is OK for read only access but if
-  // we write to a buffer, how would we know which one had been written to?
-  // Should unused allocations be updated afterwards? We have no way of knowing
-  // any of these things in the current API design.
+UR_APIEXPORT ur_result_t UR_APICALL
+urMemGetNativeHandle(ur_mem_handle_t hMem, ur_device_handle_t Device,
+                     ur_native_handle_t *phNativeMem) {
+#if defined(__HIP_PLATFORM_NVIDIA__)
+  if (sizeof(BufferMem::native_type) > sizeof(ur_native_handle_t)) {
+    // Check that all the upper bits that cannot be represented by
+    // ur_native_handle_t are empty.
+    // NOTE: The following shift might trigger a warning, but the check in the
+    // if above makes sure that this does not underflow.
+    BufferMem::native_type UpperBits =
+        std::get<BufferMem>(hMem->Mem).getPtr(Device) >>
+        (sizeof(ur_native_handle_t) * CHAR_BIT);
+    if (UpperBits) {
+      // Return an error if any of the remaining bits is non-zero.
+      return UR_RESULT_ERROR_INVALID_MEM_OBJECT;
+    }
+  }
+  *phNativeMem = reinterpret_cast<ur_native_handle_t>(
+      std::get<BufferMem>(hMem->Mem).getPtr(hDevice));
+#elif defined(__HIP_PLATFORM_AMD__)
+  *phNativeMem = reinterpret_cast<ur_native_handle_t>(
+      std::get<BufferMem>(hMem->Mem).getPtr(hDevice));
+#else
+#error("Must define exactly one of __HIP_PLATFORM_AMD__ or __HIP_PLATFORM_NVIDIA__");
+#endif
+  return UR_RESULT_SUCCESS;
+}
+
   return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
 }
 
--- a/source/adapters/hip/memory.cpp
+++ b/source/adapters/hip/memory.cpp
@@ -297,17 +297,14 @@ urMemGetNativeHandle(ur_mem_handle_t hMem, ur_device_handle_t Device,
     }
   }
   *phNativeMem = reinterpret_cast<ur_native_handle_t>(
-      std::get<BufferMem>(hMem->Mem).getPtr(hDevice));
+      std::get<BufferMem>(hMem->Mem).getPtr(Device));
 #elif defined(__HIP_PLATFORM_AMD__)
   *phNativeMem = reinterpret_cast<ur_native_handle_t>(
-      std::get<BufferMem>(hMem->Mem).getPtr(hDevice));
+      std::get<BufferMem>(hMem->Mem).getPtr(Device));
 #else
 #error("Must define exactly one of __HIP_PLATFORM_AMD__ or __HIP_PLATFORM_NVIDIA__");
 #endif
   return UR_RESULT_SUCCESS;
-}
-
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL urMemBufferCreateWithNativeHandle(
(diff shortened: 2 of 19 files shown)


hdelan:
Gentle ping @GeorgeWeb @oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-level-zero-write @oneapi-src/unified-runtime-opencl-write 

Would be great to get some reviews

muhammad-tanvir-1211:
This PR is blocking the following two PRs on oneMKL as all the buffer tests (with `accessor interop`) are failing for the `rocblas` backend with `PI_ERROR_INVALID_OPERATION`.
https://github.com/oneapi-src/oneMKL/pull/428
https://github.com/oneapi-src/oneMKL/pull/425

Logs:
[PR_425.txt](https://github.com/oneapi-src/unified-runtime/files/13918171/PR_425.txt)
[PR_428.txt](https://github.com/oneapi-src/unified-runtime/files/13918196/PR_428.txt)

source/adapters/hip/memory.cpp
```diff
@@ -282,16 +282,34 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
...
+    // Check that all the upper bits that cannot be represented by
+    // ur_native_handle_t are empty.
+    // NOTE: The following shift might trigger a warning, but the check in the
+    // if above makes sure that this does not underflow.
+    BufferMem::native_type UpperBits =
+        std::get<BufferMem>(hMem->Mem).getPtr(Device) >>
+        (sizeof(ur_native_handle_t) * CHAR_BIT);
+    if (UpperBits) {
+      // Return an error if any of the remaining bits is non-zero.
+      return UR_RESULT_ERROR_INVALID_MEM_OBJECT;
```
GeorgeWeb, line 301: It may be good to extend the UR API declaration's documentation of this API with this return code (`UR_RESULT_ERROR_INVALID_MEM_OBJECT`) as it was not part of the original version. However, since it is `__HIP_PLATFORM_NVIDIA__`-specific it may or may not be useful readers if we chop the HIP on NVIDIA support completely but it is still a valid case today.

I am not adding this as a blocker to the PR because it is so niche due to not being related to AMD hip target devices but I think because we currently still care to support this one case we ought to document all possible result codes accordingly.

scripts/core/memory.yml
```diff
@@ -432,6 +432,10 @@ params:
       name: hMem
       desc: |
             [in] handle of the mem.
+    - type: $x_device_handle_t
+      name: hDevice
+      desc: |
+            [in] handle of the device that the native handle will be resident on.
```
ldrumm, line 438:
```suggestion
            [in] handle of the device on which the native handle will reside.
```

scripts/core/HIP.rst
```diff
@@ -91,6 +91,22 @@ take the extra global offset argument. Use of the global offset is not
...
+for UR objects. The native handle may refer to a number of different native HIP
+types, depending on the entry point called.
+
+- ${x}MemGetNativeHandle stores a ``hipDeviceptr_t`` in the
+  ${x}_native_handle_t argument passed. The ``hipDeviceptr_t`` will be resident
+  on the ${x}_device_handle_t provided as an argument, if the device shares a
+  context with the ${x}_mem_handle_t. If the ${x}_device_handle_t argument
+  passed to ${x}MemGetNativeHandle is ``nullptr``,
+  ``UR_RESULT_ERROR_UNSUPPORTED_FEATURE`` will be returned.
+- TODO: add more stuff
```
ldrumm, line 107: What "more stuff"?

scripts/core/CUDA.rst
```diff
@@ -148,6 +148,21 @@ take the extra global offset argument. Use of the global offset is not
...
+for UR objects. The native handle may refer to a number of different native CUDA
+types, depending on the entry point called.
+
+- ${x}MemGetNativeHandle stores a ``CUdeviceptr`` in the ${x}_native_handle_t
+  argument passed. The ``CUdeviceptr`` will be resident on the
+  ${x}_device_handle_t provided as an argument, if the device shares a context
+  with the ${x}_mem_handle_t. If the ${x}_device_handle_t argument passed to
+  ${x}MemGetNativeHandle is ``nullptr``,
+  ``UR_RESULT_ERROR_UNSUPPORTED_FEATURE`` will be returned.
+- TODO: add more stuff
```
ldrumm, line 164: What "more stuff"?

GeorgeWeb (approved):
HIP and CUDA adapters changes look good to me.

See my inline comment regarding the new return code in the hip-nvidia path of the entry point, which is not a blocker, but I'd like to see your view on it.

source/adapters/hip/memory.cpp
```diff
@@ -282,16 +282,34 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemGetInfo(ur_mem_handle_t hMemory,
...
+UR_APIEXPORT ur_result_t UR_APICALL
+urMemGetNativeHandle(ur_mem_handle_t hMem, ur_device_handle_t Device,
+                     ur_native_handle_t *phNativeMem) {
+  if (!Device)
+    return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+#if defined(__HIP_PLATFORM_NVIDIA__)
+  if (sizeof(BufferMem::native_type) > sizeof(ur_native_handle_t)) {
+    // Check that all the upper bits that cannot be represented by
+    // ur_native_handle_t are empty.
+    // NOTE: The following shift might trigger a warning, but the check in the
```
ldrumm, line 294: If you want to do bitwise arithmetic on a pointer, and not have the compiler - or language lawyers - complain, you need to convert it first to a suitable type, such as `uintptr_t`.

ldrumm: Also, can you document why we need to check these bits? Is this something to do with how HIP encodes device info? If so: is it documented by NVIDIA?

ldrumm (changes requested):
Your commit messages lack any info into what parameter is added and why. Please include context so future you can actually understand motivation

scripts/core/CUDA.rst
aarongreig, line 164: could you make this (and [this](https://github.com/oneapi-src/unified-runtime/pull/1226/files#diff-91c197bfc7627108e5d24b5e37090bf16106ecfe1348119fe0712a15844f1f16R107)) a comment so it doesn't render, and also elaborate a bit on stuff

source/adapters/hip/memory.cpp
hdelan: Yeah I'm not sure about this. At the moment there isn't a precedent for explaining return codes specific to the HIP and CUDA adapter, much less for the HIP Nvidia adapter, so I would be inclined to let this slide.

Also bear in mind that this code is merely copy pasted from what was there before the multi dev ctx PR https://github.com/oneapi-src/unified-runtime/pull/999/files#diff-cdda61bf8ead17f141255a7bceb4231d050653676e565fde5aa6449a30a428b2L321

scripts/core/HIP.rst
hdelan: More `blahGetNativeHandle` stuff. So that we know what is being stored in the `ur_native_handle`. Ie when we do `urDeviceGetNativeHandle` a `CUdevice` is stored in the `ur_native_handle`

source/adapters/hip/memory.cpp
GeorgeWeb: Yeah, it's why I mentioned I wasn't too sure either. I'll leave it to your best judgement then. Thanks!

source/adapters/hip/memory.cpp
hdelan: This was pre-existing code from before the multi device context PR so I really don't know why we needed to do this.

source/adapters/hip/memory.cpp
hdelan: Thanks @GeorgeWeb

scripts/core/HIP.rst
hdelan: Have elaborated in comment

PietroGhg (approved):
Native CPU lgtm

scripts/core/CUDA.rst
hdelan: Have elaborated in comment

hdelan: Change made thanks

hdelan:
> Your commit messages lack any info into what parameter is added and why. Please include context so future you can actually understand motivation

Have changed the commit message

source/adapters/hip/memory.cpp
hdelan: Aha https://github.com/intel/llvm/pull/6207

ldrumm: Oh right. It's a little suspect, so we can track down the meaning and update in another patch that'd be good.

ldrumm:
> Your commit messages lack any info into what parameter is added and why. Please include context so future you can actually understand motivation

Not any more!

Thanks and LGTM

hdelan:
Friendly ping @oneapi-src/unified-runtime-level-zero-write

nrspruit (approved):
LGTM for level-zero

kbenzie (approved):

PR #1229, opened 2024-01-05
Revert "Revert "[HIP] Implement ext_oneapi_queue_priority""

kbenzie:
Reverts oneapi-src/unified-runtime#1228

### DIFF
--- a/source/adapters/hip/queue.cpp
+++ b/source/adapters/hip/queue.cpp
@@ -38,8 +38,8 @@ hipStream_t ur_queue_handle_t_::getNextComputeStream(uint32_t *StreamToken) {
       // The second check is done after mutex is locked so other threads can not
       // change NumComputeStreams after that
       if (NumComputeStreams < ComputeStreams.size()) {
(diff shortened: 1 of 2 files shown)


source/adapters/hip/queue.cpp
```diff
@@ -118,6 +118,19 @@ urQueueCreate(ur_context_handle_t hContext, ur_device_handle_t hDevice,
     std::unique_ptr<ur_queue_handle_t_> QueueImpl{nullptr};
 
     unsigned int Flags = 0;
+    ur_queue_flags_t URFlags = 0;
+    int Priority = 0; // Not guaranteed, but, in ROCm 5.7, 0 is the default
```
al42and, line 122:
```suggestion
    int Priority = 0; // Not guaranteed, but, in ROCm 5.0-6.0, 0 is the default
```

al42and (changes requested):

kbenzie:
Thanks @al42and

PR #1230, opened 2024-01-08
[OPENCL] Retain context for urMemBufferCreate on opencl

omarahmed1111:
Testing [PR](https://github.com/intel/llvm/pull/12321)

### DIFF
--- a/source/adapters/opencl/memory.cpp
+++ b/source/adapters/opencl/memory.cpp
@@ -264,6 +264,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemBufferCreate(
       *phBuffer = reinterpret_cast<ur_mem_handle_t>(FuncPtr(
           CLContext, PropertiesIntel.data(), static_cast<cl_mem_flags>(flags),
           size, pProperties->pHost, cl_adapter::cast<cl_int *>(&RetErr)));
+      if (RetErr == CL_SUCCESS) {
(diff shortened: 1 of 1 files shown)


source/adapters/opencl/memory.cpp
```diff
@@ -411,6 +412,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemRetain(ur_mem_handle_t hMem) {
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL urMemRelease(ur_mem_handle_t hMem) {
+  cl_context CLContext;
+  CL_RETURN_ON_FAILURE(clGetMemObjectInfo(cl_adapter::cast<cl_mem>(hMem), CL_MEM_CONTEXT,
+                         sizeof(CLContext), &CLContext, nullptr));
+  CL_RETURN_ON_FAILURE(clReleaseContext(CLContext));
```
kbenzie, line 418: We should be consistent about using either the UR or CL entry points, at the moment above used `urContextRetain` but this uses `clReleaseContext`.

omarahmed1111: I made it using cl api everywhere as that avoid some casts and would be more consistent across the entry-point implementation.

omarahmed1111:
After adding reference counting to UR layer in OpenCL adapter in this [PR](https://github.com/oneapi-src/unified-runtime/pull/1176), this should be resolved in a cleaner way. So, closing this for now.

PR #1233, opened 2024-01-08
[CTS] Add more USM testing

callumfare:
* Expand USMMemcpy2D testing to test all USM allocation types
* Add USM linked list kernel execution test

The linked list test is based on USM/smemll.cpp from the SYCL-E2E tests and checks that USM allocations are actually usable from a kernel.

### DIFF
--- a/test/conformance/enqueue/urEnqueueKernelLaunch.cpp
+++ b/test/conformance/enqueue/urEnqueueKernelLaunch.cpp
@@ -263,3 +263,102 @@ TEST_P(urEnqueueKernelLaunchMultiDeviceTest, KernelLaunchReadDifferentQueues) {
         ASSERT_EQ(val, output) << "Result on queue " << i << " did not match!";
     }
 }
+
+struct urEnqueueKernelLaunchUSMLinkedList
+    : uur::urKernelTestWithParam<uur::BoolTestParam> {
+    struct Node {
+        Node() : next(nullptr), num(0xDEADBEEF) {}
+
+        Node *next;
+        uint32_t num;
+    };
+
+    void SetUp() override {
+        program_name = "usm_ll";
+        UUR_RETURN_ON_FATAL_FAILURE(
+            uur::urKernelTestWithParam<uur::BoolTestParam>::SetUp());
+
+        use_pool = getParam().value;
+        ASSERT_SUCCESS(urQueueCreate(context, device, 0, &queue));
+        ur_usm_pool_desc_t pool_desc{UR_STRUCTURE_TYPE_USM_POOL_DESC, nullptr,
+                                     0};
+        if (use_pool) {
+            ASSERT_SUCCESS(urUSMPoolCreate(this->context, &pool_desc, &pool));
+        }
+    }
+
+    void TearDown() override {
+        auto *list_cur = list_head;
+        while (list_cur) {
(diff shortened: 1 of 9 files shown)


test/conformance/enqueue/urEnqueueKernelLaunch.cpp
```diff
@@ -262,3 +262,72 @@ TEST_P(urEnqueueKernelLaunchMultiDeviceTest, KernelLaunchReadDifferentQueues) {
...
+    size_t global_offset = 0;
+    Node *list_head = nullptr;
+    const int num_nodes = 4;
+};
+
+UUR_INSTANTIATE_DEVICE_TEST_SUITE_P(urEnqueueKernelLaunchUSMLinkedList);
+
+TEST_P(urEnqueueKernelLaunchUSMLinkedList, Success) {
+    // Build linked list with USM allocations
+    ASSERT_SUCCESS(urUSMSharedAlloc(context, device, nullptr, nullptr,
```
aarongreig, line 302: could you add a case that uses a pool with the allocations as well?

callumfare: Done now, I ended up discovering a bug in the original test which is fixed too

test/conformance/enqueue/urEnqueueUSMMemcpy2D.cpp
```diff
@@ -25,15 +23,25 @@ struct urEnqueueUSMMemcpy2DTestWithParam
...
+        const auto [in2DParams, inSrcKind, inDstKind] = getParam();
+        std::tie(pitch, width, height, src_kind, dst_kind) =
+            std::make_tuple(in2DParams.pitch, in2DParams.width,
+                            in2DParams.height, inSrcKind, inDstKind);
+
+        ur_device_usm_access_capability_flags_t device_usm = 0;
+        ASSERT_SUCCESS(uur::GetDeviceUSMDeviceSupport(device, device_usm));
+        if (!device_usm && (src_kind == uur::USMKind::Device ||
+                            dst_kind == uur::USMKind::Device)) {
+            GTEST_SKIP() << "Device USM is not supported";
```
aarongreig, line 35: we should move the memcpy2d support check from [here](https://github.com/oneapi-src/unified-runtime/pull/1233/files#diff-1b1b9f128a556f23bc1fbc8dbcabaa852e78b6d5b17cd2a76ef6994cd3be1d21R22) to after this one since failing that check implies we're going to fail this as well

aarongreig (approved):

test/conformance/testing/include/uur/fixtures.h
```diff
@@ -1071,7 +1071,8 @@ struct urProgramTest : urQueueTest {
 template <class T> struct urProgramTestWithParam : urContextTestWithParam<T> {
     void SetUp() override {
         UUR_RETURN_ON_FATAL_FAILURE(urContextTestWithParam<T>::SetUp());
-        uur::KernelsEnvironment::instance->LoadSource("foo", 0, il_binary);
+        uur::KernelsEnvironment::instance->LoadSource(program_name, 0,
+                                                      il_binary);
```
kbenzie, line 1075: We'll need to handle the HIP case here, since #1184 merged.

kbenzie (approved):

PR #1234, opened 2024-01-08
[HIP] USM-P2P extension implementation

JackAKirk:
Uses the latest non deprecated hip APIs. This enables the DPC++ hip backend to use the P2P extension. Full support is enabled.


JackAKirk:
Note that proper testing of this PR cannot be made via intel/llvm CI, since a system with two gpus is required.
However both MI250x and MI100 systems have been testing locally using the updated tests here:
https://github.com/intel/llvm/compare/sycl...JackAKirk:llvm:p2p-make-test-portable

Eventually I will open a PR to update these tests in intel/llvm. However the form of these tests depends on https://github.com/intel/llvm/pull/10737 getting merged.

cc @jinz2014

source/adapters/hip/usm_p2p.cpp
```diff
@@ -9,25 +9,59 @@
...
 }
 
-UR_APIEXPORT ur_result_t UR_APICALL
-urUsmP2PDisablePeerAccessExp(ur_device_handle_t, ur_device_handle_t) {
-  detail::ur::die(
-      "urUsmP2PDisablePeerAccessExp is not implemented for HIP adapter.");
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+UR_APIEXPORT ur_result_t UR_APICALL urUsmP2PDisablePeerAccessExp(
+    ur_device_handle_t commandDevice, ur_device_handle_t peerDevice) {
+  ur_result_t result = UR_RESULT_SUCCESS;
```
hdelan, line 28: Can you use the pattern where instead of using this `result` var you just return `err` in the `catch` and otherwise return `UR_RESULT_SUCCESS`?

JackAKirk: OK

source/adapters/hip/usm_p2p.cpp
```diff
@@ -9,25 +9,59 @@
...
-    void *pPropValue, size_t *pPropSizeRet) {
+    ur_device_handle_t commandDevice, ur_device_handle_t peerDevice,
+    ur_exp_peer_info_t propName, size_t propSize, void *pPropValue,
+    size_t *pPropSizeRet) {
   UrReturnHelper ReturnValue(propSize, pPropValue, pPropSizeRet);
-  // Zero return value indicates that all of the queries currently return false.
-  return ReturnValue(uint32_t{0});
+
+  int value;
+  hipDeviceP2PAttr hip_attr;
```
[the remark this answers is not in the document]
JackAKirk: Sure

hdelan:
Should we enable some tests in DPC++ to check this behaviour? Can we test for this in UR?

JackAKirk:
> Should we enable some tests in DPC++ to check this behaviour? Can we test for this in UR?

See https://github.com/oneapi-src/unified-runtime/pull/1234#issuecomment-1881447246

hdelan:
> See [#1234 (comment)](https://github.com/oneapi-src/unified-runtime/pull/1234#issuecomment-1881447246)

Aha sorry missed this

JackAKirk:
> > Should we enable some tests in DPC++ to check this behaviour? Can we test for this in UR?
> 
> See [#1234 (comment)](https://github.com/oneapi-src/unified-runtime/pull/1234#issuecomment-1881447246)

For UR testing it is a good question. I remember some discussion around having tests like this in UR:
https://github.com/JackAKirk/llvm/blob/4ab6215b48d80376670200cdd98216964609d530/sycl/unittests/Extensions/USMP2P.cpp

There could also in principle be e2e integration testing in UR, that for example could test integration of features, such as enabling p2p, and then performing usm copies via p2p. I do not know if tests like this exist or are desired in the future.

hdelan (approved):
LGTM

JackAKirk:
> > > Should we enable some tests in DPC++ to check this behaviour? Can we test for this in UR?
> > 
> > 
> > See [#1234 (comment)](https://github.com/oneapi-src/unified-runtime/pull/1234#issuecomment-1881447246)
> 
> For UR testing it is a good question. I remember some discussion around having (or moving) tests like this in UR: https://github.com/JackAKirk/llvm/blob/4ab6215b48d80376670200cdd98216964609d530/sycl/unittests/Extensions/USMP2P.cpp
> 
> There could also in principle be e2e integration testing in UR, that for example could test integration of features, such as enabling p2p, and then performing usm copies via p2p. I do not know if tests like this exist or are desired in the future.

I thought about it a bit more. I could add a test at this point and happy to do so if people want it. I think that it would require a system with two or more gpus in each platform for it to be executed. I think such a test would be most sensible at the "conformance" level (although my reading as atm such tests aren't written for extensions?), such that it only uses UR apis portably and tests correct e2e behavior. Doing any kind of testing with just a single gpu would have to be platform specific I think and would only be able to test that the correct native errors are returned when you try to do any p2p with only one gpu.

hdelan:
There are similar tests in DPC++ to test for buffer migration https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/Basic/buffer/buffer_migrate.cpp#L25 that only test the desired behaviour if using a multi gpu system. Something similar could be done here. Therefore the tests would always pass on a single gpu system but would test the desired behaviour by default on a multi gpu system

JackAKirk:
> There are similar tests in DPC++ to test for buffer migration https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/Basic/buffer/buffer_migrate.cpp#L25 that only test the desired behaviour if using a multi gpu system. Something similar could be done here. Therefore the tests would always pass on a single gpu system but would test the desired behaviour by default on a multi gpu system

Yeah it would be identical to https://github.com/JackAKirk/llvm/blob/4ab6215b48d80376670200cdd98216964609d530/sycl/unittests/Extensions/USMP2P.cpp
but with UR apis instead of DPC++ ones.

source/adapters/cuda/usm_p2p.cpp
```diff
@@ -13,28 +13,24 @@
...
-
-  ur_result_t result = UR_RESULT_SUCCESS;
   try {
     ScopedContext active(commandDevice->getContext());
     UR_CHECK_ERROR(cuCtxEnablePeerAccess(peerDevice->getContext(), 0));
   } catch (ur_result_t err) {
-    result = err;
+    return err;
   }
-  return result;
```
ldrumm, line 24: These changes appear unrelated. Please submit them as a separate cleanup

ldrumm (changes requested):

source/adapters/cuda/usm_p2p.cpp
hdelan: Because of the overhead involved in getting UR PRs merged I would be inclined to make an exception and keep these changes in this PR. They will still appear as a separate commit since we don't squash in UR

ldrumm: That's not a good reason. The hardest bug I ever had to solve was because of "just an extra bit of code" unrelated to the commit message.

hdelan: I defer to your judgement

JackAKirk: New PR for those format changes is here: https://github.com/oneapi-src/unified-runtime/pull/1311

JackAKirk:
> > There are similar tests in DPC++ to test for buffer migration https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/Basic/buffer/buffer_migrate.cpp#L25 that only test the desired behaviour if using a multi gpu system. Something similar could be done here. Therefore the tests would always pass on a single gpu system but would test the desired behaviour by default on a multi gpu system
> 
> Yeah it would be identical to [intel/llvm@sycl...JackAKirk:llvm:p2p-make-test-portable](https://github.com/intel/llvm/compare/sycl...JackAKirk:llvm:p2p-make-test-portable) but with UR apis instead of DPC++ ones.

Also I think (but not certain) in UR it should be easier to make a portable test for hip/cuda/l0, even when cuda doesn't have the multi-device context patch. I will have a go, and plan to update this PR with the test.

kbenzie:
> @kbenzie
> 
> Is there a plan to have tests for unified-runtime extensions conformance? atm I don't see such tests in the UR repo. I've been staring at the test folder and I think the only testing of value for this P2P extension would be a backend agnostic conformance test, but atm there isn't an existing example I can see for extensions, only core spec. If you think such a thing would be valuable I am happy to introduce such testing. However I do not want to do something new like this without consulting you first.

Experimental features are not required to have CTS tests but adding tests for them is very welcome. One of the criteria for making an experimental feature core is to have CTS tests. Please go ahead with adding P2P experimental feature tests.

PR #1241, opened 2024-01-10
[CL] Fix Windows CMake, remove explicit Linux path

### LINKED ISSUE #1171: Visual Studio OpenCL adapter build error
When building the OpenCL adatper on Windows when an OpenCL driver using Visual Studio and the Ninja  generator the build dependency on the OpenCL-ICD-Loader target appears to be using a Linux specific filename:

```
unified-runtime\out\build\x64-Debug\ninja : error : 'lib/libOpenCL.so', needed by 'bin/Debug/ur_adapter_opencl.dll', missing and no known rule to make it
```

kbenzie:
Fixes #1171 by replacing an explicit Linux path used as the value for the internal `OpenCLICDLoaderLibrary` CMake varaible and replacing it with the `OpenCL::OpenCL` target name. This is an target alias provided by the https://github.com/KhronosGroup/OpenCL-ICD-Loader.

### DIFF
--- a/source/adapters/opencl/CMakeLists.txt
+++ b/source/adapters/opencl/CMakeLists.txt
@@ -67,16 +67,14 @@ if(UR_OPENCL_ICD_LOADER_LIBRARY)
     set(OpenCLICDLoaderLibrary ${UR_OPENCL_ICD_LOADER_LIBRARY})
 else()
     find_package(OpenCL 3.0)
-    if(OpenCL_FOUND)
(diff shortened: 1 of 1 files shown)


fabiomestre (approved):

wlemkows (approved):
I hope this will help on Coverity build for OpenCL, I tested it on my fork and it works fine :)

kbenzie:
> I hope this will help on Coverity build for OpenCL, I tested it on my fork and it works fine :)

It has a similar error, right? I think it should help if so.

PR #1245, opened 2024-01-11
[L0] Only Override max allocation limits given env

nrspruit:
- Change the defaults from always allowing > 4GB allocations to making the user have to request > 4GB allocation support given the max allocation allowed on that system is less than 4GB.
- This ensures performance is maintained on systems that dont handle > 4GB allocations natively and avoids breaking Ahead of Time (AOT) binaries that were built without > 4GB resource support.
- By setting UR_L0_ENABLE_RELAXED_ALLOCATION_LIMITS=1 , the L0 Adapter will force the modules to be built with stateless or > 4GB support and will allow for the allocations to exceed the max single allocation size limit for that device.

### DIFF
--- a/source/adapters/level_zero/device.cpp
+++ b/source/adapters/level_zero/device.cpp
@@ -10,6 +10,7 @@
 
 #include "device.hpp"
 #include "ur_level_zero.hpp"
+#include "ur_util.hpp"
 #include <algorithm>
 #include <climits>
 #include <optional>
@@ -268,9 +269,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
     return ReturnValue(uint32_t{64});
   }
   case UR_DEVICE_INFO_MAX_MEM_ALLOC_SIZE:
-    // if not optimized for 32-bit access, return total memory size.
-    // otherwise, return only maximum allocatable size.
-    if (Device->useOptimized32bitAccess() == 0) {
+    // if the user wishes to allocate large allocations on a system that usually
+    // does not allow that allocation size, then we return the max global mem
+    // size as the limit.
+    if (Device->useRelaxedAllocationLimits()) {
       return ReturnValue(uint64_t{calculateGlobalMemSize(Device)});
     } else {
       return ReturnValue(uint64_t{Device->ZeDeviceProperties->maxMemAllocSize});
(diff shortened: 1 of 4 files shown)


MichalMrozek (approved):
lgtm

source/adapters/level_zero/device.hpp
```diff
@@ -160,7 +160,7 @@ struct ur_device_handle_t_ : _ur_object {
   // provide support for only one, like for Intel(R)
   // Data Center GPU Max, for which L0 driver only
   // supports stateless.
-  int32_t useOptimized32bitAccess();
+  int32_t useRelaxedAllocationLimits();
```
pbalcer, line 163: `bool`?

nrspruit: Previously we had the possible values of 0, 1, and -1, but with the change you are right the new values are 0,1 true/false so I can change that logic to be cleaner.

source/adapters/level_zero/device.cpp
```diff
@@ -923,20 +924,15 @@ ur_device_handle_t_::useImmediateCommandLists() {
...
-  static const int32_t Optimize32bitAccessMode = [this] {
-    // If device is Intel(R) Data Center GPU Max,
-    // use default provided by L0 driver.
-    // TODO: Use IP versioning to select based on range of devices
-    if (this->isPVC())
-      return -1;
-    const char *UrRet = std::getenv("UR_L0_USE_OPTIMIZED_32BIT_ACCESS");
+int32_t ur_device_handle_t_::useRelaxedAllocationLimits() {
+  static const int32_t EnableRelaxedAllocationLimits = [] {
+    const char *UrRet = std::getenv("UR_L0_ENABLE_RELAXED_ALLOCATION_LIMITS");
```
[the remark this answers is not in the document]
nrspruit: Fixed in the latest patch.

source/adapters/level_zero/device.hpp
nrspruit: Updated to use bool instead of the int32_t and updated the checks to simplify to true/false.

pbalcer (approved):

source/adapters/level_zero/device.cpp
```diff
@@ -13,6 +13,7 @@
 #include <algorithm>
 #include <climits>
 #include <optional>
+#include "ur_util.hpp"
```
nrspruit, line 16: Need to add the util header otherwise this build fails in intel/llvm.

nrspruit:
Pre-merge checks have passed here: https://github.com/intel/llvm/pull/12375 This change should be ready to be merged to main, then cherry-picked to 0.8.x release. @kbenzie , please merge this when able.

PR #1246, opened 2024-01-12
[UR] Add default implementation for cooperative kernel functions

0x12CC:
Cooperative kernels can synchronize using device-scope barriers. These kernels use `urKernelSuggestMaxCooperativeGroupCountExp` to ensure that all work groups can run concurrently. When the maximum number of work groups is 1, these kernels behave the same as regular kernels.

This PR adds a default implementation for `urKernelSuggestMaxCooperativeGroupCountExp` that returns 1. Also, it adds a default implementation for `urEnqueueCooperativeKernelLaunchExp` that calls `urEnqueueKernelLaunch`.


kbenzie (approved):
OpenCL changes LGTM

JackAKirk:
I think the main design considerations for this interface are the following

- Which backends currently support the feature? _I know cuda does from cuda 11 for at least ampere. amd does too: https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/structhip_device_prop__t.html#a7e99d56223ca8e08a3b0281a8dc07a46 although from what I can tell amd only supports grid synchronization from mi300x https://www.amd.com/content/dam/amd/en/documents/instinct-tech-docs/instruction-set-architectures/amd-instinct-mi300-cdna3-instruction-set-architecture.pdf : see 2.2.2. Global Wave Sync (GWS) _
- Do the backends that support the feature have interfaces supported by this unified interface? _I think that hip interface matches cuda and I guess this doesn't change for mi300x: which them simplifies the problem to knowing whether cuda case is supported by this unified interface, so long as only those two backends support this feature. But I'm not 100% on these points. Do you know?_

nrspruit (commented):
LGTM for level-zero

JackAKirk:
> > * Which backends currently support the feature?
> 
> I think CUDA and Level Zero support device-wide synchronization. I don't know about HIP and OpenCL would require an extension.
> 
> > * Do the backends that support the feature have interfaces supported by this unified interface?
> 
> I'm not sure I understand this question. CUDA has the following two functions:
> 
>     * [`cudaLaunchCooperativeKernel`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EXECUTION.html#group__CUDART__EXECUTION_1g504b94170f83285c71031be6d5d15f73)
> 
>     * [`cudaOccupancyMaxActiveBlocksPerMultiprocessor`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__HIGHLEVEL.html#group__CUDART__HIGHLEVEL_1g5a5d67a3c907371559ba692195e8a38c)
> 
> 
> L0 has two similar functions:
> 
>     * [`zeCommandListAppendLaunchCooperativeKernel`](https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=coop#zecommandlistappendlaunchcooperativekernel)
> 
>     * [`zeKernelSuggestMaxCooperativeGroupCount`](https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=coop#ze__api_8h_1af0b050a6cc08132ef84a8618942ce125)
> 
> 
> In addition to the kernel function, the CUDA occupancy query also requires a block size and dynamic memory usage. I'm not sure how L0 implements a similar behavior without these parameters. The two API pairs seem otherwise equivalent. I believe these functions can be used to provide implementations for the CUDA and L0 UR adapters.

OK thanks, I'll look into the HIP case.

scripts/core/exp-cooperative-kernels.yml
```diff
@@ -78,6 +78,12 @@ params:
     - type: $x_kernel_handle_t
       name: hKernel
       desc: "[in] handle of the kernel object"
+    - type: size_t
+      name: localWorkSize
+      desc: "[in] number of local work-items that will form a work-group when the kernel is launched"
```
JackAKirk, line 83: This maps directly to the CUDA interface. However the HIP interface expresses the local work size as a three dimensional range (DIM3 in hip parlance): https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___module.html#gaa516e011bb07d01550102c98adb57ec2

You get get a 1D work-group size from a 3D range, but not vice versa (or at least the mapping is not 1:1). So it would appear it is more natural that this parameter corresponds to a vector of ints/size_t/3d range etc. This is assuming that the triple is not redundant with regards to behaviour for the hip backend, such that it would make a difference if you e.g. map workgroup size 512 to (16,8,4) vs (8,8,8). I don't know the answer to this question.

I think the same issue applies to the global range parameter.

JackAKirk: You are right. Sorry my mistake.

source/adapters/cuda/kernel.cpp
```diff
@@ -169,6 +169,16 @@ UR_APIEXPORT ur_result_t UR_APICALL urKernelGetNativeHandle(
   return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
 }
 
+UR_APIEXPORT ur_result_t UR_APICALL urKernelSuggestMaxCooperativeGroupCountExp(
+    ur_kernel_handle_t hKernel, size_t localWorkSize,
+    size_t dynamicSharedMemorySize, uint32_t *pGroupCountRet) {
+  (void)hKernel;
+  (void)localWorkSize;
+  (void)dynamicSharedMemorySize;
+  *pGroupCountRet = 1;
```
JackAKirk, line 178: nit: Personally I'd just mark these guys unimplemented for now, since it could be confusing to return 1 for all input cases.

But probably not a big deal.

JackAKirk (approved):

aarongreig:
since these are new entry point implementations you'll need to add them to the interface loader file for each adapter, like here for l0 https://github.com/oneapi-src/unified-runtime/blob/1732c6b12929ab5d99b54e316fed16f374d3eda7/source/adapters/level_zero/ur_interface_loader.cpp#L449

aarongreig (approved):
I think the fail was a CI issue we were having, it should go away now

kbenzie:
> @kbenzie, can this PR be merged? I think it's ready.

We've got a bit of a backlog of [ready to merge](https://github.com/oneapi-src/unified-runtime/pulls?q=is%3Aopen+is%3Apr+label%3A%22ready+to+merge%22) we'll aim to get this merge ASAP.

PR #1248, opened 2024-01-12
Add common pool handle and memory provider classes to the common umf helpers

aarongreig:


aarongreig:
@kswiecicki obviously this patch isn't finished (especially on the l0 side) but just a heads up that I started working on adding pool support to the CL adapter and due to the greater need there to support arbitrary allocation flags I ended up making a bunch of changes to the UMF helper stuff in `common/`, which turned into this PR. No pressure to properly review it yet but if you find the time I would value your feedback on the approach I've taken to the common pool handle and memory provider classes

source/common/ur_pool_manager.hpp
```diff
@@ -158,63 +194,83 @@ inline bool pool_descriptor::operator==(const pool_descriptor &other) const {
         }
     }
 
+    if (lhs.allocLocation.has_value()) {
```
igchor, line 197: nit: I think you can simplify it to: `if (lhs.allocLocation != rhs.allocLocation) return false;`

igchor: Ah, nevermind, `pNext` might be different. I'm wondering if we should perhaps implement operator== for ur_usm_alloc_location_desc_t ...

source/common/ur_pool_manager.hpp
```diff
@@ -158,63 +194,83 @@ inline bool pool_descriptor::operator==(const pool_descriptor &other) const {
...
+            descriptors.emplace_back(poolHandle, hContext, device,
+                                     UR_USM_TYPE_SHARED, &desc);
         }
     }
 
     return {ret, descriptors};
 }
 
+inline usm::DisjointPoolMemType
+urTypeToDisjointPoolType(const ur_usm_type_t &type) {
```
igchor, line 260: What about SharedReadOnly?

source/common/ur_pool_manager.hpp
```diff
@@ -158,63 +194,83 @@ inline bool pool_descriptor::operator==(const pool_descriptor &other) const {
...
+            return false;
+        }
+
+        if (lhs.allocLocation.value().location !=
+            rhs.allocLocation.value().location) {
+            return false;
+        }
+    }
+
     return lhsNative == rhsNative && lhs.type == rhs.type &&
```
igchor, line 208: What about isSharedAllocationReadOnlyOnDevice? Is there no distinction between shared and sharedReadOnly pools anymore?

source/common/ur_pool_handle.hpp
```diff
@@ -0,0 +1,152 @@
...
+        auto PoolType = usm::urTypeToDisjointPoolType(desc.type);
+        std::tie(Ret, newPool) =
+            createUMFPoolForDesc<usm::DisjointPool, HostProvider,
+                                 DeviceProvider, SharedProvider>(
+                desc, this->DisjointPoolConfigs.Configs[PoolType]);
+        if (Ret) {
+            throw umf::UsmAllocationException(Ret);
+        }
+
+        PoolManager.addPool(desc, newPool);
```
igchor, line 147: nit: I think it would be less error-prone to make addPool accept uniquq_ptr by value so that the caller would need to std::move() the argument.

source/common/ur_pool_manager.hpp
aarongreig: the distinction is now encoded in `deviceFlags`, the only difference before was that ReadOnlyOnDevice pools passed the `UR_USM_DEVICE_MEM_FLAG_DEVICE_READ_ONLY` flag to the shared allocation entry point

source/common/ur_pool_manager.hpp
aarongreig: I think only having these three types makes sense because they correspond to the three allocation entry points in UR, read only pools will still be distinct by the fact that they're TYPE_SHARED and pass the appropriate read only alloc flag

source/common/ur_pool_manager.hpp
igchor: Right, that makes sense.

source/common/ur_pool_manager.hpp
igchor: The problem is that Shared and SharedReadOnly pools have different defaults (specifically, for Shared pooling is disabled by [default](https://github.com/oneapi-src/unified-runtime/blob/main/source/common/umf_pools/disjoint_pool_config_parser.cpp#L53)) and we also allow users to change the pooling behavior for each of those 4 types separately through UR_L0_USM_ALLOCATOR (aka SYCL_PI_LEVEL_ZERO_USM_ALLOCATOR). This env variable is not very well documented and is still considered 'experimental' but it's been useful for some applications.

aarongreig: Oh I see.. I was worried that the extra type was more than it first seemed. This aspect probably needs a re-design in that case, I'll be revisiting this PR once the pool manager PRs merge and I can rebase

PR #1252, opened 2024-01-16
[L0] move platform cache into the adapter structure

### LINKED ISSUE #824: Platform conformance test hangs on L0 adapter

pbalcer:
The platform cache is a global variable used exclusively by the L0 adapter, and it's protected by a loosely-associated spin lock. However, its destruction is associated with the lifetime of the adapter structure and is deleted the first time adapter refcount reaches 0. This was causing issues whenever the adapter was initialized and destroyed multiple time inside of a single process, which, for example, happens during tests.

This patch fixes the above problem by moving the platform cache from the global state into the adapter structure. This allows for a simpler implementation that no longer requires an explicit lock and instead uses lazy loading (std::call_once).

With this patch, all platform tests are now passing for L0. Closes #824

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -213,15 +213,7 @@ jobs:
       working-directory: ${{github.workspace}}/build
       run: ctest -C ${{matrix.build_type}} --output-on-failure -L "adapter-specific" --timeout 180
 
-    # Temporarily disabling platform test for L0, because of hang
-    # See issue: #824
-    - name: Test L0 adapter
-      if: matrix.adapter.name == 'L0'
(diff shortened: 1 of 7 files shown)


pbalcer:
https://github.com/intel/llvm/pull/12403

nrspruit (approved):
LGTM for the level-zero changes, thank you!

PR #1254, opened 2024-01-16
[EXP][CMDBUF] HIP adapter support for command buffers

mfrancepillois:
- command buffer functions for HIP.

- Some additional methods and refactoring to reuse existing adapter code.
(This code was mainly written by Andrey Alekseenko @al42and) 

Co-authored-by: Andrey Alekseenko andrey.alekseenko@scilifelab.se

### DIFF
--- a/source/adapters/hip/command_buffer.cpp
+++ b/source/adapters/hip/command_buffer.cpp
@@ -9,185 +9,1093 @@
 //===----------------------------------------------------------------------===//
 
 #include "command_buffer.hpp"
-#include "common.hpp"
 
-/// Stub implementations of UR experimental feature command-buffers
+#include <hip/hip_runtime.h>
+
+#include "common.hpp"
+#include "enqueue.hpp"
+#include "event.hpp"
+#include "kernel.hpp"
+#include "memory.hpp"
+#include "queue.hpp"
+
+#include <cstring>
+
+namespace {
+ur_result_t
+commandBufferReleaseInternal(ur_exp_command_buffer_handle_t CommandBuffer) {
+  if (CommandBuffer->decrementInternalReferenceCount() != 0) {
+    return UR_RESULT_SUCCESS;
+  }
+
+  delete CommandBuffer;
+  return UR_RESULT_SUCCESS;
+}
+
+ur_result_t
+commandHandleReleaseInternal(ur_exp_command_buffer_command_handle_t Command) {
+  if (Command->decrementInternalReferenceCount() != 0) {
+    return UR_RESULT_SUCCESS;
+  }
+
+  // Decrement parent command-buffer internal ref count
+  commandBufferReleaseInternal(Command->CommandBuffer);
+
+  delete Command;
+  return UR_RESULT_SUCCESS;
+}
+} // end anonymous namespace
+
+ur_exp_command_buffer_handle_t_::ur_exp_command_buffer_handle_t_(
+    ur_context_handle_t hContext, ur_device_handle_t hDevice, bool IsUpdatable)
+    : Context(hContext), Device(hDevice),
+      IsUpdatable(IsUpdatable), HIPGraph{nullptr}, HIPGraphExec{nullptr},
+      RefCountInternal{1}, RefCountExternal{1} {
+  urContextRetain(hContext);
+  urDeviceRetain(hDevice);
+}
+
+/// The ur_exp_command_buffer_handle_t_ destructor releases
+/// all the memory objects allocated for command_buffer managment
+ur_exp_command_buffer_handle_t_::~ur_exp_command_buffer_handle_t_() {
+  // Release the memory allocated to the Context stored in the command_buffer
+  UR_TRACE(urContextRelease(Context));
+
+  // Release the device
+  UR_TRACE(urDeviceRelease(Device));
+
+  // Release the memory allocated to the HIPGraph
+  UR_CHECK_ERROR(hipGraphDestroy(HIPGraph));
+
+  // Release the memory allocated to the HIPGraphExec
+  if (HIPGraphExec) {
+    UR_CHECK_ERROR(hipGraphExecDestroy(HIPGraphExec));
+  }
+}
+
+ur_exp_command_buffer_command_handle_t_::
+    ur_exp_command_buffer_command_handle_t_(
+        ur_exp_command_buffer_handle_t CommandBuffer, ur_kernel_handle_t Kernel,
+        std::shared_ptr<hipGraphNode_t> &&Node, hipKernelNodeParams Params,
+        uint32_t WorkDim, const size_t *GlobalWorkOffsetPtr,
+        const size_t *GlobalWorkSizePtr, const size_t *LocalWorkSizePtr)
+    : CommandBuffer(CommandBuffer), Kernel(Kernel), Node(std::move(Node)),
+      Params(Params), WorkDim(WorkDim), RefCountInternal(1),
+      RefCountExternal(1) {
+  CommandBuffer->incrementInternalReferenceCount();
+
+  const size_t CopySize = sizeof(size_t) * WorkDim;
+  std::memcpy(GlobalWorkOffset, GlobalWorkOffsetPtr, CopySize);
+  std::memcpy(GlobalWorkSize, GlobalWorkSizePtr, CopySize);
+  // Local work size may be nullptr
+  if (LocalWorkSizePtr) {
+    std::memcpy(LocalWorkSize, LocalWorkSizePtr, CopySize);
+  } else {
+    std::memset(LocalWorkSize, 0, sizeof(size_t) * 3);
+  }
+
+  if (WorkDim < 3) {
+    const size_t ZeroSize = sizeof(size_t) * (3 - WorkDim);
+    std::memset(GlobalWorkOffset + WorkDim, 0, ZeroSize);
+    std::memset(GlobalWorkSize + WorkDim, 0, ZeroSize);
+  }
+}
+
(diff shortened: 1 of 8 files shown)


mfrancepillois:
Liked DPC++ PR: https://github.com/intel/llvm/pull/12230

source/adapters/hip/enqueue.cpp
```diff
@@ -17,11 +18,9 @@
 
 extern size_t imageElementByteSize(hipArray_Format ArrayFormat);
 
-namespace {
-
-ur_result_t enqueueEventsWait(ur_queue_handle_t, hipStream_t Stream,
-                              uint32_t NumEventsInWaitList,
-                              const ur_event_handle_t *EventWaitList) {
+UR_APIEXPORT ur_result_t UR_APICALL enqueueEventsWait(
```
EwanC, line 21: I don't think we should have `UR_APIEXPORT ` and `UR_APICALL ` here, as this isn't a UR spec defined entry-point. Same for `simpleGuessLocalWorkSize`

source/adapters/hip/enqueue.hpp
```diff
@@ -0,0 +1,37 @@
...
+                        ur_kernel_handle_t Kernel, uint32_t LocalSize);
+
+ur_result_t
+setKernelParams(const ur_device_handle_t Device, const uint32_t WorkDim,
+                const size_t *GlobalWorkOffset, const size_t *GlobalWorkSize,
+                const size_t *LocalWorkSize, ur_kernel_handle_t &Kernel,
+                hipFunction_t &HIPFunc, size_t (&ThreadsPerBlock)[3],
+                size_t (&BlocksPerGrid)[3]);
+
+void setCopyRectParams(ur_rect_region_t region, const void *SrcPtr,
```
EwanC, line 32: mixing snake_case and CamelCase in the parameter variable naming, same in .cpp definition

EwanC (approved):

source/adapters/hip/enqueue.hpp
mfrancepillois: I updated the variable names to make them CamelCase compliant.

source/adapters/hip/enqueue.cpp
mfrancepillois: You're right. This file has been updated accordingly.

source/adapters/hip/command_buffer.hpp
```diff
@@ -9,7 +9,222 @@
...
 #include <ur/ur.hpp>
+#include <ur_api.h>
 
-/// Stub implementation of command-buffers for HIP
+#include "context.hpp"
+#include <hip/hip_runtime.h>
+#include <memory>
 
-struct ur_exp_command_buffer_handle_t_ {};
+static inline const char *getUrResultString(ur_result_t Result) {
```
jchlanda, line 18: Is there no way of sharing this string-ifier, a quick grep through the repo reveals it's been done in 3 other places:
* https://github.com/oneapi-src/unified-runtime/blob/b2c7e83521949cd6039cdd1c3874ed0bbd238939/include/ur_print.hpp#L328
* https://github.com/oneapi-src/unified-runtime/blob/b2c7e83521949cd6039cdd1c3874ed0bbd238939/source/adapters/level_zero/common.hpp#L30
* https://github.com/oneapi-src/unified-runtime/blob/b2c7e83521949cd6039cdd1c3874ed0bbd238939/source/adapters/cuda/command_buffer.hpp#L18

source/adapters/hip/command_buffer.hpp
```diff
@@ -9,7 +9,222 @@
...
+    SyncPoints[SyncPoint] = HIPNode;
+    NextSyncPoint++;
+  }
+
+  ur_exp_command_buffer_sync_point_t GetNextSyncPoint() const {
+    return NextSyncPoint;
+  }
+
+  // Helper to register next sync point
+  // @param HIPNode Node to register as next sycn point
```
jchlanda, line 196:
```suggestion
  // @param HIPNode Node to register as next sync point
```

source/adapters/hip/command_buffer.cpp
```diff
@@ -9,156 +9,756 @@
...
+  UR_TRACE(urDeviceRelease(Device));
+
+  // Release the memory allocated to the HIPGraph
+  UR_CHECK_ERROR(hipGraphDestroy(HIPGraph));
+
+  // Release the memory allocated to the HIPGraphExec
+  UR_CHECK_ERROR(hipGraphExecDestroy(HIPGraphExec));
+}
+
+/// Helper function for finding the HIP Nodes associated with the
```
jchlanda, line 48: Seems like this could be re-formatted:
```
/// Helper function for finding the HIP Nodes associated with the commands in a
/// command-buffer, each event is pointed to by a sync-point in
```

source/adapters/hip/command_buffer.cpp
```diff
@@ -9,156 +9,756 @@
...
+/// @param[in] SyncPointWaitList List of sync points in \p CommandBuffer
+/// to find the events for.
+/// @param[out] HipNodesList Return parameter for the HIP Nodes associated with
+/// each sync-point in \p SyncPointWaitList.
+///
+/// @return UR_RESULT_SUCCESS or an error code on failure
+static ur_result_t getNodesFromSyncPoints(
+    const ur_exp_command_buffer_handle_t &CommandBuffer,
+    size_t NumSyncPointsInWaitList,
+    const ur_exp_command_buffer_sync_point_t *SyncPointWaitList,
```
jchlanda, line 63: Is `SuncPointWaitList` passed as a pointer (as opposed to by reference )because it's an array? If so (and it's a general comment, not just for this func) would it be good to assert on it having a value set prior to using it. Or are we sure that the values will never be null?

source/adapters/hip/command_buffer.cpp
```diff
@@ -9,156 +9,756 @@
...
+      // into 1 byte values, and set the buffer using multiple strided calls.
+      // This means that one hipGraphAddMemsetNode call is made for every 1
+      // bytes in the pattern.
+
+      // List to handle inter-node dependencies
+      std::vector<hipGraphNode_t> HIPNodesList = {};
+      // List shared pointer that will point to the last node created
+      std::shared_ptr<hipGraphNode_t> GraphNodePtr;
+
+      size_t NumberOfSteps = PatternSize / sizeof(uint8_t);
```
jchlanda, line 133: If want 4 bytes, should we be not dividing by size of `uint32_t`?

source/adapters/hip/command_buffer.cpp
```diff
@@ -9,156 +9,756 @@
...
+            UR_RESULT_ERROR_INVALID_KERNEL);
+  UR_ASSERT(workDim > 0, UR_RESULT_ERROR_INVALID_WORK_DIMENSION);
+  UR_ASSERT(workDim < 4, UR_RESULT_ERROR_INVALID_WORK_DIMENSION);
+
+  ur_result_t Result = UR_RESULT_SUCCESS;
+  hipGraphNode_t GraphNode;
+
+  std::vector<hipGraphNode_t> DepsList;
+
+  UR_CALL(getNodesFromSyncPoints(hCommandBuffer, numSyncPointsInWaitList,
```
jchlanda, line 266: This might be a personal preference, but I find accumulation of `Result` introducing unnecessary complexity to the logic. The value is only really important at one place, here: https://github.com/oneapi-src/unified-runtime/blob/b2c7e83521949cd6039cdd1c3874ed0bbd238939/source/adapters/hip/command_buffer.cpp#L266
It would be cleaner to return `Err` in the `catch` and `UR_RESULT_SUCCESS` at the end of the function. 

The same patter of needles result-accumulation is repeated in many functions.

source/adapters/hip/command_buffer.cpp
mfrancepillois: A step consists in adding a node that performs a `uint8_t` memset. 
So, the `NumberOfSteps` must be computed using `uint8_t`.
However, I agree that some fields of `NodeParamsStepFirst` were not set correctly. I updated them.

source/adapters/hip/command_buffer.cpp
mfrancepillois: This has be re-formatted.

source/adapters/hip/command_buffer.hpp
mfrancepillois: This makes sense. Thanks for the suggestions. However, we prefer to keep this PR focusing only on graph support for the HIP backend and open another refactoring PR that will touch all implemented backends.

source/adapters/hip/command_buffer.cpp
mfrancepillois: `SyncPointWaitList` is provided by the UR API and we forward it directly to this internal function without reformatting it.
We added asserts entry points functions to check the value of `pSyncPointWaitList`.

source/adapters/hip/command_buffer.cpp
mfrancepillois: The code has been updated to avoid this.

source/adapters/hip/command_buffer.hpp
jchlanda: OK, do we need a ticket or will it be handled internally?

jchlanda (approved):

source/adapters/hip/command_buffer.hpp
EwanC: I'll create an internal ticket to do this

jchlanda: Thanks @EwanC 🏅

aarongreig (approved):
CTS changes LGTM

PR #1256, opened 2024-01-16
[EXP][CMDBUF] Fix dependency handling for large buffer fill ops in CUDA graph support

mfrancepillois:
The CUDA backend does not handle buffer fill larger than 32 bits. 
The implemented workaround is to add several node that perform 32 bits buffer fill ops. 
In the previous implementation, all these nodes took all previous node(s) as predecessors. 
In this version, the first node takes all previous node(s) as predecessors, then the subsequent node takes the newly added node as a predecessor which ensures that the whole fill ops is completed when the last node in this sequence is completed.

### DIFF
--- a/source/adapters/cuda/command_buffer.cpp
+++ b/source/adapters/cuda/command_buffer.cpp
@@ -148,6 +148,9 @@ static ur_result_t enqueueCommandBufferFillHelper(
 
       size_t NumberOfSteps = PatternSize / sizeof(uint32_t);
 
+      // List shared pointer that will point to the last node created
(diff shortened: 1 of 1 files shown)


EwanC (approved):

mfrancepillois:
Linked DPC++ PR: https://github.com/intel/llvm/pull/12405

Bensuo (approved):

mfrancepillois:
PR https://github.com/oneapi-src/unified-runtime/pull/1319 includes this bugfix and other improvements for the buffer fill op in CUDA (Command-Buffer).
Consequently, this PR can be closed.

PR #1257, opened 2024-01-17
[L0] fix waiting on non-owned native handle events

pbalcer:
L0 adapter makes optimizations based on whether an event has any external references. The adapter keeps track of this based on a reference count. This reference count wasn't being properly set when first creating an event based on a native handle, leading to code incorrectly assuming that the event was internal and had no external references, which ultimately led to segfaults.

This patch sets the refcount to 1 when first creating an event, solving the problem. It also adds a missing null-check to make user native handle events even work.


pbalcer:
https://github.com/intel/llvm/pull/12425

nrspruit (approved):
LGTM

PR #1258, opened 2024-01-17
[UTIL] Make sure Windows.h doesn't define min

kbenzie:
Define `NOMINMAX` before including Windows.h to fix the building in
https://github.com/intel/llvm/pull/12375.

### DIFF
--- a/source/common/ur_util.hpp
+++ b/source/common/ur_util.hpp
@@ -22,6 +22,9 @@
 #include <vector>
 
 #ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
 #include <windows.h>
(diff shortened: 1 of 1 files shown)


callumfare (approved):

pbalcer:
All targets should already have this problem fixed:
https://github.com/oneapi-src/unified-runtime/blob/main/cmake/helpers.cmake#L87

kbenzie:
> All targets should already have this problem fixed:
> https://github.com/oneapi-src/unified-runtime/blob/main/cmake/helpers.cmake#L87

This PI plugins aren't inheriting this build option. Saying that, might be possible to fix it in that CMake instead of here.

kbenzie:
Superceeded by https://github.com/oneapi-src/unified-runtime/pull/1260, closing.

PR #1259, opened 2024-01-17
[L0] do not ignore returned values from zeHostSynchronize

igchor:
Instead, return the error immediately. Ignoring the errors, resulted in silent failure and unexpected errors from next operations.


igchor:
This PR fixes a specific issue but I've noticed that some other functions are also called without checking return value, eg. here: https://github.com/oneapi-src/unified-runtime/blob/536f31a8eeb9ac60314149e88e8772d8b5249058/source/adapters/level_zero/queue.cpp#L712C5-L712C22 and in some other places. Are some of those by design or is this just an oversight? If it's the latter then perhaps we should mark every (or most?) functions that return ur_result_t as [[nodiscard]]?

pbalcer (approved):
+1 to adding nodiscard, but I'm not sure if we can just add it en masse without then having to spend weeks trying to solve issues.

source/adapters/level_zero/queue.cpp
```diff
@@ -438,7 +438,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urQueueRelease(
       return Res;
 
     // Make sure all commands get executed.
-    Queue->synchronize();
+    UR_CALL(Queue->synchronize());
```
pbalcer, line 441: I'm not sure if this isn't leaking the queue object. But other similar checks also don't do anything with it so dunno :)

nrspruit (approved):
LGTM, I think this might be related to another bug being encountered that delayed the error to the next kernel after the previous event failed. Please merge this.

pbalcer:
@igchor please create an intel/llvm testing PR.

source/adapters/level_zero/queue.cpp
igchor: Yeah, I didn't see any cleanup in other parts of the code as well.

If we actually end up failing here (on this call or any other in this function) it will lead to abort on SYCL level because urQueueRelease is called inside a queue dtor and PI translates UR errors into exceptions (dtor is noexcept) so we don't need to worry about leaks. I don't know is this is desired but that's how it is right now. 

However, I guess it would nice to have some tests on UR level with error injection perhaps so we could check for leaks.

igchor:
> @igchor please create an intel/llvm testing PR (after #1260 is merged).

Rebased and created a testing PR: https://github.com/intel/llvm/pull/12427

PR #1260, opened 2024-01-18
[Common] Don't include Windows.h in ur_util.hpp

kbenzie:
Avoid transitively including Windows.h in headers which include ur_util.hpp.


pbalcer (approved):
are the added casts in opencl related?

kbenzie:
> are the added casts in opencl related?

They turned up in my Windows L0+CL build, I just checked and they are also present on main so no. I'll split those out into a separate PR.

PR #1262, opened 2024-01-18
[CI] Add CI config for Native CPU

martygrant:


.github/workflows/cmake.yml
```diff
@@ -170,6 +170,7 @@ jobs:
           {name: HIP, triplet: amdgcn-amd-amdhsa, platform: ""},
           {name: L0, triplet: spir64, platform: ""},
           {name: OPENCL, triplet: spir64, platform: "Intel(R) OpenCL"}
+          {name: NATIVECPU, triplet: spir64, platform: ""}
```
lukaszstolarczuk, line 173: just want to confirm if you're aware - the `name` is also used as a label for self-hosted runner (see `runs-on: ${{matrix.adapter.name}}` below). I don't believe we have such a runner available at the moment.

lukaszstolarczuk: ok, the new runner should be up and running now - please verify if it works as needed 😉

martygrant: Thank you! I started a new job but realised the label for the runner is `NATIVECPU`, the naming in the workflow and CMake is `NATIVE_CPU`. Could you please edit the runner label? Thanks!

lukaszstolarczuk: Done

.github/workflows/cmake.yml
```diff
@@ -170,6 +170,7 @@ jobs:
           {name: HIP, platform: ""},
           {name: L0, platform: ""},
           {name: OPENCL, platform: "Intel(R) OpenCL"}
```
lukaszstolarczuk, line 172: you're missing a coma at the end of this line, and the workflow won't start

PR #1265, opened 2024-01-19
[Bindless][Exp] Add support for device to device copies between CuArrays

cppchedy:
This PR:
* Introduce support for device to device copy in 1D, 2D and 3D.
* target only `Cuarray`s

corresponding PR in https://github.com/intel/llvm/pull/12449

### DIFF
--- a/source/adapters/cuda/image.cpp
+++ b/source/adapters/cuda/image.cpp
@@ -898,8 +898,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
         UR_CHECK_ERROR(cuMemcpy3DAsync(&cpy_desc, Stream));
       }
       // Synchronization is required here to handle the case of copying data
-      // from
-      // host to device, then device to device and finally device to host.
+      // from host to device, then device to device and finally device to host.
       // Without it, there is a risk of the copies not being executed in the
       // intended order.
       cuStreamSynchronize(Stream);
--- a/source/adapters/cuda/image.cpp
+++ b/source/adapters/cuda/image.cpp
@@ -659,8 +659,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
     enqueueEventsWait(hQueue, Stream, numEventsInWaitList, phEventWaitList);
 
     // We have to use a different copy function for each image dimensionality.
-    // All the async copy function should be treated as synchronous because of
-    // the explicit call to cuStreamSynchronize at the end
 
     if (imageCopyFlags == UR_EXP_IMAGE_COPY_FLAG_HOST_TO_DEVICE) {
       if (pImageDesc->type == UR_MEM_TYPE_IMAGE1D) {
@@ -834,6 +832,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
       }
     } else {
       // imageCopyFlags == UR_EXP_IMAGE_COPY_FLAG_DEVICE_TO_DEVICE
+
+      // All the following async copy function calls should be treated as
+      // synchronous because of the explicit call to cuStreamSynchronize at
+      // the end
       if (pImageDesc->type == UR_MEM_TYPE_IMAGE1D) {
         CUDA_MEMCPY2D cpy_desc = {};
         cpy_desc.srcXInBytes = srcOffset.x;
--- a/source/adapters/cuda/image.cpp
+++ b/source/adapters/cuda/image.cpp
@@ -690,12 +690,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
         }
       } else if (pImageDesc->type == UR_MEM_TYPE_IMAGE2D) {
         CUDA_MEMCPY2D cpy_desc = {};
-        cpy_desc.srcMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_HOST;
-        cpy_desc.srcHost = pSrc;
         cpy_desc.srcXInBytes = srcOffset.x * PixelSizeBytes;
         cpy_desc.srcY = srcOffset.y;
         cpy_desc.dstXInBytes = dstOffset.x * PixelSizeBytes;
         cpy_desc.dstY = dstOffset.y;
+        cpy_desc.srcMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_HOST;
+        cpy_desc.srcHost = pSrc;
         cpy_desc.srcPitch = hostExtent.width * PixelSizeBytes;
         if (pImageDesc->rowPitch == 0) {
           cpy_desc.dstMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_ARRAY;
@@ -781,8 +781,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
         cpy_desc.srcY = srcOffset.y;
         cpy_desc.dstXInBytes = dstOffset.x;
         cpy_desc.dstY = dstOffset.y;
-        cpy_desc.dstMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_HOST;
-        cpy_desc.dstHost = pDst;
         if (pImageDesc->rowPitch == 0) {
           cpy_desc.srcMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_ARRAY;
           cpy_desc.srcArray = (CUarray)pSrc;
@@ -792,6 +790,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
           cpy_desc.srcPitch = pImageDesc->rowPitch;
           cpy_desc.srcDevice = (CUdeviceptr)pSrc;
         }
+        cpy_desc.dstMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_HOST;
+        cpy_desc.dstHost = pDst;
         cpy_desc.WidthInBytes = PixelSizeBytes * copyExtent.width;
         cpy_desc.Height = copyExtent.height;
         UR_CHECK_ERROR(cuMemcpy2DAsync(&cpy_desc, Stream));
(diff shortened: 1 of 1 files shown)


Seanst98 (approved):

isaacault (approved):

DBDuncan (approved):

source/adapters/cuda/image.cpp
```diff
@@ -816,9 +816,49 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
...
+        cpy_desc.dstZ = dstOffset.z;
+        cpy_desc.srcMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_ARRAY;
+        cpy_desc.srcArray = (CUarray)pSrc;
+        cpy_desc.dstMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_ARRAY;
+        cpy_desc.dstArray = (CUarray)pDst;
+        cpy_desc.WidthInBytes = PixelSizeBytes * copyExtent.width;
+        cpy_desc.Height = copyExtent.height;
+        cpy_desc.Depth = copyExtent.depth;
+        UR_CHECK_ERROR(cuMemcpy3DAsync(&cpy_desc, Stream));
+      }
```
isaacault, line 861: What about image arrays?

isaacault (changes requested):

source/adapters/cuda/image.cpp
cppchedy: implemented now.

source/adapters/cuda/image.cpp
```diff
@@ -675,14 +675,15 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
...
       } else if (pImageDesc->type == UR_MEM_TYPE_IMAGE2D) {
         CUDA_MEMCPY2D cpy_desc = {};
-        cpy_desc.srcMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_HOST;
-        cpy_desc.srcHost = pSrc;
         cpy_desc.srcXInBytes = srcOffset.x * PixelSizeBytes;
         cpy_desc.srcY = srcOffset.y;
         cpy_desc.dstXInBytes = dstOffset.x * PixelSizeBytes;
         cpy_desc.dstY = dstOffset.y;
+        cpy_desc.srcMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_HOST;
+        cpy_desc.srcHost = pSrc;
```
isaacault, line 686: Was this an accident? Could it be reverted please?

source/adapters/cuda/image.cpp
```diff
@@ -814,11 +821,73 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
         cpy_desc.Height = std::max(uint64_t{1}, copyExtent.height);
         cpy_desc.Depth = pImageDesc->arraySize;
         UR_CHECK_ERROR(cuMemcpy3DAsync(&cpy_desc, Stream));
+        cuStreamSynchronize(Stream);
       }
     } else {
-      /// imageCopyFlags == UR_EXP_IMAGE_COPY_FLAG_DEVICE_TO_DEVICE
```
isaacault, line 819: Could we keep this comment? Always helpful on an `else` like this.

source/adapters/cuda/image.cpp
```diff
@@ -675,14 +675,15 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
           // This should be unreachable.
           return UR_RESULT_ERROR_INVALID_VALUE;
         }
+        cuStreamSynchronize(Stream);
```
isaacault, line 678: Can we move this out of the conditional statements and just have it once after?

isaacault (changes requested):

source/adapters/cuda/image.cpp
cppchedy: >Was this an accident?

No, it was changed to make the it consistent with other if blocks.

source/adapters/cuda/image.cpp
isaacault: Not sure where the weird formatting came on the comment, but while we're here can you fix it to start with only two "//"?

cppchedy: fixed.

source/adapters/cuda/image.cpp
cppchedy: done

isaacault (approved):

isaacault (changes requested):
Sorry for just noticing this now, but can the title and description of this PR be updated to remove mention of the mem handle? 

This PR doesn't touch any mem handles, and any comments about it don't belong in this project but rather in intel/llvm. This PR is strictly about implementing D2D copies.

source/adapters/cuda/image.cpp
```diff
@@ -816,10 +816,70 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
...
+        cpy_desc.srcArray = (CUarray)pSrc;
+        cpy_desc.dstMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_ARRAY;
+        cpy_desc.dstArray = (CUarray)pDst;
+        cpy_desc.WidthInBytes = PixelSizeBytes * copyExtent.width;
+        cpy_desc.Height = std::max(uint64_t{1}, copyExtent.height);
+        cpy_desc.Depth = pImageDesc->arraySize;
+        UR_CHECK_ERROR(cuMemcpy3DAsync(&cpy_desc, Stream));
+      }
     }
+    // required sync to fix the random failure in device_to_device_copy tests
```
isaacault, line 881: I think `the random failure in device_to_deivce_copy tests` is quite vague and implicative of an over-arching issue that must be obvious. I don't know about the failure, but depending on the nature can I suggest one of 

- Create an issue to track it, and link it in this thread
- Explain in the comment why this is necessary (what difference it makes, what's observed without it)
- Explain the failure observed in this thread, and reword the comment to omit mention of the failure

isaacault (changes requested):

source/adapters/cuda/image.cpp
cppchedy: The is tracked internally with details on what have been attempted to address the issue.

I updated the comments to include a brief description of the test case that trigger the random failure and what happens without the sync.

isaacault (approved):

source/adapters/cuda/image.cpp
```diff
@@ -831,10 +831,74 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
...
+        cpy_desc.srcY = 0;
+        cpy_desc.dstXInBytes = dstOffset.x;
+        cpy_desc.dstY = 0;
+        cpy_desc.srcMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_ARRAY;
+        cpy_desc.srcArray = (CUarray)pSrc;
+        cpy_desc.dstMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_ARRAY;
+        cpy_desc.dstArray = (CUarray)pDst;
+        cpy_desc.WidthInBytes = PixelSizeBytes * copyExtent.width;
+        cpy_desc.Height = 1;
+        UR_CHECK_ERROR(cuMemcpy2DAsync(&cpy_desc, Stream));
```
przemektmalon, line 847: I'm wondering whether we should include comments for all of these async calls that we are actually synchronizing immediately below, because at first glance someone might think that these are actually asynchronous without looking lower down.

source/adapters/cuda/image.cpp
```diff
@@ -831,10 +831,74 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
...
+        cpy_desc.Height = std::max(uint64_t{1}, copyExtent.height);
+        cpy_desc.Depth = pImageDesc->arraySize;
+        UR_CHECK_ERROR(cuMemcpy3DAsync(&cpy_desc, Stream));
+      }
     }
+    // Synchronization is required here to handle the case of copying data from
+    // host to device, then device to device and finally device to host.
+    // Without it, there is a risk of the copies not being executed in the
+    // intended order.
+    cuStreamSynchronize(Stream);
```
przemektmalon, line 901: Is this synchronization in the right scope? Right now it's synchronizing for all directions of copy, do we not only need it to synchronize inside the Device to Device `if` branch?

If we really need to synchronize for all directions of copies to fix this issue, then this isn't a viable solution. In that case we would need to just use the default stream for all copies.

Thinking about it, I wonder if that is just a better solution performance wise than synchronizing, as using the default stream asynchronously might allow for CUDA to better schedule operations among it's other streams. It also means that `sycl::event`'s returned from all copies including Device to Device are valid and meaningful, which is not the case if we always synchronize the Device to Device copies.

source/adapters/cuda/image.cpp
```diff
@@ -831,10 +831,74 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
...
+        cpy_desc.srcZ = srcOffset.z;
+        cpy_desc.dstXInBytes = dstOffset.x;
+        cpy_desc.dstY = dstOffset.y;
+        cpy_desc.dstZ = dstOffset.z;
+        cpy_desc.srcMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_ARRAY;
+        cpy_desc.srcArray = (CUarray)pSrc;
+        cpy_desc.dstMemoryType = CUmemorytype_enum::CU_MEMORYTYPE_ARRAY;
+        cpy_desc.dstArray = (CUarray)pDst;
+        cpy_desc.WidthInBytes = PixelSizeBytes * copyExtent.width;
+        cpy_desc.Height = std::max(uint64_t{1}, copyExtent.height);
```
przemektmalon, line 892: Is `copyExtent.height == 0` ever a valid value? 

I would think that if the user passes `0` as the copy extent height then this should get caught somewhere further up in the runtime and throw an error. If we're not doing that check somewhere else, then having the `std::max` here might mask the invalid value passed by the user.

source/adapters/cuda/image.cpp
cppchedy: > In that case we would need to just use the default stream for all copies.

this works for the device to device test but causes failures in  other tests. I didn't investigate the cause of the failure for the later and switched to sync.

source/adapters/cuda/image.cpp
cppchedy: I think in the case of `UR_MEM_TYPE_IMAGE1D_ARRAY` we set the height to be 0 when we create the descriptor. so I think it's valid here.

source/adapters/cuda/image.cpp
cppchedy: >If we really need to synchronize for all directions of copies to fix this issue, then this isn't a viable solution.

I tested now and the only solution is to a global sync. if we limit it to the device to device block it fails.

>If we really need to synchronize for all directions of copies to fix this issue, then this isn't a viable solution. 

I agree with that. currently, this is tracked internally to provide a solution that take full advantage of the Async nature if CUDA copy APIs.

cppchedy: >I tested now and the only solution is to a global sync. if we limit it to the device to device block it fails.

actually this is wrong, moving it to the device to device scope works. I applied your suggestion.

source/adapters/cuda/image.cpp
cppchedy: good point. I added a comment saying that the CUDA APIs copy calls, in the device to device block, should be thought of as synchronous calls.

source/adapters/cuda/image.cpp
```diff
@@ -831,10 +831,80 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
...
+        cpy_desc.Depth = pImageDesc->arraySize;
+        UR_CHECK_ERROR(cuMemcpy3DAsync(&cpy_desc, Stream));
+      }
+      // Synchronization is required here to handle the case of copying data
+      // from
```
DBDuncan, line 901: Formatting is odd here.

cppchedy: fixed. thanks for pointing that out.

przemektmalon (approved):

PR #1280, opened 2024-01-24
Fix typo in FindRocmAgentEnumerator.cmake

aarongreig:
Also reset HIP compiler args between targets in device code generation
to avoid generating unused argument warnings.


kbenzie (approved):

lukaszstolarczuk (commented):
👍

PR #1281, opened 2024-01-24
[urinfo] Fix stray NULLs in stdout

al42and:
We were putting null-terminated strings to std::string data, which made the terminator part of the string content.


kbenzie (approved):

al42and:
HIP test failure seems unrelated?

kbenzie:
> HIP test failure seems unrelated?

Yes, they were resolved in https://github.com/oneapi-src/unified-runtime/pull/1283

PR #1282, opened 2024-01-24
Fix size confusion for several device property queries

al42and:
- `UR_DEVICE_INFO_SUB_GROUP_SIZES_INTEL`: The spec says it's an array of uint32_t, but CUDA, HIP, OpenCL, and LevelZero adapters were returning arrays of size_t. As a result, we had extra dummy zeroes in the output. Native CPU adapter was correct.
- `UR_DEVICE_INFO_ATOMIC_*_CAPABILITIES`: Should be `ur_memory_scope_capability_flags_t`, was `uint64_t` in CUDA/HIP.

### DIFF
--- a/source/adapters/opencl/device.cpp
+++ b/source/adapters/opencl/device.cpp
@@ -910,7 +910,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   case UR_EXT_DEVICE_INFO_OPENCL_C_VERSION:
   case UR_DEVICE_INFO_BUILT_IN_KERNELS:
   case UR_DEVICE_INFO_MAX_WORK_ITEM_SIZES:
-  case UR_DEVICE_INFO_SUB_GROUP_SIZES_INTEL:
   case UR_DEVICE_INFO_IP_VERSION: {
     /* We can just use the OpenCL outputs because the sizes of OpenCL types
      * are the same as UR.
@@ -928,6 +927,19 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
                         propSize, pPropValue, pPropSizeRet));
 
     return UR_RESULT_SUCCESS;
+  }
+  case UR_DEVICE_INFO_SUB_GROUP_SIZES_INTEL: {
+    // Have to convert size_t to uint32_t
+    size_t SubGroupSizesSize = 0;
+    CL_RETURN_ON_FAILURE(
+        clGetDeviceInfo(cl_adapter::cast<cl_device_id>(hDevice), CLPropName, 0,
+                        nullptr, &SubGroupSizesSize));
+    std::vector<size_t> SubGroupSizes(SubGroupSizesSize / sizeof(size_t));
+    CL_RETURN_ON_FAILURE(
+        clGetDeviceInfo(cl_adapter::cast<cl_device_id>(hDevice), CLPropName,
+                        SubGroupSizesSize, SubGroupSizes.data(), nullptr));
+    return ReturnValue.template operator()<uint32_t>(SubGroupSizes.data(),
+                                                     SubGroupSizes.size());
   }
   case UR_DEVICE_INFO_EXTENSIONS: {
     cl_device_id Dev = cl_adapter::cast<cl_device_id>(hDevice);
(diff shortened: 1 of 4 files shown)


nrspruit (approved):
LGTM for level-zero

al42and:
Added more fixes of the same nature.

al42and:
PR for testing: https://github.com/intel/llvm/pull/12488

source/adapters/opencl/device.cpp
```diff
@@ -907,6 +906,19 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
 
     return UR_RESULT_SUCCESS;
   }
+  case UR_DEVICE_INFO_SUB_GROUP_SIZES_INTEL: {
+    // Have to convert size_t to uint32_t
```
jchlanda, line 910: Do we need to check that the values fit in 32 bits when potentially downcasting from `size_t`?

al42and: > Do we need to check that the values fit in 32 bits when potentially downcasting from size_t?

For GPUs, I am not aware of any hardware with sub-group size of >64. We can add an assert if you think it's helpful. But, IMO, if we get a sub-group size of >2^31, a lot of other assumptions will be broken too :)

jchlanda: 😄

jchlanda (approved):

aarongreig (approved):

al42and:
I'm stupefied by the test failures:

```

/home/test-user/actions-runner/_work/unified-runtime/unified-runtime/test/conformance/device/urDeviceGetInfo.cpp:273: Failure

Expected equality of these values:

  uur::Result(result)

    Which is: UR_RESULT_ERROR_INVALID_VALUE

  uur::Result(UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION)

    Which is: UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION

[  FAILED  ] urDeviceGetInfoTest.Success/UR_DEVICE_INFO_MAX_MEMORY_BANDWIDTH, where GetParam() = UR_DEVICE_INFO_MAX_MEMORY_BANDWIDTH (0 ms)
```

But the code was explicitly returning the "invalid value" error even prior to this PR:

```
  case UR_DEVICE_INFO_MAX_MEMORY_BANDWIDTH:
    // currently not supported in level zero runtime
    return UR_RESULT_ERROR_INVALID_VALUE;
```

Are they expected?

aarongreig:
Unfortunately this is a quirk of how those CI jobs work. `MAX_MEMORY_BANDWIDTH` isn't the problem, the job is failing because a test is now unexpectedly passing so the output no longer matches the [match file](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/device/device_adapter_level_zero.match) for this suite/adapter combination https://github.com/oneapi-src/unified-runtime/actions/runs/7658304835/job/20870934554?pr=1282#step:8:93 

this patch should make that fail go away:
```
diff --git a/test/conformance/device/device_adapter_level_zero.match b/test/conformance/device/device_adapter_level_zero.match
index e7f102b9..b1bff737 100644
--- a/test/conformance/device/device_adapter_level_zero.match
+++ b/test/conformance/device/device_adapter_level_zero.match
@@ -1,5 +1,4 @@
 {{OPT}}urDeviceGetGlobalTimestampTest.SuccessSynchronizedTime
-urDeviceGetInfoTest.Success/UR_DEVICE_INFO_IMAGE_SUPPORTED
 urDeviceGetInfoTest.Success/UR_DEVICE_INFO_GLOBAL_MEM_FREE
 urDeviceGetInfoTest.Success/UR_DEVICE_INFO_ERROR_CORRECTION_SUPPORT
 urDeviceGetInfoTest.Success/UR_DEVICE_INFO_HOST_UNIFIED_MEMORY
```

al42and:
@aarongreig: Thanks. A few small steps towards passing CTS :)

al42and:
Removing the `bool` change, since it was merged independently.

al42and:
Can this be merged? Having multiple `0` in the returned list of supported sub-group sizes is moderately annoying.

kbenzie:
> Can this be merged? Having multiple `0` in the returned list of supported sub-group sizes is moderately annoying.

Sorry, this fell off my radar. I've now added this to the merge queue and will aim to merge ASAP.

al42and:
> Sorry, this fell off my radar. I've now added this to the merge queue and will aim to merge ASAP.

No worries 😊

PR #1284, opened 2024-01-25
[CTS] Fix resource leaks

kbenzie:
This patch introduces the `uur::raii::Wrapper<T>` object and per handle
aliases (`uur::raii::Context` for `ur_context_handle_t` etc.) to replace
instances of the following pattern in test bodies:

```cpp
TEST(Context, Example) {
    ur_context_handle_t context = nullptr;
    ASSERT_SUCCESS(urContextCreate(1, &device, nullptr, &context));

    // potential resource leak as `ASSERT_*` returns from this function
    ASSERT_TRUE(might_be_false);

    urContextRelease(context);
}
```

With the following:

```cpp
TEST(Context, Example) {
    uur::raii::Context context = nullptr;
    ASSERT_SUCCESS(urContextCreate(1, &device, nullptr, context.ptr()));

    // no leak because the destructor releases the context
    ASSERT_TRUE(might_be_false);
}
```

### DIFF
--- a/test/conformance/testing/include/uur/raii.h
+++ b/test/conformance/testing/include/uur/raii.h
@@ -1,0 +1,114 @@
+// Copyright (C) 2022-2023 Intel Corporation
+// Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+// See LICENSE.TXT
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef UUR_RAII_H_INCLUDED
+#define UUR_RAII_H_INCLUDED
+
+#include "ur_api.h"
+#include <cstddef>
+#include <utility>
+
+namespace uur {
+namespace raii {
+template <class URHandleT, ur_result_t (*retain)(URHandleT),
+          ur_result_t (*release)(URHandleT)>
+struct Wrapper {
+    using handle_t = URHandleT;
+
+    handle_t handle;
+
+    Wrapper() : handle(nullptr) {}
+    explicit Wrapper(handle_t handle) : handle(handle) {}
+    ~Wrapper() {
+        if (handle) {
+            release(handle);
+        }
+    }
+
+    Wrapper(const Wrapper &other) : handle(other.handle) { retain(handle); }
+    Wrapper(Wrapper &&other) : handle(other.handle) { other.handle = nullptr; }
+    Wrapper(std::nullptr_t) : handle(nullptr) {}
+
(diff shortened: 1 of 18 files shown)


pbalcer:
This shouldn't be needed I think. Coverity is not considering gtest asserts as killpaths (see https://sig-product-docs.synopsys.com/bundle/coverity-docs/page/customizing_coverity/topics/models_primitives/examples/c_models_examples_c_cpp_killpath.html). I spent a few hours trying to fix this, but it's not that simple (gtest asserts are macros).

kbenzie:
There are some simple cases like asserting the a returned handle is null isn't taken into consideration by the static analysis. However, I do think the majority of these could be a potential resource leak if the appropriate combination of assertions fail, especially in the more complex test cases.

A common pattern here is test bodies which create then release one or more handles with `ASSERT_*` between the create/release calls. If any one of those `ASSERT_*`s fail, then its a resource leak.

The alternative approach to resolving these would be to move the handles into the test fixture and release them in `TearDown()` but I decided against that as it would be a more substantial rewrite of the tests.

pbalcer:
Makes sense. I still think that coercing coverity to somewhat take the gtest asserts into account would help avoid false positives for other sort of resources, but I'm not sure whether that's even possible...

test/conformance/testing/include/uur/raii.h
```diff
@@ -0,0 +1,94 @@
...
+OVERLOAD(ur_sampler_handle_t, urSamplerRelease)
+OVERLOAD(ur_usm_pool_handle_t, urUSMPoolRelease)
+OVERLOAD(ur_physical_mem_handle_t, urPhysicalMemRelease)
+OVERLOAD(ur_program_handle_t, urProgramRelease)
+OVERLOAD(ur_kernel_handle_t, urKernelRelease)
+OVERLOAD(ur_queue_handle_t, urQueueRelease)
+OVERLOAD(ur_event_handle_t, urEventRelease)
+#undef OVERLOAD
+
+template <class URHandleT> struct Wrapper {
```
pbalcer, line 33: You could use a non-type template parameter like I did [here](https://github.com/oneapi-src/unified-runtime/pull/1286/files#diff-d0ae1ee0edc109c505db620df503c365a6daf95b5f4ef9f12ba06a7d16e0b3d8R47).

pbalcer (approved):

test/conformance/testing/include/uur/raii.h
kbenzie: macro deleted.

fabiomestre (approved):

PR #1291, opened 2024-01-29
[CUDA] Report that devices with cc >= sm_70 support seq_cst

JackAKirk:
Report that devices with cc >= sm_70 support seq_cst memory consistency capabilities. Tested with https://github.com/intel/llvm/pull/12516


GeorgeWeb (approved):
LGTM! :partying_face: 

Btw, small aside, I think at some point it may be good we cache stuff like `CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR` in the device on init, as we seem to use it for a bunch of device cap queries, so while not much it may be useful to remove the overhead (or rather move it to a one-time on init).

JackAKirk:
I've also now marked fence seq_cst supported by sm70. This is already the situation in the latest intel/llvm, see https://github.com/intel/llvm/pull/12713 for a description.
I've unmarked this as ready to merge just to get an updated review for this small update. I think it makes sense to do both changes in the one PR since.

GeorgeWeb (approved):

kbenzie:
Please pull in the main branch to have up to date testing, also update the tag in the intel/llvm PR.

PR #1296, opened 2024-01-30
[ASAN] Fix copy instead of move

kbenzie:


kbenzie:
This resolves a Coverity issue:

> COPY_INSTEAD_OF_MOVE (COPY_INSTEAD_OF_MOVE)
> 1. copy_assignment_call: AllocInfo is copied in a call to copy assignment operator =, when it could be moved instead.

fabiomestre (approved):

AllanZyne (approved):

PR #1297, opened 2024-01-30
[L0] fix caching internal events

pbalcer:
When the L0 adapter enqueues an operation, it creates an event for its own bookkeeping needs and to optionally notify the application about completion. These events are periodically cleaned up and the event objects themselves are cached to be reused. Or at least they are supposed to be. If the enqueue does not specify a user event (i.e., the event is L0 internal) the event is destroyed and the L0 event handle is simply leaked in urEventReleaseInternal. This means that every `MaxNumEventsPerPool` enqueue calls (256 by default) a new event pool is created.

This patch fixes the above problem by making such internal events own the native L0 handles, enabling them to be cached.

I found this problem in the microbenchmark from #1286. This patch nearly triples its performance on some systems.


pbalcer:
I don't think this PR makes sense in isolation, but the default behavior where the events leak by default if not created by the application isn't ideal. This should probably be revisited eventually.

PR #1299, opened 2024-01-30
[CUDA] Fix MaxRegsPerBlock check in setKernelParams

rafbiels:
Fix the computation of the block size passed to the `MaxRegsPerBlock` check. The size is a _product_ of the dimensions, not a _sum_.

Also change the error code returned if `MaxRegsPerBlock` is exceeded from `INVALID_WORK_GROUP_SIZE` to `OUT_OF_RESOURCES`. This reflects the check better and makes it consistent with the CUDA error returned in such cases. The previous fail mode produced confusing errors mentioning non-uniform work group sizes even if they were uniform (see the linked ticket below).

Fixes https://github.com/intel/llvm/issues/12363 - see comments therein for more details and a reproducer.

intel/llvm test PR: https://github.com/intel/llvm/pull/12549


source/adapters/cuda/enqueue.cpp
```diff
@@ -242,18 +242,18 @@ setKernelParams(const ur_context_handle_t Context,
...
             return Err;
           // If no error then sum the total local work size per dim.
-          KernelLocalWorkGroupSize += LocalWorkSize[Dim];
+          KernelLocalWorkGroupSize *= LocalWorkSize[Dim];
         }
 
         if (hasExceededMaxRegistersPerBlock(Device, Kernel,
                                             KernelLocalWorkGroupSize)) {
-          return UR_RESULT_ERROR_INVALID_WORK_GROUP_SIZE;
+          return UR_RESULT_ERROR_OUT_OF_RESOURCES;
```
GeorgeWeb, line 256: It is worth checking that the error handling in the sycl runtime in intel/llvm that prints the error message will still happen if you change the error code. I remember it was tied to the invalid wg size in the handleInvalidWorkgroupSizeError function.

GeorgeWeb (commented):
Thank you very much for the finding of my super silly mistake. Despite me describing it in the PR, I still did the wrong math here for some odd reason.

source/adapters/cuda/enqueue.cpp
rafbiels: In my testing the output of a program exceeding the register count changed from:
```
terminate called after throwing an instance of 'sycl::_V1::nd_range_error'
  what():  Non-uniform work-groups are not supported by the target device -54 (PI_ERROR_INVALID_WORK_GROUP_SIZE)
```
to:
```
terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  Native API failed. Native API returns: -5 (PI_ERROR_OUT_OF_RESOURCES) -5 (PI_ERROR_OUT_OF_RESOURCES)
```
It's perhaps not as informative as it could be, but at least it's not misleading any more.

We list possible reasons for "out of resources" errors in [our troubleshooting guide](https://developer.codeplay.com/products/oneapi/nvidia/2024.0.1/guides/troubleshooting#out-of-resources-on-kernel-launch), so it's potentially googlable.

I would very much support making the error message more informative, but would probably leave it for a separate PR, covering also other errors.

hdelan (approved):
Nice catch, LGTM!

source/adapters/cuda/enqueue.cpp
```diff
@@ -242,18 +242,18 @@ setKernelParams(const ur_context_handle_t Context,
           return UR_RESULT_SUCCESS;
         };
 
-        size_t KernelLocalWorkGroupSize = 0;
+        size_t KernelLocalWorkGroupSize = 1;
         for (size_t Dim = 0; Dim < WorkDim; Dim++) {
           auto Err = IsValid(Dim);
           if (Err != UR_RESULT_SUCCESS)
             return Err;
           // If no error then sum the total local work size per dim.
```
GeorgeWeb, line 250: Can we update the comment to say product here as well, please?

source/adapters/cuda/enqueue.cpp
rafbiels: Upon further discussion, removed the error code change from this PR. I agree it's better to leave it for another PR paired with intel/llvm changes in error handling.

source/adapters/cuda/enqueue.cpp
rafbiels: Good spot, updated!

GeorgeWeb (approved):

PR #1302, opened 2024-01-31
[CL] Handle INTERMEDIATE binary type

kbenzie:
If an OpenCL driver returns `CL_PROGRAM_BINARY_TYPE_INTERMEDIATE` (defined in the `cl_khr_spir` extension to enable SPIR 1.2 not SPIR-V) when querying `CL_PROGRAM_BINARY_TYPE` we should handle this gracefully. This patch maps the `CL_PROGRAM_BINARY_TYPE_INTERMEDIATE` to `UR_PROGRAM_BINARY_TYPE_NONE`. From the users perspective these are semantically equivalent as they both require compilation before the program can be used to create a kernel.

### DIFF
--- a/source/adapters/opencl/program.cpp
+++ b/source/adapters/opencl/program.cpp
@@ -259,6 +259,15 @@ static cl_int mapURProgramBuildInfoToCL(ur_program_build_info_t URPropName) {
 static ur_program_binary_type_t
 mapCLBinaryTypeToUR(cl_program_binary_type binaryType) {
   switch (binaryType) {
+  default:
+    // If we don't understand what OpenCL gave us, return NONE.
+    // TODO: Emit a warning to the user.
+    [[fallthrough]];
+  case CL_PROGRAM_BINARY_TYPE_INTERMEDIATE:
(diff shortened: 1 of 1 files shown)


source/adapters/opencl/program.cpp
```diff
@@ -259,6 +259,15 @@ static cl_int mapURProgramBuildInfoToCL(ur_program_build_info_t URPropName) {
 static ur_program_binary_type_t
 mapCLBinaryTypeToUR(cl_program_binary_type binaryType) {
   switch (binaryType) {
+  default:
+    // If we don't understand what OpenCL gave us, return NONE.
+    // TODO: Emit a warning to the user.
+    [[fallthrough]];
+  case CL_PROGRAM_BINARY_TYPE_INTERMEDIATE:
+    // The INTERMEDIOATE binary type is defined by the cl_khr_spir extension
```
aarongreig, line 267:
```suggestion
    // The INTERMEDIATE binary type is defined by the cl_khr_spir extension
```

aarongreig (approved):

PR #1303, opened 2024-01-31
Fix counting variables for CTS

szadam:
Some tests don't work on supermicro machine:
https://github.com/oneapi-src/unified-runtime/actions/runs/7559294673/job/20587325535

This PR fixes it.

### DIFF
--- a/test/conformance/CMakeLists.txt
+++ b/test/conformance/CMakeLists.txt
@@ -4,8 +4,6 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 set(UR_CONFORMANCE_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR})
-set(UR_TEST_DEVICES_COUNT 1 CACHE STRING "Count of devices on which CTS will be run")
-set(UR_TEST_PLATFORMS_COUNT 1 CACHE STRING "Count of platforms on which CTS will be run")
 
 function(add_test_adapter name adapter)
     set(TEST_TARGET_NAME test-${name})
--- a/test/conformance/CMakeLists.txt
+++ b/test/conformance/CMakeLists.txt
@@ -4,8 +4,8 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 set(UR_CONFORMANCE_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR})
-option(UR_TEST_DEVICES_COUNT "Count of devices on which CTS will be run" 1)
-option(UR_TEST_PLATFORMS_COUNT "Count of platforms on which CTS will be run" 1)
+set(UR_TEST_DEVICES_COUNT 1 CACHE STRING "Count of devices on which CTS will be run")
+set(UR_TEST_PLATFORMS_COUNT 1 CACHE STRING "Count of platforms on which CTS will be run")
 
 function(add_test_adapter name adapter)
     set(TEST_TARGET_NAME test-${name})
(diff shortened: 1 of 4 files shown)


PatKamin (approved):

test/conformance/testing/include/uur/environment.h
```diff
@@ -17,7 +17,7 @@ struct PlatformEnvironment : ::testing::Environment {
 
     struct PlatformOptions {
         std::string platform_name;
-        unsigned long platforms_count = 0;
+        unsigned long platforms_count = 1;
```
lukaszstolarczuk, line 20: can we make use of the CMake option/variable in all tests, so the change in here is not required...?

szadam: I have changed it.

test/adapters/cuda/kernel_tests.cpp
```diff
@@ -74,9 +74,9 @@ const char *threeParamsTwoLocal = "\n\
 TEST_P(cudaKernelTest, CreateProgramAndKernel) {
 
     uur::raii::Program program = nullptr;
-    ASSERT_SUCCESS(urProgramCreateWithBinary(context, device, sizeof(ptxSource),
-                                             (const uint8_t *)ptxSource,
-                                             nullptr, program.ptr()));
+    ASSERT_SUCCESS(urProgramCreateWithBinary(
```
lukaszstolarczuk, line 77: heh, pls remove Benie's commit from this PR

test/conformance/CMakeLists.txt
```diff
@@ -4,8 +4,8 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 set(UR_CONFORMANCE_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR})
-option(UR_TEST_DEVICES_COUNT "Count of devices on which CTS will be run" 1)
-option(UR_TEST_PLATFORMS_COUNT "Count of platforms on which CTS will be run" 1)
+set(UR_TEST_DEVICES_COUNT 1 CACHE STRING "Count of devices on which CTS will be run")
```
lukaszstolarczuk, line 7: soo, since this is now used in `conformance` and `adapters` tests, maybe it should be moved one CMake level up...?

and also, description could be updated (not only CTS)

kbenzie: I'm also wondering if this is the best mechanism for limiting which devices tests run on, long term I'd expect `ONEAPI_DEVICE_SELECTOR` to be a more flexible and accurate solution.

test/adapters/cuda/kernel_tests.cpp
szadam: Done

test/conformance/CMakeLists.txt
szadam: I have changed it. I think we could leave it like that for now.

lukaszstolarczuk (approved):

test/conformance/CMakeLists.txt
lukaszstolarczuk: Benie, yeah, I guess that be nice. I can see we have some issues (at least somehow) related to the `ONEAPI_DEVICE_SELECTOR ` topic, so this is a bit larger change. This is good enough for this fix, I guess.

kbenzie: Yeah, this fix makes sense for how things stand today.

pbalcer: This is pretty much blocking CTS from running in multi-device configurations, and we have at least one now.

PR #1304, opened 2024-02-01
Optimization for SYCL out of order queue 

winstonzhang-intel:
have to iron out a few more things, but the initial concept is this: 
for regular commandlist, we want to give the events device scope thus avoiding L3 flush. To achieve this, I have a list of all events with device scope being kept track on within the queue. When we get to the last one, we want that commandlist's event to have host scope and also to wait for all other device scoped events.
Note to self: need to find out how to get total number of commandlists within a queue

### DIFF
--- a/source/adapters/level_zero/kernel.cpp
+++ b/source/adapters/level_zero/kernel.cpp
@@ -209,9 +209,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
   ur_event_handle_t InternalEvent{};
   bool IsInternal = OutEvent == nullptr;
   ur_event_handle_t *Event = OutEvent ? OutEvent : &InternalEvent;
+  bool IsLastCmdList = !Queue->UsingImmCmdLists && OutEvent != nullptr;
 
   UR_CALL(createEventAndAssociateQueue(Queue, Event, UR_COMMAND_KERNEL_LAUNCH,
                                        CommandList, IsInternal));
+
   ZeEvent = (*Event)->ZeEvent;
   (*Event)->WaitList = TmpWaitList;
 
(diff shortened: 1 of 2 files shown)


source/adapters/level_zero/kernel.cpp
```diff
@@ -229,6 +233,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
   if (IndirectAccessTrackingEnabled)
     Queue->KernelsToBeSubmitted.push_back(Kernel);
 
+  if (!Queue->AppendedEvents) {
+    Queue->AppendedEvents = new ze_event_handle_t[Queue->totalCmdListCount - 1];
```
nrspruit, line 237: So, I would make this obvious or leave a comment here saying that the "AppendedEvents" are device scope.

ie maybe change to DeviceAppendedEvents or something like that.

source/adapters/level_zero/kernel.cpp
```diff
@@ -252,20 +262,31 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
   urPrint("calling zeCommandListAppendLaunchKernel() with"
           "  ZeEvent %#" PRIxPTR "\n",
           ur_cast<std::uintptr_t>(ZeEvent));
   printZeEventList((*Event)->WaitList);
-
   // Execute command list asynchronously, as the event will be used
   // to track down its completion.
   UR_CALL(Queue->executeCommandList(CommandList, false, true));
 
+  if (LastCmdList && !Queue->UsingImmCmdLists) {
```
nrspruit, line 287: The user should be waiting on this final event thru urEventWait so this should not be necessary.

If the user created n events and a provided you with those events, then what you would want to do is map the "internal events" to the host visible events. Our events have that property already to have the device event linked to the host visible event. What you will want to do is given the user calls urEventWait, if that event is device only (ie the AppendEvents), then we would wait on the final event, then signal the user's host events.

source/adapters/level_zero/kernel.cpp
```diff
@@ -229,6 +233,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
   if (IndirectAccessTrackingEnabled)
     Queue->KernelsToBeSubmitted.push_back(Kernel);
 
+  if (!Queue->AppendedEvents) {
+    Queue->AppendedEvents = new ze_event_handle_t[Queue->totalCmdListCount - 1];
+  }
+  Queue->AppendedEvents[Queue->CmdListEnqueued] = ZeEvent;
```
nrspruit, line 239: This event is host visible, you need to create a device only scoped event for each event and then set the HostVisibleEvent=ZeEvent

PR #1311, opened 2024-02-02
[CUDA] Use camelCase always (applied to usm_p2p.cpp).

JackAKirk:
Use camelCase naming convention.

This makes the naming convention consistent with the switch to camelCase in other files.


ldrumm (changes requested):
This is two separate patches.

The first one is NFC and changes only the variable names
The second changes control flow.

The second could be the source of subtle bugs, and doesn't belong with such a commit message

JackAKirk:
> This is two separate patches.
> 
> The first one is NFC and changes only the variable names The second changes control flow.
> 
> The second could be the source of subtle bugs, and doesn't belong with such a commit message

Thanks, I've removed that part and added it here: https://github.com/oneapi-src/unified-runtime/pull/1317

mmoadeli (approved):

ldrumm (approved):

PR #1318, opened 2024-02-05
[CUDA] Use appropriate return code for out of registers kernel launch

GeorgeWeb:
This PR changes the returned error code for exiting the kernel launch entry point in CUDA when exceeding the maximum available registers for execution on the SM. Previously we were returning a misleading error code.

This change requires DPC++ to adapt the specific-error handling, see PR: https://github.com/intel/llvm/pull/12604.


ldrumm (changes requested):
> Change the returned error code for exiting the kernel launch entry point in CUDA when exceeding the maximum available registers for execution on the SM. Previously we were returning a misleading error code

Put that in your commit message, please

ldrumm (approved):

PR #1320, opened 2024-02-07
Fix CI regression by generating missing bits in nullddi and ldrddi

aarongreig:


aarongreig:
I'm a little baffled as to how this could get past the original CI runs on the PR but I guess always hit that update branch button before merging...

martygrant (approved):

pbalcer (approved):

PR #1321, opened 2024-02-08
[L0] move adapter init into its constructor from urAdapterGet

pbalcer:
This restores the old behavior where urAdapterGet was essentially a noop. Hopefully fixes the issue on intel/llvm (https://github.com/intel/llvm/pull/12658).

### DIFF
--- a/source/adapters/level_zero/adapter.cpp
+++ b/source/adapters/level_zero/adapter.cpp
@@ -11,8 +11,6 @@
 #include "adapter.hpp"
 #include "ur_level_zero.hpp"
 
-ur_adapter_handle_t_ Adapter{};
-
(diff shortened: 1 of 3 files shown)


pbalcer:
intel/llvm CI is passing with this patch https://github.com/intel/llvm/pull/12658.

Please prioritize reviewing this patch, it's urgently needed to unblock the CI and make progress on 0.8.x release.

The patch itself is just moving code around (+ solves one minor issue with error handling). Prior to this patch, the adapter/level-zero was initialized on first call to urAdapterGet. With this change, it will now happen whenever the platforms are first accessed.

aarongreig (approved):

nrspruit (approved):

PR #1323, opened 2024-02-08
[CUDA][HIP] Refactor guess local worksize funcs

hdelan:
The guessLocalWorkSize func for cuda adapter was erroneously giving a factor of 2 wg size for the x dimension, even if it creates a global dim which is not the same as the user range. This may have caused out of bounds accesses in the CUDA adapter. This reverts that.

Also refactor the HIP version a little bit to remove an unnecessary check.

### DIFF
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -169,18 +169,10 @@ void guessLocalWorkSize(ur_device_handle_t Device, size_t *ThreadsPerBlock,
   ThreadsPerBlock[0] = std::min(
       MaxThreadsPerBlock[0], std::min(GlobalSizeNormalized[0], MaxBlockDim[0]));
 
-  static auto IsPowerOf2 = [](size_t Value) -> bool {
-    return Value && !(Value & (Value - 1));
-  };
-
   // Find a local work group size that is a divisor of the global
   // work group size to produce uniform work groups.
-  // Additionally, for best compute utilisation, the local size has
-  // to be a power of two.
-  while (0u != (GlobalSizeNormalized[0] % ThreadsPerBlock[0]) ||
-         !IsPowerOf2(ThreadsPerBlock[0])) {
+  while (GlobalSizeNormalized[0] % ThreadsPerBlock[0])
     --ThreadsPerBlock[0];
-  }
 }
 
 // Helper to verify out-of-registers case (exceeded block max registers).
--- a/source/adapters/hip/enqueue.cpp
+++ b/source/adapters/hip/enqueue.cpp
@@ -51,13 +51,9 @@ ur_result_t enqueueEventsWait(ur_queue_handle_t, hipStream_t Stream,
 
 void simpleGuessLocalWorkSize(size_t *ThreadsPerBlock,
                               const size_t *GlobalWorkSize,
-                              const size_t MaxThreadsPerBlock[3],
-                              ur_kernel_handle_t Kernel) {
+                              const size_t MaxThreadsPerBlock[3]) {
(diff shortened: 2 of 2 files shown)


source/adapters/cuda/enqueue.cpp
```diff
@@ -169,18 +169,10 @@ void guessLocalWorkSize(ur_device_handle_t Device, size_t *ThreadsPerBlock,
...
-    return Value && !(Value & (Value - 1));
-  };
-
   // Find a local work group size that is a divisor of the global
   // work group size to produce uniform work groups.
-  // Additionally, for best compute utilisation, the local size has
-  // to be a power of two.
-  while (0u != (GlobalSizeNormalized[0] % ThreadsPerBlock[0]) ||
-         !IsPowerOf2(ThreadsPerBlock[0])) {
+  while (GlobalSizeNormalized[0] % ThreadsPerBlock[0])
```
jchlanda, line 174: The comment here suggests that power of 2 sizes are desirable, is it not possible to modify this patch to still prefer the power of 2 when possible, and fall back to division to prevent out of bounds accesses, or are we sure that there is no performance implication?

hdelan: The work that I am doing in SYCL RT ensures that the range passed to the adapter will always be a factor of 32. 

Maybe I should change this so that it prefers something that divides the global range *and* is a power of 2, instead of the current *or* check.

hdelan: Oh wait I am stupid. This was already doing the right thing. I will close this PR

jchlanda: I wonder how realistic that would be, as in would we often end up with the size being `1`?

hdelan: I think as far as UR is concerned that's OK. I think it is the responsiblity of higher level APIs to choose a sensible range. I am doing this work atm in SYCL RT and other users of UR should do the same.

GeorgeWeb: This is the PR https://github.com/intel/llvm/pull/9736 that introduced the power of two check. That fixes the sycl::range regression that was caught in this issue: https://github.com/intel/llvm/issues/5627. In fact, you can run BableStream to check for regressions. I am certain with power of two we were getting the best performance. For example, for cuda a power of two wg size may not always be necessary, if the size is uniform and multiple of 32 (due  Due to warp granularity). 

If the global global work size allows I think we'll always end up getting a multiple of the warp size but if it does not then it was best to go with power of 2 from the perf. benchmarking hence why the condition was introduced.

Definitely check BabelStream and the above application referred to the issue I linked. It is worth because if no other changes have been made there may again be regression.

My original proposal was to keep the warp-size multiple as the preferred one where uniformity is possible, and based on performance evaluation the not super necessary power-of-two check gives actually gives more bandwidth performance, hence why I stuck with it. However, I did not do very good to handle the other case with a fallback. I think I support @jchlanda to combine both efforts and keep the power of two check preference when possible and fallback when not.

GeorgeWeb: Ignore me, you figured it all while I was slow-typing lol...

hdelan: Thanks @GeorgeWeb . I saw some out of bounds accesses with a test I was running due to something happening here, and I thought it was this thing but I think it's got to do with the Y and Z dims. Will reopen this when I have a fix

hdelan:
I couldn't reopen this PR since I force pushed to the branch so this is the new PR https://github.com/oneapi-src/unified-runtime/pull/1326 .

PR #1326, opened 2024-02-08
[CUDA][HIP] Fix bug in guess local worksize funcs and improve local worksize guessing in HIP adapter

hdelan:
The guessLocalWorkSize func for cuda adapter was erroneously giving large Y or Z factors, rounding up when it should not. This ensures that the Y and Z factors always divide the global Y or Z dimension.

It also tidies up the HIP adapter version of this func.

### DIFF
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -18,6 +18,7 @@
 
 #include <cmath>
 #include <cuda.h>
+#include <ur/ur.hpp>
 
 ur_result_t enqueueEventsWait(ur_queue_handle_t CommandQueue, CUstream Stream,
                               uint32_t NumEventsInWaitList,
@@ -140,12 +141,10 @@ ur_result_t setCuMemAdvise(CUdeviceptr DevPtr, size_t Size,
 void guessLocalWorkSize(ur_device_handle_t Device, size_t *ThreadsPerBlock,
                         const size_t *GlobalWorkSize, const uint32_t WorkDim,
                         const size_t MaxThreadsPerBlock[3],
-                        ur_kernel_handle_t Kernel, uint32_t LocalSize) {
+                        ur_kernel_handle_t Kernel) {
   assert(ThreadsPerBlock != nullptr);
   assert(GlobalWorkSize != nullptr);
   assert(Kernel != nullptr);
-  int MinGrid, MaxBlockSize;
-  size_t MaxBlockDim[3];
 
   // The below assumes a three dimensional range but this is not guaranteed by
   // UR.
@@ -154,33 +153,18 @@ void guessLocalWorkSize(ur_device_handle_t Device, size_t *ThreadsPerBlock,
     GlobalSizeNormalized[i] = GlobalWorkSize[i];
   }
 
+  size_t MaxBlockDim[3];
+  MaxBlockDim[0] = MaxThreadsPerBlock[0];
   MaxBlockDim[1] = Device->getMaxBlockDimY();
   MaxBlockDim[2] = Device->getMaxBlockDimZ();
 
-  UR_CHECK_ERROR(
-      cuOccupancyMaxPotentialBlockSize(&MinGrid, &MaxBlockSize, Kernel->get(),
-                                       NULL, LocalSize, MaxThreadsPerBlock[0]));
+  int MinGrid, MaxBlockSize;
+  UR_CHECK_ERROR(cuOccupancyMaxPotentialBlockSize(
+      &MinGrid, &MaxBlockSize, Kernel->get(), NULL, Kernel->getLocalSize(),
+      MaxThreadsPerBlock[0]));
 
-  ThreadsPerBlock[2] = std::min(GlobalSizeNormalized[2], MaxBlockDim[2]);
-  ThreadsPerBlock[1] =
-      std::min(GlobalSizeNormalized[1],
-               std::min(MaxBlockSize / ThreadsPerBlock[2], MaxBlockDim[1]));
-  MaxBlockDim[0] = MaxBlockSize / (ThreadsPerBlock[1] * ThreadsPerBlock[2]);
-  ThreadsPerBlock[0] = std::min(
-      MaxThreadsPerBlock[0], std::min(GlobalSizeNormalized[0], MaxBlockDim[0]));
-
-  static auto IsPowerOf2 = [](size_t Value) -> bool {
-    return Value && !(Value & (Value - 1));
-  };
-
-  // Find a local work group size that is a divisor of the global
-  // work group size to produce uniform work groups.
-  // Additionally, for best compute utilisation, the local size has
-  // to be a power of two.
-  while (0u != (GlobalSizeNormalized[0] % ThreadsPerBlock[0]) ||
-         !IsPowerOf2(ThreadsPerBlock[0])) {
-    --ThreadsPerBlock[0];
-  }
+  roundToHighestFactorOfGlobalSizeIn3d(ThreadsPerBlock, GlobalSizeNormalized,
+                                       MaxBlockDim, MaxBlockSize);
 }
 
 // Helper to verify out-of-registers case (exceeded block max registers).
@@ -261,7 +245,7 @@ setKernelParams(const ur_context_handle_t Context,
         }
       } else {
         guessLocalWorkSize(Device, ThreadsPerBlock, GlobalWorkSize, WorkDim,
-                           MaxThreadsPerBlock, Kernel, LocalSize);
+                           MaxThreadsPerBlock, Kernel);
       }
     }
 
--- a/source/adapters/hip/enqueue.cpp
+++ b/source/adapters/hip/enqueue.cpp
@@ -16,6 +16,8 @@
 #include "memory.hpp"
 #include "queue.hpp"
 
+#include <ur/ur.hpp>
+
 extern size_t imageElementByteSize(hipArray_Format ArrayFormat);
 
 ur_result_t enqueueEventsWait(ur_queue_handle_t, hipStream_t Stream,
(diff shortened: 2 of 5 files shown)


hdelan:
DPC++ PR: https://github.com/intel/llvm/pull/12663

keyradical (approved):
Looks good to me! :+1:

GeorgeWeb (commented):
Non-blocker, just a suggestion.
It maybe worth to add "tidies up the HIP adapter version of this func" to the commit message, or split the changes in two commits one per each adapter. The first one for the functional the changes in the Cuda adapter and the second one for the NFC/tidying-up changes in the HIP adapter.

Otherwise, thanks for addressing this inefficiency! LGTM

hdelan:
> Non-blocker, just a suggestion. It maybe worth to add "tidies up the HIP adapter version of this func" to the commit message, or split the changes in two commits one per each adapter. The first one for the functional the changes in the Cuda adapter and the second one for the NFC/tidying-up changes in the HIP adapter.
> 
> Otherwise, thanks for addressing this inefficiency! LGTM

Good shout @GeorgeWeb ! Change made

hdelan:
Friendly ping @ldrumm have a few patches in SYCL RT that depend on this being fixed :smile_cat:

npmiller:
Should the HIP adapter have the same calculation as the CUDA one?

hdelan:
> Should the HIP adapter have the same calculation as the CUDA one?

I suppose it would be no harm to put this in HIP adapter as well

source/adapters/cuda/enqueue.cpp
```diff
@@ -157,30 +157,36 @@ void guessLocalWorkSize(ur_device_handle_t Device, size_t *ThreadsPerBlock,
...
-      cuOccupancyMaxPotentialBlockSize(&MinGrid, &MaxBlockSize, Kernel->get(),
-                                       NULL, LocalSize, MaxThreadsPerBlock[0]));
+  UR_CHECK_ERROR(cuOccupancyMaxPotentialBlockSize(
+      &MinGrid, &MaxBlockSize, Kernel->get(), NULL, Kernel->getLocalSize(),
+      MaxThreadsPerBlock[0]));
+
+  // Helper lambda to make sure each x, y, z dim divide the global dimension.
+  // Can optionally specify that we want the wg size to be a power of 2 in a
+  // given dimension, which is useful for the X dim for performance reasons.
+  static auto roundToHighestFactorOfGlobalSize =
```
ldrumm, line 167: I really don't like functions that have switches to change their behaviour such that they're essentially different functions.
I'd also prefer that unless a function is a callback or a closure, it should be hoisted to a free function that it's easier to grep, keep function bodies short, and keep structure consistent.

source/adapters/hip/enqueue.cpp
```diff
@@ -49,23 +49,59 @@ ur_result_t enqueueEventsWait(ur_queue_handle_t, hipStream_t Stream,
...
+  MaxBlockDim[2] = Device->getMaxBlockDimZ();
+
+  UR_CHECK_ERROR(hipOccupancyMaxPotentialBlockSize(
+      &MinGrid, &MaxBlockSize, Kernel->get(), Kernel->getLocalSize(),
+      MaxThreadsPerBlock[0]));
+
+  // Helper lambda to make sure each x, y, z dim divide the global dimension.
+  // Can optionally specify that we want the wg size to be a power of 2 in a
+  // given dimension, which is useful for the X dim for performance reasons.
+  static auto roundToHighestFactorOfGlobalSize =
```
ldrumm, line 82: Again here. Also, since they should be identical implementations, this could be better shared if possible - we don't want to have to maintain duplicate code everywhere.

source/adapters/hip/enqueue.cpp
```diff
@@ -49,23 +49,59 @@ ur_result_t enqueueEventsWait(ur_queue_handle_t, hipStream_t Stream,
...
+// The default threadsPerBlock only require handling the first work_dim
+// dimension.
+void guessLocalWorkSize(ur_device_handle_t Device, size_t *ThreadsPerBlock,
+                        const size_t *GlobalWorkSize, const uint32_t WorkDim,
+                        const size_t MaxThreadsPerBlock[3],
+                        ur_kernel_handle_t Kernel) {
   assert(ThreadsPerBlock != nullptr);
   assert(GlobalWorkSize != nullptr);
   assert(Kernel != nullptr);
+  int MinGrid, MaxBlockSize;
```
ldrumm, line 62: These are far from their first use.

source/adapters/hip/enqueue.cpp
```diff
@@ -49,23 +49,59 @@ ur_result_t enqueueEventsWait(ur_queue_handle_t, hipStream_t Stream,
...
+  size_t MaxBlockDim[3];
+
+  // The below assumes a three dimensional range but this is not guaranteed by
+  // UR.
```
ldrumm, line 66: We assume something that's not guaranteed? Is this a FIXME?

ldrumm (changes requested):
This changes runtime behaviour for HIP, and introduces more abstract duplicate code. Thus, I don't really think it's a refactor.

I think the `hipOccupancyMaxPotentialBlockSize` change should go in as a separate fix, and that you should revisit those lambdas

source/adapters/cuda/enqueue.cpp
hdelan: Good point. I have moved this to a common place that can be shared by all adapters

source/adapters/hip/enqueue.cpp
hdelan: Thanks, have moved

source/adapters/hip/enqueue.cpp
hdelan: Thanks comment added

source/adapters/hip/enqueue.cpp
hdelan: Thanks have added to common file.

hdelan:
> This changes runtime behaviour for HIP, and introduces more abstract duplicate code. Thus, I don't really think it's a refactor.

I have changed the PR description to show all changes that have been included in this PR. 
 
> I think the `hipOccupancyMaxPotentialBlockSize` change should go in as a separate fix, and that you should revisit those lambdas

I would argue that it's OK to keep HIP changes in this PR as well since in UR the commits are not squashed when PRs are merged. I have constructed my commits such that there are two clear independent commits. One fixing a bug for the CUDA adapter and adding some common code, and another extending the functionality of the HIP adapter. In an ideal world maybe it would be good to split these into two PRs but with the slow pace of merging in UR I would argue to keep these changes together, especially since this PR is blocking [other PRs](https://github.com/intel/llvm/pull/12715) so need to be prioritized. Let me know your thoughts.

hdelan:
Friendly ping @ldrumm

ldrumm:
Now those lambdas are hoisted, I'm fine with this

ldrumm (approved):

source/adapters/cuda/enqueue.cpp
```diff
@@ -154,33 +153,18 @@ void guessLocalWorkSize(ur_device_handle_t Device, size_t *ThreadsPerBlock,
...
-  // Find a local work group size that is a divisor of the global
-  // work group size to produce uniform work groups.
-  // Additionally, for best compute utilisation, the local size has
-  // to be a power of two.
-  while (0u != (GlobalSizeNormalized[0] % ThreadsPerBlock[0]) ||
-         !IsPowerOf2(ThreadsPerBlock[0])) {
-    --ThreadsPerBlock[0];
-  }
+  int MinGrid, MaxBlockSize;
+  UR_CHECK_ERROR(cuOccupancyMaxPotentialBlockSize(
```
JackAKirk, line 162: What happens here if a kernel submits a very small range e.g. 100 and has no shared memory limitations and e.g. MaxThreadsPerBlock[0] is 1024: Won't this function return a value higher than 100? Will this then be passed to the kernel execution such that it uses a range of e.g. 1024 instead of 100?

hdelan: No the `roundToHighestFactorOfGlobalSizeIn3d` will make sure that no dims exceed the global dims in any dimension

JackAKirk: Do you understand these failures? https://github.com/intel/llvm/actions/runs/8162721434/job/22337470532
This is the only pr that seems to affect those tests from that batch: they use parallel_for with range.

hdelan: There were some bugs in the original implementation but they weren't being tested. This patch fixes the bugs and I have added some tests here https://github.com/intel/llvm/pull/12715/files#diff-0ceed90561c114d8a66c1fdfe72fdb97ac3d3af519bffabc57e62514fe117de2

But it looks like the tests failing are all just timeouts to me. I don't know what could have changed to make this happen other than hardware/CI environment related changes

hdelan: You were right there were bugs in this causing an infinite loop :stuck_out_tongue_winking_eye: Have fixed

test/conformance/enqueue/urEnqueueKernelLaunch.cpp
```diff
@@ -77,56 +77,105 @@ TEST_P(urEnqueueKernelLaunchTest, InvalidWorkDimension) {
                      UR_RESULT_ERROR_INVALID_WORK_DIMENSION);
 }
 
-struct urEnqueueKernelLaunch2DTest : uur::urKernelExecutionTest {
-    void SetUp() override {
-        program_name = "fill_2d";
-        UUR_RETURN_ON_FATAL_FAILURE(urKernelExecutionTest::SetUp());
+#define ENQUEUE_KERNEL_LAUNCH_TEST_1D_SIZES(SIZE)                              \
```
hdelan, line 80: @kbenzie let me know what you think about these added tests. Not ideal to introduce macro funcs but I think this is the most readable way to run these tests for different sizes

kbenzie: GoogleTest supports parameterized tests, it turns out similarly readable and doesn't require macros like this. I'd prefer to use GoogleTest features where possible.

hdelan: Aha I will do this instead

kbenzie: Here's one example of use using it for [urEnqueueUSMFill](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/enqueue/urEnqueueUSMFill.cpp), there are others dotted around too.

hdelan:
> How's the intel/llvm testing looking for this?

Getting some failures on AWS that I can't reproduce locally. Slow going! Will let you know when it's sorted

hdelan:
@kbenzie this is good to go

kbenzie:
> @kbenzie this is good to go

okay cool, thanks

kbenzie (approved):

yingcong-wu:
Hi there, I think there is a problem here, which is the function prototype in `source/adapters/cuda/enqueue.hpp` not updated accordingly. 

See in https://github.com/oneapi-src/unified-runtime/blob/ed1f8bf618c88eaabea6bde0f6c06fc265f3b49f/source/adapters/cuda/enqueue.hpp#L23C52-L23C70

hdelan:
> Hi there, I think there is a problem here, which is the function declaration of `guessLocalWorkSize` in `source/adapters/cuda/enqueue.hpp` is not updated accordingly.
> 
> See in https://github.com/oneapi-src/unified-runtime/blob/ed1f8bf618c88eaabea6bde0f6c06fc265f3b49f/source/adapters/cuda/enqueue.hpp#L23C52-L23C70

Thanks @yingcong-wu see here https://github.com/oneapi-src/unified-runtime/pull/1460

PR #1327, opened 2024-02-08
Add Legacy compatibility to Logger

lplewa:
Switching to the new logger changes API and log format, and requires changes in user code. To make the transition easier, this PR adds an option to the new logger that's print logs in the old format.

### DIFF
--- a/source/common/logger/ur_logger_details.hpp
+++ b/source/common/logger/ur_logger_details.hpp
@@ -11,6 +11,11 @@
 
 namespace logger {
 
+struct LegacyMessage {
+    LegacyMessage(const char *p) : message(p){};
+    const char *message;
+};
+
 class Logger {
   public:
     Logger(std::unique_ptr<logger::Sink> sink) : sink(std::move(sink)) {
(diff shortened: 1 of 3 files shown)


source/common/logger/ur_sinks.hpp
```diff
@@ -50,6 +45,14 @@ class Sink {
         flush_level = logger::Level::ERR;
     }
 
+    virtual void print(logger::Level level, std::string msg) {
```
pbalcer, line 48: I think this might allocate  the string again. `const std::string &msg` ?

source/common/logger/ur_logger_details.hpp
```diff
@@ -70,9 +114,14 @@ class Logger {
...
     }
 
+    void setLegacySink(std::unique_ptr<logger::Sink> legacySink) {
+        this->legacySink = std::move(legacySink);
+    }
+
   private:
     logger::Level level;
     std::unique_ptr<logger::Sink> sink;
+    std::unique_ptr<logger::Sink> legacySink;
```
pbalcer, line 124: why do we need a legacySink that's separate from the sink? Are they going to be used at the same time?

lplewa: Technically we don't need both - but we have to know if an legacy one is set to skip ifs checking log levels. IF you preffer i can replace legacy Sink with a Boolean value.

pbalcer: yea, a bool would be better.

lplewa: Done

source/common/logger/ur_sinks.hpp
lplewa: Done

pbalcer (approved):

source/common/logger/ur_logger_details.hpp
```diff
@@ -59,8 +64,43 @@ class Logger {
...
     void log(logger::Level level, const char *format, Args &&...args) {
+        log(logger::LegacyMessage(format), level, format,
+            std::forward<Args>(args)...);
+    }
+
+    template <typename... Args>
+    void log(const logger::LegacyMessage &p, logger::Level level,
+             const char *format, Args &&...args) {
+        if (isLegacySink) {
+            sink->log(level, p.message, std::forward<Args>(args)...);
```
PatKamin, line 101: Shouldn't you check if the sink is not a nullptr first?

lplewa: Done

PatKamin (approved):

PR #1332, opened 2024-02-12
Explicitly define which info queries are optional.

aarongreig:
This is now reflected in the spec and in the CTS tests.


aarongreig:
this is going to need #1196 and #1128 since they clean up some info related stuff

aarongreig:
I've included https://github.com/oneapi-src/unified-runtime/pull/1332/commits/08f14c15382b4667477ab3c2dbc9c3cc3c742bdc in this because there's a bit of bilateral interdependence going on, if we merged either those changes or the ones from the original commit from this PR separately we'd need to fix some stuff twice, like it would mean writing more code that then just gets immediately reverted. I can split them up again if necessary.

kbenzie (approved):

aarongreig:
ping @oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-cuda-write @oneapi-src/unified-runtime-hip-write

source/adapters/hip/device.cpp
```diff
@@ -891,6 +891,13 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     return ReturnValue(ur_bool_t{false});
   case UR_DEVICE_INFO_USM_POOL_SUPPORT:
     return ReturnValue(ur_bool_t{true});
+  case UR_DEVICE_INFO_BFLOAT16:
```
frasercrmck, line 894: I'm curious, since this sticks out to me: should `BFLOAT16`s not ideally have capabilities, like other floating-point formats?

Also, would bfloat16 support on HIP not depend on some device-specific features?

source/adapters/hip/program.cpp
```diff
@@ -410,7 +410,12 @@ urProgramGetInfo(ur_program_handle_t hProgram, ur_program_info_t propName,
   case UR_PROGRAM_INFO_KERNEL_NAMES:
     return getKernelNames(hProgram);
   case UR_PROGRAM_INFO_IL:
-    return UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION;
+    // Hip only supports urProgramCreateWithBinary, so we can always return
```
frasercrmck, line 413: I can only see two instances of `Hip` in comments in the adapter. `HIP` is far more common so I propose that.

source/adapters/hip/device.cpp
JackAKirk: `UR_DEVICE_INFO_BFLOAT16` I think should have been removed after https://github.com/intel/llvm/pull/13351 but think maybe it was lost in the UR changes when the PI was removed.

bfloat16 is supported on all backends on all devices via software implementations as a minimmum.

aarongreig: I can make an issue to remove it, that isn't a change for this PR.

aarongreig: https://github.com/oneapi-src/unified-runtime/issues/2385

frasercrmck (approved):
CUDA/HIP LGTM

source/adapters/opencl/device.cpp
```diff
@@ -1066,7 +1065,15 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     return ReturnValue(UUID);
   }
 
-  case UR_DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS: {
+  case UR_DEVICE_INFO_COMPONENT_DEVICES:
+  case UR_DEVICE_INFO_COMPOSITE_DEVICE:
+    // These two are exclusive to L0.
```
martygrant, line 1070: just wondering if we have a number of things exclusive to a particular adapter if we should have something in the spec or generation stage to say its not needed for the other adapters, just skip it?

aarongreig: Not really sure why I decided these aren't optional, looking at it in the spec and how it's used by sycl I think they should be. The big `UNSUPPORTED_ENUM` switch fall-through that we have in all the DeviceInfo implementations does look very automate-able although I'm not sure the relatively small maintenance burden justifies spending the time to be that clever about it

martygrant: yeah that makes sense

martygrant (approved):

aarongreig:
@oneapi-src/unified-runtime-native-cpu-write please take a look, the adapter changes are pretty trivial

aarongreig:
LLVM PR: https://github.com/intel/llvm/pull/16381

GeorgeWeb (approved):
Cuda and HIP adapter changes look good to me.
As I am not part of the cuda and hip adapters reviewers' group anymore, after moving teams internally, it may be beneficial for @oneapi-src/unified-runtime-cuda-write and @oneapi-src/unified-runtime-hip-write to double check and confirm that.

martygrant:
Pulled into intel/llvm by https://github.com/intel/llvm/pull/16256

PR #1340, opened 2024-02-14
[HIP][CUDA][Command-Buffer] Fix Coverity issues in HIP/CUDA command-buffer code

EwanC:
Address issues in the CUDA adapter code added by https://github.com/oneapi-src/unified-runtime/pull/1089 which have been flagged by Coverity.

* Uninitialized struct member of `CUDA_KERNEL_NODE_PARAMS` struct
* copying instead of moving a shared pointer to a node when constructing a command-handle.

As well as issues flagged by coverity in the HIP adapter code added in https://github.com/oneapi-src/unified-runtime/pull/1254
* Initialize `NextSyncPoint`
* Move rather than copy node shared-pointers during sync-point registration
* Don't throw exceptions from a class destructor 

DPC++ PR https://github.com/intel/llvm/pull/12723

### DIFF
--- a/source/adapters/cuda/command_buffer.cpp
+++ b/source/adapters/cuda/command_buffer.cpp
@@ -66,7 +66,9 @@ ur_exp_command_buffer_handle_t_::~ur_exp_command_buffer_handle_t_() {
   cuGraphDestroy(CudaGraph);
 
   // Release the memory allocated to the CudaGraphExec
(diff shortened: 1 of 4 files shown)


mfrancepillois (approved):
LGTM

mfrancepillois (approved):

Issue #1191, opened 2023-12-14
Ensure in CTS getInfo tests that Object creation handles is the same as queried


omarahmed1111:
Some of the tests doesn't check that the handles that create the tested object is the same as the queried ones. For example: this [test](https://github.com/oneapi-src/unified-runtime/blob/4424195c7d5d990901f7343fa9e7ac500b03e38d/test/conformance/event/urEventGetInfo.cpp#L32) checks that the queried context associated with the event is the same context that the event was created from. But, here in this [test](https://github.com/oneapi-src/unified-runtime/blob/4424195c7d5d990901f7343fa9e7ac500b03e38d/test/conformance/context/urContextGetInfo.cpp#L45), it didn't check that the queried devices are the same as the devices that creates the context handle. So, we need to ensure consistency in this behaviour in all the getInfo UR handles tests.

PR #1346, opened 2024-02-15
Add output validation checks for getInfo tests and a small USM spec fix

omarahmed1111:
This PR fixes: 
* Some genInfo tests was only checking that the query was returning successfully, however wasn't validating the output. This PR add some validation to some of the queries to validate the output. Some queries like device queries couldn't be validated as they depend on the device properties that could be queried through UR backend, so the CTS won't be able to validate those. The queries that is validated in this PR is only the ones that their info was passed on the creation of the handle so the CTS should already have a copy of the output, like querying the context devices or the allocated buffer size.
* USM allocation UR entries was not mentioning in the spec that allocations could be bigger than the size that was specified, as different backends have different behaviour on this regard. So, add a mention for that in the spec.
* Native CPU backend was not reporting `Reference counting` for context handles, so add that as it should be already have a reference handling mechanism.
* Change hip adapter `urProgramGetInfo` to return the context device when queried about the program devices.

resolves [#1191](https://github.com/oneapi-src/unified-runtime/issues/1191)

[intel/llvm](https://github.com/intel/llvm/pull/12782) testing for the small adapters change

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -3449,7 +3449,7 @@ urUSMHostAlloc(
     ur_context_handle_t hContext,  ///< [in] handle of the context object
     const ur_usm_desc_t *pUSMDesc, ///< [in][optional] USM memory allocation descriptor
     ur_usm_pool_handle_t pool,     ///< [in][optional] Pointer to a pool created using urUSMPoolCreate
-    size_t size,                   ///< [in] size in bytes of the USM memory object to be allocated
+    size_t size,                   ///< [in] minimum size in bytes of the USM memory object to be allocated
     void **ppMem                   ///< [out] pointer to USM host memory object
 );
 
@@ -3497,7 +3497,7 @@ urUSMDeviceAlloc(
     ur_device_handle_t hDevice,    ///< [in] handle of the device object
     const ur_usm_desc_t *pUSMDesc, ///< [in][optional] USM memory allocation descriptor
     ur_usm_pool_handle_t pool,     ///< [in][optional] Pointer to a pool created using urUSMPoolCreate
-    size_t size,                   ///< [in] size in bytes of the USM memory object to be allocated
+    size_t size,                   ///< [in] minimum size in bytes of the USM memory object to be allocated
     void **ppMem                   ///< [out] pointer to USM device memory object
 );
 
@@ -3546,7 +3546,7 @@ urUSMSharedAlloc(
     ur_device_handle_t hDevice,    ///< [in] handle of the device object
     const ur_usm_desc_t *pUSMDesc, ///< [in][optional] Pointer to USM memory allocation descriptor.
     ur_usm_pool_handle_t pool,     ///< [in][optional] Pointer to a pool created using urUSMPoolCreate
-    size_t size,                   ///< [in] size in bytes of the USM memory object to be allocated
+    size_t size,                   ///< [in] minimum size in bytes of the USM memory object to be allocated
     void **ppMem                   ///< [out] pointer to USM shared memory object
 );
 
@@ -4117,6 +4117,8 @@ urProgramCreateWithIL(
 ///     - Following a successful call to this entry point, `phProgram` will
 ///       contain a binary of type ::UR_PROGRAM_BINARY_TYPE_COMPILED_OBJECT or
 ///       ::UR_PROGRAM_BINARY_TYPE_LIBRARY for `hDevice`.
+///     - The device specified by `hDevice` must be device associated with
+///       context.
 ///
 /// @remarks
 ///   _Analogues_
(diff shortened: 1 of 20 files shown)


fabiomestre (approved):

test/conformance/context/urContextGetInfo.cpp
```diff
@@ -56,6 +57,34 @@ TEST_P(urContextGetInfoTestWithInfoParam, Success) {
...
+            reinterpret_cast<uint32_t *>(info_data.data());
+        ASSERT_EQ(*returned_num_of_devices,
+                  uur::DevicesEnvironment::instance->devices.size());
+        break;
+    }
+    case UR_CONTEXT_INFO_DEVICES: {
+        auto returned_devices =
+            reinterpret_cast<ur_device_handle_t *>(info_data.data());
+        for (uint32_t i = 0;
+             i < uur::DevicesEnvironment::instance->devices.size(); i++) {
```
aarongreig, line 73: I'd be surprised if this broke any time soon but I don't think we make any guarantees about the order of devices returned for this query, and this test is now enforcing that. We already verify that the number of devices is what we expect in another test, so I think it would be fine to just verify that we can find each device from `uur::DevicesEnvironment::instance->devices` in the output.

test/conformance/memory/urMemGetInfo.cpp
```diff
@@ -31,6 +31,22 @@ TEST_P(urMemGetInfoTest, Success) {
...
+    switch (info) {
+    case UR_MEM_INFO_CONTEXT: {
+        auto returned_context =
+            reinterpret_cast<ur_context_handle_t *>(info_data.data());
+        ASSERT_EQ(context, *returned_context);
+        break;
+    }
+    case UR_MEM_INFO_SIZE: {
+        auto returned_size = reinterpret_cast<size_t *>(info_data.data());
+        ASSERT_EQ(allocation_size, *returned_size);
```
aarongreig, line 44: I think the thing with the USM sizes applies to buffers as well, even if it isn't breaking on the hardware we test I don't think there's any guarantees about it.

test/conformance/program/urProgramGetInfo.cpp
```diff
@@ -32,6 +32,46 @@ TEST_P(urProgramGetInfoTest, Success) {
...
+    }
+    case UR_PROGRAM_INFO_NUM_DEVICES: {
+        auto returned_num_of_devices =
+            reinterpret_cast<uint32_t *>(property_value.data());
+        ASSERT_EQ(*returned_num_of_devices, 1U);
+        break;
+    }
+    case UR_PROGRAM_INFO_DEVICES: {
+        auto returned_devices =
+            reinterpret_cast<ur_device_handle_t *>(property_value.data());
```
aarongreig, line 56: same comment as context devices

test/conformance/context/urContextGetInfo.cpp
omarahmed1111: Thanks for raising this point, you are right. I also changed that to address the fact that we could create a context with not all the available devices.

test/conformance/program/urProgramGetInfo.cpp
omarahmed1111: fixed

test/conformance/memory/urMemGetInfo.cpp
omarahmed1111: fixed that.

include/ur_api.h
```diff
@@ -3396,7 +3396,7 @@ urUSMHostAlloc(
     ur_context_handle_t hContext,  ///< [in] handle of the context object
     const ur_usm_desc_t *pUSMDesc, ///< [in][optional] USM memory allocation descriptor
     ur_usm_pool_handle_t pool,     ///< [in][optional] Pointer to a pool created using urUSMPoolCreate
-    size_t size,                   ///< [in] size in bytes of the USM memory object to be allocated
+    size_t size,                   ///< [in] minumum size in bytes of the USM memory object to be allocated
```
jchlanda, line 3399: s/minumum/minimu/g

Here and in a bunch of places below.

jchlanda (approved):
LGTM

GeorgeWeb (changes requested):
Ideally would prefer two separate commits if it is possible to split the UR spec fix from the output validation checks changes. If that's too much, at least just adding a summary to the current commit body about this spec fix will still be good.

Otherwise, the changes look good to me.

include/ur_api.h
omarahmed1111: Fixed, Thanks for noting that!

omarahmed1111:
> Ideally would prefer two separate commits if it is possible to split the UR spec fix from the output validation checks changes. If that's too much, at least just adding a summary to the current commit's message body about this spec fix will still be good.
> 
> Otherwise, the changes look good to me.

I was just thinking that it is an associated change to the checks added so it is included with the checks implicitly. But you are right, it might be confused and more clarification is better. So, Changed that.

omarahmed1111:
@oneapi-src/unified-runtime-native-cpu-write  could I have a review on this when possible, Thanks!

PietroGhg (approved):
Change looks good for Native CPU, thank you

PR #1349, opened 2024-02-15
[CI] Fix failing nightly build

PatKamin:
Changes from commit 019551c in `urDeviceGet()` validation resulted in a failure to create devices successfully in fuzz tests. This resulted in generating plenty of calls to `urPlatformGet()`. The size of the input generated by fuzzer most probably led to going out of memory (available to the fuzzer).

### DIFF
--- a/test/fuzz/urFuzz.cpp
+++ b/test/fuzz/urFuzz.cpp
@@ -38,14 +38,18 @@ int ur_device_get(TestState &state) {
         return -1;
     }
 
-    ur_result_t res = urDeviceGet(state.platforms[state.platform_num],
-                                  state.device_type, state.num_entries,
-                                  state.devices.data(), &state.num_devices);
+    ur_result_t res = UR_RESULT_SUCCESS;
+    if (state.devices.size() == 0) {
+        res = urDeviceGet(state.platforms[state.platform_num],
+                          state.device_type, 0, nullptr, &state.num_devices);
+        state.devices.resize(state.num_devices);
+    } else {
+        res =
+            urDeviceGet(state.platforms[state.platform_num], state.device_type,
+                        state.num_entries, state.devices.data(), nullptr);
+    }
     if (res != UR_RESULT_SUCCESS) {
         return -1;
-    }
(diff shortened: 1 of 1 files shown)


test/fuzz/urFuzz.cpp
```diff
@@ -38,15 +38,19 @@ int ur_device_get(TestState &state) {
...
     }
 
-    ur_result_t res = urDeviceGet(state.platforms[state.platform_num],
-                                  state.device_type, state.num_entries,
-                                  state.devices.data(), &state.num_devices);
+    ur_result_t res = UR_RESULT_SUCCESS;
+    if (state.devices.size() == 0) {
+        res = urDeviceGet(state.platforms[state.platform_num],
+                          state.device_type, 0, nullptr, &state.num_devices);
+        state.devices.resize(state.num_devices);
```
pbalcer, line 45: but this will just create invalid `num_devices` entries in t he devices vector. how does this fix the problem?

PatKamin: This leads to two step `urDeviceGet()` usage. If the devices vector is empty, get the number of devices. If the devices vector was already resized, fill it with valid devices handles. This "unblocks" the test so it generates API calls beyond `urDeviceGet()`. So with this we have a working nightly build, fuzz tests do not fail.

However, it doesn't resolve the root cause of the OOM. The case where it is occurring now just won't be generated by the fuzzer. I've found that the commit with which the fuzz test had started to OOM is 2a82754. Running fuzz tests on this commit with tracing **disabled** also fixes the OOM issue. I'm debugging it further to find the exact root cause.

There is also one more thing (perhaps for another PR) that I see locally when running the "unblocked" fuzz tests - the kernel doesn't run. It looks like the missing `UR_CONFORMANCE_TARGET_TRIPLES` env var in the nightly build is the cause for this.

pbalcer: makes sense, thanks.

pbalcer (approved):

PR #1353, opened 2024-02-16
[L0] Support updating kernel commands in command buffers

againull:
Initial support for updating command buffers.

Also applied temporary fixes for conformance tests:
* Execution range update is not supported by the L0 driver right now, so skipping those tests for L0 backend.

* There is a synchronization issue with immediate submission when used for command buffers. It is reproducible even without changes of this PR, so should be fixed separately. For now use batched submission for command buffer update tests.

intel/llvm PR: https://github.com/intel/llvm/pull/12897

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -23,7 +23,8 @@ ur_exp_command_buffer_handle_t_::ur_exp_command_buffer_handle_t_(
     : Context(Context), Device(Device), ZeCommandList(CommandList),
       ZeCommandListResetEvents(CommandListResetEvents),
       ZeCommandListDesc(ZeDesc), ZeFencesList(), QueueProperties(),
-      SyncPoints(), NextSyncPoint(0) {
+      SyncPoints(), NextSyncPoint(0),
+      IsUpdatable(Desc ? Desc->isUpdatable : false) {
   (void)Desc;
   urContextRetain(Context);
   urDeviceRetain(Device);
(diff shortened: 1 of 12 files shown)


test/conformance/exp_command_buffer/fixtures.h
```diff
@@ -112,24 +112,52 @@ struct urUpdatableCommandBufferExpExecutionTest
             GTEST_SKIP() << "Updating EXP command-buffers is not supported.";
         }
 
+        // Currently level zero driver doesn't support updating execution range.
+        // Also disable immediate command lists because there is synchronization issue causing test failures.
+        if (backend == UR_PLATFORM_BACKEND_LEVEL_ZERO) {
+            setenv("UR_L0_USE_IMMEDIATE_COMMANDLISTS", "0", 0);
```
EwanC, line 118: Do we need to set this environment variable if we're setting `UR_QUEUE_FLAG_SUBMISSION_BATCHED` later on when creating the queue?

Using env vars is a bit messy if users that setting their own too, so would like to avoid it in preference of setting properties on the queues if possible.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -982,20 +1050,275 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
...
+  }
+
+  // Check if new global size is provided and we need to update group count.
+  size_t *NewGlobalWorkSize = CommandDesc->pNewGlobalWorkSize;
+  if (NewGlobalWorkSize && Dim > 0) {
+    ze_group_count_t ZeThreadGroupDimensions{1, 1, 1};
+    uint32_t WG[3];
+    // If new global work size is provided but new local work size is not
+    // provided then we still need to update local work size based on size
+    // suggested by the driver for the kernel.
```
EwanC, line 1141: In my mind the expected semantics is that if the global work size changes, and local work size is not provided then the local work size stays the same and it's an error if the global/local sizes are not compatible. This is based on similar OpenCL semantics in [cl_mutable_dispatch_config_khr](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_mutable_command_types)

> local_work_size Points to an array of work_dim unsigned values that describe the number of work-items that make up a work-group that will execute the kernel. If local_work_size is NULL then the number of local work-items in the dispatch is not changed. See clEnqueueNDRangeKernel for valid usage.

I'm not sure if that's the best semantics though, in our [SYCL-Graph Update API](https://github.com/intel/llvm/pull/12486) we can update the range based on a `sycl::range` or a `sycl::nd_range`. Using a `sycl::nd_range` would mean that both global & local work sizes are set, while using a `sycl::range` implies the user doesn't care about the local size.

test/conformance/exp_command_buffer/fixtures.h
againull: Oh, sorry, I forgot to remove that, will do. You are right, it is not needed anymore.

source/adapters/level_zero/command_buffer.cpp
EwanC: To recap my own perspective on this from the call. A SYCL user can't go from a `sycl::range` to an `sycl::nd_range` during update an vice versa. So they will either care about the local size or not. The current semantics of picking the best local size possible when a user hasn't set the local size I think is desirable, because it corresponds to the usecase where a user indicates they don't care about local size so would want the implementation to pick for them.

I'll take the action to refine the UR API to remove the ambiguity about what happens if a user creates a kernel with a local size, and then only updates the global size. But for this purpose of this PR I think current the implementation is fine.

source/adapters/level_zero/device.cpp
```diff
@@ -918,7 +918,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
   case UR_DEVICE_INFO_COMMAND_BUFFER_SUPPORT_EXP:
     return ReturnValue(true);
   case UR_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP:
-    return ReturnValue(false);
+    return ReturnValue(Device->Platform->ZeMutableCmdListExt.Supported);
```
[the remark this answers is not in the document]
againull: Indeed, thank you for pointing that out. Added check for that.

test/conformance/exp_command_buffer/fixtures.h
againull: Removed.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -93,7 +93,8 @@ ur_exp_command_buffer_handle_t_::ur_exp_command_buffer_handle_t_(
     const ur_exp_command_buffer_desc_t *Desc)
     : Context(Context), Device(Device), ZeCommandList(CommandList),
       ZeCommandListDesc(ZeDesc), ZeFencesList(), QueueProperties(),
-      SyncPoints(), NextSyncPoint(0) {
+      SyncPoints(), NextSyncPoint(0),
+      IsUpdatable(Desc ? Desc->isUpdatable : false) {
   (void)Desc;
```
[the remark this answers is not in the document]
againull: Indeed, removed.

source/adapters/level_zero/command_buffer.cpp
againull: Thank you for clarifying!

EwanC (approved):
LGTM - Ideally we could test this with the DPC++ PR in https://github.com/intel/llvm/pull/12840 on a setup with this driver feature and see if the new tests pass (apart from range update) when enabled for L0. But with the CTS tests I wouldn't say this is required and we could fix up any issues in a follow-up PR.

nrspruit (approved):
LGTM for Level Zero

kbenzie:
Please pull in the main branch to have up to date testing, also update the tag in the intel/llvm PR.

againull:
> Please pull in the main branch to have up to date testing, also update the tag in the intel/llvm PR.

Done.

againull:
Hi @kbenzie @nrspruit, Is it possible to merge this PR?

kbenzie:
> Hi @kbenzie @nrspruit, Is it possible to merge this PR?

@againull yes, I'll be getting to this soon. It's on our [v0.9.x](https://github.com/oneapi-src/unified-runtime/pulls?q=is%3Aopen+is%3Apr+label%3Av0.9.x) merge list.

PR #1355, opened 2024-02-18
[DeviceSanitizerLayer] Demangle the kernel name in error reporting

yingcong-wu:
Try to demangle the kernel name in error reporting to improve user experience.

### DIFF
--- a/source/loader/layers/sanitizer/asan_interceptor.cpp
+++ b/source/loader/layers/sanitizer/asan_interceptor.cpp
@@ -13,7 +13,6 @@
 
 #include "asan_interceptor.hpp"
 #include "ur_sanitizer_layer.hpp"
-#include <cxxabi.h>
 
 namespace ur_sanitizer_layer {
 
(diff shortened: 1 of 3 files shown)


source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -13,6 +13,7 @@
 
 #include "asan_interceptor.hpp"
 #include "ur_sanitizer_layer.hpp"
+#include <cxxabi.h>
```
AllanZyne, line 16: IMO, it'd better to extract these codes as a function, which is defined in `unified-runtime/source/loader/layers/sanitizer/linux/san_utils.cpp`, as it's specific to Linux

AllanZyne (commented):
IMO, it'd better to extract these codes as a function, which is defined in `unified-runtime/source/loader/layers/sanitizer/linux/san_utils.cpp`, as it's specific to Linux

yingcong-wu:
> IMO, it'd better to extract these codes as a function, which is defined in `unified-runtime/source/loader/layers/sanitizer/linux/san_utils.cpp`, as it's specific to Linux

Updated.

source/loader/layers/sanitizer/linux/san_utils.cpp
```diff
@@ -84,4 +86,15 @@ void *GetMemFunctionPointer(const char *FuncName) {
...
+std::string DemangleName(const std::string &name) {
+    std::string result = name;
+    char *demangled =
+        abi::__cxa_demangle(name.c_str(), nullptr, nullptr, nullptr);
+    if (demangled) {
+        result = demangled;
+        free(demangled);
+    }
+    return result;
+}
```
AllanZyne, line 98: If "DemangleName" failed, the "KernelName" will be an empty string. 
How about "void DemangleName(std::string& name)"? We aren't interested in whether demangling succeeds or fails anyway.

yingcong-wu: `std::string result = name;` makes sure if demangling failed, then the name is unchanged.

AllanZyne: Okay. Does "abi::__cxa_demangle" need to link a library?

AllanZyne (approved):

AllanZyne (approved):
LGTM

PR #1356, opened 2024-02-19
[OCL] Fix use of CL_RETURN_ON_FAILURE instead of UR_RETURN_ON_FAILURE

Naghasan:
`usmSetIndirectAccess` returns a UR error code and not an opencl one

### DIFF
--- a/source/adapters/opencl/kernel.cpp
+++ b/source/adapters/opencl/kernel.cpp
@@ -320,7 +320,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urKernelSetExecInfo(
   switch (propName) {
   case UR_KERNEL_EXEC_INFO_USM_INDIRECT_ACCESS: {
     if (*(static_cast<const ur_bool_t *>(pPropValue)) == true) {
-      CL_RETURN_ON_FAILURE(usmSetIndirectAccess(hKernel));
+      UR_RETURN_ON_FAILURE(usmSetIndirectAccess(hKernel));
     }
     return UR_RESULT_SUCCESS;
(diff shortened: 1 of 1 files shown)


aarongreig (approved):

Naghasan:
The CI issue is a cmake unable to fetch `xpti` properly, have you seen that on other PRs ?

pbalcer:
> The CI issue is a cmake unable to fetch `xpti` properly, have you seen that on other PRs ?

Yes, this just happens sometimes, not sure why. I've restarted the failed jobs.

kbenzie (approved):

Naghasan:
All tests pass, can it be merged ? (it is not needed by another patch in intel/llvm, so no need to sync)

kbenzie:
They are not direct dependencies but this is related to https://github.com/intel/llvm/pull/12780, right?

Since this does change an adapter we should be testing the change in intel/llvm as per the [adapter change process](https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#adapter-change-process). Easiest way would be to update https://github.com/intel/llvm/pull/12780 to point at your fork and the HEAD of this branch.

Naghasan:
> They are not direct dependencies but this is related to https://github.com/intel/llvm/pull/12780, right?

The only link is that I discovered this issue by investigating 12780, but there is no relation between the 2. I don't see any issue to test the change via this intel/llvm PR as long as it doesn't block it.

Naghasan:
@kbenzie the intel/llvm PR passes (has only expected failures)

PR #1357, opened 2024-02-19
[HIP] Check for null arg to urKernelSetArgMemObj

EwanC:
The `pProperties` parameter to [urKernelSetArgMemObj](https://oneapi-src.github.io/unified-runtime/core/api.html#urkernelsetargmemobj) is optional, and so may be `nullptr`. However the HIP adapter assumes it is set currently. Updated this to check for nullptr, and if so assume that the buffer is read/write.

Updated CTS match file so that the test executable doesn't segfault early ([CTS passes nullptr properties](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/testing/include/uur/fixtures.h#L1181)) and more tests run but fail. Discovered during testing of https://github.com/oneapi-src/unified-runtime/pull/1254 and is required for those CTS tests to pass.

DPC++ PR - https://github.com/intel/llvm/pull/12758


hdelan (approved):
LGTM

PR #1360, opened 2024-02-19
[CTS] Refactor fixtures to remove duplicated code

fabiomestre:
The `fixtures.h` header of the CTS has a lot of duplicated code because we are using separate structs for tests with parameters and tests without parameters. This PR refactors the CTS to use variadic templates which allows using the same `struct` for both cases.

List of changes to help with reviewing:
- Change all `WithParam` structs in `testing/include/uur/fixtures.h`  to be variadic templates and delete the non-templated structs
- Move the event fixtures from `events/fixtures.h` to `testing/include/uur/fixtures.h`
- Remove `WithParam` from the name of all tests that had it (and update the `.match` files)
-  Move some helper functions from `testing/include/uur/fixtures.h` to `testing/include/uur/utils.h`


test/conformance/testing/include/uur/fixtures.h
```diff
@@ -75,168 +120,39 @@ struct urAllDevicesTest : urPlatformTest {
...
-        UUR_RETURN_ON_FATAL_FAILURE(urContextTest::SetUp());
-    }
-
-    void TearDown() override {
-        if (image) {
-            EXPECT_SUCCESS(urMemRelease(image));
-        }
-        UUR_RETURN_ON_FATAL_FAILURE(urContextTest::TearDown());
+    template <size_t Index = 1> auto &getParam() const {
+        return std::get<Index>(this->GetParam());
```
fabiomestre, line 133: There is an old `TODO` here about `getParam` and `GetParam` being very similar. I spent some time thinking about it and still don't know what else to name it. Open to suggestions :)

test/adapters/cuda/context_tests.cpp
```diff
@@ -9,7 +9,7 @@
 #include "uur/raii.h"
 #include <thread>
 
-using cudaUrContextCreateTest = uur::urDeviceTest;
+using cudaUrContextCreateTest = uur::urDeviceTest<>;
```
kbenzie, line 12: I don't love having to add `<>` everywhere.

fabiomestre: I don't love it either but thought it was the lesser evil. 

I don't like the "WithParam" naming scheme for structs / tests (they all have parameters even if it's just the device). This was the only way to make the naming more straightforward and avoid adding extra abstractions. Maybe I got a bit carried away with the changes :man_shrugging: 

If it makes it more palatable, I can revert the changes to test / struct names and add a `using urDeviceTest = urDeviceTestWithParam<>` inside `fixtures.h`?

kbenzie: `WithParam` comes from gtest itself. The device paramter was intentially hidden to make writing tests more straight forwards. I've not dug into this enough to form a proper opinion yet, I'll come back to it.

fabiomestre:
Closing since the team prefers to approach this issue in a different way.

PR #1363, opened 2024-02-19
[CUDA] Refactor device initialization

hdelan:
Some logic for device initialization was split across platform init and device init. This duplicated some calls to get info for max block dim in Y and Z dimension. This rectifies that and makes sure the max block dims are only queried once, which is called from the device constructor.

### DIFF
--- a/source/adapters/cuda/device.hpp
+++ b/source/adapters/cuda/device.hpp
@@ -41,15 +41,19 @@ public:
         Platform(platform) {
 
     UR_CHECK_ERROR(cuDeviceGetAttribute(
-        &MaxBlockDimY, CU_DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_Y, cuDevice));
-    UR_CHECK_ERROR(cuDeviceGetAttribute(
-        &MaxBlockDimZ, CU_DEVICE_ATTRIBUTE_MAX_BLOCK_DIM_Z, cuDevice));
-    UR_CHECK_ERROR(cuDeviceGetAttribute(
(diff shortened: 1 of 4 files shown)


hdelan:
Friendly ping @konradkusiak97

konradkusiak97:
Should there be an intel/llvm CI for this?

hdelan:
Oops forgot to link it. Here it is https://github.com/intel/llvm/pull/12762

konradkusiak97 (approved):
Aha that's strange how it was done before, it could've been partially my fault for not seeing this duplication with my caching PRs. Nice refactor, LGTM!

Just a side note, it seems some of those changes could also be applied to the HIP adapter, do you think it would be feasible to squeeze it together here?

hdelan:
> Just a side note, it seems some of those changes could also be applied to the HIP adapter, do you think it would be feasible to squeeze it together here?

The HIP adapter is only checking this stuff once. It does it in a different manner, ie it doesnt use the ur helper, but I think there isn't anything wrong with the current approach of the HIP adapter so I would be inclined to leave it as it is. Let me know if you agree!

konradkusiak97:
> > Just a side note, it seems some of those changes could also be applied to the HIP adapter, do you think it would be feasible to squeeze it together here?
> 
> The HIP adapter is only checking this stuff once. It does it in a different manner, ie it doesnt use the ur helper, but I think there isn't anything wrong with the current approach of the HIP adapter so I would be inclined to leave it as it is. Let me know if you agree!

Sounds good!

PR #1369, opened 2024-02-21
[EXP][usm p2p] Add test for ur_exp_usm_p2p and impl for hip

JackAKirk:
- Adds USM_P2P impl for hip backend.
- Introduces UR_USM_P2P_EXTENSION_STRING_EXP macro to the spec which is used to check for usm_p2p experimental extension support in UR.
- The existing supported backend for this extension (CUDA) now correctly returns this device value string.
- This device value string is used in the new test that is introduced to check the basic extension behavior for all implementations.

This PR matches the precedent set in https://github.com/oneapi-src/unified-runtime/pull/1089 for experimental extension conformance testing.


source/adapters/cuda/device.cpp
```diff
@@ -612,6 +612,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     SupportedExtensions += "pi_ext_intel_devicelib_assert ";
     // Return supported for the UR command-buffer experimental feature
     SupportedExtensions += "ur_exp_command_buffer ";
+    SupportedExtensions += "ur_exp_usm_p2p ";
```
aarongreig, line 615: could do something like
```
SupportedExtensions += std::string(UR_USM_P2P_EXTENSION_STRING_EXP) + " ";
```
just to re-use the macro definition, although that does come at the cost of the extra std::string constructor

aarongreig (approved):

scripts/core/EXP-USM-P2P.rst
```diff
@@ -48,14 +52,23 @@ Functions
...
-| Revision  | Changes                |
-+===========+========================+
-| 1.0       | Initial Draft          |
-+-----------+------------------------+
++-----------+---------------------------------------------+
+| Revision  | Changes                                     |
++===========+=============================================+
+| 1.0       | Initial Draft                               |
++-----------+---------------------------------------------+
+| 1.1       | Added USM_P2P_EXTENSION_STRING_EXP ID Macro |
```
mmoadeli, line 70: not sure if that should revision `2.0`?

scripts/core/EXP-USM-P2P.rst
```diff
@@ -27,16 +27,20 @@ or copying the memory located on a separate "peer" device.
...
-Two such examples that SYCL supports are Pytorch and Gromacs.
-This experimental extension to UR aims to provide a portable interface that can
-call appropriate driver functions to query and control peer memory access
-across the CUDA, HIP and L0 adapters.
+Several important projects that programming models such as SYCL or OpenMP aim
+to support use fine-grained peer to peer memory access controls.
+This experimental extension to the Unified-Runtime API aims to provide a
+portable interface that can call appropriate driver functions to query and
+control peer memory access within different adapters such as CUDA, HIP and
+Level Zero.
```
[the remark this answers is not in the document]
JackAKirk: I'll switch the first sentence, I think this is an improvement. I think the second one is not as good as the original though.

scripts/core/EXP-USM-P2P.rst
JackAKirk: I'm following other extensions that iterate through 1.1,1.2 etc. This is a minor change to the spec.

source/adapters/cuda/device.cpp
JackAKirk: It does seem to make more sense to use the macro directly, otherwise I don't see much point in having the macro (but I guess there are other reasons). I don't mind although if I did that it would make sense to also change the "ur_exp_command_buffer " case also to use the corresponding macro.
It doesn't really seem important in any case.

mmoadeli (approved):

source/adapters/cuda/device.cpp
JackAKirk: Let me know if you want me to make this change though. I just want confirmation that it is OK to change the command buffer one too before making such a change.

JackAKirk: OK I'll change them both in this PR.

JackAKirk: OK the problem is that I should make the switch the macro usage in all the plugins e.g. https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/device.cpp#L244 is not using the macros, I'm not even sure they have the macros for all strings.
I could make a separate PR to do this prior to this one, however I want to get this PR merged ASAP because it is a prerequisite for https://github.com/oneapi-src/unified-runtime/pull/1234 which we want in the next release.
I suggest:
1. merging this PR as is
2. fixing all the macro strings in this device query for all plugins in a later PR.

aarongreig: alright, I'm ok with this course of action

JackAKirk:
> Please merge in the main branch so we have up to date test runs, it should also have an intel/llvm PR for testing.

I didn't add a intel/llvm PR for testing here because it doesn't affect any intel/llvm code at all. The only code anywhere that this uses is in the test that is added.
I can merge ur if you want but FYI the situation here is basically the same: no existing tests are affected by the code change.

PR #1370, opened 2024-02-22
[L0] Support for counter-based events using L0 driver

winstonzhang-intel:
Counter-based events implementation.
Counter-based events can be enabled via the flag UR_L0_USE_DRIVER_COUNTER_BASED_EVENTS=1

Below are the three conformance tests that ensures that this implementation is working as expected:
```
PASS: SYCL :: InorderQueue/in_order_buffs.cpp (1 of 1)

Testing Time: 6.09s

Total Discovered Tests: 1
  Passed: 1 (100.00%)

PASS: SYCL :: InorderQueue/in_order_kernels.cpp (1 of 1)

Testing Time: 6.29s

Total Discovered Tests: 1
  Passed: 1 (100.00%)


PASS: SYCL :: Basic/in_order_queue_status.cpp (1 of 1)

Testing Time: 7.55s

Total Discovered Tests: 1
  Passed: 1 (100.00%)
```
LLVM Draft with CI passing: https://github.com/intel/llvm/pull/12848
Rebased against Raiyan's in-order list [implementation](https://github.com/oneapi-src/unified-runtime/pull/1372/)

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -933,6 +933,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
       MustSignalWaitEvent = false;
     }
   }
+  // Given WaitEvent was created without specifying Counting Events, then this
+  // event can be signalled on the host.
   if (MustSignalWaitEvent) {
     ZE2UR_CALL(zeEventHostSignal, (CommandBuffer->WaitEvent->ZeEvent));
   }
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -471,7 +471,8 @@ static const uint32_t MaxNumEventsPerPool = [] {
 
 ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
     ze_event_pool_handle_t &Pool, size_t &Index, bool HostVisible,
-    bool ProfilingEnabled, ur_device_handle_t Device) {
+    bool ProfilingEnabled, ur_device_handle_t Device,
+    bool CounterBasedEventEnabled, bool UsingImmCmdList) {
   // Lock while updating event pool machinery.
   std::scoped_lock<ur_mutex> Lock(ZeEventPoolCacheMutex);
 
@@ -481,7 +482,8 @@ ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
     ZeDevice = Device->ZeDevice;
   }
   std::list<ze_event_pool_handle_t> *ZePoolCache =
-      getZeEventPoolCache(HostVisible, ProfilingEnabled, ZeDevice);
+      getZeEventPoolCache(HostVisible, ProfilingEnabled,
+                          CounterBasedEventEnabled, UsingImmCmdList, ZeDevice);
 
   if (!ZePoolCache->empty()) {
     if (NumEventsAvailableInEventPool[ZePoolCache->front()] == 0) {
@@ -506,15 +508,27 @@ ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
   Index = 0;
   // Create one event ZePool per MaxNumEventsPerPool events
   if (*ZePool == nullptr) {
+    ze_event_pool_counter_based_exp_desc_t counterBasedExt = {
+        ZE_STRUCTURE_TYPE_COUNTER_BASED_EVENT_POOL_EXP_DESC};
     ZeStruct<ze_event_pool_desc_t> ZeEventPoolDesc;
     ZeEventPoolDesc.count = MaxNumEventsPerPool;
     ZeEventPoolDesc.flags = 0;
+    ZeEventPoolDesc.pNext = nullptr;
     if (HostVisible)
       ZeEventPoolDesc.flags |= ZE_EVENT_POOL_FLAG_HOST_VISIBLE;
     if (ProfilingEnabled)
       ZeEventPoolDesc.flags |= ZE_EVENT_POOL_FLAG_KERNEL_TIMESTAMP;
     logger::debug("ze_event_pool_desc_t flags set to: {}",
                   ZeEventPoolDesc.flags);
+    if (CounterBasedEventEnabled) {
+      if (UsingImmCmdList) {
+        counterBasedExt.flags = ZE_EVENT_POOL_COUNTER_BASED_EXP_FLAG_IMMEDIATE;
+      } else {
+        counterBasedExt.flags =
+            ZE_EVENT_POOL_COUNTER_BASED_EXP_FLAG_NON_IMMEDIATE;
+      }
+      ZeEventPoolDesc.pNext = &counterBasedExt;
+    }
 
     std::vector<ze_device_handle_t> ZeDevices;
     if (ZeDevice) {
@@ -540,7 +554,8 @@ ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
 }
 
 ur_event_handle_t ur_context_handle_t_::getEventFromContextCache(
-    bool HostVisible, bool WithProfiling, ur_device_handle_t Device) {
+    bool HostVisible, bool WithProfiling, ur_device_handle_t Device,
+    bool CounterBasedEventEnabled) {
   std::scoped_lock<ur_mutex> Lock(EventCacheMutex);
   auto Cache = getEventCache(HostVisible, WithProfiling, Device);
   if (Cache->empty())
@@ -548,6 +563,9 @@ ur_event_handle_t ur_context_handle_t_::getEventFromContextCache(
 
   auto It = Cache->begin();
   ur_event_handle_t Event = *It;
+  if (Event->CounterBasedEventsEnabled != CounterBasedEventEnabled) {
+    return nullptr;
+  }
   Cache->erase(It);
   // We have to reset event before using it.
   Event->reset();
@@ -579,13 +597,16 @@ ur_context_handle_t_::decrementUnreleasedEventsInPool(ur_event_handle_t Event) {
   }
 
   ze_device_handle_t ZeDevice = nullptr;
+  bool UsingImmediateCommandlists =
+      !Event->UrQueue || Event->UrQueue->UsingImmCmdLists;
 
   if (!Event->IsMultiDevice && Event->UrQueue) {
     ZeDevice = Event->UrQueue->Device->ZeDevice;
   }
 
   std::list<ze_event_pool_handle_t> *ZePoolCache = getZeEventPoolCache(
-      Event->isHostVisible(), Event->isProfilingEnabled(), ZeDevice);
+      Event->isHostVisible(), Event->isProfilingEnabled(),
+      Event->CounterBasedEventsEnabled, UsingImmediateCommandlists, ZeDevice);
 
   // Put the empty pool to the cache of the pools.
   if (NumEventsUnreleasedInEventPool[Event->ZeEventPool] == 0)
@@ -676,6 +697,12 @@ ur_result_t ur_context_handle_t_::getAvailableCommandList(
   // command list is available for reuse.
   ur_result_t ur_result = UR_RESULT_ERROR_OUT_OF_RESOURCES;
 
+  // As a limitation of regular command list, when counter based events are used
+  // the command list in the cache still has the potential of having events that
+  // are still associated with another command list. We disable cache for
+  // regular command list when counter based events are enabled to avoid race
+  // condition.
+  if (!Queue->CounterBasedEventsEnabled)
   // Initally, we need to check if a command list has already been created
   // on this device that is available for use. If so, then reuse that
   // Level-Zero Command List and Fence for this PI call.
@@ -683,8 +710,8 @@ ur_result_t ur_context_handle_t_::getAvailableCommandList(
     // Make sure to acquire the lock before checking the size, or there
     // will be a race condition.
     std::scoped_lock<ur_mutex> Lock(Queue->Context->ZeCommandListCacheMutex);
-    // Under mutex since operator[] does insertion on the first usage for every
-    // unique ZeDevice.
+    // Under mutex since operator[] does insertion on the first usage for
+    // every unique ZeDevice.
     auto &ZeCommandListCache =
         UseCopyEngine
             ? Queue->Context->ZeCopyCommandListCache[Queue->Device->ZeDevice]
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -697,12 +697,6 @@ ur_result_t ur_context_handle_t_::getAvailableCommandList(
   // command list is available for reuse.
   ur_result_t ur_result = UR_RESULT_ERROR_OUT_OF_RESOURCES;
 
-  // As a limitation of regular command list, when counter based events are used
-  // the command list in the cache still has the potential of having events that
-  // are still associated with another command list. We disable cache for
-  // regular command list when counter based events are enabled to avoid race
-  // condition.
-  if (!Queue->CounterBasedEventsEnabled)
   // Initally, we need to check if a command list has already been created
   // on this device that is available for use. If so, then reuse that
   // Level-Zero Command List and Fence for this PI call.
--- a/source/adapters/level_zero/context.hpp
+++ b/source/adapters/level_zero/context.hpp
@@ -146,9 +146,9 @@ struct ur_context_handle_t_ : _ur_object {
   // head.
   //
   // Cache of event pools to which host-visible events are added to.
-  std::vector<std::list<ze_event_pool_handle_t>> ZeEventPoolCache{4};
+  std::vector<std::list<ze_event_pool_handle_t>> ZeEventPoolCache{12};
   std::vector<std::unordered_map<ze_device_handle_t, size_t>>
-      ZeEventPoolCacheDeviceMap{4};
+      ZeEventPoolCacheDeviceMap{12};
 
   // This map will be used to determine if a pool is full or not
   // by storing number of empty slots available in the pool.
@@ -199,48 +199,73 @@ struct ur_context_handle_t_ : _ur_object {
   ur_result_t getFreeSlotInExistingOrNewPool(ze_event_pool_handle_t &, size_t &,
                                              bool HostVisible,
                                              bool ProfilingEnabled,
-                                             ur_device_handle_t Device);
+                                             ur_device_handle_t Device,
+                                             bool CounterBasedEventEnabled,
+                                             bool UsingImmCmdList);
 
   // Get ur_event_handle_t from cache.
   ur_event_handle_t getEventFromContextCache(bool HostVisible,
                                              bool WithProfiling,
-                                             ur_device_handle_t Device);
+                                             ur_device_handle_t Device,
+                                             bool CounterBasedEventEnabled);
 
   // Add ur_event_handle_t to cache.
   void addEventToContextCache(ur_event_handle_t);
 
+  enum EventPoolCacheType {
+    HostVisibleCacheType,
+    HostInvisibleCacheType,
+    HostVisibleCounterBasedRegularCacheType,
+    HostInvisibleCounterBasedRegularCacheType,
+    HostVisibleCounterBasedImmediateCacheType,
+    HostInvisibleCounterBasedImmediateCacheType
+  };
+
   std::list<ze_event_pool_handle_t> *
   getZeEventPoolCache(bool HostVisible, bool WithProfiling,
+                      bool CounterBasedEventEnabled, bool UsingImmediateCmdList,
                       ze_device_handle_t ZeDevice) {
-    if (HostVisible) {
-      if (ZeDevice) {
-        auto ZeEventPoolCacheMap = WithProfiling
-                                       ? &ZeEventPoolCacheDeviceMap[0]
-                                       : &ZeEventPoolCacheDeviceMap[1];
-        if (ZeEventPoolCacheMap->find(ZeDevice) == ZeEventPoolCacheMap->end()) {
-          ZeEventPoolCache.emplace_back();
-          ZeEventPoolCacheMap->insert(
-              std::make_pair(ZeDevice, ZeEventPoolCache.size() - 1));
-        }
-        return &ZeEventPoolCache[(*ZeEventPoolCacheMap)[ZeDevice]];
-      } else {
-        return WithProfiling ? &ZeEventPoolCache[0] : &ZeEventPoolCache[1];
+    EventPoolCacheType CacheType;
+
+    calculateCacheIndex(HostVisible, CounterBasedEventEnabled,
+                        UsingImmediateCmdList, CacheType);
+    if (ZeDevice) {
+      auto ZeEventPoolCacheMap =
+          WithProfiling ? &ZeEventPoolCacheDeviceMap[CacheType * 2]
+                        : &ZeEventPoolCacheDeviceMap[CacheType * 2 + 1];
+      if (ZeEventPoolCacheMap->find(ZeDevice) == ZeEventPoolCacheMap->end()) {
+        ZeEventPoolCache.emplace_back();
+        ZeEventPoolCacheMap->insert(
+            std::make_pair(ZeDevice, ZeEventPoolCache.size() - 1));
       }
+      return &ZeEventPoolCache[(*ZeEventPoolCacheMap)[ZeDevice]];
     } else {
-      if (ZeDevice) {
-        auto ZeEventPoolCacheMap = WithProfiling
-                                       ? &ZeEventPoolCacheDeviceMap[2]
-                                       : &ZeEventPoolCacheDeviceMap[3];
-        if (ZeEventPoolCacheMap->find(ZeDevice) == ZeEventPoolCacheMap->end()) {
-          ZeEventPoolCache.emplace_back();
-          ZeEventPoolCacheMap->insert(
-              std::make_pair(ZeDevice, ZeEventPoolCache.size() - 1));
-        }
-        return &ZeEventPoolCache[(*ZeEventPoolCacheMap)[ZeDevice]];
-      } else {
-        return WithProfiling ? &ZeEventPoolCache[2] : &ZeEventPoolCache[3];
-      }
+      return WithProfiling ? &ZeEventPoolCache[CacheType * 2]
+                           : &ZeEventPoolCache[CacheType * 2 + 1];
     }
+  }
+
+  ur_result_t calculateCacheIndex(bool HostVisible,
+                                  bool CounterBasedEventEnabled,
+                                  bool UsingImmediateCmdList,
+                                  EventPoolCacheType &CacheType) {
+    if (CounterBasedEventEnabled && HostVisible && !UsingImmediateCmdList) {
+      CacheType = HostVisibleCounterBasedRegularCacheType;
+    } else if (CounterBasedEventEnabled && !HostVisible &&
+               !UsingImmediateCmdList) {
+      CacheType = HostInvisibleCounterBasedRegularCacheType;
+    } else if (CounterBasedEventEnabled && HostVisible &&
+               UsingImmediateCmdList) {
+      CacheType = HostVisibleCounterBasedImmediateCacheType;
+    } else if (CounterBasedEventEnabled && !HostVisible &&
+               UsingImmediateCmdList) {
+      CacheType = HostInvisibleCounterBasedImmediateCacheType;
+    } else if (!CounterBasedEventEnabled && HostVisible) {
+      CacheType = HostVisibleCacheType;
+    } else {
+      CacheType = HostInvisibleCacheType;
+    }
+    return UR_RESULT_SUCCESS;
   }
 
   // Decrement number of events living in the pool upon event destroy
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -130,7 +130,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueEventsWait(
     if (OutEvent) {
       Queue->LastCommandEvent = reinterpret_cast<ur_event_handle_t>(*OutEvent);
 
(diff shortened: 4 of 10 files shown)


mabraham:
It's create to see things starting!

source/adapters/level_zero/queue.cpp
```diff
@@ -1498,12 +1522,11 @@ ur_event_handle_t ur_queue_handle_t_::getEventFromQueueCache(bool IsMultiDevice,
...
-                                         ur_event_handle_t *Event,
-                                         ur_command_t CommandType,
-                                         ur_command_list_ptr_t CommandList,
-                                         bool IsInternal, bool IsMultiDevice,
-                                         std::optional<bool> HostVisible) {
+ur_result_t createEventAndAssociateQueue(
+    ur_queue_handle_t Queue, ur_event_handle_t *Event, ur_command_t CommandType,
+    ur_command_list_ptr_t CommandList, bool IsInternal, bool IsMultiDevice,
+    std::optional<bool> HostVisible,
+    std::optional<bool> usingCounterBasedEvents) {
```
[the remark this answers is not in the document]
winstonzhang-intel: Good catch, let me fix that.

MichalMrozek (changes requested):
1. you can only use counter based events for in order queues, they do not work for out of order queues, you need to check sycl queue parameters instead of parsing over ze queues.
2. you cannot call reset on those events
3. why you disable fences ?
4. why you create all events with immediate and non immediate flags? Driver wouldn't be able to optimize this. Events should be selected basing on usage
5. why you add synchronous mode when counter based events are present ? 
6. event pools are per context which can have in order and out of order queues, you need to differentiate when you obtain events to not get counter-based event for ooq.
7. you cannot evaluate each time ur_queue_handle_t_::usingCounterBasedEvents() , this needs to be const flag set at queue creation, in order queue cannot be made out of order queue, map browsing would kill perf

winstonzhang-intel:
@MichalMrozek Thanks for the feedback.
1. you can only use counter based events for in order queues, they do not work for out of order queues, you need to check sycl queue parameters instead of parsing over ze queues.
I do make sure that we only use counter-based events for inOrderQueue. In additionally, I have to make sure that for the ZeQueues that are being used they have the ZE_COMMAND_QUEUE_FLAG_IN_ORDER flag
2. you cannot call reset on those events
Good catch, fixed
3. why you disable fences ?
My understanding was that with counter-based events, we no longer need barriers between the different lists, therefore, I removed the fences if we are using counter-based events.
4. why you create all events with immediate and non immediate flags? Driver wouldn't be able to optimize this. Events should be selected basing on usage
I agree, at first I was taking the L0 cts as example.  But come to think of it this makes sense.
5. why you add synchronous mode when counter based events are present?
Same as above.
6. event pools are per context which can have in order and out of order queues, you need to differentiate when you obtain events to not get counter-based event for ooq.
7. you cannot evaluate each time ur_queue_handle_t_::usingCounterBasedEvents() , this needs to be const flag set at queue creation, in order queue cannot be made out of order queue, map browsing would kill perf
Agree, fixed.

source/adapters/level_zero/event.cpp
```diff
@@ -1071,14 +1072,18 @@ ur_result_t EventCreate(ur_context_handle_t Context, ur_queue_handle_t Queue,
   size_t Index = 0;
 
   if (auto Res = Context->getFreeSlotInExistingOrNewPool(
-          ZeEventPool, Index, HostVisible, ProfilingEnabled, Device))
+          ZeEventPool, Index, HostVisible, ProfilingEnabled, Device,
+          CounterBasedEventEnabled.has_value()
```
MichalMrozek, line 1076: can you use Boolean instead of std::optional ?

source/adapters/level_zero/event.cpp
```diff
@@ -1124,8 +1129,8 @@ ur_result_t ur_event_handle_t_::reset() {
 
   if (!isHostVisible())
     HostVisibleEvent = nullptr;
-
-  ZE2UR_CALL(zeEventHostReset, (ZeEvent));
+  if (!UrQueue->usingCounterBasedEvents())
```
MichalMrozek, line 1132: I think that event should know this instead of going to queue.

source/adapters/level_zero/queue.cpp
```diff
@@ -1364,6 +1374,27 @@ bool ur_queue_handle_t_::isInOrderQueue() const {
           0);
 }
 
+/*
```
MichalMrozek, line 1377: please remove commented code

source/adapters/level_zero/queue.cpp
```diff
@@ -1364,6 +1374,27 @@ bool ur_queue_handle_t_::isInOrderQueue() const {
...
+bool ur_queue_handle_t_::usingCounterBasedEvents() const {
+  return this->isInOrderQueue() && ((this->Properties &
+UR_QUEUE_FLAG_USE_DRIVER_COUNTER_BASED_EVENTS) != 0);
+}
+*/
+
+bool ur_queue_handle_t_::usingCounterBasedEvents() const {
+  /*
```
MichalMrozek, line 1385: please remove commented code

source/adapters/level_zero/queue.cpp
```diff
@@ -1818,6 +1850,9 @@ ur_queue_handle_t_::ur_queue_group_t::getZeQueue(uint32_t *QueueGroupOrdinal) {
   if (QueueIndex != 0) {
     ZeCommandQueueDesc.flags = ZE_COMMAND_QUEUE_FLAG_EXPLICIT_ONLY;
   }
+  if (Queue->usingCounterBasedEvents()) {
```
MichalMrozek, line 1853: this should be other way around, if queue is in order , then you can create counter based events.

source/adapters/level_zero/queue.cpp
```diff
@@ -1854,32 +1889,38 @@ ur_result_t ur_queue_handle_t_::createCommandList(
...
 
-  ZeStruct<ze_command_list_desc_t> ZeCommandListDesc;
   ZeCommandListDesc.commandQueueGroupOrdinal = QueueGroupOrdinal;
-
   ZE2UR_CALL(zeCommandListCreate, (Context->ZeContext, Device->ZeDevice,
                                    &ZeCommandListDesc, &ZeCommandList));
 
-  ZE2UR_CALL(zeFenceCreate, (ZeCommandQueue, &ZeFenceDesc, &ZeFence));
+  if (!usingCounterBasedEvents()) {
+    ZE2UR_CALL(zeFenceCreate, (ZeCommandQueue, &ZeFenceDesc, &ZeFence));
```
MichalMrozek, line 1913: do not disable fences, they are still needed to recycle command lists.

MichalMrozek (changes requested):
It looks like this change has dependency on in order support

source/adapters/level_zero/queue.cpp
winstonzhang-intel: ZeQueue hasn't been created at this point. If we are using counter-based event then the zequeue needs to reflect that.

source/adapters/level_zero/event.cpp
winstonzhang-intel: Agreed, I can keep track of it during the creation of event. However, if it is passed in I would have no way of querying eventpool_desc.pnext from eventpool

source/adapters/level_zero/queue.cpp
raiyanla: I believe @MichalMrozek is referring to the SYCL Queue being passed in AKA the UR Queue object in this line. You can check if this Queue was created to be in-order using the Queue->isInOrderQueue(). 

Also I believe it should also be noted that ZE_COMMAND_QUEUE_FLAG_IN_ORDER flag should only used to create immediate command lists specifically using zeCommandListCreateImmediate(), while I see this function is used to create an ordinary queue for regular command lists using zeCommandQueueCreate(). Please correct me if I'm wrong @MichalMrozek 

Reference at: https://spec.oneapi.io/level-zero/latest/core/api.html?highlight=ze_command_queue_flag_in_order#_CPPv4N23ze_command_queue_flag_t30ZE_COMMAND_QUEUE_FLAG_IN_ORDERE

winstonzhang-intel: @raiyanla Thanks for your input. Queue->usingCounterBasedEvents() does check for whether the Queue is in order with Queue->isInOrderQueue(). It also checks for ZE_COMMAND_QUEUE_FLAG_IN_ORDER, which I believe is also required for zeCommandQueueCreate(). See example in L0 coformance test [here](https://github.com/oneapi-src/level-zero-tests/blob/5990783066d8114e220ffd57061201e4f8d7cb17/conformance_tests/core/test_cmdlist/src/test_cmdlist.cpp#L823)

raiyanla: @winstonzhang-intel Gotcha, yeah I just saw your usingCounterBasedEvents() already checks for isInOrderQueue() now. Thanks for linking that L0 conformance test, that is interesting, it seems to conflict with the specification definition of it above, since the spec link indicates its ZE_COMMAND_QUEUE_FLAG_IN_ORDER should be for immediate command lists only. @MichalMrozek Can you confirm which is is appropriate? Since the spec and conformance tests are conflicting.

MichalMrozek: For regular  command lists the flag is ZE_COMMAND_LIST_FLAG_IN_ORDER , passed in command list descriptor
For immediate command lists the flag is ZE_COMMAND_QUEUE_FLAG_IN_ORDER , passed in queue descriptor

source/adapters/level_zero/event.hpp
```diff
@@ -29,9 +29,10 @@
 
 extern "C" {
 ur_result_t urEventReleaseInternal(ur_event_handle_t Event);
-ur_result_t EventCreate(ur_context_handle_t Context, ur_queue_handle_t Queue,
-                        bool IsMultiDevice, bool HostVisible,
-                        ur_event_handle_t *RetEvent);
+ur_result_t
+EventCreate(ur_context_handle_t Context, ur_queue_handle_t Queue,
+            bool IsMultiDevice, bool HostVisible, ur_event_handle_t *RetEvent,
+            std::optional<bool> CounterBasedEventEnabled = std::nullopt);
```
pbalcer, line 35:
```suggestion
            bool CounterBasedEventEnabled = false);
```

?

source/adapters/level_zero/context.cpp
```diff
@@ -468,7 +468,8 @@ static const uint32_t MaxNumEventsPerPool = [] {
 
 ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
```
pbalcer, line 469: there's ought to be a better way of allocating an event that doesn't require passing around half a dozen arguments to multiple functions. Maybe some sort of flag?

`auto event = Pool->allocateEvent(HOST_VISIBLE | ENABLE_PROFILER | COUNTER_BASED);`

having functions that accept multiple boolean values is error prone.

source/adapters/level_zero/context.cpp
```diff
@@ -510,6 +511,12 @@ ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
       ZeEventPoolDesc.flags |= ZE_EVENT_POOL_FLAG_HOST_VISIBLE;
     if (ProfilingEnabled)
       ZeEventPoolDesc.flags |= ZE_EVENT_POOL_FLAG_KERNEL_TIMESTAMP;
+    if (CounterBasedEventEnabled) {
+      ZeEventPoolDesc.flags |= ZE_EVENT_POOL_FLAG_HOST_VISIBLE;
+      ze_event_pool_counter_based_exp_desc_t counterBasedExt = {
```
pbalcer, line 516: I think you might need to update the `getZeEventPoolCache(HostVisible, ProfilingEnabled, ZeDevice);` call above. Otherwise, based on what the queue was that happened to call this function first for this combination of parameters, the event pool might or might not have this flag set.

And, again, I suggest a small refactoring here to use flags (or some other mechanism) instead of passing yet another boolean.

winstonzhang-intel: The Queue should have the flag set on creation. Because of that, all the subsequent eventpools should be cached with counter-based events enabled.

source/adapters/level_zero/context.cpp
winstonzhang-intel: @pbalcer Thanks for the feedback! We could create a new enum just for this case. However, this seems to be a bit overkill just for this one function. It is only used once event.cpp during event creation.

pbalcer: It's not just this function. All the various boolean event parameters are being passed around in multiple functions. Just from a quick search:
`createEventAndAssociateQueue, getEventFromQueueCache, EventCreate, getEventFromContextCache, getFreeSlotInExistingOrNewPool, getZeEventPoolCache`

source/adapters/level_zero/context.cpp
pbalcer: But this cache is in the context, so if you have two queues, one with counter-based events enabled and one without, the one that was first in this function will create the event pool.

source/adapters/level_zero/context.cpp
winstonzhang-intel: @pbalcer This is quite a large refactoring. Functions with these booleans are abundant. Additionally, it would make sense to add this in **ur_api**, however, we would need to add this in spec before I can do so. Perhaps I can file a new ticket and have this be its own patch?

pbalcer: Sure, this was just a suggestion.

source/adapters/level_zero/context.cpp
MichalMrozek: event pools are stored in context not in a queue
There may be different types of queues coming in and they would require different type of events.
Pool cannot return incompatible event.

winstonzhang-intel: Using the same mechanism as previously implemented, the eventpool cache returned will be guaranteed to be with counterbased enabled if it is selected, since now there is a portion of the eventpoolcache that is reserved just for counterbased event pools.

pbalcer: 👍 this desperately needs a refactor though :)

winstonzhang-intel: I'll file a ticket tomorrow for all the refactoring, should be pretty straight forward 👍

source/adapters/level_zero/context.cpp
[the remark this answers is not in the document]
winstonzhang-intel: Fixed. There will only be regular cmdlist here, immediate cmdlist is retrieved via &ur_queue_handle_t_::ur_queue_group_t::getImmCmdList() in queue.cpp

source/adapters/level_zero/queue.cpp
```diff
@@ -1983,8 +2008,10 @@ ur_command_list_ptr_t &ur_queue_handle_t_::ur_queue_group_t::getImmCmdList() {
     Priority = "High";
   }
 
-  // Evaluate performance of explicit usage for "0" index.
-  if (QueueIndex != 0) {
+  if (Queue->Device->useDriverInOrderLists() && Queue->isInOrderQueue()) {
+    ZeCommandQueueDesc.flags = ZE_COMMAND_QUEUE_FLAG_IN_ORDER;
+  } else if (QueueIndex != 0) {
```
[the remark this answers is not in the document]
winstonzhang-intel: This is part of the in-order list patch `https://github.com/oneapi-src/unified-runtime/pull/1372/.` Not related to counter-based events.

source/adapters/level_zero/context.hpp
```diff
@@ -141,9 +141,9 @@ struct ur_context_handle_t_ : _ur_object {
   // head.
   //
   // Cache of event pools to which host-visible events are added to.
-  std::vector<std::list<ze_event_pool_handle_t>> ZeEventPoolCache{4};
+  std::vector<std::list<ze_event_pool_handle_t>> ZeEventPoolCache{8};
   std::vector<std::unordered_map<ze_device_handle_t, size_t>>
-      ZeEventPoolCacheDeviceMap{4};
+      ZeEventPoolCacheDeviceMap{8};
```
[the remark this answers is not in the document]
winstonzhang-intel: 2 EventPoolCacheDeviceMap reserved for each: hostVisible, not hostvisible, device and hostvisible and device and not hostvisible. Since we can't query the ext flags of eventpooldesc from eventpoolhandle, we have to reserve the space in case of each of these situations.

source/adapters/level_zero/context.hpp
```diff
@@ -206,35 +207,33 @@ struct ur_context_handle_t_ : _ur_object {
...
-        if (ZeEventPoolCacheMap->find(ZeDevice) == ZeEventPoolCacheMap->end()) {
-          ZeEventPoolCache.emplace_back();
-          ZeEventPoolCacheMap->insert(
-              std::make_pair(ZeDevice, ZeEventPoolCache.size() - 1));
-        }
-        return &ZeEventPoolCache[(*ZeEventPoolCacheMap)[ZeDevice]];
-      } else {
-        return WithProfiling ? &ZeEventPoolCache[0] : &ZeEventPoolCache[1];
-      }
+      profiling_index_a = 0;
```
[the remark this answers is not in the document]
winstonzhang-intel: yes to choose the cache index depending on hostVisible and CounterBasedEventEnabled.

source/adapters/level_zero/context.cpp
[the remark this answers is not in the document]
winstonzhang-intel: good catch, fixed

source/adapters/level_zero/context.hpp
```diff
@@ -206,36 +208,43 @@ struct ur_context_handle_t_ : _ur_object {
...
+                           : &ZeEventPoolCache[profiling_index_b];
+    }
+  }
+
+  ur_result_t calculateCacheIndex(bool HostVisible,
+                                  bool CounterBasedEventEnabled,
+                                  int *profiling_index_a,
+                                  int *profiling_index_b) {
+    if (HostVisible) {
+      *profiling_index_a = 0;
```
MichalMrozek, line 237: can you make this code more maintainable ?
assign some enum for each pool type, shift by enum sizes ?

Right now it would be very difficult to understand what is happening here and why, also profiling_index_a and profiling_index_b doesn't tell a lot what happens here.

source/adapters/level_zero/queue.cpp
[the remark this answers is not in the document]
MichalMrozek: can you re-base ?
 the in order patch is merged and has different change here

MichalMrozek (commented):
please re-base patch after in order command list change, right now it is hard to distinguish what are new additions and what comes from in order lists patch.

source/adapters/level_zero/queue.cpp
```diff
@@ -1861,19 +1877,18 @@ ur_result_t ur_queue_handle_t_::createCommandList(
   ze_command_list_handle_t ZeCommandList;
 
   uint32_t QueueGroupOrdinal;
+  ZeStruct<ze_command_list_desc_t> ZeCommandListDesc;
+  if (usingCounterBasedEvents()) {
```
nrspruit, line 1881: Does "using counter based events" override the check for:
if (Device->useDriverInOrderLists() && isInOrderQueue()) ?

Or should we be checking for all of them?

nrspruit: nvm, I found the code ie:
  this->counterBasedEventsEnabled = isInOrderQueue() &&
                                    Device->useDriverInOrderLists() &&
                                    useDriverCounterBasedEvents;
                                    
  please disregard.

winstonzhang-intel: Missed this during my rebase. But when counterbased is enabled then both inorder list and inorder queue will be enabled. Because of this we can just check for Device->useDriverInOrderLists() && isInOrderQueue()

nrspruit (approved):
LGTM, @MichalMrozek , please check to see if you have any further comments or issues with these changes.

source/adapters/level_zero/context.cpp
```diff
@@ -510,6 +511,18 @@ ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
       ZeEventPoolDesc.flags |= ZE_EVENT_POOL_FLAG_HOST_VISIBLE;
     if (ProfilingEnabled)
       ZeEventPoolDesc.flags |= ZE_EVENT_POOL_FLAG_KERNEL_TIMESTAMP;
+    if (CounterBasedEventEnabled) {
+      ZeEventPoolDesc.flags |= ZE_EVENT_POOL_FLAG_HOST_VISIBLE;
```
MichalMrozek, line 515: why HOST_VISIBLE even if HostVisible is not set?

source/adapters/level_zero/context.hpp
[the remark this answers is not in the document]
MichalMrozek: What if you need profiling on top of that?
Which pool would you choose?

If you want to permute all combination this would quickly be out of control.
- host visible
- counter based for immediate
- counter based for non immediate
- profiling

we are already at 16

source/adapters/level_zero/context.hpp
```diff
@@ -204,38 +206,49 @@ struct ur_context_handle_t_ : _ur_object {
...
-        return &ZeEventPoolCache[(*ZeEventPoolCacheMap)[ZeDevice]];
-      } else {
-        return WithProfiling ? &ZeEventPoolCache[2] : &ZeEventPoolCache[3];
-      }
+      return WithProfiling ? &ZeEventPoolCache[CacheType]
+                           : &ZeEventPoolCache[CacheType + 1];
+    }
+  }
+
+  ur_result_t calculateCacheIndex(bool HostVisible,
```
MichalMrozek, line 239: how do you differentiate between counter based events for immediate and regular command lists?

source/adapters/level_zero/context.hpp
```diff
@@ -204,38 +206,49 @@ struct ur_context_handle_t_ : _ur_object {
...
-          ZeEventPoolCache.emplace_back();
-          ZeEventPoolCacheMap->insert(
-              std::make_pair(ZeDevice, ZeEventPoolCache.size() - 1));
-        }
-        return &ZeEventPoolCache[(*ZeEventPoolCacheMap)[ZeDevice]];
-      } else {
-        return WithProfiling ? &ZeEventPoolCache[2] : &ZeEventPoolCache[3];
-      }
+      return WithProfiling ? &ZeEventPoolCache[CacheType]
+                           : &ZeEventPoolCache[CacheType + 1];
```
MichalMrozek, line 235: shouldn't you multiply by 2? 

Cache type +1 gives next enum value
for HostVisibleCacheType you will get HostInvisibleCacheType

winstonzhang-intel: +1 is for with profiling and without profiling for each hostvisible/!hostvisible pair for each regular, regular with counterbase and imm with counterbase.

source/adapters/level_zero/context.hpp
winstonzhang-intel: added immediate detection

source/adapters/level_zero/context.cpp
winstonzhang-intel: removed

nrspruit:
> 1. you can only use counter based events for in order queues, they do not work for out of order queues, you need to check sycl queue parameters instead of parsing over ze queues.
> 2. you cannot call reset on those events
> 3. why you disable fences ?
> 4. why you create all events with immediate and non immediate flags? Driver wouldn't be able to optimize this. Events should be selected basing on usage
> 5. why you add synchronous mode when counter based events are present ?
> 6. event pools are per context which can have in order and out of order queues, you need to differentiate when you obtain events to not get counter-based event for ooq.
> 7. you cannot evaluate each time ur_queue_handle_t_::usingCounterBasedEvents() , this needs to be const flag set at queue creation, in order queue cannot be made out of order queue, map browsing would kill perf

So, the CI failure here:
urUSMHostAllocTest.Success/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___Intel_R__Arc_TM__A750_Graphics

is because the reused counting event is coming up as signaled so it immediately completes. Either this is a bug in the L0 gpu driver if the spec says we cannot reset or we are missing something in this implementation.

nrspruit:
> > 1. you can only use counter based events for in order queues, they do not work for out of order queues, you need to check sycl queue parameters instead of parsing over ze queues.
> > 2. you cannot call reset on those events
> > 3. why you disable fences ?
> > 4. why you create all events with immediate and non immediate flags? Driver wouldn't be able to optimize this. Events should be selected basing on usage
> > 5. why you add synchronous mode when counter based events are present ?
> > 6. event pools are per context which can have in order and out of order queues, you need to differentiate when you obtain events to not get counter-based event for ooq.
> > 7. you cannot evaluate each time ur_queue_handle_t_::usingCounterBasedEvents() , this needs to be const flag set at queue creation, in order queue cannot be made out of order queue, map browsing would kill perf
> 
> So, the CI failure here: urUSMHostAllocTest.Success/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___Intel_R__Arc_TM__A750_Graphics
> 
> is because the reused counting event is coming up as signaled so it immediately completes. Either this is a bug in the L0 gpu driver if the spec says we cannot reset or we are missing something in this implementation.

Issue is fixed in the latest patch, the issue was the desc was not being used.

winstonzhang-intel:
> > > 1. you can only use counter based events for in order queues, they do not work for out of order queues, you need to check sycl queue parameters instead of parsing over ze queues.
> > > 2. you cannot call reset on those events
> > > 3. why you disable fences ?
> > > 4. why you create all events with immediate and non immediate flags? Driver wouldn't be able to optimize this. Events should be selected basing on usage
> > > 5. why you add synchronous mode when counter based events are present ?
> > > 6. event pools are per context which can have in order and out of order queues, you need to differentiate when you obtain events to not get counter-based event for ooq.
> > > 7. you cannot evaluate each time ur_queue_handle_t_::usingCounterBasedEvents() , this needs to be const flag set at queue creation, in order queue cannot be made out of order queue, map browsing would kill perf
> > 
> > 
> > So, the CI failure here: urUSMHostAllocTest.Success/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___Intel_R__Arc_TM__A750_Graphics
> > is because the reused counting event is coming up as signaled so it immediately completes. Either this is a bug in the L0 gpu driver if the spec says we cannot reset or we are missing something in this implementation.
> 
> Issue is fixed in the latest patch, the issue was the desc was not being used.

Thank you @nrspruit for looking into this. Once this patch is merged I will post another for the refactoring that @pbalcer requested.

source/adapters/level_zero/context.hpp
MichalMrozek: But you will get cache types as values from 0..5
So +1 values would be in range 1..6
1. Code would never use 12 cache entries.
2. Code would incorrectly overlap some entries

winstonzhang-intel: Oversight on my part, fixed now. Thanks.

source/adapters/level_zero/queue.hpp
```diff
@@ -234,6 +234,9 @@ struct ur_queue_handle_t_ : _ur_object {
   // Keeps the properties of this queue.
   ur_queue_flags_t Properties;
 
+  // Keeps track of whether we are using Counter-based Events
+  bool counterBasedEventsEnabled = false;
```
EwanC, line 238: nit: variable naming not quite consistent
```suggestion
  bool CounterBasedEventsEnabled = false;
```

source/adapters/level_zero/queue.cpp
```diff
@@ -955,6 +955,17 @@ ur_queue_handle_t_::ur_queue_handle_t_(
   ComputeCommandBatch.QueueBatchSize =
       ZeCommandListBatchComputeConfig.startSize();
   CopyCommandBatch.QueueBatchSize = ZeCommandListBatchCopyConfig.startSize();
+
+  static const bool useDriverCounterBasedEvents = [] {
```
EwanC, line 959: defining a lambda that only has one call site feels like extra complexity for little benefit.

source/adapters/level_zero/event.hpp
```diff
@@ -222,6 +223,11 @@ struct ur_event_handle_t_ : _ur_object {
 
   // Get the host-visible event or create one and enqueue its signal.
   ur_result_t getOrCreateHostVisibleEvent(ze_event_handle_t &HostVisibleEvent);
+
+  // Keeps track of whether we are using Counter-based Events.
+  bool CounterBasedEventsEnabled = false;
+
+  bool usingCounterBasedEvents() const { return CounterBasedEventsEnabled; }
```
EwanC, line 230: `CounterBasedEventsEnabled` looks like a public struct member, what's the benefit of defining a getter function to return it?

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -933,7 +933,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
       MustSignalWaitEvent = false;
     }
   }
-  if (MustSignalWaitEvent) {
+  if (MustSignalWaitEvent &&
+      !CommandBuffer->WaitEvent->CounterBasedEventsEnabled) {
```
EwanC, line 937: If we didn't set the `CounterBasedEventEnabled` parameter to true when `WaitEvent` is [created](https://github.com/oneapi-src/unified-runtime/blob/4ef258c85ea39c5c5b4269ee9ff1768a0d11192b/source/adapters/level_zero/command_buffer.cpp#L438)
```cpp
UR_CALL(EventCreate(Context, nullptr, false, false,
                      &RetCommandBuffer->WaitEvent));
```
Why do we need to check here that `CounterBasedEventsEnabled` isn't set? Can this property of the event change after the event has been created?

source/adapters/level_zero/queue.cpp
nrspruit: When it comes to querying environment variables, we define a lambda in all uses in the code. This provides consistency even if right now this is only used once.

MichalMrozek (approved):

nrspruit (approved):
LGTM with the updated changes.

raiyanla (approved):

PR #1371, opened 2024-02-22
[L0] fix a deadlock in queue sync and event status query

pbalcer:


pbalcer:
https://github.com/intel/llvm/pull/12795

kbenzie (approved):

nrspruit (approved):
So, by skipping the "open" execution, we chance the scenario where the open command list has not been executed and thus not make progress on the open command list, but as you say that is optional to progress on the open command list.

Thanks for the fix! I will have to look at how we can avoid the open command list case for an event more deterministically in the future if it does not apply.

PR #1375, opened 2024-02-22
[NFC] Remove USM memset entry points from level zero adapter.

aarongreig:
These were removed from the spec a while ago so this is dead code.


nrspruit (approved):
LGTM

PR #1380, opened 2024-02-23
[L0] fix implementation of device-related program queries

pbalcer:
This fixes the implementation of UR_PROGRAM_INFO_DEVICES and UR_PROGRAM_INFO_NUM_DEVICES when using multi-devices programs.

### DIFF
--- a/source/adapters/level_zero/program.cpp
+++ b/source/adapters/level_zero/program.cpp
@@ -81,7 +81,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramCreateWithBinary(
     ur_program_handle_t
         *Program ///< [out] pointer to handle of Program object created.
 ) {
-  std::ignore = Device;
   std::ignore = Properties;
   // In OpenCL, clCreateProgramWithBinary() can be used to load any of the
   // following: "program executable", "compiled program", or "library of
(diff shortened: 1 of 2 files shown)


pbalcer:
https://github.com/intel/llvm/pull/12813

source/adapters/level_zero/program.cpp
```diff
@@ -601,12 +601,16 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramGetInfo(
...
+  case UR_PROGRAM_INFO_NUM_DEVICES: {
+    return ReturnValue(static_cast<uint32_t>(Program->ZeModuleMap.size()));
+  }
+  case UR_PROGRAM_INFO_DEVICES: {
+    std::vector<ze_device_handle_t> devices;
+    for (const auto &entry : Program->ZeModuleMap) {
+      devices.push_back(entry.first);
+    }
+    return ReturnValue(devices.data(), devices.size());
+  }
```
omarahmed1111, line 613: I think that should report the devices inside the context. From what I could find `ZeModuleMap` will only be accumulated when either `urProgramBuildExp` is called or `urProgramLinkExp` and we could call this query before calling these entries (I might be missing something here).

pbalcer: Yea, this was wrong :D My fault for rushing.
From what I can see, all paths to build a program go through the exp functions except for urProgramCreateWithBinary. In which case `ZeModuleMap` is going to be empty.

Thanks.

nrspruit (approved):

PR #1382, opened 2024-02-26
[UR][L0] Implement bindless image functions from ZE_experimental_bindless_image

wenju-he:
Function zeMemGetPitchFor2dImage and zeImageGetDeviceOffsetExp from
L0 extension ZE_experimental_bindless_image are used.

In urBindlessImagesUnsampledImageCreateExp, if image description is
different from that of image allocated by urBindlessImagesImageAllocateExp,
an image view is created from the image.

### DIFF
--- a/source/adapters/level_zero/device.cpp
+++ b/source/adapters/level_zero/device.cpp
@@ -924,7 +924,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
   case UR_DEVICE_INFO_BINDLESS_IMAGES_SHARED_USM_SUPPORT_EXP:
     return ReturnValue(true);
   case UR_DEVICE_INFO_BINDLESS_IMAGES_1D_USM_SUPPORT_EXP:
-    return ReturnValue(true);
+    return ReturnValue(false);
   case UR_DEVICE_INFO_BINDLESS_IMAGES_2D_USM_SUPPORT_EXP:
     return ReturnValue(true);
   case UR_DEVICE_INFO_IMAGE_PITCH_ALIGN_EXP:
--- a/source/adapters/level_zero/device.cpp
+++ b/source/adapters/level_zero/device.cpp
@@ -919,6 +919,31 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
     return ReturnValue(true);
   case UR_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP:
     return ReturnValue(false);
+  case UR_DEVICE_INFO_BINDLESS_IMAGES_SUPPORT_EXP:
+    return ReturnValue(true);
+  case UR_DEVICE_INFO_BINDLESS_IMAGES_SHARED_USM_SUPPORT_EXP:
+    return ReturnValue(true);
+  case UR_DEVICE_INFO_BINDLESS_IMAGES_1D_USM_SUPPORT_EXP:
+    return ReturnValue(true);
+  case UR_DEVICE_INFO_BINDLESS_IMAGES_2D_USM_SUPPORT_EXP:
+    return ReturnValue(true);
+  case UR_DEVICE_INFO_IMAGE_PITCH_ALIGN_EXP:
+  case UR_DEVICE_INFO_MAX_IMAGE_LINEAR_WIDTH_EXP:
+  case UR_DEVICE_INFO_MAX_IMAGE_LINEAR_HEIGHT_EXP:
+  case UR_DEVICE_INFO_MAX_IMAGE_LINEAR_PITCH_EXP:
+    urPrint("Unsupported ParamName in urGetDeviceInfo\n");
+    urPrint("ParamName=%d(0x%x)\n", ParamName, ParamName);
+    return UR_RESULT_ERROR_INVALID_VALUE;
+  case UR_DEVICE_INFO_MIPMAP_SUPPORT_EXP:
+    return ReturnValue(true);
+  case UR_DEVICE_INFO_MIPMAP_ANISOTROPY_SUPPORT_EXP:
+    return ReturnValue(true);
+  case UR_DEVICE_INFO_MIPMAP_MAX_ANISOTROPY_EXP:
+  case UR_DEVICE_INFO_MIPMAP_LEVEL_REFERENCE_SUPPORT_EXP:
+  case UR_DEVICE_INFO_INTEROP_MEMORY_IMPORT_SUPPORT_EXP:
+  case UR_DEVICE_INFO_INTEROP_MEMORY_EXPORT_SUPPORT_EXP:
+  case UR_DEVICE_INFO_INTEROP_SEMAPHORE_IMPORT_SUPPORT_EXP:
+  case UR_DEVICE_INFO_INTEROP_SEMAPHORE_EXPORT_SUPPORT_EXP:
   default:
     urPrint("Unsupported ParamName in urGetDeviceInfo\n");
     urPrint("ParamName=%d(0x%x)\n", ParamName, ParamName);
--- a/source/adapters/level_zero/image.cpp
+++ b/source/adapters/level_zero/image.cpp
@@ -552,6 +552,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urUSMPitchedAllocExp(
     size_t *pResultPitch) {
   std::shared_lock<ur_shared_mutex> Lock(hContext->Mutex);
 
+  UR_ASSERT(hContext && hDevice, UR_RESULT_ERROR_INVALID_NULL_HANDLE);
+  UR_ASSERT(widthInBytes != 0, UR_RESULT_ERROR_INVALID_USM_SIZE);
+  UR_ASSERT(ppMem && pResultPitch, UR_RESULT_ERROR_INVALID_NULL_POINTER);
+
   static std::once_flag InitFlag;
   std::call_once(InitFlag, [&]() {
     ze_driver_handle_t DriverHandle = hContext->getPlatform()->ZeDriver;
@@ -606,6 +610,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageAllocateExp(
     ur_exp_image_mem_handle_t *phImageMem) {
   std::shared_lock<ur_shared_mutex> Lock(hContext->Mutex);
 
+  UR_ASSERT(hContext && hDevice, UR_RESULT_ERROR_INVALID_NULL_HANDLE);
+  UR_ASSERT(pImageFormat && pImageDesc && phImageMem,
+            UR_RESULT_ERROR_INVALID_NULL_POINTER);
+
   ZeStruct<ze_image_desc_t> ZeImageDesc;
   UR_CALL(ur2zeImageDesc(pImageFormat, pImageDesc, ZeImageDesc));
 
@@ -641,6 +649,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesUnsampledImageCreateExp(
     ur_exp_image_handle_t *phImage) {
   std::shared_lock<ur_shared_mutex> Lock(hContext->Mutex);
 
+  UR_ASSERT(hContext && hDevice && hImageMem,
+            UR_RESULT_ERROR_INVALID_NULL_HANDLE);
+  UR_ASSERT(pImageFormat && pImageDesc && phMem && phImage,
+            UR_RESULT_ERROR_INVALID_NULL_POINTER);
+
   ZeStruct<ze_image_desc_t> ZeImageDesc;
   UR_CALL(ur2zeImageDesc(pImageFormat, pImageDesc, ZeImageDesc));
 
@@ -669,7 +682,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesUnsampledImageCreateExp(
       ZeImage = UrImage->ZeImage;
       *phMem = nullptr;
     }
-  } else {
+  } else if (MemAllocProperties.type == ZE_MEMORY_TYPE_DEVICE) {
     ze_image_pitched_exp_desc_t PitchedDesc;
     PitchedDesc.stype = ZE_STRUCTURE_TYPE_PITCHED_IMAGE_EXP_DESC;
     PitchedDesc.pNext = nullptr;
@@ -688,6 +701,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesUnsampledImageCreateExp(
                (hContext->ZeContext, hDevice->ZeDevice, ZeImage));
     UR_CALL(createUrMemFromZeImage(hContext, ZeImage, /*OwnZeMemHandle*/ true,
                                    ZeImageDesc, phMem));
+  } else {
+    return UR_RESULT_ERROR_INVALID_VALUE;
   }
 
   static std::once_flag InitFlag;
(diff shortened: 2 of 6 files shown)


wenju-he:
@HoppeMateusz could you please review? thank you

wenju-he:
@jsji @YuriPlyakhin could you please review, thank you.

source/adapters/level_zero/device.cpp
```diff
@@ -919,6 +919,31 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
     return ReturnValue(true);
   case UR_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP:
     return ReturnValue(false);
+  case UR_DEVICE_INFO_BINDLESS_IMAGES_SUPPORT_EXP:
+    return ReturnValue(true);
+  case UR_DEVICE_INFO_BINDLESS_IMAGES_SHARED_USM_SUPPORT_EXP:
+    return ReturnValue(true);
+  case UR_DEVICE_INFO_BINDLESS_IMAGES_1D_USM_SUPPORT_EXP:
```
HoppeMateusz, line 926: ZE experimental bindless image is only defined for 2D image - this is what cuda support - and we didnt decide to support 1D.

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +595,178 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
                                              ur_exp_image_handle_t hImage) {
-  std::ignore = hContext;
-  std::ignore = hDevice;
-  std::ignore = hImage;
-  urPrint("[UR][L0] %s function not implemented!\n", __FUNCTION__);
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+  // Sampled image is a combination of unsampled image and sampler.
+  urBindlessImagesUnsampledImageHandleDestroyExp(hContext, hDevice, hImage);
+
+  return UR_RESULT_SUCCESS;
```
HoppeMateusz, line 609: we do not support sampled images - shoudl this function return SUCCESS ?

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +595,178 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
+      }
+    } else {
+      *phMem = nullptr;
+    }
+  } else {
+    ze_image_pitched_exp_desc_t ZeImagePitchedDesc;
+    ZeImagePitchedDesc.stype = ZE_STRUCTURE_TYPE_PITCHED_IMAGE_EXP_DESC;
+    ZeImagePitchedDesc.pNext = nullptr;
+    ZeImagePitchedDesc.ptr = hImageMem;
+    ZeImageDesc.pNext = &ZeImagePitchedDesc;
```
HoppeMateusz, line 711: we need BINDLESS FLAG here too - image from pitched memory must be created with bindless descriptor:

https://spec.oneapi.io/level-zero/latest/core/api.html#ze-image-pitched-exp-desc-t

struct ze_image_pitched_exp_desc_t
**Image descriptor for bindless images** created from pitched allocations

wenju-he: So it would be a chain of pNext like your test code:
```
    ze_image_pitched_exp_desc_t pitchedDesc = {};
    pitchedDesc.stype = ZE_STRUCTURE_TYPE_PITCHED_IMAGE_EXP_DESC;
    pitchedDesc.ptr = reinterpret_cast<void *>(0x1234000);

    ze_image_bindless_exp_desc_t bindlessExtDesc = {};
    bindlessExtDesc.stype = ZE_STRUCTURE_TYPE_BINDLESS_IMAGE_EXP_DESC;
    bindlessExtDesc.pNext = &pitchedDesc;
    bindlessExtDesc.flags = ZE_IMAGE_BINDLESS_EXP_FLAG_BINDLESS;
```
ZE_STRUCTURE_TYPE_PITCHED_IMAGE_EXP_DESC is only used for bindless image, right?
Alternatively can we set bindless flag to true when ZE_STRUCTURE_TYPE_PITCHED_IMAGE_EXP_DESC  type is present? This would save us a pNext.
IMHO the chain of two pNext make more sense if ZE_STRUCTURE_TYPE_PITCHED_IMAGE_EXP_DESC has multiple purposes.

`Image descriptor for bindless images created from pitched allocations` also indicates that ZE_STRUCTURE_TYPE_PITCHED_IMAGE_EXP_DESC is only used for bindless image.

source/adapters/level_zero/image.cpp
wenju-he: @ph0b, bindless sampler isn't supported in IGC and L0 runtime. For instance, if there is an array of sampled images used in a kernel, the case can't work correctly. It is a technical blocker at the moment. There is an on-going discussion of the issue involving @HoppeMateusz and IGC people.

@HoppeMateusz the reason I return UR_RESULT_SUCCESS here is that I'm able to make SYCL bindless sampled image tests that only use sampled image as kernel argument, e.g [sampling_1D.cpp](https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/bindless_images/sampling_1D.cpp) [sampling_2D.cpp](https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/bindless_images/sampling_2D.cpp) and [sampling_3D.cpp](https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/bindless_images/sampling_3D.cpp), working correctly on DG2, with the help of a few draft changes in IGC and L0 runtime. Basically sampler still needs a sampler_index encoded by IGC and L0 runtime saves sampler offset to a place according to sampler_index.

Return UR_RESULT_SUCCESS here also would allow us to test bindless sampler support as the above-mentioned technical blocker is in the process of being resolved.

source/adapters/level_zero/image.cpp
HoppeMateusz: yes, this has to be a chain.

For now, ZE_STRUCTURE_TYPE_PITCHED_IMAGE_EXP_DESC  is only defined for BINDLESS images - so we need to pass  ze_image_bindless_exp_desc_t   too. 

There could be possible etnesion that will allow creating non-bindless images with pitched pointer - something like OpenCL's:

cl_mem clCreateImageWithProperties(
    cl_context context,
    const cl_mem_properties* properties,
    cl_mem_flags flags,
    const cl_image_format* image_format,
    const cl_image_desc* image_desc,
    void* host_ptr,
    cl_int* errcode_ret);

source/adapters/level_zero/image.cpp
HoppeMateusz: ok, thanks for explanation.

Wouldnt it be better to return error code from the function call urBindlessImagesUnsampledImageHandleDestroyExp() then ? Or Is it failing?

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +586,186 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
+  try {
+    auto UrImage = new _ur_image(hContext, ZeImage, /*OwnZeMemHandle*/ true);
+    UrImage->ZeImageDesc = ZeImageDesc;
+    *phImageMem = reinterpret_cast<ur_exp_image_mem_handle_t>(UrImage);
+  } catch (const std::bad_alloc &) {
+    return UR_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+  } catch (...) {
+    return UR_RESULT_ERROR_UNKNOWN;
+  }
+  return UR_RESULT_SUCCESS;
```
jsji, line 633: Should we try to share the code with `urMemImageCreate` in memory.cpp?

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +586,186 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
+  ze_image_bindless_exp_desc_t ZeImageBindlessDesc;
+  ZeImageBindlessDesc.stype = ZE_STRUCTURE_TYPE_BINDLESS_IMAGE_EXP_DESC;
+  ZeImageBindlessDesc.pNext = nullptr;
+  ZeImageBindlessDesc.flags = ZE_IMAGE_BINDLESS_EXP_FLAG_BINDLESS;
+  ZeImageDesc.pNext = &ZeImageBindlessDesc;
+
+  ze_image_handle_t ZeImage;
+  ZE2UR_CALL(zeImageCreate,
+             (hContext->ZeContext, hDevice->ZeDevice, &ZeImageDesc, &ZeImage));
+  ZE2UR_CALL(zeContextMakeImageResident,
```
jsji, line 622: Did not see this in L0 spec example, is this required for bindless image only?

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +586,186 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
+    return UR_RESULT_SUCCESS;
+
+  if (UrImage->OwnNativeHandle) {
+    auto ZeResult = ZE_CALL_NOCHECK(zeImageDestroy, (UrImage->ZeImage));
+    // Gracefully handle the case that L0 was already unloaded.
+    if (ZeResult && ZeResult != ZE_RESULT_ERROR_UNINITIALIZED)
+      return ze2urResult(ZeResult);
+  }
+  delete UrImage;
+  return UR_RESULT_SUCCESS;
```
jsji, line 652: Similarly, should we try to share the code with `urMemRelease`?

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +586,186 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
+        (void **)&zeImageGetDeviceOffsetExpFunctionPtr);
+    if (Result != ZE_RESULT_SUCCESS)
+      urPrint("zeDriverGetExtensionFunctionAddress zeImageGetDeviceOffsetExp "
+              "failed, err = %d\n",
+              Result);
+  });
+  if (!zeImageGetDeviceOffsetExpFunctionPtr)
+    return UR_RESULT_ERROR_INVALID_OPERATION;
+
+  uint64_t DeviceOffset;
```
jsji, line 741: Should we init it?

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +586,186 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
                                              ur_exp_image_handle_t hImage) {
-  std::ignore = hContext;
-  std::ignore = hDevice;
-  std::ignore = hImage;
-  urPrint("[UR][L0] %s function not implemented!\n", __FUNCTION__);
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+  // Sampled image is a combination of unsampled image and sampler.
+  UR_CALL(urBindlessImagesUnsampledImageHandleDestroyExp(hContext, hDevice,
+                                                         hImage));
+
```
jsji, line 600: Do we need to call `zeSamplerDestroy` for `ZeSampler`?

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +586,186 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
+  struct combined_sampled_image_handle {
+    uint64_t raw_image_handle;
+    uint64_t raw_sampler_handle;
+  };
+  combined_sampled_image_handle *sampledImageHandle =
+      reinterpret_cast<combined_sampled_image_handle *>(phImage);
+  sampledImageHandle->raw_image_handle = reinterpret_cast<uint64_t>(*phImage);
+  sampledImageHandle->raw_sampler_handle =
+      reinterpret_cast<uint64_t>(hSampler->ZeSampler);
```
jsji, line 766: Do we want to assert `hSampler->ZeSampler`  is not null?

source/adapters/level_zero/image.cpp
```diff
@@ -114,32 +775,149 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
...
-  std::ignore = srcOffset;
-  std::ignore = dstOffset;
-  std::ignore = copyExtent;
-  std::ignore = hostExtent;
-  std::ignore = numEventsInWaitList;
-  std::ignore = phEventWaitList;
-  std::ignore = phEvent;
-  urPrint("[UR][L0] %s function not implemented!\n", __FUNCTION__);
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+  std::scoped_lock<ur_shared_mutex> Lock(hQueue->Mutex);
```
jsji, line 778: Should we try to share some code with `enqueueMemImageCommandHelper` in memory.cpp?

source/adapters/level_zero/device.cpp
HoppeMateusz: thanks

source/adapters/level_zero/image.cpp
HoppeMateusz: yes, memory resdiency of bindless images must be controlled by user of L0 driver , calling it here is correct.

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +586,151 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
+                                     ZeImageDesc, phMem));
+    } else {
+      ZeImage = UrImage->ZeImage;
+      *phMem = nullptr;
+    }
+  } else {
+    ze_image_pitched_exp_desc_t PitchedDesc;
+    PitchedDesc.stype = ZE_STRUCTURE_TYPE_PITCHED_IMAGE_EXP_DESC;
+    PitchedDesc.pNext = nullptr;
+    PitchedDesc.ptr = hImageMem;
```
HoppeMateusz, line 676: is this correct ? 

ptr is expected to be raw pointer returned from  urUSMPitchedAllocExp -> urUSMDeviceAlloc(hContext, hDevice, pUSMDesc, pool, Size, ppMem) ->  ppMem

source/adapters/level_zero/image.cpp
```diff
@@ -36,75 +586,151 @@ urBindlessImagesUnsampledImageHandleDestroyExp(ur_context_handle_t hContext,
...
+
+  UR_CALL(urBindlessImagesUnsampledImageCreateExp(
+      hContext, hDevice, hImageMem, pImageFormat, pImageDesc, phMem, phImage));
+
+  struct combined_sampled_image_handle {
+    uint64_t raw_image_handle;
+    uint64_t raw_sampler_handle;
+  };
+  combined_sampled_image_handle *sampledImageHandle =
+      reinterpret_cast<combined_sampled_image_handle *>(phImage);
```
HoppeMateusz, line 728: there is no mechanism to prevent wrong casts ?
 
we rely on application calling  XXXSampledImageXXX() function with only Sampled Images

and UnsampledImage - with non sampled images? 

pasign Unsampled image handle to this function will crash  / or be UB

source/adapters/level_zero/image.cpp
HoppeMateusz: why cant we use Image function   to copy from HOST  to Bindless Image ?

ze_image_handle_t  returned from zeImageCreate should work with standard image API

wenju-he: if the underlying memory is non-usm, i.e. created by zeImageCreate, zeCommandListAppendImageCopyFromMemory is called as shown below, so it is image function.
I was intending to use zeCommandListAppendImageCopyFromMemoryExt/zeCommandListAppendImageCopyToMemoryExt, but then find they are not implemented in level-zero runtime. Is this a gap? @HoppeMateusz

source/adapters/level_zero/image.cpp
wenju-he: ur_exp_image_handle_t size is 64bit for unsampled image and 128bit for sampled image for L0 adaptor.
Unfortunately size info isn't passed to this function.
In the long run the ideal solution is likely to support 64bit sampled image handle. For instance, can we always set bindless sampler offset to be the same as bindless image offset when zeImageCreate accepts a sampler argument in the future like cuda api cuTexObjectCreate? Then only one 64bit handle is enough for sampled image.

source/adapters/level_zero/image.cpp
wenju-he: hImageMem can be either an image created by zeImageCreate or usm created by zeMemAllocDevice. So a query is needed at the beginning of this function.
added a check and return error if MemAllocProperties.type isn't device usm

HoppeMateusz: ok, 
it wasnt clear to me that  ur_exp_image_mem_handle_t   is interpreted as image handle or just a raw pointer.

HoppeMateusz (approved):

ph0b (approved):

source/adapters/level_zero/image.cpp
jsji: Thanks @wenju-he  and @HoppeMateusz !

source/adapters/level_zero/image.cpp
jsji: Sounds good.

jsji (approved):
Thanks!

nrspruit (approved):
LGTM

raiyanla (approved):

kbenzie:
@nrspruit this should not have been merged without the approval of @oneapi-src/unified-runtime-bindless-images-write. It should also not have been merged before other PR's in the [ready to merge](https://github.com/oneapi-src/unified-runtime/pulls?q=is%3Aopen+is%3Apr+label%3A%22ready+to+merge%22) queue. Please reach out before pressing merge next time.

kbenzie:
@nrspruit and there's always room to accomodate deliverable priorities while keeping in the queue.

PR #1390, opened 2024-02-27
Add enqueue timestamp recording extension

steffenlarsen:
This commit adds a new extension feature for recording timestamps into events, the information from which can be queried using the existing profiling queries.


steffenlarsen:
Re-opened as https://github.com/oneapi-src/unified-runtime/pull/1400 for a cleaner history.

PR #1394, opened 2024-02-28
Add some missing validations in CTS

omarahmed1111:
This PR add tests for:
* `urDeviceGet` with variants on the device type.
* `UR_MAP_FLAG_WRITE_INVALIDATE_REGION` for `urEnqueueMemBufferMap`
* `urMemBufferCreate` with properties and assigned host ptr.
* `UR_PLATFORM_INFO_BACKEND` info for `urPlatformGetInfo`
* `urProgramBuild` with empty options argument.
* `urProgramCreateWithIL` with properties enabled.
* `urQueueCreate` with different flags.


test/conformance/memory/memory_adapter_opencl.match
```diff
@@ -1 +1,23 @@
 urMemImageCreateTest.InvalidImageDescStype/Intel_R__OpenCL___{{.*}}
+urMemImageCreateTestWithChannelOrderParam.Success/Intel_R__OpenCL___{{.*}}___UR_IMAGE_CHANNEL_ORDER_A
```
aarongreig, line 2: This is something we need more spec for, it's unlikely many (or maybe any) backends will support all of these. We're missing something like https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clGetSupportedImageFormats (also see table number 19 under https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#minimum-list-of-supported-image-formats for what we should probably expect for our minimum)

omarahmed1111: Yeah, good point, I am inclined to make all of them skip if unsupported and change that when we agree on a list that would match opencl and other current adapters too. What do you think?

aarongreig: I think that makes sense, although I'm worried that in the absence of a way to query which formats are supported there wont be a consistent way to tell if a fail is because the format is unsupported or if it's happening for another reason. Are the adapters returning `UR_RESULT_ERROR_IMAGE_FORMAT_NOT_SUPPORTED` when these tests fail or are the native errors getting translated into something more generic?

omarahmed1111: From my checking, CUDA/HIP/OPENCL should be returning `UR_RESULT_ERROR_IMAGE_FORMAT_NOT_SUPPORTED`. I think I could skip on that error code return only and for any other failures, it shouldn't skip.

aarongreig: Ok yeah if it'll work consistently then that's a good short term solution

omarahmed1111: I have added a skip condition on `UR_RESULT_ERROR_IMAGE_FORMAT_NOT_SUPPORTED` but it still fails on this tests on adapters with another error code `UR_RESULT_ERROR_UNSUPPORTED_FEATURE` is returned. The adapters after investigation seem to return `UR_RESULT_ERROR_IMAGE_FORMAT_NOT_SUPPORTED` on `image.cpp` file but the `memory.cpp` was only using `die` and `UR_RESULT_ERROR_UNSUPPORTED_FEATURE` for unsupported image format. So, I think we could leave the match code like that to force the correction of adapters behaviour regarding reporting unsupported image format. and work on that in another PR that would fix adapters behaviour but would need also `INTEL/LLVM` testing with it.

aarongreig: I don't think it's worth adding the tests in this PR if adapter changes are needed to make them work properly, might as well add them alongside the necessary fixes in another PR

omarahmed1111: Yeah, after a second thought, these tests might as well need to be done in a different way to test every combination of `ur_image_channel_order_t` with `ur_image_channel_type_t` and define the primary and non-primary combinations in the spec and in the CTS tests as mentione the OPENCL spec. So, I will move all `urMemImageCreate` tests in a new PR that will be dedicated for that.

omarahmed1111: Done

aarongreig (approved):

kbenzie (approved):

PR #1395, opened 2024-02-28
[HIP] Implement workaround for hipMemset2D

konradkusiak97:
There is an issue with `hipMemset2D` in ROCm prior to **6.0.0** version and this PR adds a workaround for it in `commonMemSetLargePattern`.

The issue appears only when using a pointer to host pinned memory from `hipHostMalloc`. I believe such a case hasn't been used until trying to refactor the USM fill (https://github.com/intel/llvm/pull/12702).

**Testing intel/llvm**: https://github.com/intel/llvm/pull/12898

### DIFF
--- a/source/adapters/hip/enqueue.cpp
+++ b/source/adapters/hip/enqueue.cpp
@@ -769,29 +769,19 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferCopyRect(
   return Result;
 }
 
-// HIP has no memset functions that allow setting values more than 4 bytes. UR
-// API lets you pass an arbitrary "pattern" to the buffer fill, which can be
-// more than 4 bytes. We must break up the pattern into 1 byte values, and set
-// the buffer using multiple strided calls.  The first 4 patterns are set using
-// hipMemsetD32Async then all subsequent 1 byte patterns are set using
-// hipMemset2DAsync which is called for each pattern.
-ur_result_t commonMemSetLargePattern(hipStream_t Stream, uint32_t PatternSize,
-                                     size_t Size, const void *pPattern,
-                                     hipDeviceptr_t Ptr) {
-  // Calculate the number of patterns, stride, number of times the pattern
-  // needs to be applied, and the number of times the first 32 bit pattern
-  // needs to be applied.
+static inline void memsetRemainPattern(hipStream_t Stream, uint32_t PatternSize,
+                                       size_t Size, const void *pPattern,
+                                       hipDeviceptr_t Ptr) {
+
+  // Calculate the number of patterns, stride and the number of times the
+  // pattern needs to be applied.
   auto NumberOfSteps = PatternSize / sizeof(uint8_t);
   auto Pitch = NumberOfSteps * sizeof(uint8_t);
   auto Height = Size / NumberOfSteps;
-  auto Count32 = Size / sizeof(uint32_t);
 
-  // Get 4-byte chunk of the pattern and call hipMemsetD32Async
-  auto Value = *(static_cast<const uint32_t *>(pPattern));
-  UR_CHECK_ERROR(hipMemsetD32Async(Ptr, Value, Count32, Stream));
   for (auto step = 4u; step < NumberOfSteps; ++step) {
     // take 1 byte of the pattern
-    Value = *(static_cast<const uint8_t *>(pPattern) + step);
+    auto Value = *(static_cast<const uint8_t *>(pPattern) + step);
 
     // offset the pointer to the part of the buffer we want to write to
     auto OffsetPtr = reinterpret_cast<void *>(reinterpret_cast<uint8_t *>(Ptr) +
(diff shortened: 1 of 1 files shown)


source/adapters/hip/enqueue.cpp
```diff
@@ -770,29 +770,68 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferCopyRect(
...
+  };
+  // There is a bug in ROCm prior to 6.0.0 version which causes hipMemset2D and
+  // hipMemset3D to behave incorrectly when acting on host pinned memory. In
+  // such a case the following part of memsetting the remaining part of the
+  // pattern is emulated with memcpy.
+#if HIP_VERSION < 60000000
+  hipPointerAttribute_t ptrAttribs{};
+  UR_CHECK_ERROR(hipPointerGetAttributes(&ptrAttribs, (const void *)Ptr));
+
+  const bool ptrIsHost{ptrAttribs.memoryType == hipMemoryTypeHost};
```
GeorgeWeb, line 808: Since ROCm 5.7.1 the `memoryType`  for shared allocations will override `hipMemoryTypeHost` and `hipMemoryTypeDevice` with `hipMemoryTypeManaged`, so it's not ideal to rely on this way. I see you have checked explicitly `!ptrAttribs.isManaged` below, which is sufficient to tell you what  you need in this case, but I just wanted to let you know to be aware of those changes.
I think another possible way to check for a _HIP host allocation_ is `ptrAttribs.hostPointer && !ptrAttribs.isManaged`, but as long as `!isManaged` then your check will work anyways, so not blocking the PR.blocking

source/adapters/hip/enqueue.cpp
```diff
@@ -770,29 +770,68 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferCopyRect(
...
+    // Loop through the memory area to memset, advancing each time by the
+    // PatternSize and memcpy the left over pattern bits.
+    for (uint32_t i = 0; i < NumOfCopySteps; ++i) {
+      auto OffsetDstPtr = reinterpret_cast<void *>(
+          reinterpret_cast<uint8_t *>(Ptr) + Offset + i * PatternSize);
+      UR_CHECK_ERROR(hipMemcpyAsync(OffsetDstPtr, OffsetPatternPtr,
+                                    LeftPatternSize, hipMemcpyHostToHost,
+                                    Stream));
+    }
+  } else
```
GeorgeWeb, line 830: Preferably stick to one style for the `if-else` here, so adding the `{}` in the `else` as well.

source/adapters/hip/enqueue.cpp
```diff
@@ -770,29 +770,68 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferCopyRect(
...
   auto Value = *(static_cast<const uint32_t *>(pPattern));
   UR_CHECK_ERROR(hipMemsetD32Async(Ptr, Value, Count32, Stream));
-  for (auto step = 4u; step < NumberOfSteps; ++step) {
-    // take 1 byte of the pattern
-    Value = *(static_cast<const uint8_t *>(pPattern) + step);
 
-    // offset the pointer to the part of the buffer we want to write to
-    auto OffsetPtr = reinterpret_cast<void *>(reinterpret_cast<uint8_t *>(Ptr) +
-                                              (step * sizeof(uint8_t)));
+  auto memsetRemainPattern = [&Stream, &pPattern,
```
GeorgeWeb, line 779: I like that this logic has been abstracted in a helper but would prefer it to be a self-contained free function instead of a lambda. It is easier to read and find, and also that lambda isn't free of cost anymore since it is a little complex and captures state anyways.

GeorgeWeb (commented):
Generally looks good to me. Would just prefer the lambda be a free function.

source/adapters/hip/enqueue.cpp
```diff
@@ -770,29 +770,68 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferCopyRect(
...
+      // set all of the pattern chunks
+      UR_CHECK_ERROR(hipMemset2DAsync(OffsetPtr, Pitch, Value, sizeof(uint8_t),
+                                      Height, Stream));
+    }
+  };
+  // There is a bug in ROCm prior to 6.0.0 version which causes hipMemset2D and
+  // hipMemset3D to behave incorrectly when acting on host pinned memory. In
+  // such a case the following part of memsetting the remaining part of the
+  // pattern is emulated with memcpy.
+#if HIP_VERSION < 60000000
```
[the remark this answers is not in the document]
konradkusiak97: Changed to `HIP_VERSION_MAJOR < 6`

source/adapters/hip/enqueue.cpp
konradkusiak97: Aha yes, I changed it to `hostPointer && !isManaged`, thanks for the suggestion!

source/adapters/hip/enqueue.cpp
konradkusiak97: Done

source/adapters/hip/enqueue.cpp
konradkusiak97: I moved the logic to a standalone function

source/adapters/hip/enqueue.cpp
GeorgeWeb: That working as expected? I never tested this suggestion myself.

GeorgeWeb (approved):

source/adapters/hip/enqueue.cpp
GeorgeWeb: Yeah that's the right behaviour. Nice, good to know.

PR #1397, opened 2024-02-29
[CUDA][HIP] Catch and report bad_alloc errors for event object creation

GeorgeWeb:
This PR is just to make sure there errors are reported accordingly if new fails to allocate the requested space when creating UR event object from a native event.

inte/llvm CI: https://github.com/intel/llvm/pull/12895


keyradical (approved):
LGTM.

Issue #1235, opened 2024-01-09
Info queries for interop objects fail silently


alexbatashev:
Objects that have been created with interop APIs often lack important pieces of information, required for info queries. For example, `urEventCreateWithNativeHandle` expects a native CUDA handle, and the resulting event handle will lack a reference to a UR queue. However, `urEventGetInfo` has no checks for whether an object was created with a native API or not, neither it will return an error in that case. Instead the returned queue handle will be invalid, i.e. `== nullptr`:

https://github.com/oneapi-src/unified-runtime/blob/2b7b827cfb8c87d84039188ed85d4e7d9fb738de/source/adapters/cuda/event.cpp#L160-L183

alycm:
Expect that this also applies to HIP, but not OpenCL, initial thought on Level Zero is that this should issue should not apply but need to be confirmed.

Possible solutions:
* Could track events created with handles and return an error on queries that cannot be supported.
* Could always fully initialize the event, but is this possible?

alexbatashev:
> Expect that this also applies to HIP, but not OpenCL, initial thought on Level Zero is that this should issue should not apply but need to be confirmed.

Yes, HIP is also affected by this issue. L0 is not a target of interest for me right now, so I didn't check.

> Could track events created with handles and return an error on queries that cannot be supported.

Actually, the kind of information I needed to extract from an event is the associated platform backend. This kind of info query seems to be trivial to implement for any of the handles, but that should be discussed separately.

Returning an error (while not a perfect solution) is still much better than silently returning invalid handles. 

> Could always fully initialize the event, but is this possible?

I remember early implementations of DPC++ backend interop APIs accepted a custom structure with a few more data fields to be able to sufficiently construct the object. I'll try to find a PR for a reference. This would solve some other issues, like how to capture profiling info in the UR events. However, I suspect, that it's not always possible to provide a reference to a queue when integrating with an external system.

al42and:
FYI, there is a related issue https://github.com/intel/llvm/issues/13706: for interop events, the `UR_EVENT_INFO_COMMAND_EXECUTION_STATUS` does not behave correctly, treating the event as never recorded based on default-initialized `IsRecorded` value, regardless of the actual status of the event (if it was recorded externally).

PR #1399, opened 2024-02-29
[CUDA][HIP] Return error on silently failing urEventGetInfo queries for interop created events

GeorgeWeb:
This PR addresses https://github.com/oneapi-src/unified-runtime/issues/1235 by returning an error on event queries, specifically `UR_EVENT_INFO_COMMAND_QUEUE`, that cannot be supported for the CUDA and HIP adapters. This is an effect of creating the ur_event from a native event which leads to not being able to fully initialize the UR object.

intel/llvm CI: https://github.com/intel/llvm/pull/12958

### DIFF
--- a/source/adapters/cuda/event.cpp
+++ b/source/adapters/cuda/event.cpp
@@ -17,7 +17,6 @@
 
 #include <cassert>
 #include <cuda.h>
-#include <memory>
 
 ur_event_handle_t_::ur_event_handle_t_(ur_command_t Type,
                                        ur_context_handle_t Context,
@@ -292,17 +291,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventCreateWithNativeHandle(
 
   std::unique_ptr<ur_event_handle_t_> EventPtr{nullptr};
 
-  try {
-    EventPtr =
-        std::unique_ptr<ur_event_handle_t_>(ur_event_handle_t_::makeWithNative(
-            hContext, reinterpret_cast<CUevent>(hNativeEvent)));
-  } catch (const std::bad_alloc &) {
-    return UR_RESULT_ERROR_OUT_OF_HOST_MEMORY;
-  } catch (...) {
-    return UR_RESULT_ERROR_UNKNOWN;
-  }
-
-  *phEvent = EventPtr.release();
+  *phEvent = ur_event_handle_t_::makeWithNative(
+      hContext, reinterpret_cast<CUevent>(hNativeEvent));
 
   return UR_RESULT_SUCCESS;
 }
--- a/source/adapters/cuda/event.cpp
+++ b/source/adapters/cuda/event.cpp
@@ -167,14 +167,13 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventGetInfo(ur_event_handle_t hEvent,
   case UR_EVENT_INFO_COMMAND_QUEUE: {
     // If the runtime owns the native handle, we have reference to the queue.
     // Otherwise, the event handle comes from an interop API with no RT refs.
-    if (!hEvent->backendHasOwnership()) {
+    if (!hEvent->getQueue()) {
       setErrorMessage("Command queue info cannot be queried for the event. The "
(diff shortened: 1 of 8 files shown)


source/adapters/cuda/event.cpp
```diff
@@ -165,8 +166,19 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventGetInfo(ur_event_handle_t hEvent,
...
   switch (propName) {
-  case UR_EVENT_INFO_COMMAND_QUEUE:
+  case UR_EVENT_INFO_COMMAND_QUEUE: {
+    // If the runtime owns the native handle, we have reference to the queue.
+    // Otherwise, the event handle comes from an interop API with no RT refs.
+    if (!hEvent->backendHasOwnership()) {
+      setErrorMessage("Command queue info cannot be queried for the event. The "
+                      "event object was created from a native event and has no "
+                      "valid reference to a command queue.",
+                      UR_RESULT_ERROR_INVALID_VALUE);
```
GeorgeWeb, line 176: Any opinions on the return code to use will be appreciated.

Additionally, I am willing to make it a non-backend specific error and simply return appropriate error code but I am unable to find such reasonable one. Also, I think this way we are able to relay the message to the user a bit better.

test/adapters/hip/test_event.cpp
```diff
@@ -0,0 +1,49 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "event.hpp"
+#include "fixtures.h"
+#include "uur/raii.h"
+
+#include <hip/hip_runtime.h>
+#include <tuple>
+
+struct RAIIHipEvent {
```
jchlanda, line 13: In cuda this lives in a dedicated header, would it be worth it to do the same here?

jchlanda (approved):

source/adapters/cuda/event.cpp
GeorgeWeb: It may be appropriate to just return UR_RESULT_ERROR_INVALID_VALUE but I won't be able to provide a more descriptive error message that way. Hence, adapter specific which in the SYCL Runtime is handled to report the custom error buffer accordingly.

test/adapters/hip/test_event.cpp
GeorgeWeb: Could change it, yep. I didn't solely on the basis of not using anywhere else yet, apart from the event test source file, so I thought I'd leave that work for a future PR that may need this object in other source file(s).
If you'd prefer it, I can update the PR moving the RAII wrapper it in its own header.

JackAKirk:
@GeorgeWeb I am just planning implementing the suggestion here: https://github.com/oneapi-src/unified-runtime/issues/1491#issuecomment-2083074881
which will I think make this implementation inconsistent because interop events will then have the possibility of having an associated native queue when instantiated with `make_event` interface that takes a native queue (e.g. stream). Although an error message may still be required since the existing `make_event` interface will still exist that doesn't take a native queue. So the check such I guess then be that the queue associated with the event is not null, rather than the check for ownership?

@alexbatashev FYI

GeorgeWeb:
> @GeorgeWeb I am just planning implementing the suggestion here: https://github.com/oneapi-src/unified-runtime/issues/1491#issuecomment-2083074881
> 
> which will I think make this implementation inconsistent because interop events will then have the possibility of having an associated native queue when instantiated with `make_event` interface that takes a native queue (e.g. stream). Although an error message may still be required since the existing `make_event` interface will still exist that doesn't take a native queue. So the check such I guess then be that the queue associated with the event is not null, rather than the check for ownership?
> 
> 
> 
> @alexbatashev FYI

@JackAKirk That sounds like a plan. I'll update this error on nullptr queue instead.

GeorgeWeb:
@oneapi-src/unified-runtime-maintain just a ping to have a look as that's been stale for some time. Thanks!

kbenzie (approved):

PR #1400, opened 2024-02-29
[UR][L0][CUDA][HIP] Add enqueue timestamp recording extension

steffenlarsen:
This commit adds a new extension feature for recording timestamps into events, the information from which can be queried using the existing profiling queries.

This new functionality is currently implemented for L0, CUDA and HIP.

Corresponding SYCL extension implementation: https://github.com/intel/llvm/pull/12838


source/adapters/cuda/enqueue.cpp
```diff
@@ -1756,3 +1756,33 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueWriteHostPipe(
...
 }
+
+UR_APIEXPORT ur_result_t UR_APICALL urEnqueueTimestampRecordingExp(
+    ur_queue_handle_t hQueue, bool blocking, uint32_t numEventsInWaitList,
+    const ur_event_handle_t *phEventWaitList, ur_event_handle_t *phEvent) {
+
+  ur_result_t Result = UR_RESULT_SUCCESS;
+  std::unique_ptr<ur_event_handle_t_> RetImplEvent{nullptr};
+  try {
+    ScopedContext Active(hQueue->getContext());
```
JackAKirk, line 1767:
```suggestion
    ScopedContext Active(hQueue->getDevice());
```
I think this might be necessary once https://github.com/intel/llvm/pull/10737 or similar is merged. CUcontext maps 1:1 with ur device but won't map 1:1 with ur context.

source/adapters/hip/enqueue.cpp
```diff
@@ -1895,3 +1895,34 @@ void setCopyRectParams(ur_rect_region_t Region, const void *SrcPtr,
                      : (DstType == hipMemoryTypeDevice ? hipMemcpyHostToDevice
                                                        : hipMemcpyHostToHost));
 }
+
+UR_APIEXPORT ur_result_t UR_APICALL urEnqueueTimestampRecordingExp(
+    ur_queue_handle_t hQueue, bool blocking, uint32_t numEventsInWaitList,
+    const ur_event_handle_t *phEventWaitList, ur_event_handle_t *phEvent) {
+
```
JackAKirk, line 1902: Is this function missing a scopedContext?

include/ur_api.h
```diff
@@ -1647,7 +1649,7 @@ typedef enum ur_device_info_t {
 ///     - ::UR_RESULT_ERROR_INVALID_NULL_HANDLE
 ///         + `NULL == hDevice`
 ///     - ::UR_RESULT_ERROR_INVALID_ENUMERATION
-///         + `::UR_DEVICE_INFO_INTEROP_SEMAPHORE_EXPORT_SUPPORT_EXP < propName`
```
JackAKirk, line 1650: Just checking you meant to delete this line?

source/adapters/hip/enqueue.cpp
steffenlarsen: Good catch! It has been added.

include/ur_api.h
steffenlarsen: Seems like it needs to be the last enumerator, as it's saying any value after that will result in `UR_RESULT_ERROR_INVALID_ENUMERATION`. I was skeptical at first too, but the generated script diff check insisted.

kbenzie: Yeah, this is the generator making the change. It's expected.

kbenzie: It could probably be a bit more clever to do proper validation of the enum ranges.

source/adapters/cuda/enqueue.cpp
JackAKirk: ah sorry, my mistake I think that `getDevice()` is not yet implemented in the cuda adapter (only hip), so for that one I think the original code was correct.

steffenlarsen: I see! It has been reverted. 😄

JackAKirk (approved):
hip/cuda impl and tests LGTM.

source/adapters/hip/event.cpp
```diff
@@ -58,7 +59,7 @@ ur_result_t ur_event_handle_t_::start() {
   ur_result_t Result = UR_RESULT_SUCCESS;
 
   try {
-    if (Queue->URFlags & UR_QUEUE_FLAG_PROFILING_ENABLE) {
+    if (Queue->URFlags & UR_QUEUE_FLAG_PROFILING_ENABLE || isTimestampEvent()) {
       // NOTE: This relies on the default stream to be unused.
       UR_CHECK_ERROR(hipEventRecord(EvQueued, 0));
```
steffenlarsen, line 64: @JackAKirk - In https://github.com/intel/llvm/pull/12838 it seems like the submission time on HIP is giving weird values. I did a bit of digging and it seems to me like HIP is a little different from CUDA when checking timing-differences between events. Of particular interest here is the following line for [hipEventElapsedTime()](https://docs.amd.com/projects/HIP/en/docs-5.4.0/doxygen/html/group___event.html#gad4128b815cb475c8e13c7e66ff6250b7): 

> Events which are recorded in a NULL stream will block until all commands on all other streams complete execution, and then record the timestamp.

While what we expect here is to get an event with the current time, hence using an otherwise unused stream. Fixing it might be outside the scope of this PR, but a possible solution could be to lazily have a stream specifically for recording submission time of events, tied to the context. Similar could be used in the CUDA backend to avoid the assumption noted above.

JackAKirk: @konradkusiak97 this sounds like the same thing as your hip event timings issue?

scripts/core/EXP-ENQUEUE-TIMESTAMP-RECORDING.rst
```diff
@@ -0,0 +1,70 @@
...
+    *   May be replaced, updated, or removed at any time.
+    *   Do not require maintaining API/ABI stability of their own additions over
+        time.
+    *   Do not require conformance testing of their own additions.
+
+
+Motivation
+--------------------------------------------------------------------------------
+Currently, the only way to get timestamp information is through enabling
+profiling on a queue and retrieve the information from events coming from
```
aarongreig, line 26:
```suggestion
profiling on a queue and retrieving the information from events coming from
```

scripts/core/exp-enqueue-timestamp-recording.yml
```diff
@@ -0,0 +1,66 @@
...
+            until the device timestamp recording command has executed on the
+            device.
+    - type: uint32_t
+      name: numEventsInWaitList
+      desc: "[in] size of the event wait list"
+    - type: "const $x_event_handle_t*"
+      name: phEventWaitList
+      desc: |
+            [in][optional][range(0, numEventsInWaitList)] pointer to a list of events that must be complete before the kernel execution.
+            If nullptr, the numEventsInWaitList must be 0, indicating that no wait event.
```
aarongreig, line 55:
```suggestion
            If nullptr, the numEventsInWaitList must be 0, indicating no wait events. 
```

aarongreig (approved):

source/adapters/hip/event.cpp
steffenlarsen: I have opened an issue about it here: https://github.com/intel/llvm/issues/12904.

steffenlarsen:
@oneapi-src/unified-runtime-native-cpu-write & @oneapi-src/unified-runtime-level-zero-write - Friendly ping.

PietroGhg (approved):
Native CPU lgtm, thank you

steffenlarsen:
Branch has been updated. @oneapi-src/unified-runtime-level-zero-write friendly ping.

steffenlarsen:
@oneapi-src/unified-runtime-level-zero-write Ping.

source/adapters/level_zero/event.cpp
```diff
@@ -587,6 +644,62 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventGetProfilingInfo(
...
+  UR_CALL(Queue->Context->getAvailableCommandList(
+      Queue, CommandList, UseCopyEngine, /* AllowBatching */ false));
+
+  UR_CALL(createEventAndAssociateQueue(
+      Queue, OutEvent, UR_COMMAND_TIMESTAMP_RECORDING_EXP, CommandList,
+      /* IsInternal */ false, /* HostVisible */ true));
+  ze_event_handle_t ZeEvent = (*OutEvent)->ZeEvent;
+  (*OutEvent)->WaitList = TmpWaitList;
+  (*OutEvent)->RecordEventStartTimestamp = DeviceStartTimestamp;
+
```
pbalcer, line 686: wouldn't it be better to query the `DeviceStartTimestamp` here? `getAvailableCommandList` is potentially a fairly expensive operation.

source/adapters/level_zero/event.cpp
```diff
@@ -587,6 +644,62 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventGetProfilingInfo(
...
+  // Allocate new entry in the queue's recordings.
+  uint64_t *EndTimestampPtr = (uint64_t *)malloc(sizeof(uint64_t));
```
pbalcer, line 688: does this need to be allocated? It should be possible to emplace an element into the map and then retrieve pointer to the value. Saves an extraneous heap allocation.

source/adapters/level_zero/queue.cpp
```diff
@@ -1323,6 +1352,8 @@ ur_result_t urQueueReleaseInternal(ur_queue_handle_t Queue) {
           }
   }
 
+  Queue->clearEndTimeRecordings();
```
pbalcer, line 1355: it should probably be inside of synchronize() where we are sure all prior events on that queue completed.

pbalcer (approved):
lgtm, just a few nits.

source/adapters/level_zero/event.cpp
steffenlarsen: My concern is that `std::map` doesn't guarantee that the entries don't move. I.e., if we point at the value of an element in it, an implementation could move the entry when a new entry is pushed into it. Even if most implementations don't do that, hitting an implementation that does would make for some painful debugging.

source/adapters/level_zero/queue.cpp
steffenlarsen: It is also done in `synchronize()`, but I don't see any guarantee that it is called before this function runs. Maybe I'm wrong?

source/adapters/level_zero/event.cpp
steffenlarsen: We need a command list here, in order to correctly enqueue the timestamp recording using `zeCommandListAppendWriteGlobalTimestamp`. `DeviceStartTimestamp` is the "now", while the `EdnTimestampPtr` is the "when the recording runs" in the queue, i.e. when all the work it depends on finishes.

source/adapters/level_zero/queue.cpp
pbalcer: It's called in `urQueueRelease` - https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/queue.cpp#L414.

I'm not sure whether it's possible for the queue to be unsynchronized in the urQueueReleaseInternal. urQueueRelease had to have been called already (to decrement the refcount).

source/adapters/level_zero/event.cpp
pbalcer: Right, my point is that to get more accurate duration value you want the start time to be as close to `zeCommandListAppendWriteGlobalTimestamp` as possible.

source/adapters/level_zero/queue.cpp
steffenlarsen: From what I can see, `urQueueReleaseInternal` is also called from within events: https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/event.cpp#L888

They don't seem to call `synchronize` beforehand, for all paths.

source/adapters/level_zero/event.cpp
steffenlarsen: I see! I have moved it down a bit. Good shout!

source/adapters/level_zero/event.cpp
steffenlarsen: Interesting! From what I can find about insertion as well, it seems that you are right. I have changed it and will rebase and rerun https://github.com/intel/llvm/pull/12838 ASAP.

source/adapters/level_zero/queue.cpp
pbalcer: Ok, there's no harm in calling it twice to be sure.
I think for the queue to be released in the event, it had to have been released by the user already.

pbalcer (approved):

nrspruit (approved):

kbenzie:
@steffenlarsen if you could pull in the latest changes from main and resolve the conflict I can get this merged.

steffenlarsen:
> @steffenlarsen if you could pull in the latest changes from main and resolve the conflict I can get this merged.

Thank you, @kbenzie ! Merge commit has been pushed and it should hopefully pass CI again.

source/adapters/level_zero/event.cpp
```diff
@@ -594,6 +650,62 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventGetProfilingInfo(
...
+) {
+  // Lock automatically releases when this goes out of scope.
+  std::scoped_lock<ur_shared_mutex> lock(Queue->Mutex);
+
+  ur_device_handle_t Device = Queue->Device;
+
+  bool UseCopyEngine = false;
+  _ur_ze_event_list_t TmpWaitList;
+  UR_CALL(TmpWaitList.createAndRetainUrZeEventList(
+      NumEventsInWaitList, EventWaitList, Queue, UseCopyEngine));
```
kbenzie, line 676: The interface of this member function has changed to add the event wait list arguments:

```suggestion
  UR_CALL(Queue->Context->getAvailableCommandList(
      Queue, CommandList, UseCopyEngine, NumEventsInWaitList, EventWaitList,
      /* AllowBatching */ false));
```

This should fix the build.

kbenzie (changes requested):

source/adapters/level_zero/event.cpp
steffenlarsen: Thank you, I was not aware! Was it a mistake making the suggestion to `createAndRetainUrZeEventList` or am I supposed to remove that call and refactor the use of `TmpWaitList`?

kbenzie: Sorry, looks like I've put the suggestion in the wrong place.

steffenlarsen: No problem, just needed to make sure. It's all green now! Thanks again. 😄

source/adapters/cuda/enqueue.cpp
```diff
@@ -1709,3 +1709,33 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueWriteHostPipe(
...
+    CUstream CuStream = hQueue->getNextComputeStream();
+
+    UR_CHECK_ERROR(enqueueEventsWait(hQueue, CuStream, numEventsInWaitList,
+                                     phEventWaitList));
+
+    RetImplEvent =
+        std::unique_ptr<ur_event_handle_t_>(ur_event_handle_t_::makeNative(
+            UR_COMMAND_TIMESTAMP_RECORDING_EXP, hQueue, CuStream));
+    UR_CHECK_ERROR(RetImplEvent->start());
+    UR_CHECK_ERROR(RetImplEvent->record());
```
hdelan, line 1730: Sorry for late review, but `start` and `record` record 3 separate native events. Perhaps it is useful to have a separate `EvQueued` and `EvStart` using this extension, but `EvEnd` should be unnecessary. Do you think it'd be worth having a `event::checkpoint` member func that minimizes the number of native events recorded?

hdelan: Thanks for response. Perhaps we can continue discussion here https://github.com/oneapi-src/unified-runtime/pull/1613

PR #1401, opened 2024-02-29
[UR][L0] Return device version based on DeviceIpVersion

ph0b:
When being queried a device version:
- CUDA adapter returns the compute capability version of the device,
- HIP adapter returns the architecture name.

Both of these are what define a GPU architecture to compile binaries for and can be used with GPU compilers, this is useful information for applications and developers.

On L0/Intel side, prior to this change, it returns the L0 API version. This information is already returned in BACKEND_RUNTIME_VERSION and DRIVER_VERSION queries.

With this change, we align the adapters behavior with CUDA and HIP, and return the Intel device architecture version that can be used with ocloc compiler.

https://github.com/intel/llvm/pull/13812

### DIFF
--- a/source/adapters/level_zero/device.cpp
+++ b/source/adapters/level_zero/device.cpp
@@ -1,6 +1,6 @@
 //===--------- device.cpp - Level Zero Adapter ----------------------------===//
 //
-// Copyright (C) 2023 Intel Corporation
+// Copyright (C) 2023-2024 Intel Corporation
 //
 // Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM
 // Exceptions. See LICENSE.TXT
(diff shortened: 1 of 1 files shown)


nrspruit:
How does this look between the level zero and the CUDA adapters now when it comes to the version?

Do we know if anything was relying or checking the API version from UR to determine which apis they were allowed to use? I know that direct customers of L0 were using this api version in that manner.

ph0b:
- CUDA example: "8.9"
- AMD example: "gfx1031"
- Intel example; "12.55.8" (previously, "1.3")

The good news with customer impact is that the minimum architecture version on Intel is 8.0.0, which is higher than 1.3, so if customers do check they have 1.3 or higher, it's ok... provided they correctly handle the major.min.rev versioning.

If not, returning only major.min for intel would be an option, we could even align all backends to 8.9, 10.3, 12.55 but that's a bigger breaking change, would be better suited for a real "architecture version" property, and we'd lose the actual device revision, which is needed info for using GPU compilers.

ph0b:
ci failed checking the license because adding -2024 to the header made the string go a bit further away from the beginning of the file, making the SPDX string end at char 301 while https://github.com/oneapi-src/unified-runtime/blob/908bb72865f59e7ac43bd25ebb1022510f8c1d99/scripts/verify_license.py#L15 reads the 300 first characters, you may want to slightly increase this value in the script.

ph0b:
@nrspruit now that https://github.com/oneapi-src/unified-runtime/pull/1462 is merged, can we restart the approval workflow ?

ph0b:
@nrspruit all checks have passed, is this scheduled to be merged?

kbenzie:
> @nrspruit all checks have passed, is this scheduled to be merged?

Tagging @oneapi-src/unified-runtime-level-zero-write for reviews as @nrspruit is away this week.

As for the merge process @ph0b once this is approved we'll add the *ready to merge* label and add it to our merge queue for integration into https://github.com/intel/llvm. Are there any time constrains we should be aware of?
Feel free to reach out to my via email if needed.

nrspruit (approved):
Sorry for the delay, I did not have a chance to get to this due to last minute issues I needed to review. This looks good to me, please proceed with this change.

PR #1403, opened 2024-03-01
[cuda][hip] Removed bad event wait constraint / use scopedContext correctly.

JackAKirk:
There is no reason to not allow waits from native hip/cuda events in different sycl "contexts".
Even the sycl spec does not mandate such a thing. This constraint does not appear to be applied in L0 backend.
This also fixes scopedContext to set the per device (CU/HIP)context.

This is a fix for https://github.com/intel/llvm/issues/6749#issuecomment-1971875971

### DIFF
--- a/source/adapters/cuda/event.cpp
+++ b/source/adapters/cuda/event.cpp
@@ -220,13 +220,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventSetCallback(ur_event_handle_t,
 UR_APIEXPORT ur_result_t UR_APICALL
 urEventWait(uint32_t numEvents, const ur_event_handle_t *phEventWaitList) {
   try {
-    auto Context = phEventWaitList[0]->getContext();
-    ScopedContext Active(Context);
+    ScopedContext Active(phEventWaitList[0]->getContext());
 
-    auto WaitFunc = [Context](ur_event_handle_t Event) -> ur_result_t {
+    auto WaitFunc = [](ur_event_handle_t Event) -> ur_result_t {
       UR_ASSERT(Event, UR_RESULT_ERROR_INVALID_EVENT);
-      UR_ASSERT(Event->getContext() == Context,
-                UR_RESULT_ERROR_INVALID_CONTEXT);
 
       return Event->wait();
     };
--- a/source/adapters/hip/event.cpp
+++ b/source/adapters/hip/event.cpp
@@ -190,14 +190,9 @@ urEventWait(uint32_t numEvents, const ur_event_handle_t *phEventWaitList) {
   UR_ASSERT(numEvents > 0, UR_RESULT_ERROR_INVALID_VALUE);
 
   try {
-
-    auto Context = phEventWaitList[0]->getContext();
     ScopedContext Active(phEventWaitList[0]->getDevice());
-
-    auto WaitFunc = [Context](ur_event_handle_t Event) -> ur_result_t {
+    auto WaitFunc = [](ur_event_handle_t Event) -> ur_result_t {
       UR_ASSERT(Event, UR_RESULT_ERROR_INVALID_EVENT);
-      UR_ASSERT(Event->getContext() == Context,
-                UR_RESULT_ERROR_INVALID_CONTEXT);
 
       return Event->wait();
     };


source/adapters/cuda/event.cpp
```diff
@@ -220,13 +220,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventSetCallback(ur_event_handle_t,
...
   try {
-    auto Context = phEventWaitList[0]->getContext();
-    ScopedContext Active(Context);
-
-    auto WaitFunc = [Context](ur_event_handle_t Event) -> ur_result_t {
+    auto WaitFunc = [](ur_event_handle_t Event) -> ur_result_t {
       UR_ASSERT(Event, UR_RESULT_ERROR_INVALID_EVENT);
-      UR_ASSERT(Event->getContext() == Context,
-                UR_RESULT_ERROR_INVALID_CONTEXT);
+      ScopedContext Active(Event->getContext());
```
hdelan, line 225: I don't think we need to set the context every time this lambda is run. It's valid to wait on an event that was recorded in another context, so setting the context just once at the start of the `try catch` should be enough

hdelan (commented):
Think we should only have a context set once per `urEventWait`, not once per event

JackAKirk:
> Think we should only have a context set once per `urEventWait`, not once per event

urEventWait waits on a list of events which can be associated with different device CUcontexts. This is the reason for the error that motivated this patch. If you only set the context once it can generally be set for a CUcontext that doesn't correspond to the CUcontext used for the stream that the event is associated with.

source/adapters/cuda/event.cpp
JackAKirk: Is it valid to wait for an event that was set in another CUcontext (on a different device)?
I assumed this was not the case.

hdelan:
> urEventWait waits on a list of events which can be associated with different device CUcontexts. This is what the code was doing (using different devices) that led to the error that motivated this patch. If you only set the context once it could generally be set for a CUcontext that doesn't correspond to the CUcontext used for the stream that the event is associated with.

These are docs for the CUDA runtime which says we don't need to set a context to use `cudaEventSynchronize`. 
https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#stream-and-event-behavior

I think it must be the same for the driver API as well since if we have a `CUevent` associated with context A and a stream associated with context B, since in the CUDA runtime it says we can successfully call `cudaStreamWaitEvent`, this must also be possible in the CUDA driver since there is no way to emulate having two active contexts to make this call. 

However this is sort of on the fringes of documentation so I think we should run some tests to make sure that this is OK in cuda driver API.

JackAKirk:
> > urEventWait waits on a list of events which can be associated with different device CUcontexts. This is what the code was doing (using different devices) that led to the error that motivated this patch. If you only set the context once it could generally be set for a CUcontext that doesn't correspond to the CUcontext used for the stream that the event is associated with.
> 
> These are docs for the CUDA runtime which says we don't need to set a context to use `cudaEventSynchronize`. https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#stream-and-event-behavior
> 
> I think it must be the same for the driver API as well since if we have a `CUevent` associated with context A and a stream associated with context B, since in the CUDA runtime it says we can successfully call `cudaStreamWaitEvent`, this must also be possible in the CUDA driver since there is no way to emulate having two active contexts to make this call.
> 
> However this is sort of on the fringes of documentation so I think we should run some tests to make sure that this is OK in cuda driver API.

Yeah makes sense, nice find. I'll run some tests.

JackAKirk:
> > > urEventWait waits on a list of events which can be associated with different device CUcontexts. This is what the code was doing (using different devices) that led to the error that motivated this patch. If you only set the context once it could generally be set for a CUcontext that doesn't correspond to the CUcontext used for the stream that the event is associated with.
> > 
> > 
> > These are docs for the CUDA runtime which says we don't need to set a context to use `cudaEventSynchronize`. https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#stream-and-event-behavior
> > I think it must be the same for the driver API as well since if we have a `CUevent` associated with context A and a stream associated with context B, since in the CUDA runtime it says we can successfully call `cudaStreamWaitEvent`, this must also be possible in the CUDA driver since there is no way to emulate having two active contexts to make this call.
> > However this is sort of on the fringes of documentation so I think we should run some tests to make sure that this is OK in cuda driver API.
> 
> Yeah makes sense, nice find. I'll run some tests.

Made some multidevice/multiqueue tests and they pass fine on nvidia/hip systems so I think you must be right. I've made the change you suggested.

hdelan:
> Made some multidevice/multiqueue tests and they pass fine on nvidia/hip systems so I think you must be right. I've made the change you suggested.

Great thanks

hdelan (approved):

PR #1405, opened 2024-03-01
[conformance] Don't give tests latitude to segfault

ldrumm:
There is no point in running a test if it's allowed to segfault (unless maybe we're testing an operating system kernel). If a test can optionally segfault, it's worse than not running the test at all and only serves the illusion of test coverage and progress.


ldrumm:
FYI this patch is here to show us what's broken so we have visibility on these problems. Thus I'll link tickets here that are relevant to remaining failures

npmiller:
Note: this has been resolved for CUDA and HIP, but there are still some segfaults for Level Zero and Native CPU

RossBrunton:
@ldrumm We recently updated our testing framework so that segfaults (and things like abort) are now detected and treated as "regular" failures in the match files.

Does that solve the issue this MR was created to solve?

RossBrunton:
@ldrumm 

https://github.com/oneapi-src/unified-runtime/pull/2199/files
https://github.com/oneapi-src/unified-runtime/pull/2220/files

Basically, the test runner now runs each failing test in its own invocation, so that if it segfaults then it doesn't bring down the rest of the tests.

ldrumm:
> @ldrumm
> 
> https://github.com/oneapi-src/unified-runtime/pull/2199/files https://github.com/oneapi-src/unified-runtime/pull/2220/files
> 
> Basically, the test runner now runs each failing test in its own invocation, so that if it segfaults then it doesn't bring down the rest of the tests.

Thanks Ross. Looks great. I'll close this now

PR #1408, opened 2024-03-04
[CI] Skip E2E tests on forks

lukaszstolarczuk:
Added missing skip for HW-related (E2E) jobs. They will always fail on forks without this `if`.


pbalcer (approved):

wlemkows (approved):

lukaszstolarczuk:
Ok, so apparently e2e workflows don't work currently per PR comment, only work on nightly schedule. I'll try and fix that in a separate PR.

PR #1409, opened 2024-03-05
[CTS] Add CTS tests for urMemImageCreate entry-point

omarahmed1111:
This PR:  
* Adds CTS tests for different image types.
* Adds CTS tests for different image formats
* Defines the primary and optional supported image formats.
* Deletes redundant `ERROR_IMAGE_FORMAT_NOT_SUPPORTED` error code.
* Deletes UR_MEM_TYPE_BUFFER as it should be redundant to `urMemBufferCreate` entry-point.
* Deletes `URMEM_TYPE_IMAGE1D_BUFFER` which is only used in opencl and  opencl will need cl_image_desc.buffer/mem_object value when `CL_MEM_OBJECT_IMAGE1D_BUFFER` is used, and we don't have a way in urMemImageCreate to pass it a buffer object that it will use to initialize an image with it and use this type. And if we initialized a new buffer inside the entry-point, we don't have a way of clearing the buffer memory at the end.
* Make sure the adapters return the right error-codes for `urMemImageCreate` entry-point.

These changes moved on its own PR based on this [discussion](https://github.com/oneapi-src/unified-runtime/pull/1394#discussion_r1510958287)

[Intel/llvm testing](https://github.com/intel/llvm/pull/13058)


scripts/core/memory.yml
```diff
@@ -213,6 +209,15 @@ name: ImageCreate
...
     - "**clCreateImage**"
+details:
+    - |
+      The primary $x_image_format_t that must be supported by all the adapters are {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_UNORM_INT8}, 
+      {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_UNORM_INT16}, {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_SNORM_INT8}, 
+      {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_SNORM_INT16}, {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_SIGNED_INT8}, 
+      {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_SIGNED_INT16}, {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_SIGNED_INT32}, 
+      {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_UNSIGNED_INT8}, {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_UNSIGNED_INT16}, 
+      {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_UNSIGNED_INT32}, {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_HALF_FLOAT}, 
+      {UR_IMAGE_CHANNEL_ORDER_RGBA, UR_IMAGE_CHANNEL_TYPE_FLOAT}.
```
omarahmed1111, line 220: This is a good candidate to be added in a table but not sure if that would be simple or complicated.

omarahmed1111, line 220: This list was picked by looking at the intersection of the supported image formats on the current adapters except native_cpu.

omarahmed1111: We could also use this sycl spec defined image format list in [here](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_sampled_image_interface) but not sure if we should base our main formats on this?

isaacault (approved):
`source/adapters/*/image.cpp` LGTM.

konradkusiak97 (approved):
`source/adapters/cuda|hip` LGTM

nrspruit (approved):
LGTM for level zero

kbenzie (approved):

test/conformance/memory/urMemImageCreateWithImageFormatParam.cpp
```diff
@@ -0,0 +1,120 @@
+// Copyright (C) 2023 Intel Corporation
```
RossBrunton, line 1: Should be 2024

RossBrunton (approved):

omarahmed1111:
@oneapi-src/unified-runtime-command-buffer-write Could I have a review on that when available, Thanks!

test/conformance/memory/urMemImageCreateWithImageFormatParam.cpp
omarahmed1111: Fixed

PR #1410, opened 2024-03-05
[CMake] Support external adapter source dirs

kbenzie:
Allow the configuration of external source directory for any of the existing adapters. This is done via a CMake cache variable which if set will is used as the `source_dir` argument to `add_subdirectory()`, the `binary_dir` argument will also be set to the same location as a normal in-tree build. This is done via the addition of the new `add_ur_adapter_subdirectory()` helper function which also hoists the logic to enable/disable each adapter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,16 @@ set(UR_SYCL_LIBRARY_DIR "" CACHE PATH
 set(UR_CONFORMANCE_TARGET_TRIPLES "" CACHE STRING
     "List of sycl targets to build CTS device binaries for")
 set(UR_CONFORMANCE_AMD_ARCH "" CACHE STRING "AMD device target ID to build CTS binaries for")
+set(UR_ADAPTER_LEVEL_ZERO_SOURCE_DIR "" CACHE PATH
+    "Path to external 'level_zero' adapter source dir")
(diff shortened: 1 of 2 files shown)


pbalcer (commented):
Makes sense. I'm wondering though whether it wouldn't be simpler to disable building the adapter in the main UR clone, and then add/build just that adapter/target from the adapter-specific clone. This would allow us to contain this logic to just SYCL CMake.

kbenzie:
> Makes sense. I'm wondering though whether it wouldn't be simpler to disable building the adapter in the main UR clone, and then add/build just that adapter/target from the adapter-specific clone. This would allow us to contain this logic to just SYCL CMake.

The adapter subdirectories use CMake utilities so to actually support that would be a more invasive change to either remove those uses or add `include`s. This has the potential to also be quite brittle if/when an adapter unwittingly introduces an incompatible change in that CMake use.

My main goal with this approach is to ensure that there is one source of truth for the UR API while still enabling decoupling of forward progress of the adapter implementations. By hooking a this point I feel we have the smallest chance of something unintentially breaking another adapter. Or at least if an adapter does break, should be the one using this mechanism rather than one of the other adapters. It should also cause the build to fail and thus never merge.

pbalcer (approved):

callumfare (approved):

PR #1411, opened 2024-03-05
[cmake] Always generate compile_commands.json

ldrumm:
I can't think of a valid reason not to generate compile_commands.json and several reasons why we should

1. I don't like typing configure options that much.
2. I like editor completion to work without me having to regenerate my build.
3. It's a good way to inspect final build flags without having to query the build system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,9 @@ set(UR_CONFORMANCE_TARGET_TRIPLES "" CACHE STRING
     "List of sycl targets to build CTS device binaries for")
 set(UR_CONFORMANCE_AMD_ARCH "" CACHE STRING "AMD device target ID to build CTS binaries for")
 
+# There's little reason not to generate the compile_commands.json
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 include(Assertions)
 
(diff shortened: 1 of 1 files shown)


kbenzie:
CMake will initialize `CMAKE_EXPORT_COMPILE_COMMANDS` from an environment variable of the same name: https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html

Perhaps adding this to your shell init would best fit your needs?

ldrumm:
> CMake will initialize `CMAKE_EXPORT_COMPILE_COMMANDS` from an environment variable of the same name: https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html
> 
> Perhaps adding this to your shell init would best fit your needs?

That would work too. I just prefer a zero-configuration approach, and LLVM and CA set this by default so I figured that might be precedent enough.

Either way, I'm not particularly fussed, but I like ergonomics without extra configuration.

kbenzie:
> That would work too. I just prefer a zero-configuration approach, and LLVM and CA set this by default so I figured that might be precedent enough.

Given the env var support is relatively new this was the only option prior to CMake 3.17

> Either way, I'm not particularly fussed, but I like ergonomics without extra configuration.

This won't cause any trouble to include, was just pointing this out as an alternative.

kbenzie (approved):

PR #1421, opened 2024-03-07
Introduce a utility to walk polymorphic linked lists

ldrumm:
Add `find_stype_node` for walking polymorphic linked lists looking for a particular type.

The implementation here works with an auto-generated compile time map, that links a given type to the structure enumeration tag. The implementation simply walks the list looking for the .stype implied by the template parameter type and casts it to the expected type, then returning it.

It's one of those unfortunate cases where you can write pretty nasty implementation code that makes the user code much much nicer. In this case the user doesn't need to worry about the `.types` at all, and the const-void casts can be eliminated from user code.


source/common/ur_util.hpp
```diff
@@ -266,6 +266,59 @@ inline ur_result_t exceptionToResult(std::exception_ptr eptr) {
...
+/// if (!node)
+///   printf("node of expected type not found!\n");
+/// ```
+///
+/// To find multiple nodes of a given type:
+///
+/// ```cpp
+/// std::vector<ur_usm_host_desc_t> nodes;
+///
+/// for (auto *node = find_stype_node<ur_usm_host_desc_t>(mylist); node; node =
```
pbalcer, line 299: Doesn't this assume that all the elements following the first `ur_usm_host_desc_t` are of that same type?

ldrumm: This snippet is an example which finds all `ur_usm_host_desc_t` in a linked list. It's artificial and probably would never occur in real code. The linked list in question can have nodes of any type

pbalcer: Right, but the comment and the code is wrong, because, as you note, the list can have nodes of any types. So, if I read this right, this will find the first node of the type `ur_usm_host_desc_t` and then assume everything that comes after is is also `ur_usm_host_desc_t`. Or I'm missing something?

ldrumm: e.g.
in the following imaginary list
```
(a) ur_usm_alloc_location_desc_t -> (b )ur_usm_host_desc_t -> (c) ur_usm_device_desc_t -> (d) ur_usm_host_desc_t
```

The example loop above will push (b) and (d) onto the `nodes` vector

ldrumm: I see what you're saying.
I guess there technically needs to be a cast in the example, but given how these lists work in practise, I don't think we need to care about type-safety in example code

pbalcer: > e.g. in the following imaginary list
> 
> ```
> (a) ur_usm_alloc_location_desc_t -> (b )ur_usm_host_desc_t -> (c) ur_usm_device_desc_t -> (d) ur_usm_host_desc_t
> ```
> 
> The example loop above will push (b) and (d) onto the `nodes` vector

I don't think that's true. It would push (b), (c) an (d) onto the vector, incorrectly casting (c) to `ur_usm_host_desc_t`.

ldrumm: I see the bug!
+/// ```cpp
+/// std::vector<ur_usm_host_desc_t> nodes;
+///
+/// for (ur_usm_host_desc_t *node; node = find_stype_node<ur_usm_host_desc_t>(mylist); node; node = static_cast<ur_usm_host_desc_t>(node->pNext)) {
+///   nodes.push_back(node);
+/// }

pbalcer: this would be an infinite loop because you never update mylist ;)

Maybe this would be simpler:
```
auto *node = mylist;
while (node = find_stype_node<ur_usm_host_desc_t>(node)) {
  ...
  node = node->pNext; // potentially casts incorrectly from void *. Not sure if this isn't UB.
}
```

This is surprisingly tricky given the `find_stype_node` as written :D

ldrumm: OH jeez. Perhaps I should just remove this example, seeing as I can't get it right, and it's a very unlikely use-case?

pbalcer: Sure, that's the simplest solution ;)

ldrumm: My brain isn't working this morning, clearly. Thanks for sticking with me!

pbalcer (approved):

ldrumm:
@oneapi-src/unified-runtime-maintain please merge

PR #1428, opened 2024-03-11
Add match file entries for inconsistent L0 segfaults on clang.

aarongreig:


pbalcer:
This appears to be a problem with packages on Ubuntu 22.04, and there's little we can do apart from maybe reverting to some older version of level-zero packages. For now let's add the segfaults to match files.

PR #1429, opened 2024-03-11
[L0] Support for urUsmP2PPeerAccessGetInfoExp to query p2p access info

nrspruit:


source/adapters/level_zero/usm_p2p.cpp
```diff
@@ -35,11 +35,36 @@ UR_APIEXPORT ur_result_t UR_APICALL urUsmP2PPeerAccessGetInfoExp(
...
-  std::ignore = peerDevice;
-  std::ignore = propName;
-
   UrReturnHelper ReturnValue(propSize, pPropValue, pPropSizeRet);
-  // Zero return value indicates that all of the queries currently return false.
-  return ReturnValue(uint32_t{0});
+
+  bool propertyValue = false;
+  switch (propName) {
+  case UR_EXP_PEER_INFO_UR_PEER_ACCESS_SUPPORTED: {
```
pbalcer, line 42: If `zeDeviceCanAccessPeer` returns the same information as `ZE_DEVICE_P2P_PROPERTY_FLAG_ACCESS`, this can be simplified to:
```
ze_device_p2p_property_flags_t ZeFlag;
switch (propName) {
  case UR_EXP_PEER_INFO_UR_PEER_ACCESS_SUPPORTED: {
    ZeFlag = ZE_DEVICE_P2P_PROPERTY_FLAG_ACCESS;
  } break;
  case UR_EXP_PEER_INFO_UR_PEER_ATOMICS_SUPPORTED: {
    ZeFlag = ZE_DEVICE_P2P_PROPERTY_FLAG_ATOMICS;
  } break;
  default: return UR_RESULT_ERROR_INVALID_ENUMERATION;
}
    
ze_device_p2p_properties_t p2pProperties;
ZE2UR_CALL(zeDeviceGetP2PProperties, (commandDevice->ZeDevice, peerDevice->ZeDevice, &p2pProperties));
propertyValue = p2pProperties.flags & ZeFlag;

return ReturnValue(propertyValue);
```

nrspruit: So, ZE_DEVICE_P2P_PROPERTY_FLAG_ACCESS reports if the device has the ability to be accessed by any peer (ie HW capability), but zeDeviceCanAccessPeer verifies that command and peer device can access each other's resources (ie is ready and stable).

We want to do both to ensure that we report true only when the access to that device is supported by the specific set of devices.

pbalcer: Then I don't think this is the right interface. The [CUDA adapter](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/cuda/usm_p2p.cpp#L53) returns the "capability", not "readiness" for p2p access. And they also do have another interface `cudaCanAccessPeer`. I'm not sure whether there's a difference of that platform though. But if there is, we probably should have a dedicated API for both.

@JackAKirk ping

pbalcer: Also, the UR interface is called `UR_EXP_PEER_INFO_UR_PEER_ACCESS_SUPPORTED` which suggest that it queries the capability.

nrspruit: > Also, the UR interface is called `UR_EXP_PEER_INFO_UR_PEER_ACCESS_SUPPORTED` which suggest that it queries the capability.

That is the reason why for L0 I query both the property which is  `ZE_DEVICE_P2P_PROPERTY_FLAG_ACCESS` And the can access peer because it should encompass both to satisfy `UR_EXP_PEER_INFO_UR_PEER_ACCESS_SUPPORTED` which suggests that it confirms that it is supported.

pbalcer: Ok, if in cuda `cudaCanAccessPeer` is the same as `CU_DEVICE_P2P_ATTRIBUTE_ACCESS_SUPPORTED`, then I think this implementation makes sense. Thanks.

pbalcer (approved):

raiyanla (approved):

PR #1432, opened 2024-03-12
[CTS] Don't segfault when failing to load a kernel

RossBrunton:
If LoadSource fails, we don't exit immediately and continue running the test. This then attempts to dereference a null shared pointer and segfaults.

This by itself won't fix buggy test environments, but will make it so that they fail the test with a sensible message rather than crashing everything.


aarongreig (approved):

kbenzie:
Looks like the native cpu match files need updated after this bug fix.

PR #1442, opened 2024-03-14
[EXP][Command-Buffer] Optimize L0 command buffer submission

Bensuo:
- Adds command buffer property to explicitly enable profiling
- Add ability to enforce use of in-order command lists

Draft DPC++ PR: https://github.com/intel/llvm/pull/13565

### DIFF
--- a/scripts/core/EXP-COMMAND-BUFFER.rst
+++ b/scripts/core/EXP-COMMAND-BUFFER.rst
@@ -77,9 +77,13 @@ Command-Buffer Creation
 Command-Buffers are tied to a specific ${x}_context_handle_t and
 ${x}_device_handle_t. ${x}CommandBufferCreateExp optionally takes a descriptor
 to provide additional properties for how the command-buffer should be
-constructed. The only unique member defined in ${x}_exp_command_buffer_desc_t
-is ``isUpdatable``, which should be set to ``true`` to support :ref:`updating
+constructed. The members defined in ${x}_exp_command_buffer_desc_t are:
+* ``isUpdatable``, which should be set to ``true`` to support :ref:`updating
 command-buffer commands`.
+* ``isInOrder``, which should be set to ``true`` to enable commands enqueued to
+a command-buffer to be executed in an in-order fashion where possible.
+* ``enableProfiling``, which should be set to ``true`` to enable profiling of
+the command-buffer.
 
 Command-buffers are reference counted and can be retained and released by
 calling ${x}CommandBufferRetainExp and ${x}CommandBufferReleaseExp respectively.
@@ -123,7 +127,8 @@ Sync-Points
 A sync-point is a value which represents a command inside of a command-buffer
 which is returned from command-buffer append function calls. These can be
 optionally passed to these functions to define execution dependencies on other
-commands within the command-buffer.
+commands within the command-buffer. Sync-points passed to functions may be
+ignored if the command-buffer was created in-order.
 
 Sync-points are unique and valid for use only within the command-buffer they
 were obtained from.
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -15,17 +15,44 @@
 https://github.com/intel/llvm/blob/sycl/sycl/doc/design/CommandGraph.md#level-zero
 */
 
+namespace {
+/// Checks the version of the level-zero driver.
+/// @param Context Execution context
+/// @param VersionMajor Major verion number to compare to.
+/// @param VersionMinor Minor verion number to compare to.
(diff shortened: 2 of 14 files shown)


scripts/core/EXP-COMMAND-BUFFER.rst
```diff
@@ -77,9 +77,12 @@ Command-Buffer Creation
...
 ${x}_device_handle_t. ${x}CommandBufferCreateExp optionally takes a descriptor
 to provide additional properties for how the command-buffer should be
-constructed. The only unique member defined in ${x}_exp_command_buffer_desc_t
-is ``isUpdatable``, which should be set to ``true`` to support :ref:`updating
+constructed. The members defined in ${x}_exp_command_buffer_desc_t are:
+* ``isUpdatable``, which should be set to ``true`` to support :ref:`updating
 command-buffer commands`.
+* ``isInOrder``, which should be set to ``true`` to enable the graph workload
+  to be submitted to an in-order command-list.
+* ``enableProfiling``, which should be set to ``true`` to enable graph profling.
```
EwanC, line 85: Think we could improve the language here a bit:
* `graph workload` - We probably shouldn't be talking about graphs at this abstraction layer, would say "command-buffer" workload.
* `command-list` - "command-list" is L0 terminology, should be in terms of "command-buffer".
* `graph profling.` - "profiling of the command-buffer"

source/adapters/level_zero/command_buffer.hpp
```diff
@@ -82,6 +82,14 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
   bool IsUpdatable = false;
   // Indicates if command buffer was finalized.
   bool IsFinalized = false;
+  // Command-buffer profiling is enabled.
+  bool IsProfilingEnabled = false;
+  // Command-buffer can be submitted to an in-order command-list.
+  bool IsInOrderCmdList;
```
EwanC, line 88: To be consistent with other members
```suggestion
  bool IsInOrderCmdList = false;
```

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -251,27 +303,33 @@ static ur_result_t enqueueCommandBufferMemCopyHelper(
...
-                                  SyncPointWaitList, ZeEventList));
-
-  ur_event_handle_t LaunchEvent;
-  UR_CALL(
-      EventCreate(CommandBuffer->Context, nullptr, false, false, &LaunchEvent));
-  LaunchEvent->CommandType = CommandType;
+  if (CommandBuffer->IsInOrderCmdList) {
+    ZE2UR_CALL(
+        zeCommandListAppendMemoryCopy,
+        (CommandBuffer->ZeCommandList, Dst, Src, Size, nullptr, 0, nullptr));
```
EwanC, line 310: wondering if it's an issue in these code paths than we never set the return `SyncPoint` parameter to the entry-point, we could maybe set it but not register it with the command-buffer

Bensuo: Hmm, returning a sync-point value that is silently ignored seems a little unintuitive to me. 

Perhaps the issue is that we don't have a way to report that it is unused/not set? Since sync points are a typedef for `uint32_t` and `0` is a valid value for a sync point. If it was signed we could return `-1` to signify the sync point is unused, or swap the typedef for a simple struct/class that could indicate whether it is a valid sync point. This might be overkill though for this issue.

If we are ignoring sync points passed to these entry points it might also be better to make that explicit and return an error if the user tries to pass sync points when appending to an in-order command-buffer? That would make it more explicit but also increase the burden on the user side whether to store/pass the sync points or not.

EwanC: > If we are ignoring sync points passed to these entry points it might also be better to make that explicit and return an error if the user tries to pass sync points when appending to an in-order command-buffer? That would make it more explicit but also increase the burden on the user side whether to store/pass the sync points or not.

Agree with adding some spec wording as a solution, would make the obligations on caller/impl explicit. Would probably also be okay with the spec saying an in-order command-buffer will ignore any sync point parameters rather than making it a hard error if that makes some usage code simpler.

EwanC (approved):
LGTM

aarongreig (approved):

nrspruit (approved):
LGTM, @raiyanla , can you give this a check since you authored the in order lists changes?

raiyanla (approved):

EwanC (commented):
There's conflicts that needs resolved in the rebase of this PR by using replacing `urPrint` with `logger::debug`

Bensuo:
Updated DPC++ PR to remove dependency on SYCL runtime changes.

Issue #1362, opened 2024-02-19
CL adapter: add consistent and useful fail state if entry point extension isn't supported.


aarongreig:
We have a few entry point implementations in the CL adapter that rely on functions from OpenCL extensions, such as the [USM](https://github.com/oneapi-src/unified-runtime/blob/3fd11f1d23e7634fd7a847074a4e752273a19223/source/adapters/opencl/usm.cpp#L98) functionality. Right now if we find that the CL extension function needed for an entry point isn't available we [return UR_RESULT_ERROR_INVALID_VALUE](https://github.com/oneapi-src/unified-runtime/blob/3fd11f1d23e7634fd7a847074a4e752273a19223/source/adapters/opencl/common.hpp#L404), which isn't helpful and doesn't fit with the few instances where we have a spec definition of what should happen in these cases (for instance the UR_RESULT_ERROR_INVALID_OPERATION return from [usUSMHostAlloc](https://oneapi-src.github.io/unified-runtime/core/api.html#urusmhostalloc)).

We should for sure just change the return code from that helper function, and I think this should be accompanied by a spec pass to make sure that, wherever appropriate, we have a return code defined for when a backend doesn't support an entry point with an associated device info query. For example the [host pipe functions](https://oneapi-src.github.io/unified-runtime/core/api.html#urenqueuewritehostpipe) are similar to the USM ones in that there is an associated device info query for support, but their spec doesn't define the error code that should be returned if that support isn't there.

alycm:
Discussed 2024-02-28: An error code in this case makes sense.

PR #1448, opened 2024-03-15
[OpenCL] Make extension function lookup return unusupported error

steffenlarsen:
This commit changes the error code of failing to look up extension pointers in OpenCL return UR_RESULT_ERROR_UNSUPPORTED_FEATURE instead of UR_RESULT_ERROR_INVALID_VALUE. Likewise, some of the places this lookup was used, error would return UR_RESULT_ERROR_INVALID_OPERATION. The motivation behind this change is to make the error codes for cases where extensions are not supported for the OpenCL target clearer.


source/adapters/opencl/command_buffer.cpp
```diff
@@ -26,8 +26,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferCreateExp(
           CLContext, cl_ext::ExtFuncPtrCache->clCreateCommandBufferKHRCache,
           cl_ext::CreateCommandBufferName, &clCreateCommandBufferKHR);
 
-  if (!clCreateCommandBufferKHR || Res != CL_SUCCESS)
-    return UR_RESULT_ERROR_INVALID_OPERATION;
+  if (Res != CL_SUCCESS)
+    return Res;
```
aarongreig, line 30: this is returning a CL error code where there should be a UR one, instead we should remove this altogether and wrap the `getExtFunc` calls in `UR_RETURN_ON_FAILURE` (applies elsewhere)

aarongreig:
relates to #1362

source/adapters/opencl/command_buffer.cpp
steffenlarsen: Good eye! This should have been addressed.

EwanC (approved):

steffenlarsen:
> relates to #1362

Interesting! This was an ambiguity we ran into when debugging an issue with an unexpected device being picked up in DPC++. I have made a patch to also make a mapping for the unsupported error in intel/llvm as well: https://github.com/intel/llvm/pull/13036

The hope is that this will make it easier to differentiate when the backend is unhappy with a call and when it is simply just not supporting it.

source/adapters/opencl/enqueue.cpp
```diff
@@ -409,21 +403,21 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueReadHostPipe(
...
-  ur_result_t RetVal =
+  UR_RETURN_ON_FAILURE(
       cl_ext::getExtFuncFromContext<cl_ext::clEnqueueReadHostPipeINTEL_fn>(
           CLContext, cl_ext::ExtFuncPtrCache->clEnqueueReadHostPipeINTELCache,
-          cl_ext::EnqueueReadHostPipeName, &FuncPtr);
+          cl_ext::EnqueueReadHostPipeName, &FuncPtr));
 
   if (FuncPtr) {
-    RetVal = mapCLErrorToUR(
+    CL_RETURN_ON_FAILURE(mapCLErrorToUR(
```
aarongreig, line 412:
```suggestion
    CL_RETURN_ON_FAILURE(
```
`CL_RETURN_ON_FAILURE` does `mapCLErrorToUR` for you (same [here](https://github.com/oneapi-src/unified-runtime/pull/1448/files#diff-c80bbd3e96dbcb2b2ca1d99af816a310b1720decee46fdb0e97d5fd3075e5cb3R444))

aarongreig (approved):

steffenlarsen:
@kbenzie - I think this is ready.

hvdijk:
Hi, there were some checks for functions that were followed by a `if (FuncPtr)` check and properly handled the optional function not being implemented. It looks like they now return an error for that, is that right? If I am reading this right (and I may well not, please correct me if so) this change basically means that OpenCL implementations other than Intel's own can no longer be used. I'm seeing this with OCK where we now get e.g. when running SYCL-CTS
```
---> piMemBufferCreate(
	<unknown> : 0xa204c80
	<unknown> : 9
	<unknown> : 40
	<unknown> : 0x7ffff86c0910
	<unknown> : 0x7ffff86bf5a0
	<unknown> : 0
) ---> 	pi_result : -995
	[out]void * : 0x7ffff86c0910
	[out]pi_mem * : 0x7ffff86bf5a0[ 0 ... ]


~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
The SYCL 2020 Conformance Test Suite is a Catch2 v3.2.1 host application.
Run with -? for options

-------------------------------------------------------------------------------
requisite for the entire underlying buffer for sycl::accessor 
-------------------------------------------------------------------------------
/builds/ComputeAorta/ci-github/SYCL-CTS/tests/accessor/accessor_requisite_entire_buffer.cpp:32
...............................................................................

/builds/ComputeAorta/ci-github/SYCL-CTS/tests/accessor/accessor_requisite_entire_buffer.cpp:32: FAILED:
due to unexpected exception with message:
  SYCL exception
  with category name: 'sycl'
  with code: 'runtime'
  with code raw value: '1'
  with code message: 'SYCL Error'
  with what: 'Native API failed. Native API returns: -995 (The plugin or device
  does not support the called function) -995 (The plugin or device does not
  support the called function)'

===============================================================================
test cases: 1 | 1 failed
assertions: 1 | 1 failed
```

coldav:
note that OCK was passing (fast) SYCL CTS before this point, so it suggests that perhaps we have made something optional an error.

steffenlarsen:
Thank you for letting me know, @hvdijk & @coldav ! You are absolutely right, the changes to {{urMemBufferCreate}} were not right. They can be safely reverted, as done in https://github.com/oneapi-src/unified-runtime/pull/1496.

PR #1450, opened 2024-03-15
[CL] Atomic fence scope capability for Intel FPGA driver

kbenzie:
This patch implements a workaround for the Intel FPGA driver which is currently an OpenCL 1.2 device which also supports `UR_MEMORY_SCOPE_CAPABILITY_FLAG_DEVICE` rather than the baseline for OpenCL 1.2 devices.

The workaround unconditionally queries for `CL_DEVICE_ATOMIC_FENCE_CAPABILITIES`, if it is successful the value is used. Otherwise the error code is ignored and the baseline capabilities are returned.

### DIFF
--- a/source/adapters/opencl/device.cpp
+++ b/source/adapters/opencl/device.cpp
@@ -569,6 +569,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     return ReturnValue(
         static_cast<ur_memory_order_capability_flags_t>(URCapabilities));
   }
+
   case UR_DEVICE_INFO_ATOMIC_MEMORY_SCOPE_CAPABILITIES: {
     /* Initialize result to minimum mandated capabilities according to
      * SYCL2020 4.6.3.2. Because scopes are hierarchical, wider scopes support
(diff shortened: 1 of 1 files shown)


source/adapters/opencl/device.cpp
```diff
@@ -563,6 +563,24 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
...
+                          CL_DEVICE_NAME, 0, nullptr, &deviceNameSize));
+      std::string deviceName(deviceNameSize, '\0');
+      CL_RETURN_ON_FAILURE(clGetDeviceInfo(
+          cl_adapter::cast<cl_device_id>(hDevice), CL_DEVICE_NAME,
+          deviceNameSize, deviceName.data(), nullptr));
+      if (deviceName.find("Intel(R) FPGA SDK for OpenCL(TM)") !=
+              std::string::npos ||
+          deviceName.find("Intel(R) FPGA Emulation Platform for OpenCL(TM)") !=
+              std::string::npos) {
+        URCapabilities = UR_MEMORY_SCOPE_CAPABILITY_FLAG_DEVICE;
```
aarongreig, line 582: should maybe be
```suggestion
        URCapabilities |= UR_MEMORY_SCOPE_CAPABILITY_FLAG_DEVICE;
```

kbenzie: Thanks, I refactored this a bit because I implemented the wrong workaround. I've made sure to `|=` the value. Please check again.

aarongreig (approved):

PR #1451, opened 2024-03-15
[CTS] Modify various urEnqueueMemBuffer* conformance tests to use parameterized fixtures

martygrant:
Added new `urMemBufferQueueTestWithParam` template fixture to accept parameters so different values for the buffer size and type can be easily changed. This is specifically for `urEnqueueMemBuffer*` tests.

TODO: Will need to add new match file entries for the parameterized versions of the tests so CI passes.


test/conformance/enqueue/urEnqueueMemBufferMap.cpp
```diff
@@ -2,12 +2,52 @@
...
+static std::vector<uur::mem_buffer_test_parameters_t> test_parameters{
+    {8, UR_MEM_FLAG_READ_WRITE},
+    {16, UR_MEM_FLAG_READ_WRITE},
+    {32, UR_MEM_FLAG_READ_WRITE},
+    {64, UR_MEM_FLAG_READ_WRITE},
+    {128, UR_MEM_FLAG_READ_WRITE},
+    {256, UR_MEM_FLAG_READ_WRITE},
+    {512, UR_MEM_FLAG_READ_WRITE},
+    {1024, UR_MEM_FLAG_READ_WRITE},
+    {8, UR_MEM_FLAG_WRITE_ONLY},
```
martygrant, line 20: The idea here was so a range of buffer sizes could be tested but something I didn't initially consider was just how many tests would be spun up from the amount of variations here. Before this change this particular test suite had 11 tests, but this change would increase that up to 384 which would mean a lot more entries to the match files unless using a wildcard to include them all in one line. Would this be fine or better to choose less buffer sizes?

test/conformance/enqueue/urEnqueueMemBufferMap.cpp
```diff
@@ -21,22 +61,30 @@ TEST_P(urEnqueueMemBufferMapTest, SuccessRead) {
...
     ASSERT_SUCCESS(urEnqueueMemBufferWrite(queue, buffer, true, 0, size,
                                            input.data(), 0, nullptr, nullptr));
 
     uint32_t *map = nullptr;
-    ASSERT_SUCCESS(urEnqueueMemBufferMap(queue, buffer, true, getParam(), 0,
-                                         size, 0, nullptr, nullptr,
-                                         (void **)&map));
+    ASSERT_SUCCESS(urEnqueueMemBufferMap(queue, buffer, true,
+                                         getParam().map_flag, 0, size, 0,
+                                         nullptr, nullptr, (void **)&map));
```
martygrant, line 87: @omarahmed1111 We both created a similar `urMemBufferQueueTestWithParam` fixture around the same time. The difference in mine was I was accessing parameters in the fixture to facilitate creating different buffer types. I've attempted to merge them together but not sure it's a great solution, particularly from adding a `ur_mem_flag_t` that you needed in this test, but I didn't need it for the tests I modified although probably not a big deal given it's just a single enum. Does this seem okay or would it be better to separate the fixtures?

omarahmed1111: Merging them does make sense to me, the test that I added only needed to inherit from the parameter fixture chain and initialize a buffer, so any fixture that does these 2 things would work, so it doesn't need to be a special one for it.

test/conformance/enqueue/urEnqueueMemBufferMap.cpp
martygrant: I was just asking @callumfare about this. I didn't know that the smaller sizes would probably be rounded up so didn't realise many of these entries are redundant. I will change this to just be a couple different larger sizes instead, so it's still similar to what the E2E tests were doing.

test/conformance/enqueue/urEnqueueMemBufferMap.cpp
martygrant: Thanks for this. I guess I was thinking more about having to change your test so the template is based off the new struct I added `mem_buffer_map_write_test_parameters_t` instead of just `ur_map_flag_t` because in the fixture I'm setting the buffer size and flags from parameters and this wasn't needed for yours. But if you think it looks okay I'll just go ahead with that change, thanks!

martygrant:
Ignore the 2 failing Coverage jobs, failure appears to be random.

test/conformance/enqueue/helpers.h
```diff
@@ -96,6 +96,58 @@ print2DTestString(const testing::TestParamInfo<typename T::ParamType> &info) {
...
+struct mem_buffer_map_write_test_parameters_t {
+    size_t count;
+    ur_mem_flag_t mem_flag;
+    ur_map_flag_t map_flag;
+};
+
+template <typename T>
+inline std::string printMemBufferTestString(
+    const testing::TestParamInfo<typename T::ParamType> &info) {
+    // ParamType will be std::tuple<ur_device_handle_t, test_parameters_t>
```
aarongreig, line 113:
```suggestion
    // ParamType will be std::tuple<ur_device_handle_t, mem_buffer_test_parameters_t>
```

test/conformance/enqueue/urEnqueueMemUnmap.cpp
```diff
@@ -2,45 +2,74 @@
...
+        uur::urMemBufferQueueTestWithParam<
+            uur::mem_buffer_test_parameters_t>::TearDown();
+    }
 
     uint32_t *map = nullptr;
 };
-UUR_INSTANTIATE_DEVICE_TEST_SUITE_P(urEnqueueMemUnmapTest);
 
-TEST_P(urEnqueueMemUnmapTest, Success) {
+static std::vector<uur::mem_buffer_test_parameters_t> test_parameters{
```
aarongreig, line 27: since this same set of params gets used a few times could we move it into the header? as like a common buffer ops parameter set

aarongreig (approved):

test/conformance/enqueue/urEnqueueMemUnmap.cpp
martygrant: Good shout, have moved them into `helpers.h`. Thanks!

PR #1455, opened 2024-03-18
[HIP] Fix memory type detection in allocation info queries and USM copy2D

GeorgeWeb:
This PR fixes detection of the pointer type for sycl pointer allocation info queries and USM copy2D for the HIP backend.

With regards to `urUSMGetMemAllocInfo` for `UR_USM_ALLOC_INFO_TYPE`:
Takes extra care identifying unknown/unregistered memory allocations since ROCm 6.0.0 introduces `hipMemoryTypeUnregistered`, which we now use to identify system allocations and the attributes query will still return success in this case.

With regards to `urEnqueueUSMMemcpy2D`:
This is a little awkward workaround. Ideally we'd just use `hipPointerGetAttribute` but it crashes when used on system / pageable host memory hence using `hipPointerGetAttributes` to first determine that and ideally use the memory type value. From ROCm 5.7.1 we cannot rely on the memory type member of the attributes struct to detect whether it is device or host as managed allocations finally update it to managed memory type, hence we need to query the type specifically via `hipPointerGetAttribute` now knowing that the memory handle is not pageable host memory.
Ideally we'd just use allocation flags to determine the memory type and call it a day, but ROCm does not update the user-facing flags member from the attributes struct, so sadly we have no way of relying on allocation flags.

intel/llvm CI: https://github.com/intel/llvm/pull/13059

### DIFF
--- a/source/adapters/hip/enqueue.cpp
+++ b/source/adapters/hip/enqueue.cpp
@@ -1621,6 +1621,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
     // Determine if pSrc and/or pDst are system allocated pageable host memory.
     bool srcIsSystemAlloc{false};
     bool dstIsSystemAlloc{false};
+
+    hipError_t hipRes{};
     // Error code hipErrorInvalidValue returned from hipPointerGetAttributes
     // for a non-null pointer refers to an OS-allocation, hence we can work
     // with the assumption that this is a pointer to a pageable host memory.
@@ -1630,11 +1632,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
     // the pointer handle as system allocated pageable host memory.
     // The HIP runtime can handle the registering/unregistering of the memory
     // as long as the right copy-kind (direction) is provided to hipMemcpy2D*.
-    hipError_t hipRet = hipPointerGetAttributes(&srcAttribs, pSrc);
-    if (pSrc && hipRet == hipErrorInvalidValue)
+    hipRes = hipPointerGetAttributes(&srcAttribs, pSrc);
+    if (hipRes == hipErrorInvalidValue && pSrc)
       srcIsSystemAlloc = true;
-    hipRet = hipPointerGetAttributes(&dstAttribs, (const void *)pDst);
-    if (pDst && hipRet == hipErrorInvalidValue)
+    hipRes = hipPointerGetAttributes(&dstAttribs, (const void *)pDst);
+    if (hipRes == hipErrorInvalidValue && pDst)
       dstIsSystemAlloc = true;
 #if HIP_VERSION_MAJOR >= 6
     srcIsSystemAlloc |= srcAttribs.type == hipMemoryTypeUnregistered;
@@ -1653,7 +1655,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
     // NOTE: The hipPointerGetAttribute API is marked as [BETA] and fails with
     // exit code -11 when passing a system allocated pointer to it.
     if (!srcIsSystemAlloc && srcAttribs.isManaged) {
-      UR_ASSERT(srcAttribs.hostPointer && srcAttribs.hostPointer,
+      UR_ASSERT(srcAttribs.hostPointer && srcAttribs.devicePointer,
                 UR_RESULT_ERROR_INVALID_VALUE);
       UR_CHECK_ERROR(hipPointerGetAttribute(
           &srcMemType, HIP_POINTER_ATTRIBUTE_MEMORY_TYPE,
--- a/source/adapters/hip/enqueue.cpp
+++ b/source/adapters/hip/enqueue.cpp
@@ -1618,25 +1618,55 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
     hipPointerAttribute_t srcAttribs{};
     hipPointerAttribute_t dstAttribs{};
 
+    // Determine if pSrc and/or pDst are system allocated pageable host memory.
     bool srcIsSystemAlloc{false};
     bool dstIsSystemAlloc{false};
-
-    hipError_t hipRes{};
-    // hipErrorInvalidValue returned from hipPointerGetAttributes for a non-null
-    // pointer refers to an OS-allocation, hence pageable host memory. However,
-    // this means we cannot rely on the attributes result, hence we mark system
-    // pageable memory allocation manually as host memory. The HIP runtime can
-    // handle the registering/unregistering of the memory as long as the right
-    // copy-kind (direction) is provided to hipMemcpy2DAsync for this case.
-    hipRes = hipPointerGetAttributes(&srcAttribs, (const void *)pSrc);
-    if (hipRes == hipErrorInvalidValue && pSrc)
+    // Error code hipErrorInvalidValue returned from hipPointerGetAttributes
+    // for a non-null pointer refers to an OS-allocation, hence we can work
+    // with the assumption that this is a pointer to a pageable host memory.
+    // Since ROCm version 6.0.0, the enum hipMemoryType can also be marked as
+    // hipMemoryTypeUnregistered explicitly to relay that information better.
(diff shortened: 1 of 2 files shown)


source/adapters/hip/enqueue.cpp
```diff
@@ -1618,25 +1618,47 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
...
-    if (hipRes == hipErrorInvalidValue && pSrc)
+    // hipErrorInvalidValue or hipErrorMemoryAllocation returned from
+    // hipPointerGetAttributes, for a non-null pointer, refers to an
+    // OS-allocation, hence pageable host memory. However, this means we cannot
+    // rely on any attribute result, hence we mark it as system allocated
+    // pageable host memory. The HIP runtime can handle the
+    // registering/unregistering of the memory as long as the right copy-kind
+    // (direction) is provided to hipMemcpy2DAsync for this case.
+    if (pSrc &&
+        hipPointerGetAttributes(&srcAttribs, (const void *)pSrc) != hipSuccess)
```
hdelan, line 1632: Should we be checking that the return error is actually `hipErrorInvalidValue` or `hipErrorMemoryAllocation`, instead of just checking if it's not `hipSuccess`?

source/adapters/hip/enqueue.cpp
```diff
@@ -1618,25 +1618,47 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
...
-    const unsigned int srcMemType{srcAttribs.type};
-    const unsigned int dstMemType{dstAttribs.type};
+    unsigned int srcMemType{srcAttribs.type};
+    unsigned int dstMemType{dstAttribs.type};
+
+    // ROCm 5.7.1 finally started updating the type attribute member to
+    // hipMemoryTypeManaged for shared memory allocations(hipMallocManaged).
+    // Hence, we use a separate query that verifies the pointer use via flags.
+#if HIP_VERSION >= 50700001
+    // Determine the source/destionation memory type for shared allocations.
```
hdelan, line 1645:
```suggestion
    // Determine the source/destination memory type for shared allocations.
```

source/adapters/hip/enqueue.cpp
```diff
@@ -1618,25 +1618,47 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
...
+    //
+    // NOTE: The hipPointerGetAttribute API is marked as [BETA] and fails with
+    // exit code -11 when passing a system allocated pointer to it.
+    if (!srcIsSystemAlloc && srcAttribs.isManaged) {
+      assert(srcAttribs.hostPointer && srcAttribs.hostPointer);
```
hdelan, line 1650: Can we use a `UR_ASSERT` here, instead?

source/adapters/hip/enqueue.cpp
```diff
@@ -1618,25 +1618,47 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
...
+      UR_CHECK_ERROR(hipPointerGetAttribute(
+          &srcMemType, HIP_POINTER_ATTRIBUTE_MEMORY_TYPE,
+          reinterpret_cast<hipDeviceptr_t>(const_cast<void *>(pSrc))));
+    }
+    if (!dstIsSystemAlloc && dstAttribs.isManaged) {
+      assert(dstAttribs.hostPointer && dstAttribs.devicePointer);
```
hdelan, line 1656: Also here

source/adapters/hip/enqueue.cpp
GeorgeWeb: Yep

source/adapters/hip/enqueue.cpp
GeorgeWeb: @hdelan That condition was actually wrong for ROCm 6, so thanks for checking!!

Also, I found that `hipErrorMemoryAllocation` is irrelevant to figuring out unregistered memory (it is for invalid obtained device memory handle). Since ROCm 6 there is now a new value for the hipMemoryType attribute which can relay that information accurately - `hipMemoryTypeUnregistered` which equals `0` and I was originally confused as I assumed `0` was still refering to `hipMemoryTypeHost` but it appears no more.

The new changes introduced to this PR fix pointer type detections for both this copy2d workaround and the sycl allocation info pointer-type query. Would you mind re-reviewing this again, please? Apologies for that as the new changes are sort of a rework. Thank you!

source/adapters/hip/usm.cpp
```diff
@@ -170,19 +169,27 @@ urUSMGetMemAllocInfo(ur_context_handle_t hContext, const void *pMem,
...
-        // pointer to managed memory
-        return ReturnValue(UR_USM_TYPE_SHARED);
+      // ROCm 6.0.0 introduces hipMemoryTypeUnregistered in the hipMemoryType
+      // enum to mark unregistered allocations (i.e., via system allocators).
+#if HIP_VERSION_MAJOR >= 6
+      if (hipPointerAttributeType.type == hipMemoryTypeUnregistered) {
+        // pointer not known to the HIP subsystem
+        return ReturnValue(UR_USM_TYPE_UNKNOWN);
       }
-      UR_CHECK_ERROR(hipPointerGetAttributes(&hipPointerAttributeType, pMem));
```
GeorgeWeb, line 178: This line was removed as I couldn't figure out the need for re-querying if we've already validated the return value of the first `hipError_t Ret = hipPointerGetAttributes(&hipPointerAttributeType, pMem);` call. The first call does not throw as it is used to check the reutrn value but we call `checkErrorUR` on that return value a few lines later which will still throw on non-success.

Not sure if I am missing something but I don't think it is needed.

hdelan: That looks OK to me

source/adapters/hip/enqueue.cpp
```diff
@@ -1618,25 +1618,55 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMMemcpy2D(
...
+    // Error code hipErrorInvalidValue returned from hipPointerGetAttributes
+    // for a non-null pointer refers to an OS-allocation, hence we can work
+    // with the assumption that this is a pointer to a pageable host memory.
+    // Since ROCm version 6.0.0, the enum hipMemoryType can also be marked as
+    // hipMemoryTypeUnregistered explicitly to relay that information better.
+    // This means we cannot rely on any attribute result, hence we just mark
+    // the pointer handle as system allocated pageable host memory.
+    // The HIP runtime can handle the registering/unregistering of the memory
+    // as long as the right copy-kind (direction) is provided to hipMemcpy2D*.
+    hipError_t hipRet = hipPointerGetAttributes(&srcAttribs, pSrc);
```
hdelan, line 1633: Is there a reason we no longer need to cast `pSrc` to `const void *`? 

Also is it necessary to change `hipRes` to `hipRet`? I would avoid naming changes if possible.

I also think changing the order of checks from `(hipRes == blah && pSrc)` to `(pSrc && hipRes == blah)` is unnecessary.

GeorgeWeb: > Is there a reason we no longer need to cast pSrc to const void *?
`pSrc` is passed as `const void *` argument to the `urEnqueueUSMMemcpy2D` function, so I just avoided the unnecessary cast.

> Also is it necessary to change hipRes to hipRet? I would avoid naming changes if possible.
My bad with the renaming. I meant to leave it as it was, meaning `hipRes`. Will change to how it was before, i.e. `(hipRes == blah && pSrc)`.

> I also think changing the order of checks from (hipRes == blah && pSrc) to (pSrc && hipRes == blah) is unnecessary.
I agree the current renamings and change of order are unnecessary noise, I just didn't pay enough attention after reworking. Thanks!

hdelan: Thanks Georgi, I agree with removing the cast if unnecessary

GeorgeWeb: The unnecessary noise should be resolved in commit [83c5eee](https://github.com/oneapi-src/unified-runtime/pull/1455/commits/83c5eeed097f7930a5003c8ba024abcb8d9eee43).

hdelan (approved):
LGTM

PR #1456, opened 2024-03-19
[CTS] Don't use mangled global variable name

kbenzie:
In the `urGlobalVariableTest` fixture, replace the mangled variable name with the non-mangled version. This enables the test pass on Intel OpenCL Graphis driver.


kbenzie:
Naming of global variables for us in this entry point is not as well defined as it should be. The SYCL requirements are apparently wider that the CUDA entry-point or (unspecified) OpenCL extension expect for reasons I do not comprehend yet.

PR #1458, opened 2024-03-19
[UR] Improve handling of error cases in urProgramLink

RossBrunton:
Note that this change includes a specification change: urProgramLink now requires the output parameter to contain either nullptr or some unspecified binary on failure.

As well as this change, a number of bugs have been fixed:
* The Level Zero adapter now correctly returns `UR_RESULT_ERROR_PROGRAM_LINK_FAILURE` when linking fails, rather than `UR_RESULT_ERROR_UNKNOWN`.
* A workaround has been added for some OpenCL devices that return `CL_INVALID_BINARY` rather than `CL_LINK_PROGRAM_FAILURE` on linker failure.
* The `phProgram` handle is wrapped in a loader handle by the loader even if an error would be returned. This is required by Level Zero, which outputs a "dummy" program to store the linker log.

Conformance tests have also been added.


scripts/core/exp-multi-device-compile.yml
```diff
@@ -93,7 +93,7 @@ analogue:
     - "**clLinkProgram**"
 details:
     - "The application may call this function from simultaneous threads."
-    - "Following a successful call to this entry point the program returned in `phProgram` will contain a binary of the $X_PROGRAM_BINARY_TYPE_EXECUTABLE type for each device in `phDevices`."
+    - "Following a successful call to this entry point the program returned in `phProgram` will contain a binary of the $X_PROGRAM_BINARY_TYPE_EXECUTABLE type for each device in `phDevices`. If a non-success code is returned and `phProgram` is not `nullptr`, `phProgram` will contain an unspecified binary or `nullptr`."
```
kbenzie, line 96: I wonder if we should be more explicit about the state of the `phProgram`, i.e. that it would be valid to query its build log?

kbenzie: We only have 3 error codes defined for this entry point, would if make sense to limit this behaviour to only when `UR_RESULT_ERROR_PROGRAM_LINK_FAILURE` is returned?

RossBrunton: I've updated the documentation to specifically call out that implementations may use the log file of the program to indicate linker failures. However, for consistency's sake, I'll just have all errors require it being set to nullptr (the work is already done anyway).

include/ur_api.h
```diff
@@ -4211,6 +4211,11 @@ urProgramCompile(
 ///       in `phProgram` will contain a binary of the
 ///       ::UR_PROGRAM_BINARY_TYPE_EXECUTABLE type for each device in
 ///       `hContext`.
+///     - If a non-success code is returned and `phProgram` is not `nullptr`, it
+///       will contain an unspecified program or `nullptr`. Implementations may
+///       use the build log of this program (accessable via
```
steffenlarsen, line 4216:
```suggestion
///       use the build log of this program (accessible via
```
Same applies to the 7 other cases of this text.

test/conformance/device_code/linker_error.cpp
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2024 Intel Corporation
+// Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+// See LICENSE.TXT
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <CL/sycl.hpp>
```
steffenlarsen, line 6:
```suggestion
#include <sycl/sycl.hpp>
```

test/conformance/device_code/linker_error.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+SYCL_EXTERNAL void this_function_does_not_exist();
+
+int main() {
+    cl::sycl::queue deviceQueue;
```
steffenlarsen, line 11:
```suggestion
    sycl::queue deviceQueue;
```

We should probably be using SYCL 2020 syntax for good measure.

steffenlarsen (approved):
HIP and CUDA changes LGTM. Just a few nits.

nrspruit (approved):
LGTM for level zero.

kbenzie (approved):

kbenzie:
this has merged conflicts, please fix

RossBrunton:
> this has merged conflicts, please fix

Unfortunately the job in https://github.com/intel/llvm/pull/13085 is failing, so I'll need to look into that first when I get the chance.

kbenzie:
> > this has merged conflicts, please fix
> 
> Unfortunately the job in [intel/llvm#13085](https://github.com/intel/llvm/pull/13085) is failing, so I'll need to look into that first when I get the chance.

I don't see how rebasing this PR requires investigation into the intel/llvm failures. Sure, to get it merged they would need to not fail but that shouldn't be a blocker for rebasing this PR.

kbenzie:
@oneapi-src/unified-runtime-native-cpu-write please review

PietroGhg (approved):
lgtm, thank you

PR #1468, opened 2024-03-22
[L0] fix a deadlock on a recursive event rwlock

pbalcer:
L0, when creating a list of events to wait on, in some cases was first grabbing a lock on a potentially completed event, and then tried to get a command list, which sometimes needs to cleanup all completed events. This caused a deadlock.

This patch moves getting a command list to before the event lock. But because the lock is required to decide whether this command list actually needed, we might be wasting time here.

### DIFF
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -1307,6 +1307,15 @@ ur_result_t _ur_ze_event_list_t::createAndRetainUrZeEventList(
           }
         }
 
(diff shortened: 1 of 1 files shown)


nrspruit (approved):
This matches what I was seeing in the logs during my debug.

Thanks for finding the right place to move the recursive call!

nrspruit (approved):

pbalcer:
https://github.com/intel/llvm/pull/13112

PR #1469, opened 2024-03-22
[Spec Constants] Improved handling of invalid spec. constants

RossBrunton:
Two main changes to how `Kernel/ProgramSetSpecializationConstants` are handled:
* They may now output either `INVALID_VALUE` or the new `INVALID_SPEC_ID` when the provided list is invalid.
* The OpenCL and level 0 adapters now respond to `UR_DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS` with `false` rather than erroring out. This fixes some tests that were incorrectly not being skipped.
* `urKernelSetSpecializationConstants` now "implemented" (as a function that returns `UNSUPPORTED_FEATURE` for opencl and cuda.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -510,6 +510,7 @@ typedef enum ur_result_t {
     UR_RESULT_ERROR_LAYER_NOT_PRESENT = 67,                                   ///< A requested layer was not found by the loader.
     UR_RESULT_ERROR_IN_EVENT_LIST_EXEC_STATUS = 68,                           ///< An event in the provided wait list has ::UR_EVENT_STATUS_ERROR.
     UR_RESULT_ERROR_DEVICE_NOT_AVAILABLE = 69,                                ///< Device in question has `::UR_DEVICE_INFO_AVAILABLE == false`
+    UR_RESULT_ERROR_INVALID_SPEC_ID = 70,                                     ///< A specialization constant identifier is not valid.
     UR_RESULT_ERROR_INVALID_COMMAND_BUFFER_EXP = 0x1000,                      ///< Invalid Command-Buffer
     UR_RESULT_ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_EXP = 0x1001,           ///< Sync point is not valid for the command-buffer
     UR_RESULT_ERROR_INVALID_COMMAND_BUFFER_SYNC_POINT_WAIT_LIST_EXP = 0x1002, ///< Sync point wait list is invalid
@@ -4649,6 +4650,11 @@ typedef struct ur_specialization_constant_info_t {
 ///         + `NULL == pSpecConstants`
 ///     - ::UR_RESULT_ERROR_INVALID_SIZE
 ///         + `count == 0`
+///     - ::UR_RESULT_ERROR_INVALID_VALUE
+///         + A pSpecConstant entry contains a size that does not match that of the specialization constant in the module.
+///         + A pSpecConstant entry contains a nullptr pValue.
+///     - ::UR_RESULT_ERROR_INVALID_SPEC_ID
+///         + Any id specified in a pSpecConstant entry is not a valid specialization constant identifier.
 UR_APIEXPORT ur_result_t UR_APICALL
 urProgramSetSpecializationConstants(
     ur_program_handle_t hProgram,                           ///< [in] handle of the Program object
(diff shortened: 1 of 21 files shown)


test/conformance/program/program_adapter_cuda.match
```diff
@@ -22,3 +22,6 @@
 {{OPT}}urProgramGetInfoTest.InvalidNullHandleProgram/NVIDIA_CUDA_BACKEND___{{.*}}___UR_PROGRAM_INFO_KERNEL_NAMES
 {{OPT}}urProgramLinkTest.Success/NVIDIA_CUDA_BACKEND___{{.*}}_
 {{OPT}}urProgramSetSpecializationConstantsTest.Success/NVIDIA_CUDA_BACKEND___{{.*}}_
```
aarongreig, line 24: I don't think any of the `ProgramSetSpecializationConstants` tests should be `{{OPT}}` for cuda and hip, neither support the entry point and we haven't gotten around to adding a device query to allow the tests to skip so they always fail.

aarongreig (approved):

ldrumm:
@jchlanda is our spec constants expert. Jakub, please take over this review

source/adapters/cuda/kernel.cpp
```diff
@@ -377,3 +377,8 @@ urKernelSetArgSampler(ur_kernel_handle_t hKernel, uint32_t argIndex,
   }
   return Result;
 }
+
+UR_APIEXPORT ur_result_t UR_APICALL urKernelSetSpecializationConstants(
```
jchlanda, line 381: Would it be worth it to do the same for HIP?

jchlanda (approved):

source/adapters/cuda/kernel.cpp
RossBrunton: HIP already does this.

include/ur_api.h
```diff
@@ -4535,6 +4536,11 @@ typedef struct ur_specialization_constant_info_t {
 ///         + `NULL == pSpecConstants`
 ///     - ::UR_RESULT_ERROR_INVALID_SIZE
 ///         + `count == 0`
+///     - ::UR_RESULT_ERROR_INVALID_VALUE
+///         + Any size specified in a pSpecConstant entry does not match the specialization constant in the module.
```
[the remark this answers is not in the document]
RossBrunton: Reworded it to "A pSpecConstant entry contains a size that does not match that of the specialization constant in the module."

source/adapters/cuda/kernel.cpp
jchlanda: Thank you

include/ur_api.h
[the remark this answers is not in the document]
jchlanda: Thank you.

PietroGhg (approved):
Native CPU lgtm, thank you

kbenzie:
> @RossBrunton I think the HIP match files need updated on this.

Forgot you were off. I would make the changes to this PR myself but the checkbox to allow project maintainers to make changes appears to have been disabled.

RossBrunton:
Can someone from @oneapi-src/unified-runtime-level-zero-write have a quick look at this? The changes for level_zero are changing an optional function from `DIE_NO_IMPLEMENTATION` to `  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;`

source/adapters/level_zero/device.cpp
```diff
@@ -861,6 +861,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(
     return ReturnValue(ur_bool_t{false});
   case UR_DEVICE_INFO_IMAGE_SRGB:
     return ReturnValue(ur_bool_t{false});
+  case UR_DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS:
+    return ReturnValue(false);
```
nrspruit, line 865: Shouldn't this be ur_bool_t{false}?

RossBrunton: In the time between creating this MR and now, it looks as if this change has been added via 646cc9ce7 , so this MR no longer touches level zero. There's a lot of device infos that return `bool`s rather than `ur_bool_t` here, so it might be worth having a separate change that updates all of them at once.

omarahmed1111:
@RossBrunton Please rebase this when possible, I removed the ready to merge label and please readd it when rebased.

RossBrunton:
> @RossBrunton Please rebase this when possible, I removed the ready to merge label and please readd it when rebased.

Updated. Sorry for the delay.

PR #1475, opened 2024-03-26
[SPEC] Add more clarification to invalid size return code

omarahmed1111:
Adding more context on when `UR_RESULT_INVALID_BUFFER_SIZE` could be returned in the spec.


aarongreig (approved):

martygrant (approved):

omarahmed1111:
@oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-level-zero-write Could I have a review on this when available? It is only fixing some tests for the adapters.

PR #1482, opened 2024-03-28
[CTS] Add new urUSMDeviceAllocAlignmentTest.SuccessAlignedAllocations test

martygrant:
Adds a new test for USM device allocations, `urUSMDeviceAllocAlignmentTest.SuccessAlignedAllocations` which tests for successful allocations using a range of different size and alignment values. 

I experimented with trying to create a subclass out of `urUSMDeviceAllocTest` to handle this but it was a bit too complicated involving templated parameters at multiple levels so decided to just create a new tuple `using USMDeviceAllocParams = std::tuple<uur::BoolTestParam, uint32_t, size_t>;`, where originally it was just using `uur::BoolTestParam`. For the original set of `urUSMDeviceAllocTest` tests simply provide a single 0 value for the extra two params so the tests don't get run more than they originally were.

Also @aarongreig you mentioned to test for non power of 2 alignment values but there was already a test for this (`InvalidValueAlignPowerOfTwo`) so have just made this PR focusing on a range of different valid values.


test/conformance/usm/urUSMDeviceAlloc.cpp
```diff
@@ -2,21 +2,47 @@
...
 // See LICENSE.TXT
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-
 #include <uur/fixtures.h>
 
-struct urUSMDeviceAllocTest : uur::urQueueTestWithParam<uur::BoolTestParam> {
+using USMDeviceAllocParams = std::tuple<uur::BoolTestParam, uint32_t, size_t>;
+
+template <typename T>
+inline std::string printUSMDeviceAllocTestString(
```
aarongreig, line 10: I think this could equally be used for the other USM alloc entry points, maybe it should be in a header as just `USMAllocParams`/`printUSMAllocTestString`?

martygrant: No bother to add these into a common header. Should I also add a `SuccessAlignedAllocations` test for Host and Shared?

aarongreig: yeah they should work exactly the same way wrt alignment

test/conformance/usm/urUSMDeviceAlloc.cpp
```diff
@@ -27,16 +28,20 @@ struct urUSMDeviceAllocTest : uur::urQueueTestWithParam<uur::BoolTestParam> {
...
         UUR_RETURN_ON_FATAL_FAILURE(
-            uur::urQueueTestWithParam<uur::BoolTestParam>::TearDown());
+            uur::urQueueTestWithParam<uur::USMDeviceAllocParams>::TearDown());
     }
 
     ur_usm_pool_handle_t pool = nullptr;
+    bool usePool = std::get<0>(getParam()).value;
 };
 
+// The 0 value parameters are not relevant for urUSMDeviceAllocTest tests, they are used below in urUSMDeviceAllocAlignmentTest for allocation size and alignment values
```
aarongreig, line 38: looks like formatting missed these comments

test/conformance/usm/usm_adapter_native_cpu.match
```diff
@@ -66,3 +126,33 @@ urUSMSharedAllocTest.InvalidUSMSize/SYCL_NATIVE_CPU___SYCL_Native_CPU___UsePoolE
 urUSMSharedAllocTest.InvalidUSMSize/SYCL_NATIVE_CPU___SYCL_Native_CPU___UsePoolDisabled
 urUSMSharedAllocTest.InvalidValueAlignPowerOfTwo/SYCL_NATIVE_CPU___SYCL_Native_CPU___UsePoolEnabled
 urUSMSharedAllocTest.InvalidValueAlignPowerOfTwo/SYCL_NATIVE_CPU___SYCL_Native_CPU___UsePoolDisabled
+ urUSMSharedAllocAlignmentTest.SuccessAlignedAllocations/SYCL_NATIVE_CPU___SYCL_Native_CPU___UsePoolEnabled_4_8
```
aarongreig, line 129:
```suggestion
urUSMSharedAllocAlignmentTest.SuccessAlignedAllocations/SYCL_NATIVE_CPU___SYCL_Native_CPU___UsePoolEnabled_4_8
```

test/conformance/usm/urUSMDeviceAlloc.cpp
martygrant: Trying to format this locally and I get no change, should it be done in a particular way?

aarongreig: ah right, I guess just split the line so it's under the 80 column maximum width

martygrant: Cool updated it now

aarongreig (approved):

PR #1483, opened 2024-03-28
[L0] Fix regular in order command list reuse given inorder queue

nrspruit:
- Fix command lists cache usage such that given an inorder queue and in order lists is requested, then only in order lists can be used from the cache.

### DIFF
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -712,6 +712,11 @@ ur_result_t ur_context_handle_t_::getAvailableCommandList(
 
     for (auto ZeCommandListIt = ZeCommandListCache.begin();
          ZeCommandListIt != ZeCommandListCache.end(); ++ZeCommandListIt) {
+      // If this is an InOrder Queue, then only allow lists which are in order.
+      if (Queue->Device->useDriverInOrderLists() && Queue->isInOrderQueue() &&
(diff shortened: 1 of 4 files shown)


source/adapters/level_zero/context.cpp
```diff
@@ -712,6 +712,11 @@ ur_result_t ur_context_handle_t_::getAvailableCommandList(
 
     for (auto ZeCommandListIt = ZeCommandListCache.begin();
          ZeCommandListIt != ZeCommandListCache.end(); ++ZeCommandListIt) {
+      // If this is an InOrder Queue, then only allow lists which are in order.
+      if (Queue->Device->useDriverInOrderLists() && Queue->isInOrderQueue() &&
+          !(ZeCommandListIt->second.flags & ZE_COMMAND_QUEUE_FLAG_IN_ORDER)) {
```
raiyanla, line 717: Should be ZE_COMMAND_LIST_FLAG_IN_ORDER ? Instead of QUEUE?

nrspruit: fixed in the latest update.

raiyanla (approved):
LGTM now, assuming CI/tests all pass still.

nrspruit:
This change is a blocker for https://github.com/oneapi-src/unified-runtime/pull/1370 which exposed this bug.

winstonzhang-intel (approved):

PR #1484, opened 2024-03-29
[L0] fetch and build L0 using GCC

pbalcer:
This is a workaround for a crash when the L0 loader is built using clang. To be reverted once the driver is fixed.


aarongreig (approved):

pbalcer:
This doesn't appear to be working as expected. The tests don't seem to be able to find the platforms now... I'll look into this once I'm back.

PR #1487, opened 2024-04-01
[CTS] Add more CTS coverage for urKernelSetArgSampler

omarahmed1111:
This PR adds more CTS tests to `urKernelSetArgSampler` entry-point by testing setting the kernel argument with different types of samplers.


test/conformance/kernel/urKernelSetArgSampler.cpp
```diff
@@ -5,6 +5,65 @@
...
+                                    uur::SamplerCreateParamTy>::TearDown());
+    }
+
+    ur_sampler_handle_t sampler = nullptr;
+};
+
+UUR_TEST_SUITE_P(
+    urKernelSetArgSamplerTestWithParam,
+    ::testing::Combine(
+        ::testing::Values(true, false),
```
martygrant, line 53: I've seen in other test suites for parameterising tests with enabling/disabling of flags the use of `testing::ValuesIn(uur::BoolTestParam::makeBoolParam("UsePool")` which could maybe could be of use here and with the test name printing function.

test/conformance/kernel/urKernelSetArgSampler.cpp
```diff
@@ -36,10 +95,13 @@ struct urKernelSetArgSamplerTest : uur::urBaseKernelTest {
...
 };
+
 UUR_INSTANTIATE_KERNEL_TEST_SUITE_P(urKernelSetArgSamplerTest);
 
-TEST_P(urKernelSetArgSamplerTest, Success) {
-    ASSERT_SUCCESS(urKernelSetArgSampler(kernel, 2, nullptr, sampler));
+TEST_P(urKernelSetArgSamplerTest, SuccessWithProps) {
+    ur_kernel_arg_sampler_properties_t props{
+        UR_STRUCTURE_TYPE_KERNEL_ARG_SAMPLER_PROPERTIES, nullptr};
+    ASSERT_SUCCESS(urKernelSetArgSampler(kernel, 2, &props, sampler));
```
martygrant, line 104: What's the meaning of the `2` here and could it be a named variable?

test/conformance/testing/include/uur/fixtures.h
```diff
@@ -1068,6 +1068,14 @@ std::string deviceTestWithParamPrinter<BoolTestParam>(
     const ::testing::TestParamInfo<
         std::tuple<ur_device_handle_t, BoolTestParam>> &info);
 
+using SamplerCreateParamTy =
```
martygrant, line 1071: Is the `Ty` at the end of this name a typo?

martygrant (approved):
LGTM, couple small queries.

test/conformance/kernel/urKernelSetArgSampler.cpp
omarahmed1111: I think `makeBoolParam` was used mainly to generate meaningful names for tests that we use true and false for the param only like `usepool_true` / `usepool_false`. But in our case `normalized` and `unnormalized` name should be more clear I think, so I modified it to reflect that in the device printer.

test/conformance/kernel/urKernelSetArgSampler.cpp
omarahmed1111: Thanks for pointing that, it should be the `arg_index` but I didn't notice that it is a magic number so changed that.

test/conformance/testing/include/uur/fixtures.h
omarahmed1111: It should reflect an alias for a tuple that holds the sampler creation parameters. So, I guess `Ty` should reflect that it is like an alias type. But I could change that if you think that shouldn't be accurate.

test/conformance/kernel/urKernelSetArgSampler.cpp
martygrant: Fair enough!

test/conformance/testing/include/uur/fixtures.h
martygrant: Ah is this a naming convention? Never came across it before.

kbenzie: `Ty` is usually form `llvm::Type*` variables, though. I don't think its UR naming convention for an actual type. That would either be `T` in PascalCase names or `_t` in snake_case names.

omarahmed1111: I changed it to a `T`. I didn't notice that `Ty` is not common in UR codebase.

kbenzie (approved):

PR #1488, opened 2024-04-02
Bump jinja and its dependency version

lukaszstolarczuk:
- [ ] I still have to verify if docs are built and published properly


third_party/deps.yml
```diff
@@ -33,6 +33,7 @@ dependencies:
   - openssl=3.1.1
   - pkg-config=0.29.2
   - rhash=1.4.3
+  # don't upgrade xz utils due to sec. issue
```
PatKamin, line 36: Could you link to some info on CVE?

lukaszstolarczuk: Right, done.

pbalcer (approved):

PR #1492, opened 2024-04-04
[L0] Enable Disabling of Queue lock during L0 Sync calls

nrspruit:
- Prevent Queue Lock deadlock while L0 queue sync or host sync is in progress.

### DIFF
--- a/source/adapters/level_zero/queue.cpp
+++ b/source/adapters/level_zero/queue.cpp
@@ -1400,7 +1400,13 @@ ur_result_t ur_queue_handle_t_::synchronize() {
       return UR_RESULT_SUCCESS;
 
     // wait for all commands previously submitted to this immediate command list
-    ZE2UR_CALL(zeCommandListHostSynchronize, (ImmCmdList->first, UINT64_MAX));
+    if (UrL0QueueSyncNonBlocking) {
+      Queue->Mutex.unlock();
+      ZE2UR_CALL(zeCommandListHostSynchronize, (ImmCmdList->first, UINT64_MAX));
+      Queue->Mutex.lock();
+    } else {
+      ZE2UR_CALL(zeCommandListHostSynchronize, (ImmCmdList->first, UINT64_MAX));
+    }
 
     // Cleanup all events from the synced command list.
     CleanupEventListFromResetCmdList(ImmCmdList->second.EventList, true);
@@ -1414,7 +1420,13 @@ ur_result_t ur_queue_handle_t_::synchronize() {
     // zero handle can have device scope, so we can't synchronize the last
     // event.
     if (isInOrderQueue() && !LastCommandEvent->IsDiscarded) {
-      ZE2UR_CALL(zeHostSynchronize, (LastCommandEvent->ZeEvent));
+      if (UrL0QueueSyncNonBlocking) {
+        this->Mutex.unlock();
+        ZE2UR_CALL(zeHostSynchronize, (LastCommandEvent->ZeEvent));
+        this->Mutex.lock();
+      } else {
+        ZE2UR_CALL(zeHostSynchronize, (LastCommandEvent->ZeEvent));
+      }
 
       // clean up all events known to have been completed as well,
       // so they can be reused later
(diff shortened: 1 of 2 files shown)


source/adapters/level_zero/queue.cpp
```diff
@@ -1313,10 +1312,22 @@ ur_result_t urQueueReleaseInternal(ur_queue_handle_t Queue) {
       for (auto &QueueGroup : QueueMap)
         for (auto &ZeQueue : QueueGroup.second.ZeQueues)
           if (ZeQueue) {
-            auto ZeResult = ZE_CALL_NOCHECK(zeCommandQueueDestroy, (ZeQueue));
-            // Gracefully handle the case that L0 was already unloaded.
-            if (ZeResult && ZeResult != ZE_RESULT_ERROR_UNINITIALIZED)
-              return ze2urResult(ZeResult);
+            if (UrL0QueueSyncNonBlocking) {
```
pbalcer, line 1315: we can only get here if the Queue's refcount reaches 0. Which shouldn't happen while there's a synchronize happening on that queue in parallel.

source/adapters/level_zero/queue.cpp
```diff
@@ -1414,7 +1435,16 @@ ur_result_t ur_queue_handle_t_::synchronize() {
     // zero handle can have device scope, so we can't synchronize the last
     // event.
     if (isInOrderQueue() && !LastCommandEvent->IsDiscarded) {
-      ZE2UR_CALL(zeHostSynchronize, (LastCommandEvent->ZeEvent));
+      if (UrL0QueueSyncNonBlocking) {
+        this->Mutex.unlock();
+        {
+          std::scoped_lock<ur_shared_mutex> Lock(this->zeQueueInUseMutex);
```
pbalcer, line 1441: dropping the `Mutex` and acquiring the `zeQueueInUseMutex` would have to be atomic for this to be safe. Otherwise, assuming RefCount can be 0 here, a thread calling `urQueueReleaseInternal` could delete the queue object in the unsynchronized gap between these locks.

But, again, I don't believe `zeQueueInUseMutex` is even necessary.

source/adapters/level_zero/queue.hpp
```diff
@@ -548,4 +550,4 @@ ur_result_t createEventAndAssociateQueue(
 // list.
 ur_result_t CleanupEventListFromResetCmdList(
     std::vector<ur_event_handle_t> &EventListToCleanup,
-    bool QueueLocked = false);
+    bool QueueLocked = false);
```
pbalcer, line 553: missing empty line :)

source/adapters/level_zero/queue.cpp
```diff
@@ -1441,8 +1471,19 @@ ur_result_t ur_queue_handle_t_::synchronize() {
               UR_CALL(syncImmCmdList(this, ImmCmdList));
           } else {
             for (auto &ZeQueue : QueueGroup.second.ZeQueues)
-              if (ZeQueue)
-                ZE2UR_CALL(zeHostSynchronize, (ZeQueue));
+              if (ZeQueue) {
```
pbalcer, line 1474: This happens 3 times. Maybe this should be a function in Queue?

source/adapters/level_zero/queue.cpp
nrspruit: I had added the mutex because originally I was not sure, but let me check again if there is any possibility. If there isn't, then I will remove the mutex all together and instead have the "queue sync" be in a "lockless" mode while awaiting the return from the sync.

source/adapters/level_zero/queue.cpp
nrspruit: I will recheck to see if there is any possibility of the queue to be deleted while a thread is calling the sync, if not, then I will remove the mutex entirely.

source/adapters/level_zero/queue.hpp
nrspruit: Will fix, weird formatting change.

source/adapters/level_zero/queue.cpp
nrspruit: I checked the templating and it seems to add to much complexity for this case. I left this out to make the change simpler for now.

pbalcer (approved):
If we merge this, we should also be able to revert https://github.com/oneapi-src/unified-runtime/pull/1371 (but I'm not sure whether we should given the opt-out `UrL0QueueSyncNonBlocking` flag)

kbenzie:
@pbalcer has approved this but this is still pending review from the `@oneapi-src/unified-runtime-level-zero-write` team. Should @pbalcer (and others?) be in that review team?

nrspruit:
> @pbalcer has approved this but this is still pending review from the `@oneapi-src/unified-runtime-level-zero-write` team. Should @pbalcer (and others?) be in that review team?

@raiyanla is also part of that team so it should have caught his approval, does he need to reapprove?

kbenzie:
> @raiyanla is also part of that team so it should have caught his approval, does he need to reapprove?

Weird, perhaps this was a draft PR when that approval happened? Or GitHub being weird. Either way I think its fine, was just a thought about process moving forwards.

raiyanla (approved):

PR #1500, opened 2024-04-09
Add more coverage to CTS tests in various areas

omarahmed1111:
This PR increase CTS coverage by adding tests for:
* `urContextCreateWithNativeHandle`  with owned/unowned native handle.
* `urDeviceCreateWithNativeHandle` with owned/unowned native handle.
* `urDeviceCreateWithNativeHandle` for invalid null pointer arguments.
* `urDevicePartition` for invalid null pointer props argument.
* ` urMemImageCreate` for invalid null pointer image format argument.
* `urMemImageCreate` with host ptr flags.
* `urPlatformCreateWithNativeHandle` with owned/unowned native handle.
* `urPlatformGet` for invalid null pointer output.


test/conformance/device/device_adapter_opencl.match
```diff
@@ -0,0 +1 @@
+urDeviceCreateWithNativeHandleTest.SuccessWithUnOwnedNativeHandle
```
aarongreig, line 1: we'll need to fix this, can you write up a ticket for it?

test/conformance/context/urContextCreateWithNativeHandle.cpp
```diff
@@ -30,3 +30,66 @@ TEST_P(urContextCreateWithNativeHandleTest, Success) {
 
     ASSERT_SUCCESS(urContextRelease(ctx));
 }
+
+TEST_P(urContextCreateWithNativeHandleTest, SuccessWithOwnedNativeHandle) {
+    ur_native_handle_t native_context = nullptr;
+    if (urContextGetNativeHandle(context, &native_context)) {
```
aarongreig, line 36: For this and the similar pattern in `DeviceCreateWithNativeHandle`, I think it would be better to make these ASSERT_SUCCESS instead, and update the pre-existing calls (i.e. [here](https://github.com/oneapi-src/unified-runtime/pull/1500/files#diff-948c9bd68a4224fc6d91be6b79fb7c00db0e01b180e6bebc96e7bb793f392217R13)) to the same. Looking at the match files it doesn't look like the `SKIP`/`continue` are doing anything to prevent fails, and we shouldn't really allow this call to fail anyway

test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
```diff
@@ -30,6 +30,46 @@ TEST_F(urPlatformCreateWithNativeHandleTest, Success) {
...
+        // We cannot assume anything about a native_handle, not even if it's
+        // `nullptr` since this could be a valid representation within a backend.
+        // We can however convert the native_handle back into a unified-runtime
+        // handle and perform some query on it to verify that it works.
+        ur_platform_native_properties_t props = {
+            UR_STRUCTURE_TYPE_PLATFORM_NATIVE_PROPERTIES, nullptr, true};
+        ur_platform_handle_t plat = nullptr;
+        ASSERT_SUCCESS(
+            urPlatformCreateWithNativeHandle(native_handle, &props, &plat));
+        ASSERT_NE(plat, nullptr);
```
aarongreig, line 49: could also add a query like `UR_PLATFORM_INFO_NAME` and compare the result to the original platform

test/conformance/device/device_adapter_opencl.match
omarahmed1111: Done

test/conformance/context/urContextCreateWithNativeHandle.cpp
omarahmed1111: Fixed

test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
omarahmed1111: Done

test/conformance/context/urContextCreateWithNativeHandle.cpp
aarongreig: alright looks like
> Looking at the match files it doesn't look like the SKIP/continue are doing anything to prevent fails

is not true for pretty much every object other than context. I was going to write an issue about documenting this but we actually do document returning `UR_RESULT_ERROR_UNSUPPORTED_FEATURE` "If the adapter has no underlying equivalent handle. " (you can see that under any native handle entry point). I think [urContextGetNativeHandle](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/context/urContextGetNativeHandle.cpp#L14) gets it right, with that pattern we at least make sure it's the error code we're expecting.

omarahmed1111: Yeah, good point, I didn't notice that in the spec. I changed it and covered `CreateWithNativeHandle` with that too as it also have the same possibility described in the spec.

aarongreig: cool thanks, are the match files still accurate? seems like [this](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/cuda/context.cpp#L155) should skip now but it's still here https://github.com/oneapi-src/unified-runtime/pull/1500/files#diff-27bfe80115e9399ca467c92d4bb10e9b5687728cf94bdbed791016aa0e0855d7L1

omarahmed1111: It is a deletion change as this file have only this test that was failing in the past, I have updated it to delete this line as it should skip now.

aarongreig: yeah sorry I was basically asking if you had updated the match files now that most of the native handle tests should be skipping for cuda and hip

aarongreig (approved):

PR #1502, opened 2024-04-09
[GHA] Add "Pull Request Labeler" workflow

kbenzie:
This workflow uses the [labeler](https://github.com/marketplace/actions/labeler) action to automatically add relevant labels to pull requests in order to make it easier to filter pull requests based on the files they make changes to.

Due to `pull_request_target` using the config from the base branch, this will need to be merged before it can be validated as working as stated in the [docs](https://github.com/marketplace/actions/labeler#notes-regarding-pull_request_target-event).

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,48 @@
+# Config file for .github/workflows/labeler.yml to automatically add labels to
+# pull requests based on the globs below. For documentation see:
+# https://github.com/marketplace/actions/labeler
+---
+ci-cd:
+  - changed-files:
+    - .github/**
+
+specification:
+  - changed-files:
+    - scripts/core/**
+
+experimental:
+  - changed-files:
+    - scripts/core/EXP*.rst
+    - scripts/core/exp*.yml
+
+common:
+  - changed-files:
+    - source/common/**
+
+loader:
+  - changed-files:
+    - source/loader/**
+
+opencl:
+  - changed-files:
+    - source/adapter/opencl/**
+
+level-zero:
+  - changed-files:
+    - source/adapters/level_zero/**
+
+cuda:
+  - changed-files:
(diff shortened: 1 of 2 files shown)


martygrant (approved):

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,16 @@
...
+
+on: [ pull_request_target ]
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
```
lukaszstolarczuk, line 15: I believe you first have to checkout the repo, so the labeler config file is accessible

// ref. https://github.com/actions/labeler?tab=readme-ov-file#using-configuration-path-input-together-with-the-actionscheckout-action

.github/labeler.yml
```diff
@@ -0,0 +1,48 @@
+# Config file for .github/workflows/labeler.yml to automatically add labels to
+# pull requests based on the globs below. For documentation see:
+# https://github.com/marketplace/actions/labeler
+---
+ci-cd:
```
lukaszstolarczuk, line 5: I'm not sure if `ci-cd` will match our label `ci/cd`

.github/workflows/labeler.yml
kbenzie: The checkout is not required:

> The path to the label configuration file. If the file doesn't exist at the specified path on the runner, action will read from the source repository via the Github API.

.github/labeler.yml
kbenzie: Good spot, I'll make a PR to fix this.

kbenzie: https://github.com/oneapi-src/unified-runtime/pull/1506

PR #1503, opened 2024-04-09
[Testing] Improve "program" testing to better match the DPC++ e2e tests

RossBrunton:
This adds a number of conformance tests to test things that are required for the "program" DPC++ e2e tests.

Adapter testing infrastructure for level zero has been added, and a single test has been added to check their specific handling of linker errors (they write it to the build log of the output program).

Other than that, test additions are as follows:
* A few specialization constant tests to test usage of a kernel with multiple specialization constants defined.
* Testing of the default specialization values.
* KernelGetInfo outputs the correct kernel name and context pointer.
* Compiling an (valid or invalid) program produces a valid (though unspecified) build log.
* The "num devices" program info is sensible.


aarongreig (approved):

test/conformance/kernel/urKernelGetInfo.cpp
```diff
@@ -66,3 +74,23 @@ TEST_P(urKernelGetInfoTest, InvalidNullPointerPropSizeRet) {
...
+TEST_P(urKernelGetInfoSingleTest, KernelNameCorrect) {
+    size_t name_size = 0;
+    std::vector<char> name_data;
+    ASSERT_SUCCESS(urKernelGetInfo(kernel, UR_KERNEL_INFO_FUNCTION_NAME, 0,
+                                   nullptr, &name_size));
+    name_data.resize(name_size);
+    ASSERT_SUCCESS(urKernelGetInfo(kernel, UR_KERNEL_INFO_FUNCTION_NAME,
+                                   name_size, name_data.data(), nullptr));
+    ASSERT_EQ(name_data[name_size - 1], '\0');
+    ASSERT_EQ(kernel_name, std::string{name_data.data()});
```
kbenzie, line 87: nit: there's also `ASSERT_STREQ()` which compares C-strings and produces nicer error messages if the assert fails.

test/conformance/program/urProgramSetSpecializationConstants.cpp
```diff
@@ -31,6 +43,88 @@ TEST_P(urProgramSetSpecializationConstantsTest, Success) {
...
+    uint64_t output = 0;
+
+    ur_specialization_constant_info_t info_a = {0, sizeof(uint32_t), &a};
+    ASSERT_SUCCESS(urProgramSetSpecializationConstants(program, 1, &info_a));
+
+    ur_specialization_constant_info_t info_c = {2, sizeof(bool), &c};
+    ASSERT_SUCCESS(urProgramSetSpecializationConstants(program, 1, &info_c));
+
+    ur_specialization_constant_info_t info_b = {1, sizeof(uint64_t), &b};
+    ASSERT_SUCCESS(urProgramSetSpecializationConstants(program, 1, &info_b));
```
kbenzie, line 81: any reason for spec id 2 coming before spec id 1?

test/conformance/program/program_adapter_opencl.match
```diff
@@ -1,3 +1,4 @@
+urProgramCreateWithILTest.BuildInvalidProgram/Intel_R__OpenCL___{{.*}}_
```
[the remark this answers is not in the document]
RossBrunton: I thought it did, but apparently the error was coming from `urProgramBuild` rather than create. I've added a quick test case for that.

Despite not passing in CI, this test does pass for me locally, so some devices handle it correctly.

If we aren't going to validate IR, should we amend the spec to explicitly call out that it's not required for implementations to validate that the IR is correct?

test/conformance/program/urProgramSetSpecializationConstants.cpp
RossBrunton: Just a small sanity check to ensure "out of order" sets work correctly. This would catch a naive implementation that assumed that `urProgramSetSpecializationConstant`s were always called in order.

test/conformance/program/program_adapter_opencl.match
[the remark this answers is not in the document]
aarongreig: Yeah it would be good to note that validation doesn't necessarily happen at the `ProgramCreate` step

kbenzie (approved):

PR #1510, opened 2024-04-11
Fix Coverity wrapper escape issue

kbenzie:
Issue was introduced in #1220 where a new allocation is created in a `std::unique_ptr` to enable buffer mapping. The `.get()` member function is called before the `std::unique_ptr` is passed into the `BufferMem` constructor. However, move semantics were not being explicitly followed. The `BufferMem` constructor was taking the `std::unique_ptr` by l-value refernece, then calling `std::move()` on it. This triggered a wrapper-escape, use-after-free defect in Coverity. This fix is to explicitly `std::move()` the `std::unique_ptr` into the `BufferMem` constructor, and also update the `BufferMem` constructor to take an r-value reference instead of an l-value reference.

### DIFF
--- a/source/adapters/cuda/memory.hpp
+++ b/source/adapters/cuda/memory.hpp
@@ -34,7 +34,7 @@ struct BufferMem {
           MapMem(nullptr) {}
 
     BufferMap(size_t MapSize, size_t MapOffset, ur_map_flags_t MapFlags,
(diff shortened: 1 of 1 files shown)


steffenlarsen (approved):
Good catch! 🚀

ldrumm (approved):

PR #1512, opened 2024-04-11
[Bindless][Exp] Remove phMem argument from bindless image creation functions

DBDuncan:
Remove phMem argument from bindless image creation functions. It not used outside of UR and is currently causing a minor memory leak.


DBDuncan:
Corresponding DPC++ PR: https://github.com/intel/llvm/pull/13364

DBDuncan:
@wenju-he Hey. Here is the solution I mentioned in the email I sent you.

wenju-he (approved):
LGTM. I've tested L0 adaptor change in my local repo and there is no regression on DG2.

@DBDuncan is it right L0 adaptor would need to maintain a hash map from device offset to ze_image_handle_t so that memory leak in L0 adaptor could be fixed?

cppchedy (approved):

DBDuncan:
> LGTM. I've tested L0 adaptor change in my local repo and there is no regression on DG2.
> 
> @DBDuncan is it right L0 adaptor would need to maintain a hash map from device offset to ze_image_handle_t so that memory leak in L0 adaptor could be fixed?

Yes. `ze_image_handle_t` does need to be freed at some point. (But can't be done some way via the phMem handle being removed here unless changes are made to the bindless API which we want to avoid.) Currently, yes, a hash map will be needed unless L0 is updated to have one handle instead of the current `ze_image_handle_t` handle plus bindless handle method.

wenju-he:
> Currently, yes, a hash map will be needed unless L0 is updated to have one handle instead of the current `ze_image_handle_t` handle plus bindless handle method.

Thank you @DBDuncan

DBDuncan:
Pinging @oneapi-src/unified-runtime-cuda-write, @oneapi-src/unified-runtime-cuda-write and @oneapi-src/unified-runtime-level-zero-write for approval and ready to merge tag.

Should I also ping people who review changes for the hip and native CPU backends as well? It does technically change them but only removing one argument from two functions.

ldrumm (approved):
Not sure whether we need to do anything about the ABI change at this point, but as long as we don't need that parameter for the foreseeable future, this LGTM

hdelan:
Does this mean that this entire struct is no longer needed: https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/cuda/memory.hpp#L145

DBDuncan:
> Does this mean that this entire struct is no longer needed: https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/cuda/memory.hpp#L145 ?

Yeah, I don't think it is needed unless it is used for old bound images? It makes more sense for an implementation where image memory is allocated in the same function the image is created in. But in bindless images these are done separately. Also, CUDA and the current L0 bindless specification does not have a separate sampler handle. A single sampled image handle is created instead.

hdelan:
> Also, CUDA and the current L0 bindless specification does not have a separate sampler handle. A single sampled image handle is created instead.

I don't really understand what you mean but I am happy for you to go ahead and try to get this merged

hdelan (approved):

hdelan:
https://github.com/oneapi-src/unified-runtime/pull/1565 depends on this PR. Is this ABI break likely to take a while to merge @kbenzie ?

DBDuncan:
Pinging @oneapi-src/unified-runtime-maintain @oneapi-src/unified-runtime-level-zero-write

kbenzie (approved):

kbenzie:
> #1565 depends on this PR. Is this ABI break likely to take a while to merge @kbenzie ?

Experimental features aren't subject to ABI breaking restrictions. We'll still need to follow the adapter change process though. Once all the approvals are in please add the *ready to merge* label @DBDuncan and we'll merge it ASAP.

DBDuncan:
Pinging @oneapi-src/unified-runtime-level-zero-write

DBDuncan:
@nrspruit Friendly ping.

nrspruit (approved):
LGTM for level zero.

Seanst98 (approved):

PR #1518, opened 2024-04-16
[NATIVECPU] Address coverity warnings in threadpool implementation

PietroGhg:
Addresses some warnings raised by Coverity in the Native CPU threadpool implementation:
* the `threadId` for each thread is now set when the worker thread is constructed, making it so it always initialized and `const`.
* the worker task is `std::moved`
* the `m_isRunning` member of the worker thread class is now implemented using `std::atomic`

Corresponding DPC++ PR: https://github.com/intel/llvm/pull/13431

### DIFF
--- a/source/adapters/native_cpu/threadpool.hpp
+++ b/source/adapters/native_cpu/threadpool.hpp
@@ -11,9 +11,10 @@
 #include <atomic>
 #include <condition_variable>
 #include <cstdlib>
+#include <forward_list>
 #include <functional>
 #include <future>
-#include <iostream>
+#include <iterator>
 #include <mutex>
 #include <numeric>
 #include <queue>
@@ -30,18 +31,14 @@ namespace detail {
 class worker_thread {
 public:
   // Initializes state, but does not start the worker thread
-  worker_thread() noexcept : m_isRunning(false), m_numTasks(0) {}
-
-  // Creates and launches the worker thread
-  inline void start(size_t threadId) {
+  worker_thread(size_t threadId) noexcept
+      : m_threadId(threadId), m_isRunning(false), m_numTasks(0) {
     std::lock_guard<std::mutex> lock(m_workMutex);
     if (this->is_running()) {
       return;
     }
-    m_threadId = threadId;
     m_worker = std::thread([this]() {
       while (true) {
-        // pin the thread to the cpu
         std::unique_lock<std::mutex> lock(m_workMutex);
         // Wait until there's work available
         m_startWorkCondition.wait(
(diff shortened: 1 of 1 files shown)


source/adapters/native_cpu/threadpool.hpp
```diff
@@ -121,57 +116,31 @@ class worker_thread {
...
-  }
 
   inline void schedule(const worker_task_t &task) {
     // Schedule the task on the best available worker thread
     this->best_worker().schedule(task);
   }
 
   inline bool is_running() const noexcept {
-    return m_isRunning.load(std::memory_order_acquire);
+    return m_isRunning.load();
```
pbalcer, line 140: The default ordering for atomics is seq-cst, meaning that on x86 there might be an additional fence. The previous implementation with acquire/release ordering has less overhead. Did coverity complain about this?

PietroGhg: Coverity didn't complain about this, but while working on the changes for this PR I thought that using `seq_cst` may be safer here, I'm aware that this may have more overhead. I'm not very familiar with C++'s memory ordering so I'm open to suggestions

pbalcer: In this case I believe acquire/release is safe. When you are stopping, release in store is sufficient to ensure that all writes from the calling thread are visible in the workers. Likewise, when the value is read with acquire, that ensures that all prior writes to this variable from other threads are visible in the worker.

PietroGhg: I've switched back to acquire/release, thanks for the feedback

kbenzie:
@oneapi-src/unified-runtime-native-cpu-write would be great to get this reviewed and merged soon to see if the Coverity defects are resolved by these changes.

source/adapters/native_cpu/threadpool.hpp
```diff
@@ -97,18 +90,21 @@ class worker_thread {
   }
 
   // Checks whether the thread pool is currently running threads
-  inline bool is_running() const noexcept { return m_isRunning; }
+  inline bool is_running() {
```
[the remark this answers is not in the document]
PietroGhg: Yes, thanks for pointing it out

source/adapters/native_cpu/threadpool.hpp
```diff
@@ -201,24 +171,32 @@ class simple_thread_pool {
   }
 
 private:
-  std::vector<worker_thread> m_workers;
+  size_t get_num_threads() {
```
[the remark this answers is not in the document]
PietroGhg: Yeah you are right, thank you

kbenzie:
> sorry for the delay

no worries 🙂

PR #1402, opened 2024-03-01
[SanitizerLayer] Checking "sycl::free" related errors

AllanZyne:

PR #1520, opened 2024-04-17
[DeviceSanitizer] Fix device global tests on LevelZero

AllanZyne:
LLVM: https://github.com/intel/llvm/pull/13468
Add interception on urProgramBuildExp which LevelZero uses by default.


AllanZyne:
This patch have already added in https://github.com/oneapi-src/unified-runtime/pull/1402.
Closed.

PR #1523, opened 2024-04-17
Cache events instead of destroying them for CUDA

MartinWehking:
Don't destroy events that are not needed anymore, but put them on a stack of its associated queue (If there is one).

Use events from the stack before creating new ones.

This caching mechanism ensures that the number of CUDA API calls gets significantly reduced for event creations and destructions.


MartinWehking:
https://github.com/oneapi-src/unified-runtime/pull/1538

PR #1526, opened 2024-04-17
[L0] Enable Batching out of order commands without signal events

nrspruit:
- Given a command being enqueued without dependencies or dependencies outside of a given batch of commands, then signal the event at the end of the command list allowing for parallel execution of the commands.

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -922,8 +922,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
       // Create command-list to execute before `CommandListPtr` and will signal
       // when `EventWaitList` dependencies are complete.
       ur_command_list_ptr_t WaitCommandList{};
-      UR_CALL(Queue->Context->getAvailableCommandList(Queue, WaitCommandList,
-                                                      false, false));
+      UR_CALL(Queue->Context->getAvailableCommandList(
+          Queue, WaitCommandList, false, NumEventsInWaitList, EventWaitList,
+          false));
 
       ZE2UR_CALL(zeCommandListAppendBarrier,
                  (WaitCommandList->first, CommandBuffer->WaitEvent->ZeEvent,
@@ -958,7 +959,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
   // Create a command-list to signal RetEvent on completion
   ur_command_list_ptr_t SignalCommandList{};
   UR_CALL(Queue->Context->getAvailableCommandList(Queue, SignalCommandList,
-                                                  false, false));
+                                                  false, NumEventsInWaitList,
+                                                  EventWaitList, false));
   // Reset the wait-event for the UR command-buffer that is signaled when its
   // submission dependencies have been satisfied.
   ZE2UR_CALL(zeCommandListAppendEventReset,
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -624,7 +624,8 @@ static const size_t CmdListsCleanupThreshold = [] {
 // Retrieve an available command list to be used in a PI call.
 ur_result_t ur_context_handle_t_::getAvailableCommandList(
     ur_queue_handle_t Queue, ur_command_list_ptr_t &CommandList,
-    bool UseCopyEngine, bool AllowBatching,
+    bool UseCopyEngine, uint32_t NumEventsInWaitList,
+    const ur_event_handle_t *EventWaitList, bool AllowBatching,
     ze_command_queue_handle_t *ForcedCmdQueue) {
   // Immediate commandlists have been pre-allocated and are always available.
   if (Queue->UsingImmCmdLists) {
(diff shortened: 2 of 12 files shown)


MichalMrozek (commented):
this mode needs to be device specific as for new HW generations we do not want to do this

nrspruit:
The CI failure so far is unrelated. All the e2e tests are now passing with the latest changes:
https://github.com/intel/llvm/pull/13462

winstonzhang-intel (approved):

raiyanla (approved):

source/adapters/level_zero/queue.cpp
```diff
@@ -1403,6 +1405,18 @@ ur_queue_handle_t_::executeCommandList(ur_command_list_ptr_t CommandList,
       // If we don't have host visible proxy then signal event if needed.
       this->signalEventFromCmdListIfLastEventDiscarded(CommandList);
     }
+    // Append Signalling of the inner events at the end of the batch
+    for (auto &Event : CommandList->second.EventList) {
+      if (Event->IsInnerBatchedEvent) {
+        if (AppendBarrierNeeded) {
+          ZE2UR_CALL(zeCommandListAppendBarrier,
```
pbalcer, line 1412: can you add a comment why a barrier is needed? this can only be true if this is an out-of-order queue, right?

source/adapters/level_zero/context.cpp
```diff
@@ -656,9 +657,17 @@ ur_result_t ur_context_handle_t_::getAvailableCommandList(
   // for this queue.
   if (Queue->hasOpenCommandList(UseCopyEngine)) {
     if (AllowBatching) {
-      CommandList = CommandBatch.OpenCommandList;
-      UR_CALL(Queue->insertStartBarrierIfDiscardEventsMode(CommandList));
-      return UR_RESULT_SUCCESS;
+      bool batchingAllowed = true;
```
pbalcer, line 660: it seems you've added `uint32_t NumEventsInWaitList, const ur_event_handle_t *EventWaitList` arguments just to optionally change the value of `AllowBatching`, which is also an argument to this function. Would it make sense to move checking 'allow batching' to the caller?

source/adapters/level_zero/event.cpp
```diff
@@ -935,7 +951,12 @@ ur_result_t CleanupCompletedEvent(ur_event_handle_t Event, bool QueueLocked,
   std::list<ur_event_handle_t> EventsToBeReleased;
   ur_queue_handle_t AssociatedQueue = nullptr;
   {
-    std::scoped_lock<ur_shared_mutex> EventLock(Event->Mutex);
+    // If the Event is already locked, then continue with the cleanup, otherwise
```
pbalcer, line 954: This looks like a workaround. What is causing the recursive locking?

pbalcer (approved):
LGTM, just a few minor stylistic comments. I wasn't able to fully follow all the logic, but overall the changes make sense.

source/adapters/level_zero/queue.cpp
nrspruit: correct, I can add a comment. This barrier is only needed when there are these events which are only there if this is out of order.

source/adapters/level_zero/event.cpp
nrspruit: Due to the faster completion, if one is using the host proxy event ie the event that had device scope was waited on using urEventWait, then we call GetOrCreateHostProxyEvent. This grabs the lock for the event and tries to create an event that waits on that event to complete with host scope. However, if that event is completed already, then we will call "cleanup event" which means we will take the lock again. This will be a recursive lock and fail. To fix this, we now do a try lock and also check the "bool" that is set and unset while under the lock to see if this is needed. This is because the check for lock ownership fails so another method was needed.

source/adapters/level_zero/queue.cpp
```diff
@@ -1718,6 +1732,59 @@ ur_event_handle_t ur_queue_handle_t_::getEventFromQueueCache(bool IsMultiDevice,
...
+ur_result_t setSignalEvent(ur_queue_handle_t Queue, bool UseCopyEngine,
+                           ze_event_handle_t *ZeEvent, ur_event_handle_t *Event,
+                           uint32_t NumEventsInWaitList,
+                           const ur_event_handle_t *EventWaitList,
+                           ze_command_queue_handle_t ZeQueue) {
+  if (Queue->Device->isIntegrated() &&
+      eventCanBeBatched(Queue, UseCopyEngine, NumEventsInWaitList,
+                        EventWaitList) &&
+      !Queue->isInOrderQueue() && !Queue->UsingImmCmdLists &&
+      !UrL0OutOfOrderIntegratedSignalEvent) {
```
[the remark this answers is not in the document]
nrspruit: Fixed in the latest changes.

pbalcer (approved):

Seanst98 (approved):

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -958,7 +959,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
   // Create a command-list to signal RetEvent on completion
   ur_command_list_ptr_t SignalCommandList{};
   UR_CALL(Queue->Context->getAvailableCommandList(Queue, SignalCommandList,
-                                                  false, false));
+                                                  false, NumEventsInWaitList,
+                                                  EventWaitList, false));
```
EwanC, line 963: There isn't a UR wait list for this command-list, it's ordering is enforced by calling ` Queue->executeCommandList(SignalCommandList, false, false);` after  the call on line 953
```
 ZE2UR_CALL(zeCommandQueueExecuteCommandLists,
             (ZeCommandQueue, 1, &CommandBuffer->ZeCommandList, ZeFence));
```

(see https://github.com/intel/llvm/blob/sycl/sycl/doc/design/CommandGraph.md#level-zero)

PR #1264, opened 2024-01-19
[Bindless][Exp] Add device queries for sampled image fetch

przemektmalon:
Added the following queries for device capabilities of fetching sampled images:

- DEVICE_INFO_BINDLESS_SAMPLED_IMAGE_FETCH_1D_USM_EXP
- DEVICE_INFO_BINDLESS_SAMPLED_IMAGE_FETCH_1D_EXP
- DEVICE_INFO_BINDLESS_SAMPLED_IMAGE_FETCH_2D_USM_EXP
- DEVICE_INFO_BINDLESS_SAMPLED_IMAGE_FETCH_2D_EXP
- DEVICE_INFO_BINDLESS_SAMPLED_IMAGE_FETCH_3D_USM_EXP
- DEVICE_INFO_BINDLESS_SAMPLED_IMAGE_FETCH_3D_EXP

PR #1529, opened 2024-04-18
Reinstate changes from PR #1264

aarongreig:
These were reverted in https://github.com/oneapi-src/unified-runtime/pull/1524


kbenzie:
Closing, changes will be reintroduced in #1535 instead.

PR #1530, opened 2024-04-19
[GHA] Separate out hardware workflows

kbenzie:
The goal of this patch is to make a failure of one HW build & test job only cancel jobs running on the same HW tag rather than all HW jobs for all adapters. It does this by moving the hardware jobs out of the `cmake.yml` workflow into the new `build-hw-reusable.yml` workflow, which as the name suggests is reusable. This reusable workflow is then used in by adapter specific jobs in the `cmake.yml` workflow to display jobs in a group which is collapsible.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,6 @@
 [*]
 indent_style = space
 indent_size = 4
+
+[.github/workflows/*.yml]
+indent_size = 2
--- a/.github/workflows/build-hw-reusable.yml
+++ b/.github/workflows/build-hw-reusable.yml
@@ -1,0 +1,78 @@
+---
+name: Build - Adapters on HW - Reusable
+
+on:
+  workflow_call:
+    inputs:
+      name:
+        required: true
+        type: string
+      platform:
+        required: false
+        type: string
+        default: ""
+
+permissions:
+  contents: read
+
+jobs:
+  adapter-build-hw:
+    name: Build & Test HW
+    if: github.repository == 'oneapi-src/unified-runtime'  # run only on upstream; forks won't have the HW
+    strategy:
+      matrix:
+        adapter: [
+          {name: "${{inputs.name}}", platform: "${{inputs.platform}}"},
+        ]
+        build_type: [Debug, Release]
+        compiler: [{c: gcc, cxx: g++}, {c: clang, cxx: clang++}]
+        # TODO: The latest L0 loader segfaults when built with clang.
+        exclude:
+         - adapter: {name: L0, platform: ""}
+           compiler: {c: clang, cxx: clang++}
+
+    runs-on: ${{matrix.adapter.name}}
+
+    steps:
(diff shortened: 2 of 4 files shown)


.github/workflows/build-hw-level-zero.yml
```diff
@@ -0,0 +1,75 @@
...
+on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
```
pbalcer, line 13: we should either remove it entirely and fold it into regular tests (I still think that the whole conda setup is unnecessary) or move the workflow to another file.

pbalcer (approved):

.github/workflows/build-hw-level-zero.yml
kbenzie: I think there's still value in having this test and agree about the conda setup, I hope it's not broken while it's been disabled. Having it in a separate workflow makes sense.

kbenzie: I moved it into its own workflow file: `examples-hw-level-zero.yml`

lukaszstolarczuk:
At the same time I like this change and I don't 😉 I think both of your goals are good, but I'd propose a slightly different implementation. **If you feel like you still prefer your way I'm not going to be stubborn.**

The reusable workflow is good, but in current approach it spawns a few more workflows per each PR. It's not that big of a deal, but instead of creating new files (per each adapter) we could do jobs per each adapter in `cmake.yml`, like the code below.
I think it should still achive your goals, it will omit the bloat code for each workflow file, and keep a little order in the workflows.
```
jobs:
  ...
  cuda:
    uses: ./.github/workflows/build-hw-reusable.yml
    with:
      name: CUDA
  hip:
    uses: ./.github/workflows/build-hw-reusable.yml
    with:
      name: HIP
  ...
```

Example "Checks" view on PR page:
image
vs
image
(all building workflows would be hidden as jobs in "Build and test", as it is now)

kbenzie:
> I think it should still achive your goals, it will omit the bloat code for each workflow file, and keep a little order in the workflows.

I'll try this out. I'm not 100% happy with how things got split up and displayed in my current approach either.

kbenzie:
@lukaszstolarczuk this looks good, on [here](https://github.com/oneapi-src/unified-runtime/actions/runs/8801600914/job/24155383800?pr=1530) I like how the adapter jobs are also collapsable. I'm tempted to split out the other jobs in the `cmake.yml` in a similar way if only for ease of browsing the Build and test workflow.

kbenzie:
One thing this approach doesn't do, which I liked about the separate workflows, is allow for retrying just the jobs for a single adapter. The whole `cmake.yml` workflow needs to complete before I can retry, this would be nice for when a short term network related issue pops up.

PR #1533, opened 2024-04-22
[DeviceSanitizer] Support detecting out-of-bounds errors on sycl::buffer

zhaomaosu:
As we discussed before, we intercepted buffer with usm for short-term solution. For long-term, we will draft a new ocl extension to allow user to create buffer with fixed address.

SYCLOS PR for pre-ci: https://github.com/intel/llvm/pull/13504


source/loader/layers/sanitizer/asan_buffer.cpp
```diff
@@ -0,0 +1,137 @@
...
+            context.logger.error("Failed to free buffer handle {}", Ptr);
+            return URes;
+        }
+    }
+    Allocations.clear();
+    return UR_RESULT_SUCCESS;
+}
+
+size_t MemBuffer::getAlignment() {
+    // Choose an alignment that is at most 64 and is the next power of 2
```
yingcong-wu, line 123: The `64` here seems not match the `128` in LINE 132 below.

AllanZyne (approved):

zhaomaosu:
@oneapi-src/unified-runtime-maintain, could you please help review this PR when you are available? Thanks.

source/loader/layers/sanitizer/ur_sanddi.cpp
```diff
@@ -316,6 +316,841 @@ __urdlllocal ur_result_t UR_APICALL urContextRelease(
...
+    if (nullptr == pfnRelease) {
+        return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+    }
+
+    context.logger.debug("==== urKernelRelease");
+    UR_CALL(pfnRelease(hKernel));
+
+    if (auto KernelInfo = context.interceptor->getKernelInfo(hKernel)) {
+        uint32_t RefCount;
+        UR_CALL(context.urDdiTable.Kernel.pfnGetInfo(
```
pbalcer, line 1081: This is not really thread-safe. If multiple `urKernelRelease` are done in parallel, multiple threads may read RefCount == 1. I think you might need to have your own RefCount like for MemBuffer.

source/loader/layers/sanitizer/asan_buffer.cpp
```diff
@@ -0,0 +1,137 @@
...
+    UR_CALL(context.urDdiTable.Enqueue.pfnEventsWait(Queue, Events.size(),
+                                                     Events.data(), Event));
+
+    return UR_RESULT_SUCCESS;
+}
+
+ur_result_t MemBuffer::getHandle(ur_device_handle_t Device, char *&Handle) {
+    // Sub-buffers don't maintain own allocations but rely on parent buffer.
+    if (SubBuffer) {
+        UR_CALL(SubBuffer->Parent->getHandle(Device, Handle));
```
pbalcer, line 73: Wouldn't it be easier to just store the pointer instead of Parent ptr and an offset?

source/loader/layers/sanitizer/ur_sanddi.cpp
zhaomaosu: Thanks, you are right. Updated.

source/loader/layers/sanitizer/asan_buffer.cpp
zhaomaosu: We want buffer/subbuffer do the allocation when they are actually used. If we store the pointer, this means we need to do the allocation when subbuffer is created. This is not we expected.

pbalcer (approved):

zhaomaosu:
@oneapi-src/unified-runtime-maintain, any other comments for this PR? Thanks.

kbenzie:
> @oneapi-src/unified-runtime-maintain, any other comments for this PR? Thanks.

Don't think so. If you could make https://github.com/intel/llvm/pull/13504 ready for review so to get approvals in place there before we merge here that would be great.

PR #1536, opened 2024-04-22
Unified clang format

ldrumm:
Until now we've had multiple styles throughout the project. This doesn't
buy us anything for contributors as each source file might be different - totally negating the point of having a clang-format style in the first
place - indeed I started debugging clang-format since it was apparently
not formatting my adapter code as 4 spaces after my editor had loaded
the editconfig file specifying 4 spaces. I wasted an hour on that!

Here we codify everything as LLVM style, which seems to be used by the
majority of the project, and change the placement of autogenerated doxygen comments.

n.b. if you want to see what's really happening just look at the first two commits, f68e8c0f441, and fd6b63b2a - the others are just regeneration or formatting

https://github.com/intel/llvm/pull/16672


pbalcer:
This was intentional to keep headers in a more readable format. Also, disabling `ReflowComments` was necessary to keep the output consistent, as the option is not indempotent.

ldrumm:
> This was intentional to keep headers in a more readable format.

Define "readable".

pbalcer:
> > This was intentional to keep headers in a more readable format.
> 
> Define "readable".

If I recall correctly, the primary issue was line breaks around comments and enum value definitions. Take a look at this enum: https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L415

ldrumm:
> > > This was intentional to keep headers in a more readable format.
> > 
> > 
> > Define "readable".
> 
> If I recall correctly, the primary issue was line breaks around comments and enum value definitions. Take a look at this enum: https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h#L415

That problem could be neatly solved by putting the comment on the line before and using `///` instead of `<///`

```cpp
typedef enum ur_result_t {
  /// Success
  UR_RESULT_SUCCESS = 0,
  /// Invalid operation
  UR_RESULT_ERROR_INVALID_OPERATION = 1,
  /// Invalid queue properties
  UR_RESULT_ERROR_INVALID_QUEUE_PROPERTIES = 2,
  /// Invalid queue
  UR_RESULT_ERROR_INVALID_QUEUE = 3,
  /// Invalid Value
  UR_RESULT_ERROR_INVALID_VALUE = 4,
  /// Invalid context
  UR_RESULT_ERROR_INVALID_CONTEXT = 5,
  /// Invalid platform
  UR_RESULT_ERROR_INVALID_PLATFORM = 6,
  /// Invalid binary
  UR_RESULT_ERROR_INVALID_BINARY = 7,
  /// Invalid program
  UR_RESULT_ERROR_INVALID_PROGRAM = 8,
  /// Invalid sampler
  UR_RESULT_ERROR_INVALID_SAMPLER = 9,
  /// Invalid buffer size
  UR_RESULT_ERROR_INVALID_BUFFER_SIZE = 10,
  /// Invalid memory object
  UR_RESULT_ERROR_INVALID_MEM_OBJECT = 11,
  /// Invalid event
  UR_RESULT_ERROR_INVALID_EVENT = 12,

**Comment by MartinWehking** [comment]
I agree with @ldrumm and @jchlanda.
It would definitely be better to not have 2 different styles for unified runtime and intel/llvm.

**Review by fabiomestre — APPROVED** [review]

**Review by PietroGhg — APPROVED** [review]

**Review by nrspruit — COMMENTED** [review]
LGTM for Level Zero

**Review by nrspruit — APPROVED** [review]

**Comment by ldrumm** [comment]
Thanks for the reviews everyone. I've now updated the patch with the suggestions from [this comment](https://github.com/oneapi-src/unified-runtime/pull/1536#issuecomment-2069276875)

The substantive changes are  in https://github.com/ldrumm/unified-runtime/commit/c69f330bfcf47996945271fb8dc5c5c171c08b30 [and](https://github.com/oneapi-src/unified-runtime/pull/1536/commits/feab158a9a0b6118c6d2b1ea1f2f699bc301eca0) https://github.com/ldrumm/unified-runtime/commit/00014e2f31631a46d844c54195b7a4e79dedcab0. The third commit, is simply the result of running the `generate` target, and then committing all changes so is entirely mechanical.

It's quite a lot of churn, but the separate commits should make it easy for people to resolve merge conflicts

**Comment by ldrumm** [comment]
@pbalcer Are your original concerns assuaged with the latest update?

**Comment by ldrumm** [comment]
> @pbalcer Are your original concerns assuaged with the latest update?

ping @pbalcer

**Comment by ldrumm** [comment]
> > @pbalcer Are your original concerns assuaged with the latest update?
> 
> ping @pbalcer

ping @pbalcer

**Comment by pbalcer** [comment]
> > > @pbalcer Are your original concerns assuaged with the latest update?
> > 
> > 
> > ping @pbalcer
> 
> ping @pbalcer

Apologies, I missed all your previous pings :-)

Yes, I have no objections after the changes. Thanks.

**Comment by ldrumm** [comment]
Sorry for letting this rot. I've just rebased it and we can hopefully try again.

**Comment by ldrumm** [comment]
The SPDX license identifier needs to be on its own line according to the spec: https://spdx.github.io/spdx-spec/v2.3/using-SPDX-short-identifiers-in-source-files/

python needs 4 space indents in the .editorconfig

**Comment by ldrumm** [comment]
> The SPDX license identifier needs to be on its own line according to the spec: https://spdx.github.io/spdx-spec/v2.3/using-SPDX-short-identifiers-in-source-files/
> 
> python needs 4 space indents in the .editorconfig

Fixen

**Comment by ldrumm** [comment]
I think this should be good to go if tests pass. I'll add the ready-to-merge label and then wait for something terrible to happen

File: .editorconfig [hunk 1]
```diff
@@ -1,6 +1,6 @@
 [*]
 indent_style = space
-indent_size = 4
-
-[.github/workflows/*.yml]
 indent_size = 2
+
+[*.py]
+indent_size = 4
```
kbenzie, line 6: We should also add a 4 space indent for the spec files since they are not formatted automatically and aren't changed in this patch:

```suggestion
indent_size = 4

[scripts/core/*]
indent_size = 4
```

ldrumm: That's a good idea.

You happy about https://github.com/cheshirekow/cmake_format for the cmake?

ldrumm:
> Since this is changing a lot of files all at once I do wonder if this would also be an opportune time to update the clang-format version used from 15 to the version built as part of intel/llvm. This idea is with an eye to existing within the intel/llvm repo moving forwards.

Can do. It's trivial for me to remake this patch with new versions of tools / rebase it.
Will you update the clang-format version or should I add it to this patch series?

.editorconfig
kbenzie: > You happy about https://github.com/cheshirekow/cmake_format for the cmake?

If we do switch to that I think it should be separate from the PR, its already very large.

ldrumm:
On Thu Jan 16, 2025 at 2:15 PM GMT, Kenneth Benzie (Benie) wrote:
> > You happy about https://github.com/cheshirekow/cmake_format for the cmake?
>
> If we do switch to that I think it should be separate from this PR, its
> already very large.

Makes sense. I'll include the editorconfig changes you provided and we can
reformat the cmake later

As for clang-format, it looks like [dpc++ uses
clang-format-18](https://github.com/intel/llvm/blob/f0d0f7cdbcad238bc7492c1977019777defb8996/.github/workflows/pr-code-format.yml#L63).
I'll bump the required version in the cmake and requirements.txt

kbenzie (approved):
Will aim to merge this next once the conflicts are resolved.

PR #1538, opened 2024-04-22
[CUDA][HIP] Add Event Caching

MartinWehking:
Don't destroy events that are not needed anymore, but put them on a stack
of its associated queue (If there is one).

Use events from the stack before creating new ones.

Make the operations for putting events onto the stack and retrieving them
atomic to ensure thread safety.

This caching mechanism ensures that the number of CUDA API calls gets
significantly reduced for event creations and destructions.

### DIFF
--- a/source/adapters/cuda/event.cpp
+++ b/source/adapters/cuda/event.cpp
@@ -275,13 +275,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
           urQueueRelease(Queue);
         }
         urContextRelease(Context);
-
-        return UR_RESULT_SUCCESS;
       }
-    } catch (...) {
-      return UR_RESULT_ERROR_OUT_OF_RESOURCES;
+    } catch (ur_result_t Err) {
+      return Err;
     }
-    return UR_RESULT_ERROR_INVALID_EVENT;
   }
   return UR_RESULT_SUCCESS;
 }
--- a/source/adapters/cuda/event.cpp
+++ b/source/adapters/cuda/event.cpp
@@ -42,11 +42,32 @@ ur_event_handle_t_::ur_event_handle_t_(ur_context_handle_t Context,
   urContextRetain(Context);
 }
 
+void ur_event_handle_t_::reset() {
+  RefCount = 0;
+  HasBeenWaitedOn = false;
+  IsRecorded = false;
+  IsStarted = false;
+  Queue = nullptr;
+  Context = nullptr;
+}
+
 ur_event_handle_t_::~ur_event_handle_t_() {
+  if (HasOwnership) {
+    if (EvEnd)
+      UR_CHECK_ERROR(cuEventDestroy(EvEnd));
+
+    if (EvQueued)
+      UR_CHECK_ERROR(cuEventDestroy(EvQueued));
+
+    if (EvStart)
+      UR_CHECK_ERROR(cuEventDestroy(EvStart));
+  }
   if (Queue != nullptr) {
     urQueueRelease(Queue);
   }
-  urContextRelease(Context);
+  if (Context != nullptr) {
+    urContextRelease(Context);
+  }
 }
 
 ur_result_t ur_event_handle_t_::start() {
@@ -139,22 +160,6 @@ ur_result_t ur_event_handle_t_::wait() {
   }
 
   return Result;
-}
-
-ur_result_t ur_event_handle_t_::release() {
-  if (!backendHasOwnership())
-    return UR_RESULT_SUCCESS;
-
-  assert(Queue != nullptr);
-
-  UR_CHECK_ERROR(cuEventDestroy(EvEnd));
-
-  if (Queue->URFlags & UR_QUEUE_FLAG_PROFILING_ENABLE || isTimestampEvent()) {
-    UR_CHECK_ERROR(cuEventDestroy(EvQueued));
-    UR_CHECK_ERROR(cuEventDestroy(EvStart));
-  }
-
-  return UR_RESULT_SUCCESS;
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL urEventGetInfo(ur_event_handle_t hEvent,
@@ -257,7 +262,22 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
     ur_result_t Result = UR_RESULT_ERROR_INVALID_EVENT;
     try {
       ScopedContext Active(hEvent->getContext());
-      Result = hEvent->release();
+    if (!hEvent->backendHasOwnership()) {
+      Result = UR_RESULT_SUCCESS;
+    }
+    else {
+      assert(hEvent->getQueue() != nullptr);
+      assert(hEvent->getContext() != nullptr);
+
+      auto queue = event_ptr->getQueue();
+      auto context = event_ptr->getContext();
+
+      event_ptr->reset();
+      queue->CachedEvents.emplace(event_ptr.release());
+      urQueueRelease(queue);
+      urContextRelease(context);
+      Result = UR_RESULT_SUCCESS;
+    }
     } catch (...) {
       Result = UR_RESULT_ERROR_OUT_OF_RESOURCES;
     }
--- a/source/adapters/cuda/event.cpp
+++ b/source/adapters/cuda/event.cpp
@@ -43,7 +43,9 @@ ur_event_handle_t_::ur_event_handle_t_(ur_context_handle_t Context,
 }
 
 void ur_event_handle_t_::reset() {
-  RefCount = 0;
+  detail::ur::assertion(RefCount == 0,
+                        "Attempting to reset an event that is still referenced");
+
   HasBeenWaitedOn = false;
   IsRecorded = false;
   IsStarted = false;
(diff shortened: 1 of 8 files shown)


source/adapters/cuda/event.cpp
```diff
@@ -42,11 +42,32 @@ ur_event_handle_t_::ur_event_handle_t_(ur_context_handle_t Context,
   urContextRetain(Context);
 }
 
+void ur_event_handle_t_::reset() {
+  RefCount = 0;
```
hdelan, line 46: Is this thread safe?

source/adapters/cuda/event.hpp
```diff
@@ -83,6 +83,21 @@ struct ur_event_handle_t_ {
...
              uint32_t StreamToken = std::numeric_limits<uint32_t>::max()) {
+    if (Queue->has_cached_events()) {
+      auto retEvent = Queue->get_cached_event();
+
+      retEvent->Stream = Stream;
+      retEvent->StreamToken = StreamToken;
+      retEvent->CommandType = Type;
+      retEvent->Queue = Queue;
+      retEvent->Context = Queue->Context;
+      retEvent->RefCount = 1;
```
GeorgeWeb, line 94: Is `RefCount` always going be either 1 or 0 (1 when first created or 0 when reset/destroyed)? I am asking this in relation to whether it will be worth considering incrementing and decrementing the `RefCount` (atomically, i.e. `std::atomic::fetch_add` with `std::memory_order_relaxed`) instead to safely manage copying the `ur_event` object. I haven't checked the constraints and all uses of this object, just a general observation and suggestion.

source/adapters/cuda/event.cpp
```diff
@@ -256,7 +261,22 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
     ur_result_t Result = UR_RESULT_ERROR_INVALID_EVENT;
     try {
       ScopedContext Active(hEvent->getContext());
-      Result = hEvent->release();
+    if (!hEvent->backendHasOwnership()) {
+      Result = UR_RESULT_SUCCESS;
+    }
+    else {
+      assert(hEvent->getQueue() != nullptr);
```
hdelan, line 268: This will fail for an event created with interop

source/adapters/cuda/event.cpp
```diff
@@ -256,7 +261,22 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
...
+      assert(hEvent->getContext() != nullptr);
+
+      auto queue = event_ptr->getQueue();
```
hdelan, line 271: This will fail for interop. See https://oneapi-src.github.io/unified-runtime/core/api.html#ureventcreatewithnativehandle

hdelan (changes requested):

source/adapters/cuda/event.cpp
```diff
@@ -256,7 +261,22 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
     ur_result_t Result = UR_RESULT_ERROR_INVALID_EVENT;
     try {
       ScopedContext Active(hEvent->getContext());
-      Result = hEvent->release();
+    if (!hEvent->backendHasOwnership()) {
+      Result = UR_RESULT_SUCCESS;
```
hdelan, line 265: Can we just return here?

source/adapters/cuda/event.cpp
```diff
@@ -256,7 +261,22 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
...
+      assert(hEvent->getContext() != nullptr);
+
+      auto queue = event_ptr->getQueue();
+      auto context = event_ptr->getContext();
+
+      event_ptr->reset();
+      queue->CachedEvents.emplace(event_ptr.release());
+      urQueueRelease(queue);
+      urContextRelease(context);
+      Result = UR_RESULT_SUCCESS;
```
hdelan, line 278: I think we can avoid having this `Result` var

source/adapters/cuda/event.cpp
```diff
@@ -262,20 +264,21 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
...
       event_ptr->reset();
-      queue->CachedEvents.emplace(event_ptr.release());
-      urQueueRelease(queue);
-      urContextRelease(context);
-      Result = UR_RESULT_SUCCESS;
+      if (Queue) {
+        Queue->cache_event(event_ptr.release());
+        urQueueRelease(Queue);
+      }
+      if (Context) {
```
hdelan, line 278: I don't think you need this check for Context. This should always be non null

source/adapters/cuda/event.cpp
MartinWehking: This became obsolete.
RefCount is already 0 at this point.
Making this assignment redundant.
I replaced it with an assertion to ensure that we're not resetting an event that still holds references.

source/adapters/cuda/event.hpp
MartinWehking: Good question.
RefCount is initialized to 1 in the constructor and indeed 0 when it gets destroyed.
Setting it to 1 in this particular spot should not violate thread-safety, however.
Since the retrieval of events is now atomic inside the queue and this event just lives inside the scope of this function before it's returned, there is no chance for this particular event to be accessed outside and we can therefore modify it's variables safely without creating an invalid state.

source/adapters/cuda/event.cpp
MartinWehking: Good spot.
Assertion has been replaced with an if/else

source/adapters/cuda/event.cpp
```diff
@@ -256,7 +261,22 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
     ur_result_t Result = UR_RESULT_ERROR_INVALID_EVENT;
     try {
       ScopedContext Active(hEvent->getContext());
-      Result = hEvent->release();
+    if (!hEvent->backendHasOwnership()) {
+      Result = UR_RESULT_SUCCESS;
+    }
+    else {
+      assert(hEvent->getQueue() != nullptr);
+      assert(hEvent->getContext() != nullptr);
```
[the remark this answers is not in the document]
MartinWehking: I've used detail::ur::assertion similarly to other spots where RefCount is used , The UR_ASSERT checks only for nullptr (I think)

source/adapters/cuda/event.cpp
MartinWehking: Thanks, fixed now

source/adapters/cuda/event.cpp
MartinWehking: We can indeed

source/adapters/cuda/event.cpp
```diff
@@ -256,7 +263,23 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
     ur_result_t Result = UR_RESULT_ERROR_INVALID_EVENT;
```
hdelan, line 263: Would be good to delete this var

source/adapters/hip/event.cpp
```diff
@@ -291,7 +297,21 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
     std::unique_ptr<ur_event_handle_t_> event_ptr{hEvent};
     ur_result_t Result = UR_RESULT_ERROR_INVALID_EVENT;
```
hdelan, line 298: Also here

source/adapters/cuda/event.cpp
```diff
@@ -254,16 +261,28 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
...
+        urContextRelease(Context);
+
+        return UR_RESULT_SUCCESS;
+      }
     } catch (...) {
-      Result = UR_RESULT_ERROR_OUT_OF_RESOURCES;
+      return UR_RESULT_ERROR_OUT_OF_RESOURCES;
     }
-    return Result;
+    return UR_RESULT_ERROR_INVALID_EVENT;
```
hdelan, line 284:
```suggestion
```

MartinWehking: 852d02fa049516ff388da39ba5cd006c4e17952e

source/adapters/cuda/event.cpp
```diff
@@ -254,16 +261,28 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
...
+          Queue->cache_event(event_ptr.release());
+          urQueueRelease(Queue);
+        }
+        urContextRelease(Context);
+
+        return UR_RESULT_SUCCESS;
+      }
     } catch (...) {
-      Result = UR_RESULT_ERROR_OUT_OF_RESOURCES;
+      return UR_RESULT_ERROR_OUT_OF_RESOURCES;
```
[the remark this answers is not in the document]
MartinWehking: 852d02fa049516ff388da39ba5cd006c4e17952e

source/adapters/cuda/event.cpp
```diff
@@ -254,16 +261,28 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
...
+        auto Context = event_ptr->getContext();
+
+        event_ptr->reset();
+        if (Queue) {
+          Queue->cache_event(event_ptr.release());
+          urQueueRelease(Queue);
+        }
+        urContextRelease(Context);
+
+        return UR_RESULT_SUCCESS;
```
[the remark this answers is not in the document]
MartinWehking: 852d02fa049516ff388da39ba5cd006c4e17952e

source/adapters/cuda/queue.hpp
```diff
@@ -57,6 +61,8 @@ struct ur_queue_handle_t_ {
   std::mutex ComputeStreamMutex;
   std::mutex TransferStreamMutex;
   std::mutex BarrierMutex;
+  // The event cache might be accessed in multiple threads.
+  std::mutex CacheMutex;
```
hdelan, line 65: I might be inclined to have this close in code to the CachedEvents

hdelan:
Small nit. Otherwise LGTM

hdelan (approved):

source/adapters/cuda/queue.hpp
```diff
@@ -245,4 +248,23 @@ struct ur_queue_handle_t_ {
...
+
+  void cache_event(ur_event_handle_t Event) {
+    std::lock_guard<std::mutex> CacheGuard(CacheMutex);
+    CachedEvents.push(Event);
+  }
+
+  // Returns and removes an event from the CachedEvents stack.
+  ur_event_handle_t get_cached_event() {
+    std::lock_guard<std::mutex> CacheGuard(CacheMutex);
+    assert(!CachedEvents.empty());
```
MartinWehking, line 265: @konradkusiak97 spotted this. Need to change this

source/adapters/hip/queue.hpp
```diff
@@ -242,4 +245,23 @@ struct ur_queue_handle_t_ {
...
+
+  void cache_event(ur_event_handle_t Event) {
+    std::lock_guard<std::mutex> CacheGuard(CacheMutex);
+    CachedEvents.push(Event);
+  }
+
+  // Returns and removes an event from the CachedEvents stack.
+  ur_event_handle_t get_cached_event() {
+    std::lock_guard<std::mutex> CacheGuard(CacheMutex);
+    assert(!CachedEvents.empty());
```
MartinWehking, line 262: @konradkusiak97 spotted this. Need to change this to detail::ur::assertion

MartinWehking:
Closed since it did not yield significant improvement in benchmarkings

PR #1539, opened 2024-04-22
[L0] reset and clean the command list on error unknown

nrspruit:


source/adapters/level_zero/queue.cpp
```diff
@@ -1207,16 +1207,16 @@ ur_queue_handle_t_::executeCommandList(ur_command_list_ptr_t CommandList,
...
-      }
       // Reset Command List and erase the Fence forcing the user to resubmit
       // their commands.
       std::vector<ur_event_handle_t> EventListToCleanup;
       resetCommandList(CommandList, true, EventListToCleanup, false);
       CleanupEventListFromResetCmdList(EventListToCleanup,
                                        true /* QueueLocked */);
+      if (ZeResult == ZE_RESULT_ERROR_UNKNOWN) {
+        // Turn into a more informative end-user error.
+        return UR_RESULT_ERROR_UNKNOWN;
```
raiyanla, line 1218: If we are moving this branch down to here, wouldn't the `return ze2urResult(ZeResult)` function right below this line do exactly this same error conversion for return value (based on the switch statement inside ze2urResult)?

winstonzhang-intel (approved):

source/adapters/level_zero/queue.cpp
nrspruit: It looks like that line was added afterwards, yes that branch is no longer needed, I will remove.

nrspruit: Resolved, I have removed the if entirely.

raiyanla (approved):

PR #1546, opened 2024-04-25
fix queue locking behavior when creating event lists

pbalcer:
This fixes a race when creating queue events without acquiring the appropriate locks, and also fixes a deadlock when acquiring multiple queue locks.

The deadlock scenario is roughly this:
queue1 = ...;
queue2 = ...;

for (;;) {
  queue1_event = urEnqueueKernelLaunch(queue1, ...);
  // lock order: queue2->Mutex, queue1->Mutex
  urEnqueueEventsWait(queue2, 1, [queue1_event], nullptr);
}

T2:

for (;;) {
  queue2_event = urEnqueueKernelLaunch(queue2, ...);
  // lock order: queue1->Mutex, queue2->Mutex
  urEnqueueEventsWait(queue1, 1, [queue2_event], nullptr);
}

The solution in this patch is the simplest possible fix I managed to figure out, but I strongly believe this code requires a substantial refactor to better structure the synchronization logic. Right now it's a minefield.

### DIFF
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -1261,16 +1261,26 @@ ur_result_t _ur_ze_event_list_t::createAndRetainUrZeEventList(
         }
 
         auto Queue = EventList[I]->UrQueue;
-        if (Queue) {
-          // The caller of createAndRetainUrZeEventList must already hold
-          // a lock of the CurQueue. Additionally lock the Queue if it
-          // is different from CurQueue.
-          // TODO: rework this to avoid deadlock when another thread is
-          //       locking the same queues but in a different order.
-          auto Lock = ((Queue == CurQueue)
-                           ? std::unique_lock<ur_shared_mutex>()
-                           : std::unique_lock<ur_shared_mutex>(Queue->Mutex));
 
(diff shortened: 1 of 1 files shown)


pbalcer:
https://github.com/intel/llvm/pull/13564

pbalcer:
I've been looking at this code more closely, and I'm not sure whether unlocking the CurQueue like that is a good idea, since something else can modify the state of the queue in the middle of an operation. So if an upper-level function stores some queue state on the stack or some other allocation, and then calls `createAndRetainUrZeEventList`, that state might change. This is the case [here](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/event.cpp#L1226).

Now, I don't *think* there's any bug here now, but this is such a massive footgun that it might be a good idea to consider some other solution (everything I came up so far requires a very substantial refactor, where the risk of making it this late in the release cycle might be non-trivial).

nrspruit (approved):
This was similar to what I tried, but as you said, you are dropping the lock to the current queue given the Queue in the event list is not the same. 

Thanks for finding that we needed to move the lock scope to one step above.

PR #1548, opened 2024-04-25
[L0] Use relative includes for adapter internals

kbenzie:
Fix build issues when using decoupled adapter fetch in intel/llvm by using relative include paths instead of longer explicit paths. This was causing redefinitions of objects. This patch also adds a missing `#pragma once` from the `adapter.hpp` header.


kbenzie:
This should resolve build issues like [this](https://github.com/intel/llvm/actions/runs/8830202807/job/24243132566?pr=13522).

pbalcer (approved):

nrspruit (approved):

Issue #1454, opened 2024-03-18
Control and instrospection API


pbalcer:
## Rationale

Currently Unified Runtime does not have a unified way of querying or modifying internal loader or adapter state (or statistics). For example, the L0 adapter has more than 20 individual environment variables that controls its behavior. It also collects some statistics, but the only way for see them is by setting an env variable to print it at the end.
There's no centralized location in which all knobs are stored, there's no uniform way to see which flags are set, there's no way to set a different set of defaults depending on device (e.g., PVC vs DG2), there's no way to control these things programmatically (other than using setenv and hoping for the best).

## Description

Ideally, we'd limit the number of knobs that users need to be aware of to bare minimum, and never actually require that any of them are set for optimal performance. Sadly, reality is such that setting those options is often needed to get the most of the hardware. Controllable feature flags are also a useful tool for testing out new features or optimizations without affecting everyone.

Likewise, statistics are not strictly required, but are invaluable tool for debugging and performance evaluations. Currently, the adapters do not have any real infrastructure to output statistics in a controlled way. And even if some statistics exist, they are not queryable at runtime, which is often helpful e.g., in tests.

To address these problems, I'm proposing we introduce a control and introspection API in Unified Runtime:
```
ur_result_t urLoaderCtlGet(const char *name, void *arg);
ur_result_t urLoaderCtlSet(const char *name, void *arg);
```
Inspired by similar interfaces in [jemalloc](https://jemalloc.net/jemalloc.3.html#mallctl_namespace) and [libpmemobj](https://pmem.io/pmdk/manpages/linux/master/libpmemobj/pmemobj_ctl_get.3/).

Adapters would define a tree-like structure of the ctl interface:
```
- events
  - caching (read-write)
  - discarded
    - enabled (read-write)
- stats (all read only)
  - pool
    - allocated
    - active
  - queue
      - num
      - <index>
        - command_list
          - num
          - <index>
            - events_in_flight
- debug (read-write)
```
A pointer to this structure would be then passed to the loader (through a special API in DDI, not defined here), which would include it in the global control namespaces, prepending appropriate namespace at the front (or not, depending if some nodes are meant to be shared).

This would then allow users to access the various functionality like so:
```
bool caching = false;
urLoaderCtlSet("adapters.level_zero.events.caching", &caching);
```
`UR_CTL="adapters.level_zero.events.caching=false;[more ctls]"`
```
$ cat /etc/ur.conf

adapters:
  level_zero:
    events:
[...]

PR #1324, opened 2024-02-08
[L0] Allocate event pools efficiently in multi-device scenarios

raiyanla:

PR #1549, opened 2024-04-25
Fix adding event to queue cache

igchor:
in single-device scenario.

When event is an internal event (when using queue with discard events property) the UrQueue is set to nullptr. This means that inside addEventToContextCache the event will be added to a multi-device cache which is wrong. This causes unbounded growth of the multi-device event cache if there are no multi-device events requested from the cache.

### DIFF
--- a/source/adapters/level_zero/common.cpp
+++ b/source/adapters/level_zero/common.cpp
@@ -88,7 +88,11 @@ bool setEnvVar(const char *name, const char *value) {
 ZeUSMImportExtension ZeUSMImport;
 
 // This will count the calls to Level-Zero
+// TODO: remove the ifdef once
+// https://github.com/oneapi-src/unified-runtime/issues/1454 is implemented
+#ifndef UR_L0_CALL_COUNT_IN_TESTS
 std::map<std::string, int> *ZeCallCount = nullptr;
+#endif
 
 inline void zeParseError(ze_result_t ZeError, const char *&ErrorString) {
   switch (ZeError) {
--- a/test/adapters/level_zero/event_cache_tests.cpp
+++ b/test/adapters/level_zero/event_cache_tests.cpp
@@ -1,0 +1,163 @@
+// Copyright (C) 2024 Intel Corporation
+// Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+// See LICENSE.TXT
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "ur_print.hpp"
+#include "uur/fixtures.h"
+#include "uur/raii.h"
+
+#include <map>
+#include <string>
+
+template <typename... Args> auto combineFlags(std::tuple<Args...> tuple) {
+    return std::apply([](auto... args) { return (... |= args); }, tuple);
+}
+
+extern std::map<std::string, int> *ZeCallCount;
+
+using FlagsTupleType = std::tuple<ur_queue_flags_t, ur_queue_flags_t,
+                                  ur_queue_flags_t, ur_queue_flags_t>;
+
+struct urEventCacheTest : uur::urContextTestWithParam<FlagsTupleType> {
+    void SetUp() override {
+        UUR_RETURN_ON_FATAL_FAILURE(urContextTestWithParam::SetUp());
+
+        flags = combineFlags(getParam());
+
+        ur_queue_properties_t props;
+        props.flags = flags;
+        ASSERT_SUCCESS(urQueueCreate(context, device, &props, &queue));
+        ASSERT_NE(queue, nullptr);
+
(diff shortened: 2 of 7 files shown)


igchor:
This is draft for now as I'd like to add some tests for this.

pbalcer (approved):

pbalcer:
This is a regression after #1324. We should target this for 0.9.

Given the severity of the impact, it makes me wonder whether the discard events are used by anyone and worth keeping around in their current form.  They are adding a lot of complexity to the codebase.

nrspruit (approved):

raiyanla (approved):

igchor:
I've added one test for the event caches but it's a bit hacky (I'm using ZeCallCount map from L0 adapter) - if you have a better idea how to test number of zeEventCreateCalls then I'm open for suggestions.  The testcase that verifies the issue fixed by this PR is `eventsReuseNoVisibleEvent` with `UR_QUEUE_FLAG_DISCARD_EVENTS`.

Also, there are some problems I found:
- the test seems to fail with UR_RESULT_ERROR_FEATURE_UNSUPPORTED (this is returned from either urQueueFinish or urQeueuRelease) for some machines on our CI (with older driver?). For now, I've added a GTEST_SKIP() if this error is returned but it requires further investigation
- test does not work with `UR_QUEUE_FLAG_SUBMISSION_BATCHED` - I'm not sure why
- in eventsReuseNoVisibleEvent I would expect the number of `zeEventCreate` to never be as high as `numIters * numEnqueues`  but that number is equal or higher unless `UR_QUEUE_FLAG_DISCARD_EVENTS` is set and `UR_QUEUE_FLAG_OUT_OF_ORDER_EXEC_MODE_ENABLE` is NOT set

test/adapters/level_zero/event_cache_tests.cpp
```diff
@@ -0,0 +1,170 @@
...
+template <typename... Args> auto combineFlags(std::tuple<Args...> tuple) {
+    return std::apply([](auto... args) { return (... |= args); }, tuple);
+}
+
+extern std::map<std::string, int> *ZeCallCount;
+
+using FlagsTupleType = std::tuple<ur_queue_flags_t, ur_queue_flags_t,
+                                  ur_queue_flags_t, ur_queue_flags_t>;
+
+#define ASSERT_SUCCESS_OR_UNSUPPORTED(ret)                                     \
```
pbalcer, line 22: can this be moved to some common location?

test/adapters/level_zero/event_cache_tests.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+TEST_P(urEventCacheTest, eventsReuseNoVisibleEvent) {
+    static constexpr int numIters = 16;
+    static constexpr int numEnqueues = 128;
+
+    for (int i = 0; i < numIters; i++) {
+        for (int j = 0; j < numEnqueues; j++) {
+            enqueueWork(nullptr, i * numEnqueues + j);
+        }
+        ASSERT_SUCCESS_OR_UNSUPPORTED(urQueueFinish(queue));
```
pbalcer, line 89: it's odd that the UNSUPPORTED is returned by urQueueFinish. Did you investigate what specifically is returning that?

source/adapters/level_zero/common.cpp
```diff
@@ -88,7 +88,9 @@ bool setEnvVar(const char *name, const char *value) {
 ZeUSMImportExtension ZeUSMImport;
 
 // This will count the calls to Level-Zero
+#ifndef UR_L0_CALL_COUNT_IN_TESTS
 std::map<std::string, int> *ZeCallCount = nullptr;
```
pbalcer, line 92: I really don't like this. #1454 is meant to give programmatic access to internals like this in tests. So please add a TODO.

test/adapters/level_zero/event_cache_tests.cpp
```diff
@@ -0,0 +1,170 @@
...
+    std::replace(str.begin(), str.end(), '|', '_');
+    return platform_device_name + "__" + str;
+}
+
+UUR_TEST_SUITE_P(
+    urEventCacheTest,
+    ::testing::Combine(
+        testing::Values(0, UR_QUEUE_FLAG_DISCARD_EVENTS),
+        testing::Values(0, UR_QUEUE_FLAG_OUT_OF_ORDER_EXEC_MODE_ENABLE),
+        // TODO: why the test fails with UR_QUEUE_FLAG_SUBMISSION_BATCHED?
```
pbalcer, line 166: with what combination of the flags? The discard events path doesn't seem that well tested, so it's not inconceivable that's its broken when a few different flags are set.

igchor: I rerun this on the latest driver and everything works fine there so I guess we can just uncomment this once we switch our CI to the newer driver.

PR #1550, opened 2024-04-26
[E2E] improve e2e workflow

wlemkows:


pbalcer:
I think ccache is already working. In the [last cuda e2e job](https://github.com/oneapi-src/unified-runtime/actions/runs/8840484571/job/24275908843) the SYCL build took <5 minutes. What I think was happening previously is that each runner (of which there are 4? on the system) had its own path so ccache needed at least 4 CI runs (with each runner) to be fully populated.

wlemkows:
Interesting. Are you suggesting discarding these changes?

pbalcer:
At least the ccache setup part, yes.

wlemkows:
> At least the ccache setup part, yes.

Done.

PR #1551, opened 2024-04-26
Add OpenCL e2e status label

wlemkows:


pbalcer:
opencl e2e is failing with:
```

CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
```

Can we install Ninja on the opencl e2e runners?

wlemkows:
> opencl e2e is failing with:
> 
> ```
> 
> CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
> ```
> 
> Can we install Ninja on the opencl e2e runners?



> opencl e2e is failing with:
> 
> ```
> 
> CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
> ```
> 
> Can we install Ninja on the opencl e2e runners?

Should be done now.

wlemkows:
@pbalcer looks like the ninja is already working, now failing because of the "real" problem. That's why I just decided to add a status label.

pbalcer (approved):

PR #1553, opened 2024-04-26
[HIP] Get device from queue, not event

hdelan:
If an event was constructed using some interop method it doesn't have an
associated queue. Some logic was assuming that each event has a native
device. This has been removed. It is still OK to use
`LastEventWritingToMemObj->getQueue()->getDevice()` since this is never
set to an interop event.

Also removes `event::getDevice` so that its use is discouraged.


GeorgeWeb (approved):
The new device tracking changes LGTM. I'd wait for HIP adapter CI to stabilize before moving forward.

hdelan:
CI looks good apart from Intel asan failures. @kbenzie adding `Ready to merge`

kbenzie:
> CI looks good apart from Intel asan failures. @kbenzie adding `Ready to merge`

I commented on https://github.com/intel/llvm/pull/13575 about updating the HIP UR tag rather than the main UR tag to sidestep the ASAN tests. If we do that we can merge this much sooner.

PR #1547, opened 2024-04-25
Candidate for the v0.9.1 release tag

kbenzie:
* [x] #1539
* [x] #1548
* [x] #1526
* [x] #1546
* [x] #1549

PR #1555, opened 2024-04-30
stop batching multiple CmdList execution when waiting for barriers

pbalcer:
The current implementation of queue command batching appears to assume that getAvailableCommandList is followed by a single executeCommandList, and it's not allowed to, for example, create a list of CmdLists for later execution. And there's an assert to check for this.

This makes sense, since it doesn't really make sense to batch things across many different CmdLists. However, the implementation of urEnqueueEventsWaitWithBarrier was doing exactly that.

This patch disables batching in executeCommandList, fixing a assert failure during urEnqueueEventsWaitWithBarrier when many potentially open command lists are in use.

### DIFF
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -367,6 +367,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueEventsWaitWithBarrier(
                                      IsInternal));
   }
 
+  // When executing command lists, allowing batching simply means storing
+  // the CmdLists as the current open batch. That's not what we want here.
+  // Moreover, the current implementation does not allow calling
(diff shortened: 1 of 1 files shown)


pbalcer:
https://github.com/intel/llvm/pull/13599

kbenzie:
@oneapi-src/unified-runtime-level-zero-write would be great to get some reviews on this so I can go ahead with merging everything and adding it to #1547

igchor (approved):

kbenzie:
I've removed this from [v0.9.1rc](https://github.com/oneapi-src/unified-runtime/pull/1547) due to [these issues](https://github.com/intel/llvm/pull/13599#issuecomment-2093241342)

PR #1559, opened 2024-04-30
Fix CTS test for HIP accessors

EwanC:
CTS test assumes that backend will only require a single buffer accessor offset parameter, however the HIP backend requires 3.

This should fix [sporadic failures](https://github.com/oneapi-src/unified-runtime/actions/runs/8892452470/job/24416545313) in the `BufferFillCommandTest` tests.


fabiomestre (approved):
LGTM :+1:

EwanC:
> LGTM 👍
> 
> Might be a good idea to update the title with [HIP][Command-Buffer]

I think the convention for this stopped when the labels were automatically added to PRs https://oneapi-src.github.io/unified-runtime/core/CONTRIB.html#adapter-change-process. I'll added these as labels instead

PR #1560, opened 2024-04-30
Add option to enable spec generation fast-mode

kbenzie:
The new `--fast-mode` option to the `run.py` script disables the print API section of the API listing mako template which causes the `generate` CMake target to take an order of magnitude longer to complete.

Before these changes:

```
ninja generate  199.22s user 1.80s system 99% cpu 3:21.67 total
```

After these changes:

```
ninja generate  37.63s user 1.00s system 99% cpu 38.893 total
```

This greatly improves the iterative spec writing experience and can opted by setting the CMake option `UR_ENABLE_FAST_SPEC_MODE=ON`.


RossBrunton (approved):

kbenzie:
> LGTM! Worth adding to the [README](https://github.com/oneapi-src/unified-runtime/blob/main/README.md) file?

Yup, good shout.

PR #1562, opened 2024-04-30
[HIP] Add a workaround for crashes on gfx1031

frasercrmck:
When calling hipMemGetInfo, an internal problem querying the amount of available global memory causes the function to return hipErrorInvalidValue. This is indistinguishable from us doing something wrong, and it (rightly) causes an assertion failure. This is a particularly egregious issue because we query this property and crash during the execution of `sycl-ls --verbose`.

This architecture is not officially supported by HIP, so we can't file a bug and shouldn't expect an official fix for this issue any time soon. It is possible, however, that we'll need to extend this workaround with other architectures over time.

### DIFF
--- a/source/adapters/hip/device.cpp
+++ b/source/adapters/hip/device.cpp
@@ -732,7 +732,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     hipDeviceProp_t Props;
     UR_CHECK_ERROR(hipGetDeviceProperties(&Props, hDevice->get()));
     if (strcmp(Props.gcnArchName, "gfx1031") == 0) {
-      return ReturnValue(0);
+      return ReturnValue(size_t{0});
     }
 
     size_t FreeMemory = 0;
--- a/source/adapters/hip/device.cpp
+++ b/source/adapters/hip/device.cpp
@@ -730,17 +730,14 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     // can't distinguish this condition from us doing something wrong, we can't
     // handle it gracefully.
     hipDeviceProp_t Props;
-    detail::ur::assertion(hipGetDeviceProperties(&Props, hDevice->get()) ==
-                          hipSuccess);
+    UR_CHECK_ERROR(hipGetDeviceProperties(&Props, hDevice->get()));
     if (strcmp(Props.gcnArchName, "gfx1031") == 0) {
       return ReturnValue(0);
     }
 
     size_t FreeMemory = 0;
     size_t TotalMemory = 0;
-    detail::ur::assertion(hipMemGetInfo(&FreeMemory, &TotalMemory) ==
-                              hipSuccess,
-                          "failed hipMemGetInfo() API.");
+    UR_CHECK_ERROR(hipMemGetInfo(&FreeMemory, &TotalMemory));
     return ReturnValue(FreeMemory);
   }
 
--- a/source/adapters/hip/device.cpp
+++ b/source/adapters/hip/device.cpp
@@ -724,6 +724,18 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   }
 
   case UR_DEVICE_INFO_GLOBAL_MEM_FREE: {
+    // Work around an issue on some (unsupported) architectures,
+    // where hipMemGetInfo fails internally and returns hipErrorInvalidValue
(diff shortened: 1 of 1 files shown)


source/adapters/hip/device.cpp
```diff
@@ -724,6 +724,18 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   }
 
   case UR_DEVICE_INFO_GLOBAL_MEM_FREE: {
+    // Work around an issue on some (unsupported) architectures,
+    // where hipMemGetInfo fails internally and returns hipErrorInvalidValue
+    // when trying to query the amount of available global memory. Since we
+    // can't distinguish this condition from us doing something wrong, we can't
+    // handle it gracefully.
+    hipDeviceProp_t Props;
+    detail::ur::assertion(hipGetDeviceProperties(&Props, hDevice->get()) ==
```
JackAKirk, line 733: nit: instead of using assert atm all hip calls like this should ideally be wrapped in `UR_CHECK_ERROR(` that at least points to the user to the problem code if there is an error.

JackAKirk: Think this is probably especially important with things like hip where random errors come up all the time for certain devices.

JackAKirk: Techically both ur::assertion and UR_CHECK_ERROR are not consistent with sycl spec that expects runtime errors for such failures, but I think UR_CHECK_ERROR provides at least more info for the user.

JackAKirk: I'll approve it anyway because trying to enforce a consistent way of dealing with error passing has proven impossible!

JackAKirk (approved):

source/adapters/hip/device.cpp
frasercrmck: Interesting, thanks for the in-depth explanation. Yeah I was just copying nearby code but I see that `UR_CHECK_ERROR` is used sporadically too. I will switch to `UR_CHECK_ERROR`. I see it does logging, which also sounds nice.

source/adapters/hip/device.cpp
```diff
@@ -724,11 +724,20 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
...
   case UR_DEVICE_INFO_GLOBAL_MEM_FREE: {
+    // Work around an issue on some (unsupported) architectures,
+    // where hipMemGetInfo fails internally and returns hipErrorInvalidValue
+    // when trying to query the amount of available global memory. Since we
+    // can't distinguish this condition from us doing something wrong, we can't
+    // handle it gracefully.
+    hipDeviceProp_t Props;
+    UR_CHECK_ERROR(hipGetDeviceProperties(&Props, hDevice->get()));
+    if (strcmp(Props.gcnArchName, "gfx1031") == 0) {
+      return ReturnValue(0);
```
ldrumm, line 735: You mention `hipErrorInvalidValue` in the comment above, but we're not checking for it here?

ldrumm:
I know it's not a supported architecture, but every bug reported shows people do actually want more from rocm, so if you have the mental bandwidth, please report anyway

source/adapters/hip/device.cpp
frasercrmck: Yeah, maybe I am not being clear, sorry. I'm trying to say that `hipMemGetInfo` below returns `hipErrorInvalidValue` if we let that run, and we just want to avoid that because it's ambiguous as to whether it means we're passing incorrect values, or it just internally can't do what it needs to do. Instead of falling into that situation, we just early-return a dummy value here.

GeorgeWeb: Can we instead return the dummy value here for gfx1031 _only if_ the result code of `hipMemGetInfo` is `hipErrorInvalidValue`, then call `checkErrorUR` on the result?

IIRC, we haven't always had that error. However, I may be very wrong, and if it is always an issue for all relevant ROCm versions, then I guess no need the extra complexity to this workaround, so I am happy leaving it as is.

jinz2014:
```
if (strcmp(Props.gcnArchName, "gfx1031") == 0) {
      return ReturnValue(size_t{0});
    }
```

May you add all unsupported architectures ?

frasercrmck:
> ```
> if (strcmp(Props.gcnArchName, "gfx1031") == 0) {
>       return ReturnValue(size_t{0});
>     }
> ```
> 
> May you add all unsupported architectures ?

I don't really want to add architectures without being able to test them. They might work already, in which case this change would be a regression for those architectures. I don't think we're able to say that all unsupported architectures will fail in the same way as gfx1031.

PR #1564, opened 2024-04-30
[CUDA][HIP] Replace occurrences of `size_t` with `uint64_t` to improve code portability.

mmoadeli:
Replaces the usage of `size_t` with `uint64_t` to enhance the portability of the code across Windows and Linux platforms. The discrepancy in the size of `size_t (unsigned long)` between Linux and Windows environments necessitates this change for consistent behaviour.
The issue showed as a failure testing [test-e2e/Basic/info.cpp](https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/Basic/info.cpp) on Windows.

### DIFF
--- a/source/adapters/cuda/device.cpp
+++ b/source/adapters/cuda/device.cpp
@@ -418,7 +418,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     return ReturnValue(static_cast<size_t>(Min));
   }
   case UR_DEVICE_INFO_IMAGE_MAX_ARRAY_SIZE: {
-    return ReturnValue(0lu);
+    return ReturnValue(size_t(0));
   }
   case UR_DEVICE_INFO_MAX_SAMPLERS: {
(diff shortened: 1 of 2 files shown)


kbenzie:
The types of these return values are defined in the [specification](https://oneapi-src.github.io/unified-runtime/core/api.html#ur-device-info-t) to be `size_t` so what this PR actually does is introduces bugs in the CUDA and HIP adapters.

ldrumm:
I think the test is broken rather than UR. Depending on the size of `size_t` seems buggy.

mmoadeli:
> The types of these return values are defined in the [specification](https://oneapi-src.github.io/unified-runtime/core/api.html#ur-device-info-t) to be `size_t` so what this PR actually does is introduces bugs in the CUDA and HIP adapters.

Thanks @kbenzie 
You are right. The PI seems to be problematic not the UR.

kbenzie:
> You are right. The PI seems to be problematic not the UR.

Not the first time that's happened.

If you find that there's a mismatch between the SYCL spec and UR spec that's something we should fix, though.

PR #1565, opened 2024-05-01
[CUDA] CUDA adapter multi device context

hdelan:
Implement multi device context in CUDA adapter. This allows multiple devices to share the same context.

When devices in the same context use the same ur_mem_handle_t, the SYCL runtime expects the plugin library to be able to migrate memory between allocations on different devices. This necessitates some added functionality for the ur_mem_handle_t class. Some more detailed documentation is added in memory.hpp explaining the memory migration patterns.

This patch also removes `phMem` allocation for some bindless images entry points. This is because this behaviour is untested and not useful in the bindless images API. Use of `phMem` for bindless images with multi device context is a non trivial change as it requires lazy allocation and migration of memory across devices in a context. Moreover https://github.com/oneapi-src/unified-runtime/pull/1512 removes this param entirely from function signatures, so I see no harm in removing its use here.

DPC++ testing https://github.com/intel/llvm/pull/13616


source/adapters/cuda/enqueue.cpp
```diff
@@ -410,11 +409,43 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
+
+  // phEventWaitList only contains events that are handed to UR by the SYCL
+  // runtime. However since UR handles memory dependencies within a context
+  // we may need to add more events to our dependent events list if the UR
+  // context contains multiple devices
+  if (hQueue->getContext()->Devices.size() > 1) {
+    MemMigrationLocks.reserve(hKernel->Args.MemObjArgs.size());
+    for (auto &MemArg : hKernel->Args.MemObjArgs) {
+      bool PushBack = false;
+      if (auto MemDepEvent = MemArg.Mem->LastEventWritingToMemObj;
```
JackAKirk, line 429: All this memory dependence should only be happening for buffers? is that right? If so at this point can it distinguish usm from buffers? If not then is it a performance issue for usm?

JackAKirk: Probably a simpler question is does this patch change anything perf wise from the last time you did perf analysis of the multi-dev patch?

source/adapters/cuda/memory.cpp
```diff
@@ -573,6 +419,173 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemBufferPartition(
...
+                                  ur_device_handle_t hDevice) {
+  auto &Buffer = std::get<BufferMem>(Mem->Mem);
+  if (Mem->LastEventWritingToMemObj == nullptr) {
+    // Device allocation being initialized from host for the first time
+    if (Buffer.HostPtr) {
+      UR_CHECK_ERROR(
+          cuMemcpyHtoD(Buffer.getPtr(hDevice), Buffer.HostPtr, Buffer.Size));
+    }
+  } else if (Mem->LastDeviceWritingToMemObj != hDevice) {
+    UR_CHECK_ERROR(cuMemcpyDtoD(Buffer.getPtr(hDevice),
```
JackAKirk, line 499: nit: cuMemcpyDtoD does not copy directly between devices, the semantics of it is device A to device A. I vaguely remember that it does work to copy device A to device B via host, but that if i remember correctly it was slower than manually doing cuMemcpyDtoH for device A to host then cuMemcpyHtoD for host to device B.

But once you get into this then it raises the issue that if you want best performance copies should be done P2P, which raises a whole host of other issues like max peers etc. So probably this isn't something worth worrying about at this point.

source/adapters/cuda/image.cpp
```diff
@@ -492,9 +502,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageFreeExp(
 UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesUnsampledImageCreateExp(
     ur_context_handle_t hContext, ur_device_handle_t hDevice,
     ur_exp_image_mem_handle_t hImageMem, const ur_image_format_t *pImageFormat,
-    const ur_image_desc_t *pImageDesc, ur_mem_handle_t *phMem,
+    const ur_image_desc_t *, ur_mem_handle_t *,
```
Seanst98, line 505: What's the purpose of this change? I see elsewhere in this function that you're making a change similar to what's being done in https://github.com/oneapi-src/unified-runtime/pull/1512. 

However, the change here isn't quite the same. Maybe a bad rebase? Either way, I don't think this PR needs to or should depend on the aforementioned one.

The same applies to `urBindlessImagesSampledImageCreateExp`.

source/adapters/cuda/enqueue.cpp
```diff
@@ -565,45 +618,61 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferReadRect(
...
-        hostRowPitch, hostSlicePitch);
+        hostRowPitch, hostSlicePitch));
 
     if (phEvent) {
       UR_CHECK_ERROR(RetImplEvent->record());
     }
 
     if (blockingRead) {
-      UR_CHECK_ERROR(cuStreamSynchronize(CuStream));
+      UR_CHECK_ERROR(cuStreamSynchronize(Stream));
```
JackAKirk, line 665: It seems like blockingRead is set false in dpc++ memory manager? https://github.com/intel/llvm/blob/c74cebdb19b5cd99f66afe781cd3874e88f66308/sycl/source/detail/memory_manager.cpp#L613

I didn't look thoroughly and I'm far from certain but to me it looks like possibly this change will break buffer memory consistency if you have two devices, 
- one writing to the buffer on device at same time as..
- other copying with Stream to host for this read?

If something like this is the case then I think correctness should be prioritized over performance: whatever happens unless/until until p2p is somehow enabled for multidevice buffers performance is going to be worse than using usm and P2P manually, which also in any case seems to be the priority for the majority of users. I think we should prioritize fully supporting multi device functionality with usm that this patch will enable, then having buffer multi device performance as a secondary thing. Although ideally it wouldn't make multi device buffer code any slower than the existing routes that does D2H H2D copies via the existing one device in one sycl context state of affairs.

Seanst98 (changes requested):

source/adapters/cuda/enqueue.cpp
JackAKirk: Sorry I've read more of the patch now and possibly I see this will only affect buffers. If this is the case then just ignore the above.

hdelan: Yep this is only the case for buffers/images. `Args.MemObjArgs` doesn't include USM pointers

source/adapters/cuda/image.cpp
hdelan: I didn't want to explicitly depend on https://github.com/oneapi-src/unified-runtime/pull/1512 since that would introduce a lot of noise in my PR, and make it harder to review. But this PR does indeed depend on https://github.com/oneapi-src/unified-runtime/pull/1512

source/adapters/cuda/enqueue.cpp
hdelan: This patch uses locks and events to ensure that a buffer cannot be read from at the same time as an operation is writing to it. If I want to read from a buffer I need to wait until the last event writing to that buffer has completed, and then a memcpy will happen from the device where the buffer was last written to.

hdelan: See here for a more detailed explanation https://github.com/oneapi-src/unified-runtime/pull/1565/files#diff-28b062fd8949b104c030051b807123eb1f86396e56d7524e1fdae76187d02a70R335

JackAKirk: OK, it's not obvious to me how the lock works the prevent the above hypothetical issue in all cases since these operations are asynchronous, but I'll take your word for it: I can see that you wait on the stream that does the read, so I guess that you've made it such that all other operations acting on the buffer memory  for all other hypothetical devices also wait for stream completion.
For the reasons listed above regarding performance priorities I think this is all fine IMO.

Seanst98 (changes requested):
I tested this locally and found nearly half of the bindless tests are failing:

```
********************
********************
Failed Tests (13):
  SYCL :: bindless_images/cubemap/cubemap_sampled.cpp
  SYCL :: bindless_images/mipmap/mipmap_read_2D.cpp
  SYCL :: bindless_images/mipmap/mipmap_read_3D.cpp
  SYCL :: bindless_images/read_2D.cpp
  SYCL :: bindless_images/read_2D_dynamic.cpp
  SYCL :: bindless_images/read_3D.cpp
  SYCL :: bindless_images/read_sampled.cpp
  SYCL :: bindless_images/sampling_2D.cpp
  SYCL :: bindless_images/sampling_2D_USM_shared.cpp
  SYCL :: bindless_images/sampling_2D_half.cpp
  SYCL :: bindless_images/sampling_3D.cpp
  SYCL :: bindless_images/sampling_unique_addr_modes.cpp
  SYCL :: bindless_images/user_types/mipmap_read_user_type_2D.cpp

********************
Unexpectedly Passed Tests (1):
  SYCL :: bindless_images/device_to_device_copy.cpp
```

JackAKirk:
I think that so long as this passes all the tests then it looks good to me.

hdelan:
> I tested this locally and found nearly half of the bindless tests are failing:

Not able to reproduce locally but I'm investigating.

hdelan:
> I tested this locally and found nearly half of the bindless tests are failing:

Fixed

source/adapters/cuda/image.cpp
Seanst98: Knowing that keeping `phMem` around would require non-trivial changes, I'm more than happy to include the removal here. Resolving.

Seanst98:
> > I tested this locally and found nearly half of the bindless tests are failing:
> 
> Fixed

The tests now pass for me.

Seanst98 (approved):
`image.cpp` changes LGTM

source/adapters/cuda/enqueue.cpp
hdelan: The operations are not asynchronous, the work that the operations enqueue is async. Use of locks like this enables safety in a multi threaded environment.

source/adapters/cuda/memory.cpp
hdelan: I think I'm happy with leaving this as is. If there is some perf gain to using DtoD vs DtoH + HtoD I can make a ticket to benchmark this difference.

source/adapters/cuda/enqueue.cpp
JackAKirk: Forgetting about multithreaded usage which I think is so niche in gpu projects outside of MPI (which is multi process rather than multithread within a single process so not affected by these locks anyway) to make it pretty much irrelevant.

My question is what happens if a single host thread
- 1. operates on buffer memory "mem" on device A using an async operation using e.g. stream 1 on device A
- 2. Following this it operates on buffer memory "mem" to copy it to device B using e.g. stream 2 on device B

I believe Sycl programming model mandates that the memory device B sees is consistent with the memory after the async operation on device A. Does the sycl scheduler still manage this dependency and enforce this memory consistency constraint? Or is it now the responsibility of the plugin? If it is the responsibility of the plugin then does it guarantee correctness of the example I gave (and all other possible combinations of operations)?

source/adapters/cuda/memory.cpp
JackAKirk: One other way would be to put two cuda devices into different "sycl contexts": the sycl runtime is still responsible for doing the copy for such a case, and it does it via D2H H2D. So then if that is not faster than doing it for two devs within a single "sycl context" then probs we are all good.
I think doing this performance comparison (two devices in different sycl contexts and two devices within one) would be good irrespective, just to confirm that there is no other potential performance regression issue beyond the the D2H H2D vs "D2D" one I flagged.

source/adapters/cuda/enqueue.cpp
hdelan: Each `ur_mem_handle_t` has a `ur_event_handle_t LastEventWritingToMemObj`. If an operation writes to mem A, then this event will be set. When another operation wants to read from this mem A, either on the same or another device, then the plugin will ensure that the new command doesn't start (if same device)/isn't enqueued (for different device) until `LastEventWritingToMemObj` completes.

In the case of a single device ctx we just use the normal scheduling with streams to ensure the next enqueued operation only starts after `LastEventWritingToMemObj` completes. If we need to access data on a different device in the ctx, then the enqueue of the new command will need to wait on `LastEventWritingToMemObj` to complete.

JackAKirk: Great thanks, it all makes sense to me. Just wanted to make sure.

JackAKirk (approved):
LGTM providing it is confirmed there is no performance regression wrt buffers for two devices using different sycl contexts (see my other comments).

source/adapters/cuda/enqueue.cpp
```diff
@@ -410,11 +409,43 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
 
+  std::vector<ur_event_handle_t> DepEvents(
+      phEventWaitList, phEventWaitList + numEventsInWaitList);
+  std::vector<std::pair<ur_mem_handle_t, ur_lock>> MemMigrationLocks;
+
+  // phEventWaitList only contains events that are handed to UR by the SYCL
+  // runtime. However since UR handles memory dependencies within a context
+  // we may need to add more events to our dependent events list if the UR
+  // context contains multiple devices
+  if (hQueue->getContext()->Devices.size() > 1) {
```
mmoadeli, line 425: I'd use !hQueue->getContext()->Devices.empty() to avoid un-necessary counting to size.

source/adapters/cuda/enqueue.cpp
```diff
@@ -426,26 +457,33 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
 
+    ScopedContext Active(hQueue->getDevice());
     uint32_t StreamToken;
     ur_stream_guard_ Guard;
     CUstream CuStream = hQueue->getNextComputeStream(
         numEventsInWaitList, phEventWaitList, Guard, &StreamToken);
 
-    Result = enqueueEventsWait(hQueue, CuStream, numEventsInWaitList,
-                               phEventWaitList);
+    if (DepEvents.size()) {
```
mmoadeli, line 476: !DepEvents.empty() ?

source/adapters/cuda/enqueue.cpp
```diff
@@ -426,26 +457,33 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
+      UR_CHECK_ERROR(enqueueEventsWait(hQueue, CuStream, numEventsInWaitList,
+                                       phEventWaitList));
+    }
+
+    // For memory migration across devices in the same context
+    if (hQueue->getContext()->Devices.size() > 1) {
```
mmoadeli, line 482: !empty()

source/adapters/cuda/enqueue.cpp
```diff
@@ -454,6 +492,21 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
       UR_CHECK_ERROR(RetImplEvent->start());
     }
 
+    // Once event has been started we can unlock MemoryMigrationMutex
+    if (hQueue->getContext()->Devices.size() > 1) {
```
mmoadeli, line 496: !empty() ?

mmoadeli (approved):

mmoadeli:
This is a relatively large patch. Would it be possible to add a test to show how to make use of the feature?

source/adapters/cuda/enqueue.cpp
hdelan: We need to check that the number of devices is greater than 1, not greater than 0. So i can't change the check here.

source/adapters/cuda/enqueue.cpp
hdelan: I think this is equivalent and `.size()` is more readable than `!....empty()`

source/adapters/cuda/enqueue.cpp
hdelan: Same as above

source/adapters/cuda/enqueue.cpp
hdelan: Same as above

JackAKirk:
> This is a relatively large patch. Would it be possible to add a test to show how to make use of the feature?

Since CI only has a single device, any multi-device functionality is not tested there; however there are a few tests for multi-device features. For example these p2p tests that I can update following this patch as per (https://github.com/intel/llvm/compare/sycl...JackAKirk:llvm:p2p-make-test-portable). I remember that @hdelan already manually tested these on a multi-gpu system, so if @hdelan can confirm there are no issues with this I think that this patch can be merged without any more testing so long as the existing tests for single device functionality are not broken.

source/adapters/cuda/enqueue.cpp
mmoadeli: performance wise using empty() should be preferred,  I don't have hard feeling for that though.

pbalcer:
> Since CI only has a single device, any multi-device functionality

There's a multi-gpu nvidia system in CI. The runner tag is `CUDA_E2E`.

JackAKirk:
> > Since CI only has a single device, any multi-device functionality
> 
> There's a multi-gpu nvidia system in CI. The runner tag is `CUDA_E2E`.

is that unified-runtime or intel/llvm or both?

If intel/llvm now has multi-gpu then the P2P test could be updated there and tested with this branch if people desire it.
In any case I would also update the test in a follow on pr so that it runs for multi devices in a context for cuda. I did not worry about that for this PR since I didn't believe it could be triggered to run with the CI since it used to only have a single device.

source/adapters/cuda/enqueue.cpp
mmoadeli: my impression is that empty() should be implemented as `return begin() == end()`. but size() has to count.  I have no hard feeling about that and depending on size one may be more performant.

pbalcer:
In this repo, `unified-runtime`. The default tag used to run tests is `CUDA` that includes all systems with CUDA support, including single- and multi- gpu ones.

If you'd want to create multi-gpu specific tests, we'd need to create a new workflow that uses just the `CUDA_E2E` runners and runs just those specific tests.

isaacault (approved):

JackAKirk:
> In this repo, `unified-runtime`. The default tag used to run tests is `CUDA` that includes all systems with CUDA support, including single- and multi- gpu ones.
> 
> If you'd want to create multi-gpu specific tests, we'd need to create a new workflow that uses just the `CUDA_E2E` runners and runs just those specific tests.

Thanks for the info, I didn't know this. I can look into changing the workflow of https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/exp_usm_p2p/usm_p2p.cpp to make sure it runs on CUDA_E2E if it doesn't already. However I'm not sure that test really interacts with ur context, and I think the multi-device test in intel/llvm is more appropriate.

DBDuncan (approved):

hdelan:
On a multi GPU system E2E tests are passing:
```
➜  build git:(cuda-multi-dev-ctx) LIT_FILTER_OUT="acq_rel|bfloat16_builtins.cpp" ninja check-sycl-e2e
[0/2] Re-checking globbed directories...
[0/1] Running SYCL End-to-End tests
lit.py: /home/hugh/llvm/sycl/test-e2e/lit.cfg.py:412: note: Targeted devices: ext_oneapi_cuda:gpu
lit.py: /home/hugh/llvm/sycl/test-e2e/lit.cfg.py:577: warning: Couldn't find pre-installed AOT device compiler ocloc
lit.py: /home/hugh/llvm/sycl/test-e2e/lit.cfg.py:574: note: Found pre-installed AOT device compiler opencl-aot
lit.py: /home/hugh/llvm/sycl/test-e2e/lit.cfg.py:671: note: Aspects for cuda:gpu: ext_oneapi_bindless_images_shared_usm, ext_oneapi_fixed_size_group, ext_intel_memory_bus_width, ext_oneapi_opportunistic_group, usm_shared_allocations, usm_atomic_shared_allocations, fp64, ext_oneapi_bindless_sampled_image_fetch_1d_usm, ext_intel_pci_address, ext_intel_legacy_image, online_linker, usm_atomic_host_allocations, ext_oneapi_interop_semaphore_import, ext_oneapi_mipmap_anisotropy, ext_oneapi_graph, usm_host_allocations, ext_oneapi_bindless_images_2d_usm, ext_oneapi_ballot_group, ext_intel_free_memory, ext_intel_device_id, gpu, usm_device_allocations, ext_oneapi_limited_graph, ext_oneapi_cubemap_seamless_filtering, ext_oneapi_mipmap_level_reference, ext_oneapi_bindless_sampled_image_fetch_2d_usm, ext_oneapi_bindless_sampled_image_fetch_3d, ext_oneapi_interop_memory_import, atomic64, ext_oneapi_bindless_sampled_image_fetch_2d, ext_oneapi_bindless_images, ext_oneapi_cubemap, ext_intel_memory_clock_rate, queue_profiling, fp16, ext_intel_device_info_uuid, ext_oneapi_mipmap, online_compiler
lit.py: /home/hugh/llvm/sycl/test-e2e/lit.cfg.py:683: note: SG sizes for cuda:gpu: 32

Testing Time: 1166.03s

Total Discovered Tests: 2054
  Excluded         :    3 (0.15%)
  Unsupported      : 1142 (55.60%)
  Passed           :  901 (43.87%)
  Expectedly Failed:    8 (0.39%)

1 warning(s) in tests
➜  build git:(cuda-multi-dev-ctx) clang++ --version
clang version 19.0.0git (git@github.com:hdelan/llvm.git 8330017e7e013be9448a38c8b75e2ad3987c0b85)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /home/hugh/llvm/build/bin
Build config: +assertions
➜  build git:(cuda-mulit-dev-ctx) sycl-ls
[cuda:gpu][cuda:0] NVIDIA CUDA BACKEND, NVIDIA GeForce GTX 1050 Ti 6.1 [CUDA 12.4]
[cuda:gpu][cuda:1] NVIDIA CUDA BACKEND, NVIDIA GeForce GTX 1050 Ti 6.1 [CUDA 12.4]
```

hdelan:
Microbenchmark on multi device system

| Branch | Benchmark | File | Time (us) |
| -- | -- | -- | 
| `main` | Different context, different device migration | [diff_context_diff_platform.cpp.txt](https://github.com/oneapi-src/unified-runtime/files/15389016/diff_context_diff_platform.cpp.txt) | 90969 |
| `multi-device-ctx` | Different context, different device migration | [different_context.cpp.txt](https://github.com/oneapi-src/unified-runtime/files/15389034/different_context.cpp.txt) | 90311 |
| `multi-device-ctx` | Same context, different device migration | [same_context.cpp.txt](https://github.com/oneapi-src/unified-runtime/files/15389030/same_context.cpp.txt) | 52250 |

Meaning this patch gives a no perf change for using per device contexts, and also gives a speedup for using multi device ctx.


Using system:

```
$ sycl-ls
[cuda:gpu][cuda:0] NVIDIA CUDA BACKEND, NVIDIA GeForce GTX 1050 Ti 6.1 [CUDA 12.4]
[cuda:gpu][cuda:1] NVIDIA CUDA BACKEND, NVIDIA GeForce GTX 1050 Ti 6.1 [CUDA 12.4]
```

hdelan:
Ping @oneapi-src/unified-runtime-command-buffer-write . @kbenzie once I get final approval this should be good to go

EwanC (approved):

kbenzie (approved):
Test changes LGTM

PR #1573, opened 2024-05-02
[CTS] Avoid division by zero (when there are no results)

frasercrmck:
When trying to run the CTS from a directory with no tests then the total number of tests is 0, which was causing issues when summarizing results.

### DIFF
--- a/scripts/ctest_parser.py
+++ b/scripts/ctest_parser.py
@@ -24,7 +24,7 @@ def get_cts_test_suite_names(working_directory):
     ]
 
 def percent(amount, total):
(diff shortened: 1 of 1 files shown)


frasercrmck:
Example of what happens with this fix:
```
[CTest Parser] Results:
    Total    [0]
    Passed   [0]    (0.0%) - (0.0% with skipped tests excluded)
    Skipped  [0]    (0.0%)
    Failed   [0]    (0.0%)
    Crashed  [0]    (0.0%)
```

kbenzie (approved):

PR #1576, opened 2024-05-03
Improve CTS device code generation

RossBrunton:
Some improvements to the device code generation cmake:
* Use `clang-offload-extract` instead of running the binary directly.
* Set up dependencies correctly so that re-running builds will generate missing IR files if they failed to build prior.
* Dynamically generate the list of device_code files - from now on, any .cpp file in the device_code directory will automatically be treated as device code and generate IR files.

This change drops CTS support for platforms that support multiple devices with different IR types, however the build system didn't support that anyway, nor do any of our adapters.

This change requires a `clang-offload-extract` binary to be available. By default it is assumed to be in the same directory as the provided DPC++ binary (which should work for most builds), however the CMake variable `UR_DEVICE_CODE_EXTRACTOR` is provided to override this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ List of options provided by CMake:
 | UR_HIP_PLATFORM         | Build HIP adapter for AMD or NVIDIA platform           | AMD/NVIDIA | AMD     |
 | UR_ENABLE_COMGR         | Enable comgr lib usage           | AMD/NVIDIA | AMD     |
 | UR_DPCXX | Path of the DPC++ compiler executable to build CTS device binaries | File path | `""` |
+| UR_DEVICE_CODE_EXTRACTOR | Path of the `clang-offload-extract` executable from the DPC++ package, required for CTS device binaries | File path | `"${dirname(UR_DPCXX)}/clang-offload-extract"` |
 | UR_DPCXX_BUILD_FLAGS | Build flags to pass to DPC++ when compiling device programs | Space-separated options list | `""` |
 | UR_SYCL_LIBRARY_DIR | Path of the SYCL runtime library directory to build CTS device binaries | Directory path | `""` |
 | UR_HIP_ROCM_DIR | Path of the default ROCm HIP installation | Directory path | `/opt/rocm` |
--- a/test/conformance/device_code/CMakeLists.txt
+++ b/test/conformance/device_code/CMakeLists.txt
@@ -10,6 +10,10 @@ else()
     set(AMD_ARCH "${UR_CONFORMANCE_AMD_ARCH}")
 endif()
 
+cmake_path(GET UR_DPCXX EXTENSION EXE)
+cmake_path(REPLACE_FILENAME UR_DPCXX "clang-offload-extract${EXE}" OUTPUT_VARIABLE DEFAULT_EXTRACTOR_NAME)
+set(UR_DEVICE_CODE_EXTRACTOR "${DEFAULT_EXTRACTOR_NAME}" CACHE PATH "Path to clang-offload-extract")
+
 if("${AMD_ARCH}" STREQUAL "" AND "${TARGET_TRIPLES}" MATCHES "amd")
     find_package(RocmAgentEnumerator)
     if(NOT ROCM_AGENT_ENUMERATOR_FOUND)
@@ -59,6 +63,8 @@ macro(add_device_binary SOURCE_FILE)
 
     foreach(TRIPLE ${TARGET_TRIPLES})
         set(EXE_PATH "${DEVICE_BINARY_DIR}/${KERNEL_NAME}_${TRIPLE}")
+        set(BIN_PATH "${DEVICE_BINARY_DIR}/${TRIPLE}.bin.0")
+
         if(${TRIPLE} MATCHES "amd")
             set(AMD_TARGET_BACKEND -Xsycl-target-backend=${TRIPLE})
             set(AMD_OFFLOAD_ARCH  --offload-arch=${AMD_ARCH})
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -141,6 +141,7 @@ List of options provided by CMake:
 | UR_HIP_PLATFORM         | Build HIP adapter for AMD or NVIDIA platform           | AMD/NVIDIA | AMD     |
 | UR_ENABLE_COMGR         | Enable comgr lib usage           | AMD/NVIDIA | AMD     |
 | UR_DPCXX | Path of the DPC++ compiler executable to build CTS device binaries | File path | `""` |
+| UR_DEVICE_CODE_EXTRACTOR | Path of the `clang-offload-extract` executable from the DPC++ package, required for CTS device binaries | File path | `"${dirname(UR_DPCXX)}/clang-offload-extract"` |
```
kbenzie, line 144: Does `clang-offload-extract` need to know where `libsycl.so` lives?

test/conformance/device_code/CMakeLists.txt
```diff
@@ -10,6 +10,10 @@ else()
     set(AMD_ARCH "${UR_CONFORMANCE_AMD_ARCH}")
 endif()
 
+cmake_path(GET UR_DPCXX EXTENSION EXE)
```
kbenzie, line 13: Looks like [cmake_path]() was added in CMake 3.20 but we need to support CMake 3.14 for now. Depends what llvm-project or intel/llvm's minimum requirements are actually, if those are above 3.14 then we should also update the `project()` command in the root CMakeLists.txt.

test/conformance/device_code/CMakeLists.txt
```diff
@@ -81,41 +87,26 @@ macro(add_device_binary SOURCE_FILE)
...
-add_device_binary(${CMAKE_CURRENT_SOURCE_DIR}/cpy_and_mult_usm.cpp)
-add_device_binary(${CMAKE_CURRENT_SOURCE_DIR}/multiply.cpp)
-add_device_binary(${CMAKE_CURRENT_SOURCE_DIR}/spec_constant.cpp)
-add_device_binary(${CMAKE_CURRENT_SOURCE_DIR}/spec_constant_multiple.cpp)
-add_device_binary(${CMAKE_CURRENT_SOURCE_DIR}/usm_ll.cpp)
-add_device_binary(${CMAKE_CURRENT_SOURCE_DIR}/saxpy.cpp)
-add_device_binary(${CMAKE_CURRENT_SOURCE_DIR}/saxpy_usm.cpp)
-add_device_binary(${CMAKE_CURRENT_SOURCE_DIR}/indexers_usm.cpp)
-add_device_binary(${CMAKE_CURRENT_SOURCE_DIR}/build_failure.cpp)
+file(GLOB SOURCE_FILES "*.cpp")
```
kbenzie, line 106: Using `GLOB` has caveats, if a new `.cpp` file is added it'll only get picked up when CMake is reconfigured which is not intuitive. This is generally why we've avoided it in favour of using explicit lists of source files.

test/conformance/source/environment.cpp
```diff
@@ -382,66 +382,44 @@ std::string KernelsEnvironment::getSupportedILPostfix(uint32_t device_index) {
...
+    case UR_PLATFORM_BACKEND_OPENCL:
+    case UR_PLATFORM_BACKEND_LEVEL_ZERO:
+        return "spir64";
+    case UR_PLATFORM_BACKEND_CUDA:
+        return "nvptx64-nvidia-cuda";
+    case UR_PLATFORM_BACKEND_HIP:
+        return "amdgcn-amd-amdhsa";
+    case UR_PLATFORM_BACKEND_NATIVE_CPU:
+        error = "native_cpu doesn't support kernel tests yet";
         return {};
```
kbenzie, line 397: Should we abort here to make it obvious something needs to change when support is being added?

kbenzie (commented):
Overall looks good, few minor points but the CMake version issue will deffo need fixed.

test/conformance/source/environment.cpp
RossBrunton: Aborting here would cause the entire test suite to die, which means we wouldn't be able to run test suites that have a mixture of "device code" and "non-device code" tests in them.

Any tests that use device code will check that they are loaded correctly (i.e. this function returns something non-empty), and if not they will fail quickly.

test/conformance/device_code/CMakeLists.txt
RossBrunton: Reverted this.

test/conformance/source/environment.cpp
kbenzie: Ah right. Makes sense.

test/conformance/device_code/CMakeLists.txt
kbenzie: Addressed in #1583

kbenzie (approved):

RossBrunton:
> Since `UR_DEVICE_CODE_EXTRACTOR` is the preferred method moving forwards, should we emit a CMake warning when `UR_DPCXX`/`UR_SYCL_LIBRARY_DIR` are used suggesting to switch to `UR_DEVICE_CODE_EXTRACTOR`?
> 
> I'm also wondering if the GitHub Actions need updated.

`UR_DPCXX` (and thus `UR_SYCL_LIBRARY_DIR`) are both still required for initial generation of the binary. I considered allowing the user to specify the root of the DPC++ package itself as a path, but I think the name of the compiler itself can change depending on the exact package (clang++/dpcp++/icpx).

The default logic works with the packages used by GitHub, so I'm not sure it's worth updating it.

PR #1581, opened 2024-05-04
Implement L0 cooperative kernel functions

0x12CC:
Defines `urKernelSuggestMaxCooperativeGroupCountExp` and `urEnqueueCooperativeKernelLaunchExp` to enable cooperative kernels with more than one work group.


kbenzie:
It would be great to have some UR tests for cooperative kernels now that there is an implementation, is there a plan for that @0x12CC?

FYI that doesn't block this PR which is now in the merge queue.

PR #1582, opened 2024-05-06
[NFC] Bump CMake verson to 0.10.0

kbenzie:
Bumping to v0.10.0 exposed a bug in the spec generation scripts where versions were being converted to `float` and then used doing version comparisons. Converting `0.10` to `float` results in `0.1` which is an incorrect representation of that version. To address this the scripts have been updated, including the new `Version` class which handles version comparisons, to never use `float` to represent versions.

### DIFF
--- a/scripts/run.py
+++ b/scripts/run.py
@@ -13,9 +13,12 @@ import util
 import parse_specs
 import generate_code
 import generate_docs
-import os, sys, platform
+import os
+import sys
+import platform
 import time
 import subprocess
+from version import Version
 
 """
     helper for adding mutually-exclusive boolean arguments "--name" and "--!name"
@@ -95,7 +98,7 @@ def get_version_from_cmakelists():
         for line in cmakelists_file.readlines():
             line = line.strip()
             if line.startswith('project('):
-                return re.findall(r'\d+\.\d+', line)[0]
+                return Version(re.findall(r'\d+\.\d+', line)[0])
     raise Exception(f'unable to read project version from {cmakelists_path}')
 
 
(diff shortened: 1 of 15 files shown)


aarongreig (approved):

callumfare (approved):

scripts/templates/__init__.py
```diff
@@ -1,8 +1,6 @@
 """
  Copyright (C) 2023 Intel Corporation
-
  Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
  See LICENSE.TXT
  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-
```
omarahmed1111, line 7: These 2 blank lines exist in all other files so we should keep them for consistency I think.

omarahmed1111 (approved):
LGTM, just a small nit suggestions and a question.

scripts/run.py
```diff
@@ -136,82 +144,75 @@ def main():
         'meta'   : {},
         'ref'    : {}
         }
-    
-    try:
```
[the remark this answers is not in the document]
kbenzie: When the scripts were failing this `try`/`except` block was hiding the exception back trace and simply printing *Failed to generate specification*. That was really unhelpful when trying to figure out what was going wrong. Removing that means the back trace is printed and information is not lost.

scripts/templates/__init__.py
kbenzie: yeah I accidentally deleted this, will fix

PR #1586, opened 2024-05-08
[CTS] Added test for sequencing events

RossBrunton:
This tests that the `urEnqueueEventsWaitWithBarrier` function
functions as a barrier. That is, it blocks until all events prior
to it in the queue are completed. Since this test may fail
non-deterministically, it is ran a few times.

Note that current adapters seem to always execute events
sequentially, so this is a no-op.


aarongreig (approved):

RossBrunton:
@aarongreig You had a look at this way back in May, but it's been tweaked a bit since then. Mind giving it another look over just to ensure everything is still okay?

aarongreig:
actually did you look into what's up with the level zero fail? in case it's something trivial

pbalcer:
> actually did you look into what's up with the level zero fail? in case it's something trivial

I was meaning to ask the same question. If barriers don't work in the L0 adapter, we need to fix that asap.

RossBrunton:
Looks like it was complaining about the test not failing when the match file said it should. Probably due to an earlier version of the test being broken (I was breaking this rule https://github.com/oneapi-src/unified-runtime/pull/2195 ).

Hopefully this means that it's all actually passing on non-native-cpu targets.

RossBrunton:
Only adds conformance tests so should have no effect on intel/llvm. Merging without llvm change.

PR #1596, opened 2024-05-10
[CUDA] Change `bool` variable type to `unsigned int` 

EwanC:
Seeing an error in Windows debug build about corrupted stack memory around the `IsManaged` variable which goes away if I change its type to `unsigned int`.


fabiomestre (approved):

fabiomestre:
Nice find :+1: . Have seen this kind of error in Cuda before.

hdelan (approved):
Nice one

PR #1598, opened 2024-05-13
Add more output validation for urEventGetProfilingInfo test

omarahmed1111:
Validate the output timings for `urEventGetProfilingInfo` entry-point.

[intel/llvm testing](https://github.com/intel/llvm/pull/13778)


test/conformance/event/event_adapter_cuda.match
```diff
@@ -1,3 +1,4 @@
+urEventGetProfilingInfoTest.Success/NVIDIA_CUDA_BACKEND___{{.*}}_
```
aarongreig, line 1: all the adapters apart from native cpu do implement at least some of these queries, why are all the tests failing?

omarahmed1111: I just merged them in one test to save the timing and compare it at the end of queries. The main query that fail on all adapters is `UR_EVENT_STATUS_COMPLETE`

aarongreig: I'd rather we were precise about which tests are failing, this could hide a future regression in a query that currently works

omarahmed1111: Yeah, that's a good point. I have split that to 2 tests, one concerned with testing timings, and another for testing the success for the entry-point

aarongreig (approved):

omarahmed1111:
@oneapi-src/unified-runtime-cuda-write Could I have a review on this when available. Thanks!

hdelan (approved):
CUDA adapter LGTM

PR #1412, opened 2024-03-05
[L0][OpenCL] Emulate Fill with copy when patternSize is not a power of 2

konradkusiak97:
LevelZero changes:
 - Extend the functionality of `enqueueMemFillHelper` to allow calling it with pattern sizes which are not powers of 2. In those cases filling is emulated with copying.

OpenCL changes:
- Add a condition of `isPowerOf2` to the USM fill function

Those changes are necessary for the PR: https://github.com/intel/llvm/pull/12702 which refactors `queue.fill()` to make use of the `urEnqueueUSMFill`. 

**intel/llvm** CI: https://github.com/intel/llvm/pull/12912

PR #1603, opened 2024-05-14
[OpenCL] Modify fill emulation to work for patterns which are not powers of 2

konradkusiak97:
This is a follow-up of https://github.com/oneapi-src/unified-runtime/pull/1412 which added the `isPowerOf2` condition to the OpenCL fill function. This is correct since `clEnqueueMemFillINTEL_fn` only accepts such patterns.

What was not correct was the later logic for emulating filling on the host and copying it to the destination ptr. It assumed that the pattern is greater than **128 bytes** but after adding the above `isPowerOf2` condition, it could also execute for smaller patterns which are simply not powers of 2. 

So this PR fixes my introduced bugs, **intel/llvm CI:** https://github.com/intel/llvm/pull/13779

### DIFF
--- a/source/adapters/opencl/usm.cpp
+++ b/source/adapters/opencl/usm.cpp
@@ -256,9 +256,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMFill(
     return UR_RESULT_SUCCESS;
   }
 
-  // OpenCL only supports pattern sizes as large as the largest CL type
-  // (double16/long16 - 128 bytes), anything larger we need to do on the host
-  // side and copy it into the target allocation.
+  // OpenCL only supports pattern sizes which are powers of 2 and are as large
+  // as the largest CL type (double16/long16 - 128 bytes), anything larger or
+  // not a power of 2, we need to do on the host side and copy it into the
+  // target allocation.
   clHostMemAllocINTEL_fn HostMemAlloc = nullptr;
   UR_RETURN_ON_FAILURE(cl_ext::getExtFuncFromContext<clHostMemAllocINTEL_fn>(
       CLContext, cl_ext::ExtFuncPtrCache->clHostMemAllocINTELCache,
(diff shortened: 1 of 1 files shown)


source/adapters/opencl/usm.cpp
```diff
@@ -275,14 +276,14 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMFill(
       cl_ext::MemBlockingFreeName, &USMFree));
 
   cl_int ClErr = CL_SUCCESS;
-  auto HostBuffer = static_cast<uint64_t *>(
+  auto HostBuffer = static_cast<unsigned char *>(
```
kbenzie, line 279: `uint8_t` would be preferable

source/adapters/opencl/usm.cpp
```diff
@@ -275,14 +276,14 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueUSMFill(
...
 
-  auto NumValues = size / sizeof(uint64_t);
-  auto NumChunks = patternSize / sizeof(uint64_t);
-  for (size_t i = 0; i < NumValues; i++) {
-    HostBuffer[i] = static_cast<const uint64_t *>(pPattern)[i % NumChunks];
+  auto NumChunks = size / patternSize;
+  for (size_t i = 0; i < NumChunks; i++) {
+    auto Dest = HostBuffer + i * patternSize;
+    memcpy(Dest, pPattern, patternSize);
   }
```
kbenzie, line 287: Maybe we don't need to divide `size` by `patternSize`?

```suggestion
  auto *End = HostBuffer + size;
  for (auto *Iter = HostBuffer; Iter < End; Iter += patternSize) {
    std::memcpy(Iter, pPattern, patternSize);
  }
```

Note this is completely untested so I could be totally wrong.

source/adapters/opencl/usm.cpp
konradkusiak97: done

source/adapters/opencl/usm.cpp
konradkusiak97: That's much better, thanks!

kbenzie (approved):

PR #1607, opened 2024-05-14
[L0] Optimize kernel lookup in enqueue

igchor:
Change unordered_map to vector as number of entries is expected to be low (1 in common case).

Also, do not store mapping between subdevices and the kernel handles. Instead, just look up kernel based on RootDevice handle for a sub device.

Additionally, use unique_ptr to store kernel_handle to avoid memory leaks in case zeKernelCreate fails.

### DIFF
--- a/source/adapters/level_zero/kernel.hpp
+++ b/source/adapters/level_zero/kernel.hpp
@@ -15,16 +15,36 @@
 
 struct ur_kernel_handle_t_ : _ur_object {
   ur_kernel_handle_t_(bool OwnZeHandle, ur_program_handle_t Program)
-      : Context{nullptr}, Program{Program}, ZeKernel{nullptr},
-        SubmissionsCount{0}, MemAllocs{} {
+      : Context{nullptr}, Program{Program}, SubmissionsCount{0}, MemAllocs{} {
     OwnNativeHandle = OwnZeHandle;
   }
 
   ur_kernel_handle_t_(ze_kernel_handle_t Kernel, bool OwnZeHandle,
                       ur_context_handle_t Context)
-      : Context{Context}, Program{nullptr}, ZeKernel{Kernel},
-        SubmissionsCount{0}, MemAllocs{} {
+      : Context{Context}, Program{nullptr}, SubmissionsCount{0}, MemAllocs{} {
     OwnNativeHandle = OwnZeHandle;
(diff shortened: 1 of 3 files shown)


igchor:
After https://github.com/oneapi-src/unified-runtime/pull/1611 I started to look into how rest of the functions in kernel.cpp are implemented and I think some of them might not handle kernels created from native handles and regular ones correctly (when kernel is created for more than one device). For example, here, https://github.com/oneapi-src/unified-runtime/blob/056d653264034e546d8b8f493e1d9f65c697829b/source/adapters/level_zero/kernel.cpp#L742 the properties are only set on the single ZeKernel while all others (from ZeKernelMap) are ignored. Similarly, the properties are only reported for that single `ZeKernel`: https://github.com/oneapi-src/unified-runtime/blob/056d653264034e546d8b8f493e1d9f65c697829b/source/adapters/level_zero/kernel.cpp#L876 

@nrspruit  @pbalcer Do you know if this is expected or should those functions actually set/query all ZeKernels from the ZeKernelMap?

source/adapters/level_zero/kernel.hpp
```diff
@@ -15,16 +15,36 @@
...
-      : Context{Context}, Program{nullptr}, ZeKernel{Kernel},
-        SubmissionsCount{0}, MemAllocs{} {
+      : Context{Context}, Program{nullptr}, SubmissionsCount{0}, MemAllocs{} {
     OwnNativeHandle = OwnZeHandle;
+    createdFromNativeHandle = true;
+    ZeKernels.push_back({nullptr, Kernel});
+  }
+
+  ze_kernel_handle_t getZeKernel(ur_device_handle_t Device) {
+    assert(Device);
```
pbalcer, line 31: The number of devices is basically set the moment we initialize the driver. In my event cache rewrite, I took advantage of that to give monotonically increasing ID's to the devices, allowing us to make this type of operation a direct lookup in an array.
https://github.com/pbalcer/unified-runtime/commit/147f0b4e4f2a63b2e6d0a19a7e690367f5c643bb#diff-89d38124dbcf1f91b31dc489a7e7c66bc0e6607ff77f2654810f74eae0826025R205

pbalcer (commented):
👍 the next step should be to get rid of `zeKernel` entirely :)

PR #1613, opened 2024-05-16
[CUDA][HIP] Minimize native events recorded and created by urEnqueueTimestampRecordingExp

hdelan:
Since EvStart and EvEnd are recorded directly after one another in `urEnqueueTimestampRecordingExp`, we can just copy EvStart to make EvEnd, instead of calling cuEventRecord for both `EvStart` and `EvEnd`, one after the other.

@steffenlarsen for discussion

### DIFF
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -1727,7 +1727,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueTimestampRecordingExp(
         std::unique_ptr<ur_event_handle_t_>(ur_event_handle_t_::makeNative(
             UR_COMMAND_TIMESTAMP_RECORDING_EXP, hQueue, CuStream));
     UR_CHECK_ERROR(RetImplEvent->start());
-    UR_CHECK_ERROR(RetImplEvent->record());
+    UR_CHECK_ERROR(RetImplEvent->make_end_event_same_as_start());
 
     if (blocking) {
       UR_CHECK_ERROR(cuStreamSynchronize(CuStream));
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -381,7 +381,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueEventsWaitWithBarrier(
       *phEvent = ur_event_handle_t_::makeNative(
           UR_COMMAND_EVENTS_WAIT_WITH_BARRIER, hQueue, CuStream, StreamToken);
       UR_CHECK_ERROR((*phEvent)->start());
-      UR_CHECK_ERROR((*phEvent)->record());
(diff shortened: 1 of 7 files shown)


steffenlarsen (approved):
I now get what you meant! I'm alright with this solution. 👍

hdelan:
> I now get what you meant! I'm alright with this solution. 👍

Great thanks. I'll add a DPC++ PR just to make sure no code is relying on `EvEnd` being after `EvStart`

source/adapters/cuda/event.cpp
```diff
@@ -129,6 +129,18 @@ ur_result_t ur_event_handle_t_::record() {
   return Result;
 }
 
+ur_result_t ur_event_handle_t_::make_end_event_same_as_start() {
+  if (isRecorded() || !isStarted()) {
+    return UR_RESULT_ERROR_INVALID_EVENT;
+  }
+  UR_ASSERT(Queue, UR_RESULT_ERROR_INVALID_QUEUE);
+
+  EvEnd = EvStart;
```
MartinWehking, line 138: Just a knit, but this function is and will always be called for events that are only visible inside the scope of another function (i.e., non "thread-escaping") ?
At least that's the case for urEnqueueTimestampRecordingExp if I see it correctly.
Otherwise, you might want to consider using mutexes at this point.

source/adapters/cuda/event.cpp
```diff
@@ -147,7 +159,9 @@ ur_result_t ur_event_handle_t_::release() {
 
   assert(Queue != nullptr);
 
-  UR_CHECK_ERROR(cuEventDestroy(EvEnd));
+  // Avoid double free if using timestamp
+  if (!isTimestampEvent())
+    UR_CHECK_ERROR(cuEventDestroy(EvEnd));
```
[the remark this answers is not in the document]
hdelan: No this can't happen IMO

source/adapters/cuda/enqueue.cpp
```diff
@@ -1727,7 +1727,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueTimestampRecordingExp(
         std::unique_ptr<ur_event_handle_t_>(ur_event_handle_t_::makeNative(
             UR_COMMAND_TIMESTAMP_RECORDING_EXP, hQueue, CuStream));
     UR_CHECK_ERROR(RetImplEvent->start());
-    UR_CHECK_ERROR(RetImplEvent->record());
```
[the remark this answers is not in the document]
hdelan: Good point. Have updated to add this in other commands

source/adapters/cuda/event.cpp
hdelan: This function is only called internally after a call to `start()` so there ought to never be any contention. For clarity I may combine this into a single call to `start`

hdelan: Have combined. See new diff.

npmiller:
This ended up introducing some bugs and we have refactored the base event since, so closing this.

PR #1619, opened 2024-05-16
Fix validation for urEnqueueTimestampRecordingExpTest

npmiller:
This was failing with the HIP adapter when the pointer is null and the event wait list is not 0.

### DIFF
--- a/source/loader/layers/validation/ur_valddi.cpp
+++ b/source/loader/layers/validation/ur_valddi.cpp
@@ -8975,6 +8975,14 @@ __urdlllocal ur_result_t UR_APICALL urEnqueueTimestampRecordingExp(
             return UR_RESULT_ERROR_INVALID_NULL_POINTER;
         }
 
+        if (phEventWaitList == NULL && numEventsInWaitList > 0) {
+            return UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST;
+        }
+
+        if (phEventWaitList != NULL && numEventsInWaitList == 0) {
+            return UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST;
+        }
+
         if (phEventWaitList != NULL && numEventsInWaitList > 0) {
             for (uint32_t i = 0; i < numEventsInWaitList; ++i) {
                 if (phEventWaitList[i] == NULL) {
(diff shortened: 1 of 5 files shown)


source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -8975,6 +8975,14 @@ __urdlllocal ur_result_t UR_APICALL urEnqueueTimestampRecordingExp(
...
         }
 
+        if (phEventWaitList == NULL && numEventsInWaitList > 0) {
+            return UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST;
+        }
+
+        if (phEventWaitList != NULL && numEventsInWaitList == 0) {
+            return UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST;
+        }
+
```
kbenzie, line 8985: This file is generated from the yaml specification, that's why the checks are failing. To make this change correctly, apply this patch then run the `generate` target and commit all the changes.

```diff
diff --git a/scripts/core/exp-enqueue-timestamp-recording.yml b/scripts/core/exp-enqueue-timestamp-recording.yml
index 18316f73..3dce02ae 100644
--- a/scripts/core/exp-enqueue-timestamp-recording.yml
+++ b/scripts/core/exp-enqueue-timestamp-recording.yml
@@ -63,4 +63,6 @@ params:
 returns:
     - $X_RESULT_ERROR_INVALID_NULL_HANDLE
     - $X_RESULT_ERROR_INVALID_NULL_POINTER
-    - $X_RESULT_ERROR_INVALID_EVENT_WAIT_LIST
+    - $X_RESULT_ERROR_INVALID_EVENT_WAIT_LIST:
+        - "phEventWaitList == NULL && numEventsInWaitList > 0"
+        - "phEventWaitList != NULL && numEventsInWaitList == 0"
```

source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -8975,6 +8975,14 @@ __urdlllocal ur_result_t UR_APICALL urEnqueueTimestampRecordingExp(
             return UR_RESULT_ERROR_INVALID_NULL_POINTER;
         }
 
+        if (phEventWaitList == NULL && numEventsInWaitList > 0) {
```
aarongreig, line 8978: this is a generated file, you'll want to add these [here](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/exp-enqueue-timestamp-recording.yml#L64) like [this](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/enqueue.yml#L60)

kbenzie: snap!

kbenzie (approved):

PR #1622, opened 2024-05-16
[HIP] Fix context devices property return value

npmiller:
This showed up in the UR CTS, it was returning a vector instead of the expected C array.


aarongreig (approved):

GeorgeWeb (approved):
Good catch. LGTM.

PR #1623, opened 2024-05-16
[L0] Create initial structure for supporting queue dispatcher

igchor:
This change moves all legacy queue code to a seprate class: ur_queue_handle_legacy_t_. and uses pointer to this class as a parameter in all internal functions (instead of using ur_queue_handle_t). ur_queue_handle_t is now a variant over this legacy class and a new queue dispatcher (not implemented yet). All API functions still take ur_queue_handle_t and convert it to ur_queue_handle_legacy_t using a helper function.

Once dispatcher is implemented there will be a seprate code path for each (affected) API function for legacy and new logic.

### DIFF
--- a/source/adapters/level_zero/image.cpp
+++ b/source/adapters/level_zero/image.cpp
@@ -753,12 +753,13 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesSampledImageCreateExp(
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
-    ur_queue_handle_t hQueue, void *pDst, void *pSrc,
+    ur_queue_handle_t hUrQueue, void *pDst, void *pSrc,
     const ur_image_format_t *pImageFormat, const ur_image_desc_t *pImageDesc,
     ur_exp_image_copy_flags_t imageCopyFlags, ur_rect_offset_t srcOffset,
     ur_rect_offset_t dstOffset, ur_rect_region_t copyExtent,
     ur_rect_region_t hostExtent, uint32_t numEventsInWaitList,
     const ur_event_handle_t *phEventWaitList, ur_event_handle_t *phEvent) {
+  auto hQueue = Legacy(hUrQueue);
   std::scoped_lock<ur_shared_mutex> Lock(hQueue->Mutex);
 
   UR_ASSERT(hQueue, UR_RESULT_ERROR_INVALID_NULL_HANDLE);
--- a/source/adapters/level_zero/queue.hpp
+++ b/source/adapters/level_zero/queue.hpp
@@ -27,9 +27,18 @@
 #include "common.hpp"
 #include "device.hpp"
 
+struct ur_queue_handle_legacy_t_;
+using ur_queue_handle_legacy_t = ur_queue_handle_legacy_t_ *;
+
 extern "C" {
 ur_result_t urQueueReleaseInternal(ur_queue_handle_t Queue);
 } // extern "C"
+
+namespace v2 {
+struct ur_queue_dispatcher_t {
+  // TODO
+};
+} // namespace v2
 
 struct ur_completion_batch;
 using ur_completion_batch_list = std::list<ur_completion_batch>;
@@ -70,7 +79,8 @@ struct ur_completion_batch {
 
   // Seals the event batch and appends a barrier to the command list.
   // Adding any further events after this, but before reset, is undefined.
-  ur_result_t seal(ur_queue_handle_t queue, ze_command_list_handle_t cmdlist);
+  ur_result_t seal(ur_queue_handle_legacy_t queue,
+                   ze_command_list_handle_t cmdlist);
 
   // Resets a complete batch back to an empty state. Cleanups internal state
   // but keeps allocated resources for reuse.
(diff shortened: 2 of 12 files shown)


source/common/ur_util.hpp
```diff
@@ -264,6 +264,10 @@ inline std::size_t combine_hashes(std::size_t seed, const T &v, Args... args) {
     return combine_hashes(seed ^ std::hash<T>{}(v), args...);
 }
 
+struct ur_exception_t {
```
pbalcer, line 267: what's the reason for wrapping this? Can't we just throw the ur_result_t, like it happens in other places?

pbalcer (approved):

source/common/ur_util.hpp
igchor: I was thinking about introducing additional struct and potentially inheriting from std::exception but I don't think we need it for now (it could be useful if we were to do some logging in catch()). I removed this.

igchor:
There is one function marked with `extern "C"` (EventCreate) that I'm changing the parameter type for (`ur_queue_handle_t` -> `ur_queue_handle_legacy_t`). 

@pbalcer @nrspruit do you know why this function is marked this way? I tried removing the `extern "C"` and UR compiled fine but I'm wondering if it isn't used somewhere outside UR and whether changing the parameter type won't break anything.

nrspruit (approved):
LGTM

source/adapters/level_zero/image.cpp
```diff
@@ -759,12 +759,13 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesSampledImageCreateExp(
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageCopyExp(
-    ur_queue_handle_t hQueue, void *pDst, void *pSrc,
+    ur_queue_handle_t UrQueue, void *pDst, void *pSrc,
```
isaacault, line 762: The rest of the params in this file follow the conventions of

1. camelCase, not PascalCase
2. Prefixes `p` for pointers, `h` for handles.

I'd prefer if this could remain compliant with these conventions, even if it meant changing more LOC.

I don't mind beyond that what naming you want in the various queue types, e.g.
```
auto hLegacyQueue = Legacy(hQueue);
```
or anything else.

isaacault (changes requested):

source/adapters/level_zero/queue.hpp
```diff
@@ -687,8 +699,38 @@ struct ur_queue_handle_t_ : _ur_object {
...
+struct ur_queue_handle_t_ {
+  template <typename QType, typename... Args>
+  ur_queue_handle_t_(std::in_place_type_t<QType> tag, Args &&...args)
+      : Queue(tag, std::forward<Args>(args)...) {
+    if constexpr (std::is_same_v<QType, ur_queue_handle_legacy_t_>) {
+      std::get<ur_queue_handle_legacy_t_>(Queue).UnifiedHandle = this;
+    }
+  }
+
+  std::variant<ur_queue_handle_legacy_t_, v2::ur_queue_dispatcher_t> Queue;
```
pbalcer, line 717: do we need the intermediate type `v2::ur_queue_dispatcher_t` ? Wouldn't this be simpler (and with one less cache-miss) if we just directly stored the appropriate queue types?

pbalcer (commented):
In the next PR we need to introduce the new 'dispatcher' class (or whatever new queue we want to create), a way to enable it, and a test.

I think it would be ideal if we kept the v2 implementation entirely in its own directory.

source/adapters/level_zero/queue.hpp
igchor: I assume that for the new path, we will want to create a single API (enqueue(), synchronize(), etc.) that would be implemented by the dispatcher. When, for example, enqueue() is called on a dispatcher, it would just forward the call to the specific implementation (a specific alternative in the variant).

If we make the legacy path part of the dispatcher, we would need to implement those functions for the legacy path as well (I don't think that makes much sense since the legacy queue does not have a very well-defined interface) or special-case the legacy path. I think the approach with 2 level variant just makes it a bit more explicit and clear (but there's no significant difference otherwise IMO).

pbalcer: > I assume that for the new path, we will want to create a single API (enqueue(), synchronize(), etc.) that would be implemented by the dispatcher. When, for example, enqueue() is called on a dispatcher, it would just forward the call to the specific implementation (a specific alternative in the variant).

I was thinking that a dispatcher would just have all the functions like urEnqueueKernelLaunch, and call them directly for each API entry point. Things like enqueue() would be just some shared helper between all 'new' implementations you could use to construct an enqueue function.

> I think the approach with 2 level variant just makes it a bit more explicit and clear (but there's no significant difference otherwise IMO).

Now that I understand the intention what's meant to be inside a dispatcher (I thought you wanted it to be a pointer to some object) I agree that the difference is inconsequential.

igchor: > I was thinking that a dispatcher would just have all the functions like urEnqueueKernelLaunch, and call them directly for each API entry point.

I see, yes, that would also be possible although probably a bit more verbose - we would need to define all those functions for all queue alternatives. 

Also, I think there are some interactions with different entities (kernel, event, etc.). Right now, those other entities just use variables from the queue directly so we would probably need to abstract this somehow.

pbalcer (approved):

EwanC (approved):

isaacault (approved):

cppchedy (approved):

DBDuncan (approved):

kbenzie:
@igchor if you could fix the merge conflict that would be great.

igchor:
> @igchor if you could fix the merge conflict that would be great.

@kbenzie Done. Sorry for the late reply, I was OOO for the last two weeks.

kbenzie:
> > @igchor if you could fix the merge conflict that would be great.
> 
> @kbenzie Done. Sorry for the late reply, I was OOO for the last two weeks.

No worries, thanks @igchor

PR #1624, opened 2024-05-17
[CI] Fix E2E workflow

lukaszstolarczuk:
Use test filters only when they are set, which is for L0. Otherwise we get 'error: filter did not match any tests'.


pbalcer (approved):
Might be better to set filters only if they are set, not based on the adapter. But let's merge this for now.

PR #1629, opened 2024-05-20
Use fence rather than event for sync in L0 command-buffer update 

EwanC:
The Level Zero adapter implementing `urCommandBufferUpdateKernelLaunchExp` is doing a blocking host wait with `zeEventHostSynchronize` on the executing of the command-buffer.

However, there is no guarantee that the command-buffer has been enqueued before update has been called. Resulting in a hang if this is not the case. Even if a command-buffer has been enqueued before, the L0 event primitive can't guarantee that the command-list has finished executing, only a specific command.

Instead, replace the blocking wait on a L0 event with a blocking wait on a L0 fence from the last submission of the command-buffer. This is because fences are the only safe way to track when a command-buffer execution has completed.

DPC++ PR https://github.com/intel/llvm/pull/13832

### DIFF
--- a/scripts/core/exp-command-buffer.yml
+++ b/scripts/core/exp-command-buffer.yml
@@ -900,7 +900,7 @@ returns:
     - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
 --- #--------------------------------------------------------------------------
 type: function
-desc: "Update a kernel launch command in a finalized command-buffer."
+desc: "Update a kernel launch command in a finalized command-buffer. This entry-point is synchronous and may block if the command-buffer is executing when the entry-point is called."
 class: $xCommandBuffer
 name: UpdateKernelLaunchExp
 params:
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -51,8 +51,8 @@ ur_exp_command_buffer_handle_t_::ur_exp_command_buffer_handle_t_(
     const ur_exp_command_buffer_desc_t *Desc, const bool IsInOrderCmdList)
     : Context(Context), Device(Device), ZeCommandList(CommandList),
       ZeCommandListResetEvents(CommandListResetEvents),
-      ZeCommandListDesc(ZeDesc), ZeFencesList(), QueueProperties(),
-      SyncPoints(), NextSyncPoint(0),
+      ZeCommandListDesc(ZeDesc), ZeFencesMap(), ZeActiveFence(nullptr),
+      QueueProperties(), SyncPoints(), NextSyncPoint(0),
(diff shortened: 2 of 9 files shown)


scripts/core/exp-command-buffer.yml
```diff
@@ -900,7 +900,7 @@ returns:
     - $X_RESULT_ERROR_OUT_OF_HOST_MEMORY
 --- #--------------------------------------------------------------------------
 type: function
-desc: "Update a kernel launch command in a finalized command-buffer."
+desc: "Update a kernel launch command in a finalized command-buffer. This entry-point is asynchronous and the command-buffer should not be executing when this entry-point is called, synchronization is the responsibility of the user code."
```
EwanC, line 903: typo - "synchronous"

nrspruit (approved):
LGTM, thanks for the update and finding!

EwanC:
@nrspruit Could you re-review this patch? After discussing this fix with the graphs team, we felt that we needed to synchronize on the fence here. We synchronize on L0 events elsewhere in the UR command-buffer code which is propagated to SYCL-RT, but that's not going to guarantee that the command-list has finished executing if extra commands are added by the driver when a command-list is closed. So to avoid the situation where we try to update a command-list that is actually still executing those extra commands, we need to wait on the fence.

This is also based on the assumption that, if you do the following, then although command-lists A & B will start in-order, there is no guarantee and that A will finish before B starts. Is this correct? If command-list A is guaranteed to finish execution before B starts, then using this new fence may not be needed as it corresponds to the end of A, when we are always executing B afterwards and syncing on an event in that.

```
zeCommandQueueExecuteCommandLists(command_list_A)
zeCommandQueueExecuteCommandLists(command_list_B)
```

Bensuo (approved):

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -1025,11 +1026,19 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
   uint32_t QueueGroupOrdinal;
   auto &ZeCommandQueue = QGroup.getZeQueue(&QueueGroupOrdinal);
 
-  ze_fence_handle_t ZeFence;
-  ZeStruct<ze_fence_desc_t> ZeFenceDesc;
-
-  ZE2UR_CALL(zeFenceCreate, (ZeCommandQueue, &ZeFenceDesc, &ZeFence));
-  CommandBuffer->ZeFencesList.push_back(ZeFence);
+  // If we already have created a fence for this queue reset and reuse it,
```
nrspruit, line 1029: Minor wording: "then reuse it,"

nrspruit (approved):
LGTM

PR #1630, opened 2024-05-20
[CTS] Don't depend on Mako in tests

kbenzie:
When running only the CTS it should be possible to do so without first installing the requirements for building the specification.

### DIFF
--- a/scripts/generate_kernel_header.py
+++ b/scripts/generate_kernel_header.py
@@ -12,9 +12,7 @@ import re
 import subprocess
 import sys
 
-from mako.template import Template
-
-HEADER_TEMPLATE = Template("""/*
+HEADER_TEMPLATE = """/*
  *
  * Copyright (C) 2023 Intel Corporation
  *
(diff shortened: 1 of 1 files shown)


scripts/generate_kernel_header.py
```diff
@@ -33,26 +31,33 @@
...
 def generate_header(output_file, kernel_name_dict):
     """Render the template and write it to the output file."""
     file_name = os.path.basename(output_file)
-    rendered = HEADER_TEMPLATE.render(file_name=file_name,
-                                      kernel_name_dict=kernel_name_dict)
+    device_binaries = ""
+    for program, entry_points in kernel_name_dict.items():
+        content = ""
+        for entry_point in entry_points:
+            content += ENTRY_POINT_TEMPLATE % entry_point
```
RossBrunton, line 57: Is `.format` not preferred over `%` in modern Python?

kbenzie: Since this is C++ code the `{}` position markers that `.format()` uses would necessitate escaping all of the braces in the template files. I didn't feel like doing that so chose the easy option as you see here.

RossBrunton (approved):

PR #1631, opened 2024-05-20
[CTS] Fix virtual mem enqueue test

npmiller:
`urKernelSetArgPointer` doesn't take the argument pointer directly, it takes a pointer to the argument pointer.

This fixes a crash in the CTS on CUDA.


aarongreig:
hopefully this gets fixed by https://github.com/oneapi-src/unified-runtime/pull/805 instead, I guess I need to do a match file pass there

npmiller:
Oh you're right I think #805 should fix it, I had a quick look through but missed it!

And yes if you could remove the segfault pattern in that PR that would be good, I believe both CUDA and HIP don't crash anymore.

npmiller:
Wait no, it's an enqueue test, so only CUDA should be fine, HIP is still crashing for other reasons, I'll remove the HIP segfault pattern later when all the fixes are merged

npmiller:
Re-opening this as there are some issues with #805 so we can merge this first to fix the segfault while we figure out the larger PR.

PR #1632, opened 2024-05-20
[HIP] Implement urDeviceGetNativeHandle

frasercrmck:
This is mostly just a copy of the CUDA version of this implementation.

### DIFF
--- a/source/adapters/hip/device.cpp
+++ b/source/adapters/hip/device.cpp
@@ -9,6 +9,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "device.hpp"
+#include "adapter.hpp"
 #include "context.hpp"
 #include "event.hpp"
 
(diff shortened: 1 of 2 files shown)


npmiller:
Could you also remove it from the match file (it "xfails" the test for the UR CTS scripts):
* https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/context/context_adapter_hip.match#L1

source/adapters/hip/device.cpp
```diff
@@ -950,8 +951,57 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetNativeHandle(
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL urDeviceCreateWithNativeHandle(
-    ur_native_handle_t, ur_platform_handle_t,
-    const ur_device_native_properties_t *, ur_device_handle_t *) {
+    ur_native_handle_t hNativeDevice, ur_platform_handle_t hPlatform,
+    const ur_device_native_properties_t *pProperties,
+    ur_device_handle_t *phDevice) {
+  std::ignore = pProperties;
```
npmiller, line 957: I know that's probably how it is in the CUDA adapter, but we should use `[[maybe_unused]]` instead of `std::ignore` for new code, and eventually we should probably fix the old code to use that too.

source/adapters/hip/device.cpp
```diff
@@ -950,8 +951,57 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetNativeHandle(
...
+      *phDevice = SearchRes->get();
+      return UR_RESULT_SUCCESS;
+    }
+  }
+
+  // Get list of platforms
+  uint32_t NumPlatforms = 0;
+  ur_adapter_handle_t AdapterHandle = &adapter;
+  ur_result_t Result =
+      urPlatformGet(&AdapterHandle, 1, 0, nullptr, &NumPlatforms);
```
JackAKirk, line 982: nit:

I think with the exception of opencl adapter, every adapter can only have a max of one platform.
So I think you can just check that NumPlatforms is 1 here. If not you can return `UR_RESULT_ERROR_INVALID_OPERATION` directly at this point.
Then later you don't need to have a loop over the platforms.

Although it doesn't really make much difference. So whatever you prefer.

source/adapters/hip/device.cpp
frasercrmck: Aye, good idea. Should I fix up the same situation in the CUDA adapter while I'm "here", or does that cause more trouble than it's worth?

source/adapters/hip/device.cpp
frasercrmck: Oh, thanks. So if that's the case of the CUDA adapter, I wonder why there's a loop there. Should we fix that up at some point?

JackAKirk: > Oh, thanks. So if that's the case of the CUDA adapter, I wonder why there's a loop there. Should we fix that up at some point?
Sure
I guess it is because it was originally copied from the opencl implementation where there are presumably always (at least) two platforms? cpu and gpu.

npmiller: The CUDA adapter has one platform per device, that only changed with the multi-device context patch, that has already been merged for the HIP adapter

source/adapters/hip/device.cpp
npmiller: I wouldn't fix it for the CUDA adapter in this patch, it'd probably be better as a separate NFC patch fixing all of them at once

source/adapters/hip/device.cpp
frasercrmck: Thanks for the *context*. Will do a follow-up patch.

PR #1634, opened 2024-05-20
[HIP][CUDA] Refactor using profiling events 

konradkusiak97:
HIP changes:
- To match with the current behaviour of CUDA adapter, `EvBase` in HIP was moved to `device` and                                                      `getElapsedTime` function now handles the profiling events' synchronization. Also, we were using `hipStreamDefault` flag as default, but in CUDA, we use `CU_STREAM_NON_BLOCKING`, this was also changed to match with cuda, **commits 1-3**
- Added an extra profiling stream to `Queue` which is only created when profiling is enabled and it is used to record `EvQueued`. This was necessary because before we were recording it on the `NULL` stream and this might not be the best solution for HIP, see https://github.com/intel/llvm/issues/12904, **commit 4**

CUDA changes:
- Also added the extra profiling stream for consistency, **commit 5**

**intel/llvm** CI: https://github.com/intel/llvm/pull/13861

### DIFF
--- a/source/adapters/cuda/queue.hpp
+++ b/source/adapters/cuda/queue.hpp
@@ -104,9 +104,9 @@ struct ur_queue_handle_t_ {
   native_type get() { return getNextComputeStream(); };
   ur_device_handle_t getDevice() const noexcept { return Device; };
 
-  // Function which creates the profiling stream. Called only if profiling is
-  // enabled.
-  void createProfilingStream() {
+  // Function which creates the profiling stream. Called only from makeNative
+  // event when profiling is required.
+  void createHostSubmitTimeStream() {
     static std::once_flag HostSubmitTimeStreamFlag;
     std::call_once(HostSubmitTimeStreamFlag, [&]() {
       UR_CHECK_ERROR(cuStreamCreateWithPriority(&HostSubmitTimeStream,
--- a/source/adapters/cuda/queue.hpp
+++ b/source/adapters/cuda/queue.hpp
@@ -29,8 +29,9 @@ struct ur_queue_handle_t_ {
 
   std::vector<native_type> ComputeStreams;
   std::vector<native_type> TransferStreams;
-  // Stream used for recording EvQueue. It is created only if profiling is
-  // enabled - either for the queue or per event.
+  // Stream used for recording EvQueue, which holds information about when the
+  // command in question is enqueued on host, as opposed to started. It is
+  // created only if profiling is enabled - either for queue or per event.
   native_type HostSubmitTimeStream{0};
   // delay_compute_ keeps track of which streams have been recently reused and
   // their next use should be delayed. If a stream has been recently reused it
--- a/source/adapters/cuda/queue.hpp
+++ b/source/adapters/cuda/queue.hpp
@@ -69,8 +69,8 @@ struct ur_queue_handle_t_ {
                      ur_context_handle_t_ *Context, ur_device_handle_t_ *Device,
                      unsigned int Flags, ur_queue_flags_t URFlags, int Priority,
                      bool BackendOwns = true)
-      : ComputeStreams{std::move(ComputeStreams)},
-        TransferStreams{std::move(TransferStreams)},
+      : ComputeStreams{std::move(ComputeStreams)}, TransferStreams{std::move(
+                                                       TransferStreams)},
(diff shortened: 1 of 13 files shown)


source/adapters/cuda/queue.hpp
```diff
@@ -27,6 +29,10 @@ struct ur_queue_handle_t_ {
 
   std::vector<native_type> ComputeStreams;
   std::vector<native_type> TransferStreams;
+  // Stream used solely when profiling is enabled
+  native_type ProfStream;
+  bool IsProfStreamCreated{false};
```
hdelan, line 34: I don't think we need this guy? Since profiling stream is created `if (RequiresTimings)` we can just use that same check whenever we use this guy.

source/adapters/hip/queue.hpp
```diff
@@ -22,6 +25,10 @@ struct ur_queue_handle_t_ {
 
   std::vector<native_type> ComputeStreams;
   std::vector<native_type> TransferStreams;
+  // Stream used solely when profiling is enabled
+  native_type ProfStream;
+  bool IsProfStreamCreated{false};
+  std::once_flag ProfStreamFlag;
```
hdelan, line 31: Do we need to keep this in the queue struct? Or could we have it local to `createProfilingStream`

source/adapters/cuda/queue.hpp
konradkusiak97: This was only to check if we need to [destroy](https://github.com/oneapi-src/unified-runtime/blob/55412d81af29e8b8c5715f4f1dac8f629698a7a8/source/adapters/cuda/queue.cpp#L205) that stream when the queue finished. To use `if (RequiresTimings)` we'd need to destroy it directly from the event.

hdelan:
Also I wouldn't mind if `ProfStream` had a different name. Maybe something like `HostSubmitTimeStream` with a comment to explain what it is used for

source/adapters/cuda/queue.hpp
konradkusiak97: I removed the `bool` guy and changed the condition before `destroy` to:
 ```cpp
 if (hQueue->getHostSubmitTimeStream() != CUstream{0})
```

Comment (hdelan) [reply source/adapters/cuda/queue.hpp]: Nice one

File: source/adapters/cuda/queue.hpp [hunk 3]
```diff
@@ -27,6 +29,9 @@ struct ur_queue_handle_t_ {
 
   std::vector<native_type> ComputeStreams;
   std::vector<native_type> TransferStreams;
+  // Stream used for recording EvQueue. It is created only if profiling is
```
hdelan, line 32:
```suggestion
  // Stream used for recording EvQueue, which holds information about when the command in question is enqueued on host, as opposed to started. It is created ...
```

hdelan: (Clang format needed)

source/adapters/hip/context.cpp
```diff
@@ -47,18 +47,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urContextCreate(
     // Create a scoped context.
     ContextPtr = std::unique_ptr<ur_context_handle_t_>(
         new ur_context_handle_t_{phDevices, DeviceCount});
-
```
hdelan, line 50: Why are you removing the base event?

konradkusiak97: I moved the base event to the device to be consistent with CUDA adapter. This is a bit separate of adding the new profiling stream to both adapters but I combined it into single PR. The first 3 commits are only the work on HIP - moving `EvBase` to device and making it like it is in CUDA

hdelan: Aha yes I see. It looks like we were recording two base events for HIP?

konradkusiak97: The base event for HIP used to belong to the `platform` and we were recording it twice, once here and once in `urPlatformGet`

source/adapters/cuda/queue.hpp
konradkusiak97: done

source/adapters/hip/context.cpp
hdelan: OK good catch

hdelan (approved):
Good stuff. LGTM

PR #1639, opened 2024-05-21
[Bindless][Exp] Windows & DX12 interop. Semaphore ops can take values.

przemektmalon:
The interop import API has been adapted to cater to multiple external memory and semaphore handle types.

This enables memory and semaphores to be imported by single functions irregardless of what the underlying handle type is.

The following APIs have been removed:
 - `ImportExternalOpaqueFDExp`
 - `ImportExternalSemaphoreOpaqueFDExp`

The following APIs have been added:
 - `ImportExternalMemoryExp`
 - `ImportExternalSemaphoreExp`

The following enums have been added:
 - `exp_external_mem_type_t`
 - `exp_external_semaphore_type_t`

Support has also been added to import DirectX 12 memory and fences.

Semaphore operations (wait/signal) can now take values which set the semaphore state to the value passed by the user.

### DIFF
--- a/scripts/core/registry.yml
+++ b/scripts/core/registry.yml
@@ -400,18 +400,12 @@ etors:
 - name: BINDLESS_IMAGES_MIPMAP_FREE_EXP
   desc: Enumerator for $xBindlessImagesMipmapFreeExp
   value: '142'
-- name: BINDLESS_IMAGES_IMPORT_OPAQUE_FD_EXP
-  desc: Enumerator for $xBindlessImagesImportOpaqueFDExp
-  value: '143'
 - name: BINDLESS_IMAGES_MAP_EXTERNAL_ARRAY_EXP
   desc: Enumerator for $xBindlessImagesMapExternalArrayExp
   value: '144'
 - name: BINDLESS_IMAGES_RELEASE_INTEROP_EXP
   desc: Enumerator for $xBindlessImagesReleaseInteropExp
   value: '145'
-- name: BINDLESS_IMAGES_IMPORT_EXTERNAL_SEMAPHORE_OPAQUE_FD_EXP
-  desc: Enumerator for $xBindlessImagesImportExternalSemaphoreOpaqueFDExp
-  value: '146'
 - name: BINDLESS_IMAGES_DESTROY_EXTERNAL_SEMAPHORE_EXP
   desc: Enumerator for $xBindlessImagesDestroyExternalSemaphoreExp
   value: '147'
@@ -589,6 +583,12 @@ etors:
 - name: KERNEL_GET_SUGGESTED_LOCAL_WORK_SIZE
   desc: Enumerator for $xKernelGetSuggestedLocalWorkSize
   value: '225'
+- name: BINDLESS_IMAGES_IMPORT_EXTERNAL_MEMORY_EXP
+  desc: Enumerator for $xBindlessImagesImportExternalMemoryExp
+  value: '226'
+- name: BINDLESS_IMAGES_IMPORT_EXTERNAL_SEMAPHORE_EXP
+  desc: Enumerator for $xBindlessImagesImportExternalSemaphoreExp
+  value: '227'
 ---
 type: enum
 desc: Defines structure types
--- a/source/adapters/cuda/image.cpp
+++ b/source/adapters/cuda/image.cpp
@@ -1021,8 +1021,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesMipmapFreeExp(
   return UR_RESULT_SUCCESS;
 }
 
-UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImportOpaqueFDExp(
+UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImportExternalMemoryExp(
     ur_context_handle_t hContext, ur_device_handle_t hDevice, size_t size,
+    ur_exp_external_mem_type_t memHandleType,
     ur_exp_interop_mem_desc_t *pInteropMemDesc,
     ur_exp_interop_mem_handle_t *phInteropMem) {
   UR_ASSERT(std::find(hContext->getDevices().begin(),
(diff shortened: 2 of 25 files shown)


przemektmalon:
Associated DPC++ PR: https://github.com/intel/llvm/pull/13860

JackAKirk:
I think that my comment here: https://github.com/intel/llvm/pull/13860#discussion_r1608052501
applies to all the `ur_context_handle_t` instances here. They aren't doing anything, and even hypothetically if you were to make a full impl for opencl backend in the future (where the whole context concept comes from), I don't think there would be any functional reason to add a context parameter here:

1. Because this is a USM style extension anyway.
2. Even if it mapped to usm buffers, functionally for the purpose of interop with opencl you can construct contexts separating cpu and gpu opencl platforms just from using the platform class.

I might be missing something, but to me it looks like this extension is another example of inheriting a useless concept from opencl that has no functional purpose, just because everywhere else does it. If there is a genuine reason for it to be added in all these interfaces then I'd be very happy to be proven wrong.

source/adapters/cuda/image.cpp
```diff
@@ -1039,15 +1040,29 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImportOpaqueFDExp(
...
+
+        switch (memHandleType) {
+        case UR_EXP_EXTERNAL_MEM_TYPE_WIN32_NT:
+          extMemDesc.type = CU_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32;
+          break;
+        case UR_EXP_EXTERNAL_MEM_TYPE_WIN32_NT_DX12_RESOURCE:
+          extMemDesc.type = CU_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE;
+          extMemDesc.flags = CUDA_EXTERNAL_MEMORY_DEDICATED;
+          break;
+        case UR_EXP_EXTERNAL_MEM_TYPE_OPAQUE_FD:
```
JackAKirk, line 1061: shouldn't this map to `CU_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD` instead of an error?

JackAKirk: I'm wondering if you've added all the cases you want, just because cuda has a fair few more?

source/adapters/cuda/image.cpp
```diff
@@ -1157,15 +1172,27 @@ urBindlessImagesImportExternalSemaphoreOpaqueFDExp(
...
-        return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+        auto Win32Handle = static_cast<const ur_exp_win32_handle_t *>(pNext);
+        switch (semHandleType) {
+        case UR_EXP_EXTERNAL_SEMAPHORE_TYPE_WIN32_NT:
+          extSemDesc.type = CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_WIN32;
+          break;
+        case UR_EXP_EXTERNAL_SEMAPHORE_TYPE_WIN32_NT_DX12_FENCE:
+          extSemDesc.type = CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE;
+          break;
+        case UR_EXP_EXTERNAL_SEMAPHORE_TYPE_OPAQUE_FD:
```
JackAKirk, line 1191: This looks like it should map to `CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_FD`?

source/adapters/cuda/image.cpp
```diff
@@ -1214,6 +1241,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesWaitExternalSemaphoreExp(
     enqueueEventsWait(hQueue, Stream, numEventsInWaitList, phEventWaitList);
 
     CUDA_EXTERNAL_SEMAPHORE_WAIT_PARAMS SemWaitParams = {};
+    if (hasValue) {
+      SemWaitParams.params.fence.value = waitValue;
```
JackAKirk, line 1245: If you don't set this `value` parameter then do you know if there a default value for it in constructor of `CUDA_EXTERNAL_SEMAPHORE_WAIT_PARAMS`?
There are loads of other params that this struct takes : https://docs.nvidia.com/cuda/archive/11.6.2/cuda-driver-api/structCUDA__EXTERNAL__SEMAPHORE__WAIT__PARAMS__v1.html#structCUDA__EXTERNAL__SEMAPHORE__WAIT__PARAMS__v1_188ca51398134e22437eedca531fcb420

Don't you have to set any of them?

przemektmalon: The `SemWaitParams` struct is constructed with ` = {}`, so all fields are zero initially. Currently we only support `CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_FD`, for which no values need to be set in this struct according to the documentation, and `CU_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE`, for which we need to set the `params.fence.value` field as we do here.

Any other fields are for use by semaphore types we do not currently support (e.g. keyed mutex), so we don't set any of them.

JackAKirk: Great, thanks for the clarifications.

JackAKirk:
extension/cuda impl LGTM once conflicts resolved etc.

przemektmalon:
Pinging @oneapi-src/unified-runtime-bindless-images-write @oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-opencl-write @oneapi-src/unified-runtime-level-zero-write @oneapi-src/unified-runtime-maintain 

CI failure seems to be unrelated to any changes I made to the Bindless Images APIs, could be from generating on Windows, will investigate. Either way, all code should be ready for review.

JackAKirk (approved):

DBDuncan (approved):

nrspruit (approved):
LGTM for Level Zero

scripts/core/registry.yml
```diff
@@ -400,18 +400,12 @@ etors:
...
-- name: BINDLESS_IMAGES_IMPORT_OPAQUE_FD_EXP
-  desc: Enumerator for $xBindlessImagesImportOpaqueFDExp
-  value: '143'
 - name: BINDLESS_IMAGES_MAP_EXTERNAL_ARRAY_EXP
   desc: Enumerator for $xBindlessImagesMapExternalArrayExp
   value: '144'
 - name: BINDLESS_IMAGES_RELEASE_INTEROP_EXP
   desc: Enumerator for $xBindlessImagesReleaseInteropExp
   value: '145'
-- name: BINDLESS_IMAGES_IMPORT_EXTERNAL_SEMAPHORE_OPAQUE_FD_EXP
```
aarongreig, line 412: is there a reason we can't re-use the old enum number when something gets renamed?

kbenzie: They are supposed to be globally unique for xpit I think.

PietroGhg (approved):
Native CPU LGTM, thank you

kbenzie:
@przemektmalon I've rebased this on top of main which now includes the fix from #1732 so we'll see if the e2e workflow changes are working.

PR #1640, opened 2024-05-21
Limit # of workers in E2E job through LIT_OPTS

pbalcer:
LIT_OPTS is more generic, and also covers cases where there is no direct way to modify what CMake passes to llvm-lit command.

### DIFF
--- a/.github/workflows/e2e_core.yml
+++ b/.github/workflows/e2e_core.yml
@@ -146,11 +146,14 @@ jobs:
     - name: Build SYCL
       run: cmake --build ${{github.workspace}}/sycl_build -j
 
+    - name: Set extra llvm-lit options
+      run: echo "LIT_OPTS=\"-sv \${{matrix.adapter.extra_lit_flags}}\"" >> $GITHUB_ENV
+
     - name: Run check-sycl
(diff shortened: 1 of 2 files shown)


wlemkows (approved):

.github/workflows/e2e_core.yml
```diff
@@ -146,11 +146,14 @@ jobs:
...
 
+    - name: Set extra llvm-lit options
+      run: echo "LIT_OPTS=\"-sv \${{matrix.adapter.extra_lit_flags}}\"" >> $GITHUB_ENV
+
     - name: Run check-sycl
       # Remove after fixing SYCL test :: abi/layout_handler.cpp
       # This issue does not affect further execution of e2e with UR.
       continue-on-error: true
-      run: cmake --build ${{github.workspace}}/sycl_build --target check-sycl -j6
+      run: cmake --build ${{github.workspace}}/sycl_build --target check-sycl
```
wlemkows, line 156: intentional?

pbalcer: yes, this does nothing because the number of jobs is actually determined by a separate program - llvm-lit, not cmake/ninja.

PR #1643, opened 2024-05-21
[EXP][CUDA] Add initial version of (kernel) Launch Properties extension.

JackAKirk:
Introduces an extendable extension for kernel launch properties and a full cuda implementation of the initial extension spec. Currently only three Launch Properties are supported on the cuda backend only:

- UR_EXP_LAUNCH_PROPERTY_ID_CLUSTER_DIMENSION
- UR_EXP_LAUNCH_PROPERTY_ID_COOPERATIVE
- UR_EXP_LAUNCH_PROPERTY_ID_IGNORE (does nothing, but useful for testing).

Note `UR_EXP_LAUNCH_PROPERTY_ID_IGNORE` and `UR_EXP_LAUNCH_PROPERTY_ID_COOPERATIVE` can also be supported on l0 and hip in a later PR.

For more information you can read the accompanying extension .rst. Also see my comments here: https://github.com/oneapi-src/unified-runtime/pull/1610

Compared to the proposal from https://github.com/oneapi-src/unified-runtime/pull/1610 this PR uses a simpler implementation where we only introduce one new UR function.
I also switched from using opaque UR types to using UR properties that are defined for all backends (but wouldn't have to be supported for all backends).
I made this decision because even if we use opaque types, I think programmers will have to have different code versions for each backend if their underlying property types don't match. The way it is now future backend specific properties could be be added to the UR struct/enum property types.
I decided that this would be clearer and easier, and means we only need one new UR function instead of two, but we could always change this back if people have a different opinion, or if it turns out if other backends introduce new properties that it is better to use an opaque `ur_exp_launch_property_t` type.


JackAKirk:
FYI this is highest priority.

scripts/core/exp-launch-attributes.yml
```diff
@@ -0,0 +1,131 @@
...
+      name: hQueue
+      desc: "[in] handle of the queue object"
+    - type: $x_kernel_handle_t
+      name: hKernel
+      desc: "[in] handle of the kernel object"
+    - type: uint32_t
+      name: workDim
+      desc: "[in] number of dimensions, from 1 to 3, to specify the global and work-group work-items"
+    - type: const size_t*
+      name: pGlobalWorkSize
```
hdelan, line 86: I see you have opted not to have a global offset param. Can you confirm that this was intentional?

source/adapters/cuda/enqueue.cpp
```diff
@@ -484,6 +484,133 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueCooperativeKernelLaunchExp(
...
+    ur_event_handle_t *phEvent) {
+
+  if (numAttrsInLaunchAttrList == 0) {
+    urEnqueueKernelLaunch(hQueue, hKernel, workDim, nullptr, pGlobalWorkSize,
+                          pLocalWorkSize, numEventsInWaitList, phEventWaitList,
+                          phEvent);
+  }
+
+  // Preconditions
+  UR_ASSERT(hQueue->getContext() == hKernel->getContext(),
```
hdelan, line 502: Nit: I'd be more concerned that `hQueue->getDevice() == hKernel->getProgram()->getDevice()`

scripts/core/exp-launch-attributes.yml
JackAKirk: Yup, @npmiller gave me the green light.

hdelan (approved):
LGTM just a few points

scripts/core/exp-launch-attributes.yml
hdelan: OK nice one

source/adapters/cuda/enqueue.cpp
```diff
@@ -484,6 +484,133 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueCooperativeKernelLaunchExp(
...
 }
 
+UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunchCustomExp(
+    ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel, uint32_t workDim,
+    const size_t *pGlobalWorkSize, const size_t *pLocalWorkSize,
+    uint32_t numAttrsInLaunchAttrList,
+    const ur_exp_launch_attribute_t *launchAttrList,
+    uint32_t numEventsInWaitList, const ur_event_handle_t *phEventWaitList,
+    ur_event_handle_t *phEvent) {
+
```
hdelan, line 494: In general you need to make sure there is the necessary migration of buffer and image mem between devices in a context. It would suffice to assert that the kernel has no `ur_mem_handle_t` args.

hdelan (changes requested):

source/adapters/cuda/enqueue.cpp
hdelan: See here https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/cuda/enqueue.cpp#L421 as well as other places in `urEnqueueKernelLaunch`

source/adapters/cuda/enqueue.cpp
```diff
@@ -484,6 +484,133 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueCooperativeKernelLaunchExp(
                                numEventsInWaitList, phEventWaitList, phEvent);
 }
 
+UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunchCustomExp(
+    ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel, uint32_t workDim,
+    const size_t *pGlobalWorkSize, const size_t *pLocalWorkSize,
+    uint32_t numAttrsInLaunchAttrList,
```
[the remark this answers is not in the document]
JackAKirk: Thanks, yeah I think I should change this to make it consistent with the other properties.

source/adapters/cuda/enqueue.cpp
JackAKirk: Thanks, I forgot about the multi-context patch. I've now made this consistent with it.

source/adapters/cuda/enqueue.cpp
JackAKirk: Thanks, I forgot about the multi-context patch. I've now made this consistent with it.

source/adapters/cuda/enqueue.cpp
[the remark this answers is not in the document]
JackAKirk: After some test implementations  and discussion we decided that whilst it would be nice to have a consistent properties protocol, the existing way is the best option atm:

- the most common way of passing properties via a linked list of descriptor and properties would work, but there are a couple of small things that makes it not exactly natural for this interface/property type combination. There are existing interfaces in ur that also pass properties in different ways
- The existing way is the most natural mapping to how cuda uses these properties (attributes)
- This is experimental and we could potentially update it later: Once we know about other backend requirements we will probably have to update some aspects of it anyway.

However I have changed the naming from attributes -> properties, because in UR attributes is used for things like device queries, and "properties" is consistent with UR naming for passing in properties to functions.

source/adapters/cuda/enqueue.cpp
hdelan: Nice one

hdelan (approved):
CUDA adapter LGTM. Thanks for changes

JackAKirk:
I just have to run the script again with main merged, and also I forgot to update a directory name attribute -> property

scripts/core/EXP-LAUNCH-ATTRIBUTES.rst
```diff
@@ -0,0 +1,102 @@
...
+approach is to introduce the `exp_launch_property_t` type that enables a more
+extendable API.
+
+Each `exp_launch_property_t` instance corresponds to a specific kernel launch
+property.
+Only one new function is introduced: `urEnqueueKernelLaunchCustomExp`.
+`urEnqueueKernelLaunchCustomExp` takes an array of `exp_launch_property_t` as
+an argument, and launches a kernel using these properties.
+`urEnqueueKernelLaunchCustomExp` corresponds closely to the CUDA Driver API
+`cuLaunchKernelEx`.
```
kbenzie, line 49: Single backticks render as italics in reStructuredText. Also when using `${x}` for the ur prefixes those don't render well in inline monospace because the templates generate cross-links. 

```suggestion
This extension is a future-proof and portable solution that supports these two
requirements. Instead of using a fixed set of kernel enqueue arguments, the
approach is to introduce the ${x}_exp_launch_property_t type that enables a
more extendable API.

Each ${x}_exp_launch_property_t instance corresponds to a specific kernel
launch property.
Only one new function is introduced: ${x}EnqueueKernelLaunchCustomExp.
${x}EnqueueKernelLaunchCustomExp takes an array of ${x}_exp_launch_property_t
as an argument, and launches a kernel using these properties.
${x}EnqueueKernelLaunchCustomExp corresponds closely to the CUDA Driver API
``cuLaunchKernelEx``.
```

scripts/core/EXP-LAUNCH-ATTRIBUTES.rst
```diff
@@ -0,0 +1,102 @@
...
+Functions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* ${x}EnqueueKernelLaunchCustomExp
+
+Support
+--------------------------------------------------------------------------------
+
+Adapters which support this experimental feature *must* return the valid string
+defined in ``${X}_LAUNCH_PROPERTIES_EXTENSION_STRING_EXP`` as one of the options from
```
kbenzie, line 87: The `${X}` will not render well in an inline monospace.

```suggestion
defined in ${X}_LAUNCH_PROPERTIES_EXTENSION_STRING_EXP as one of the options from
```

kbenzie (approved):

scripts/core/EXP-LAUNCH-ATTRIBUTES.rst
JackAKirk: Thanks looks good.

scripts/core/EXP-LAUNCH-ATTRIBUTES.rst
JackAKirk: I see thanks.

kbenzie:
> @kbenzie could this be merged asap?
> I've tested it on sm_90 now for the optional feature and it works.
> Thanks

Will do, there's one PR already in flight so will be after that.

PR #1645, opened 2024-05-22
[UR] Add experimental interface for native enqueue

hdelan:
Add initial API + spec for experimental native enqueue.


kbenzie (commented):
Overall the changes look good, just need to align the naming convention with the rest of the UR API.

scripts/core/exp-native-enqueue.yml
```diff
@@ -0,0 +1,92 @@
...
+return: void
+params:
+    - type: void*
+      name: pUserData
+      desc: "[in][out] pointer to data to be passed to callback"
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Immediately enqueue work through a native backend API"
+class: $x
+name: NativeEnqueueExp
```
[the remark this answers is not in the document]
hdelan: I consciously avoided the `urEnqueue` class since the interop func is being executed within the `urNativeEnqueueExp`. ie the interop func is not being enqueued, only the native work is being enqueued. A subtle difference, but not sure if this means it shouldn't be part of the `urEnqueue` class. Let me know what you think

scripts/core/exp-native-enqueue.yml
```diff
@@ -0,0 +1,92 @@
...
+# See LICENSE.TXT
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# See YaML.md for syntax definition
+#
+--- #--------------------------------------------------------------------------
+type: header
+desc: "Intel $OneApi Unified Runtime Experimental API for enqueuing work through native APIs"
+ordinal: "100"
+--- #--------------------------------------------------------------------------
```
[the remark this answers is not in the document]
hdelan: OK sure thanks will apply

scripts/core/exp-native-enqueue.yml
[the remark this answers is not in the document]
kbenzie: Its still part of the adapters queue scheduling is it not?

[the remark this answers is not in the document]
pbalcer: fwiw, I vote for `urEnqueueNative`. Yes, the name might not be obvious at a glance, but the API itself I think makes it clear that we are invoking the callback immediately to enqueue work.

include/ur_api.h
```diff
@@ -9264,6 +9269,77 @@ urUsmP2PPeerAccessGetInfoExp(
...
+                                     ///< ::UR_STRUCTURE_TYPE_NATIVE_ENQUEUE_PROPERTIES
+    void *pNext;                     ///< [in,out][optional] pointer to extension-specific structure
+    ur_native_enqueue_flags_t flags; ///< [in] native enqueue flags
+
+} ur_native_enqueue_properties_t;
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Function enqueueing work through the native API to be executed
+///        immediately.
+typedef void (*ur_native_enqueue_function_t)(
```
pbalcer, line 9305: There needs to be some way for the adapter to pass data to this callback. For example, L0 usually requires the WaitList events to be part of the "native enqueue":
https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/kernel.cpp#L256

scripts/core/exp-native-enqueue.yml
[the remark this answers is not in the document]
hdelan: So if I do:

```
urEnqueueMemBufferWrite(..., &e1);
urNativeEnqueueExp(..., func_that_reads_above_mem, e1, ...);
```
The adapter will make the dependency work, but the `func_that_reads_above_mem` is not enqueued itself. It is the enqueue-er rather than the enqueue-ee

include/ur_api.h
hdelan: That's a good idea. So maybe the callback API should take something like:

```
typedef void (ur_native_enqueue_function_t*)(ur_queue_handle_t, uint32_t numEventsInWaitList,
ur_event_handle_t *inEvents);
```

Do you think we should also have an out event? ie:

```
typedef void (ur_native_enqueue_function_t*)(ur_queue_handle_t, uint32_t numEventsInWaitList,
ur_event_handle_t *inEvents, ur_event_handle_t *outEvents);
```

pbalcer: > Do you think we should also have an out event? ie:

Yes, definitely. We also need an L0 command list (the equivalent of a cuda stream), but there's no equivalent UR type.
Igor started a refactoring work on L0 that uses a similar abstraction, see here:
https://github.com/igchor/unified-runtime/blob/queue_v2/source/adapters/level_zero/kernel.cpp#L202

```
  auto L0Enqueue = [&](ze_command_list_handle_t CommandList,
                       ze_event_handle_t SignalEvent, uint32_t WaitListLength,
                       ze_event_handle_t *WaitList) {
    ZE2UR_CALL(zeCommandListAppendLaunchKernel,
               (CommandList, ZeKernel, &ZeThreadGroupDimensions, SignalEvent,
                WaitListLength, WaitList));
    return UR_RESULT_SUCCESS;
  };

  return Queue->V2QueueDispatcher->enqueue(
      OutEvent, NumEventsInWaitList, EventWaitList,
      v2::CommandListPreference::Compute, std::move(L0Enqueue));
```

I'm not sure if there's a clean wait of passing this sort of adapter-specific stuff. Maybe we need to just pass a `void *` here. Alternatively, we could do a union to make this a bit less error-prone:
```
struct ur_enqueue_native_data {
  ur_adapter_backend_t backend;
  union {
     {
        void *CommandList;
     } level_zero;
  };
}
```

I think we need to try and create an implementation to see what's really required :)

scripts/core/exp-native-enqueue.yml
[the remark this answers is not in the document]
kbenzie: > The adapter will make the dependency work, but the `func_that_reads_above_mem` is executed immediately, not enqueued itself. It is the enqueue-er rather than the enqueue-ee

Okay, so I think using the `$xEnqueue` class. I think, to distance it from [clEnqueueNativeKernel](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clEnqueueNativeKernel.html) and to remove the potential ambiguity of the `urEnqueueNative` name, I think it should be more specific.

For example, I think `urEnqueueNativeCommands` captures the essence of what you want to do.

include/ur_api.h
hdelan: Yep I will crack on with implementation and keep you in the loop

scripts/core/exp-native-enqueue.yml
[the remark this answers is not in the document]
hdelan: Have used this naming. Thanks

source/adapters/native_cpu/enqueue.cpp
```diff
@@ -710,3 +710,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueWriteHostPipe(
 
   DIE_NO_IMPLEMENTATION;
 }
+
+UR_APIEXPORT ur_result_t UR_APICALL urEnqueueNativeCommandExp(
+    ur_queue_handle_t, ur_exp_enqueue_native_command_function_t, void *,
+    uint32_t, const ur_mem_handle_t *,
+    const ur_exp_enqueue_native_command_properties_t *, uint32_t,
+    const ur_event_handle_t *, ur_event_handle_t *) {
+  DIE_NO_IMPLEMENTATION;
```
PietroGhg, line 719: We are trying to get rid of `DIE_NO_IMPLEMENTATION`, could you use `UR_RESULT_ERROR_UNSUPPORTED_FEATURE` here?

source/adapters/hip/command_buffer.cpp
```diff
@@ -884,7 +884,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferEnqueueExp(
     std::unique_ptr<ur_event_handle_t_> RetImplEvent{nullptr};
     ScopedContext Active(hQueue->getDevice());
     uint32_t StreamToken;
-    ur_stream_quard Guard;
+    ur_stream_guard Guard;
```
EwanC, line 887: Do you know why this code worked/compiled before this change?

hdelan: The type was declared and used with this typo. See I have changed the declaration here https://github.com/oneapi-src/unified-runtime/pull/1645/files#diff-8a960acc065ee19759fae311b3e8f329529a4a0863d22b5bdaa5345d9d6e1af9R17

EwanC (approved):

source/adapters/native_cpu/enqueue.cpp
hdelan: Change made, thanks.

hdelan:
Friendly ping @MartinWehking @oneapi-src/unified-runtime-level-zero-write

source/adapters/hip/queue.hpp
```diff
@@ -261,3 +267,26 @@ struct ur_queue_handle_t_ {
...
 };
+
+// RAII object to make hQueue stream getter methods all return the same stream
+// within the lifetime of this object.
+//
+// This is useful for urEnqueueNativeCommandExp where we want guarantees that
+// the user submitted native calls will be dispatched to a known stream, which
+// must be "got" within the user submitted function.
+//
+// TODO: Add a test that this scoping works
```
MartinWehking, line 278: Nit:

To make sure that this TODO doesn't get overlooked it might make sense to also copy it to the cuda adapter, which also uses the same scoping mechanism.

MartinWehking (approved):

source/adapters/hip/queue.hpp
hdelan: Good idea thanks Martin, will add a new test tomorrow

hdelan:
Friendly ping @oneapi-src/unified-runtime-opencl-write @oneapi-src/unified-runtime-level-zero-write  this is high priority

aarongreig (approved):

nrspruit (approved):
LGTM for level zero

PR #1647, opened 2024-05-22
Add option to control building XPTI libraries

callumfare:
This allows tracing to be enabled when UR is built from intel-llvm which already builds the XPTI targets

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ option(UR_BUILD_ADAPTER_ALL "Build all currently supported adapters" OFF)
 option(UR_BUILD_EXAMPLE_CODEGEN "Build the codegen example." OFF)
 option(VAL_USE_LIBBACKTRACE_BACKTRACE "enable libbacktrace validation backtrace for linux" OFF)
 option(UR_ENABLE_ASSERTIONS "Enable assertions for all build types" OFF)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ set(UR_ADAPTER_HIP_SOURCE_DIR "" CACHE PATH
     "Path to external 'hip' adapter source dir")
 set(UR_ADAPTER_NATIVE_CPU_SOURCE_DIR "" CACHE PATH
     "Path to external 'native_cpu' adapter source dir")
+option(UR_BUILD_XPTI_LIBS "Build the XPTI libraries when tracing is enabled" ON)
```
omarahmed1111, line 66: NIT: We could move this line to line 50 to be near other options.

omarahmed1111: I think we could make it by default have the same value as `UR_ENABLE_TRACING` as so if tracing got enabled, this would be enabled too. and we could also control it manually be assigning it explicitly if we need.

callumfare: I've moved the option but kept it as defaulted to `ON`. If it defaults to the value of `UR_ENABLE_TRACING` and a user configures with tracing off, but then later enables tracing, then the option would still be `OFF` which would be surprising if they're not aware of this particular option.

omarahmed1111: Make sense.

omarahmed1111 (approved):

kbenzie (approved):

PR #1648, opened 2024-05-22
Add ability to log from the tracing layer directly rather than use XPTI

callumfare:
This adds the ability to use the tracing layer with a logger rather than relying on XPTI. The motivation for this is to provide an equivalent to `SYCL_PI_TRACE` when we remove PI from the SYCL runtime. `urtrace` and `sycl-trace` will still be supported.

### DIFF
--- a/scripts/templates/trcddi.cpp.mako
+++ b/scripts/templates/trcddi.cpp.mako
@@ -45,9 +45,15 @@ namespace ur_tracing_layer
         ${th.make_pfncb_param_type(n, tags, obj)} params = { &${",&".join(th.make_param_lines(n, tags, obj, format=["name"]))} };
         uint64_t instance = context.notify_begin(${th.make_func_etor(n, tags, obj)}, "${th.make_func_name(n, tags, obj)}", &params);
 
+        context.logger.info("---> ${th.make_func_name(n, tags, obj)}");
+
         ${x}_result_t result = ${th.make_pfn_name(n, tags, obj)}( ${", ".join(th.make_param_lines(n, tags, obj, format=["name"]))} );
 
         context.notify_end(${th.make_func_etor(n, tags, obj)}, "${th.make_func_name(n, tags, obj)}", &params, &result, instance);
+
+        std::ostringstream args_str;
+        ur::extras::printFunctionParams(args_str, ${th.make_func_etor(n, tags, obj)}, &params);
+        context.logger.info("({}) -> {};\n", args_str.str(), result);
 
         return result;
     }
@@ -111,6 +117,10 @@ namespace ur_tracing_layer
         if(!enabledLayerNames.count(name)) {
             return result;
         }
+
+        // Recreate the logger in case env variables have been modified between
+        // program launch and the call to `urLoaderInit`
+        logger = logger::create_logger("tracing", true, true);
 
         ur_tracing_layer::context.codelocData = codelocData;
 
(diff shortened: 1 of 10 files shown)


scripts/templates/trcddi.cpp.mako
```diff
@@ -49,6 +49,10 @@ namespace ur_tracing_layer
 
         context.notify_end(${th.make_func_etor(n, tags, obj)}, "${th.make_func_name(n, tags, obj)}", &params, &result, instance);
 
+        std::ostringstream args_str;
+        ur::extras::printFunctionParams(args_str, ${th.make_func_etor(n, tags, obj)}, &params);
+        context.logger.info("---> ${th.make_func_name(n, tags, obj)}({}) -> {};", args_str.str(), result);
```
callumfare, line 54: The `--->` at the start here is to make the formatting a little more like the SYCL_PI_TRACE equivalent, which makes it easier to pick out individual calls. This does differ from the `urtrace` collector though, so we might not want to do this. I don't have strong feelings either way.

scripts/templates/trcddi.cpp.mako
```diff
@@ -112,6 +116,10 @@ namespace ur_tracing_layer
             return result;
         }
 
+        // Recreate the logger in case env variables have been modified between
+        // program launch and the call to `urLoaderInit`
+        logger = std::move(logger::create_logger("tracing", true));
```
callumfare, line 121: I've done this because by default a user will need to set `UR_LOG_TRACING="level:info"` to enable tracing. To make the transition from SYCL_PI_TRACE smoother for users, this allows the SYCL runtime to provide an env var (e.g. `SYCL_UR_TRACE=1`) which acts as an alias for this. This involves programmatically setting the UR_LOG_TRACING env var, which means we have to recheck it after the global constructors for the layers, but before the layers are actually initialized.

scripts/templates/trcddi.cpp.mako
```diff
@@ -112,6 +116,10 @@ namespace ur_tracing_layer
             return result;
         }
 
+        // Recreate the logger in case env variables have been modified between
+        // program launch and the call to `urLoaderInit`
+        logger = logger::create_logger("tracing", true);
```
callumfare, line 121: I've done this because by default a user will need to set `UR_LOG_TRACING="level:info"` to enable tracing. To make the transition from SYCL_PI_TRACE smoother for users, this allows the SYCL runtime to provide an env var (e.g. `SYCL_UR_TRACE=1`) which acts as an alias for this. This involves programmatically setting the UR_LOG_TRACING env var, which means we have to recheck it after the global constructors for the layers, but before the layers are actually initialized.

scripts/templates/trcddi.cpp.mako
pbalcer: I don't have a preference either way.

pbalcer, line 54: I think we might want to print just `---> func_name` before the function call, and then the rest `(args) -> result` after it has finished. I vaguely remember SYCL_PI_TRACE doing something like that. It's useful for debugging hangs.

callumfare: That makes sense, I've made that change

pbalcer (approved):
lgtm, but please add a simple test. Something like the existing hello world tracing test: https://github.com/oneapi-src/unified-runtime/blob/main/test/layers/tracing/CMakeLists.txt#L35

kbenzie (approved):
Should we test this?

pbalcer:
Also, we might want to document this, right now the tracing layer [documentation](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/INTRO.rst#tracing) only lists the xpti entry points.

PR #1649, opened 2024-05-22
[HIP] Add support for multiple active mappings

frasercrmck:
This patch adds support for mapping the same buffer multiple times with urEnqueueMemBufferMap.

This is mostly a straight port of #1220. It brings the HIP adapter's functionality in line with that of the CUDA adapter.


source/adapters/hip/enqueue.cpp
```diff
@@ -1292,24 +1288,22 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemUnmap(
...
-                        BufferMem::AllocMode::AllocHostPtr;
+  const bool IsPinned =
+      BufferImpl.MemAllocMode == BufferMem::AllocMode::AllocHostPtr;
 
+  ur_result_t Result = UR_RESULT_SUCCESS;
   if (!IsPinned &&
-      ((std::get<BufferMem>(hMem->Mem).getMapFlags() & UR_MAP_FLAG_WRITE) ||
-       (std::get<BufferMem>(hMem->Mem).getMapFlags() &
-        UR_MAP_FLAG_WRITE_INVALIDATE_REGION))) {
+      ((Map->getMapFlags() & UR_MAP_FLAG_WRITE) ||
```
hdelan, line 1302: Nit
```suggestion
      (Map->getMapFlags() & (UR_MAP_FLAG_WRITE | UR_MAP_FLAG_WRITE_INVALIDATE_REGION))) {
```
W clang-format

source/adapters/hip/enqueue.cpp
```diff
@@ -1239,27 +1240,20 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferMap(
...
-  // Allocate a pointer in the host to store the mapped information
-  auto HostPtr = BufferImpl.mapToPtr(size, offset, mapFlags);
-  *ppRetMap = std::get<BufferMem>(hBuffer->Mem).getMapPtr();
-  if (HostPtr) {
-    Result = UR_RESULT_SUCCESS;
-  }
+  ur_result_t Result = UR_RESULT_SUCCESS;
 
   if (!IsPinned &&
       ((mapFlags & UR_MAP_FLAG_READ) || (mapFlags & UR_MAP_FLAG_WRITE))) {
```
hdelan, line 1253: Could you also combine this into a single `mapFlags & (a | b)`

frasercrmck: Sure thing buddy

source/adapters/hip/enqueue.cpp
frasercrmck: Done!

source/adapters/hip/enqueue.cpp
```diff
@@ -1239,27 +1240,20 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueMemBufferMap(
...
-  // Allocate a pointer in the host to store the mapped information
-  auto HostPtr = BufferImpl.mapToPtr(size, offset, mapFlags);
-  *ppRetMap = std::get<BufferMem>(hBuffer->Mem).getMapPtr();
-  if (HostPtr) {
-    Result = UR_RESULT_SUCCESS;
-  }
+  const bool IsPinned =
+      BufferImpl.MemAllocMode == BufferMem::AllocMode::AllocHostPtr;
 
+  ur_result_t Result = UR_RESULT_SUCCESS;
```
hdelan, line 1251: Nit: We sort of prefer using `UR_CHECK_ERROR(blah)` nowadays instead of `Result = blah`, so this var can usually be removed.

hdelan (approved):
Nice one

source/adapters/hip/enqueue.cpp
hdelan: (Unrelated to this patch but just a bit more info: if `blah` can return`UR_RESULT_ERROR_ADAPTER_SPECIFIC` we can't use `UR_CHECK_ERROR`, we need to return the return val like [here](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/cuda/enqueue.cpp#L464))

frasercrmck: Thanks - I've updated to use the newer error checking style.

kbenzie (approved):

PR #1651, opened 2024-05-22
Ensure a consistent line-endings policy

ldrumm:
I just went to change one line in a file, then found out the whole file changed since it was committed differently to my local system, and we don't do normalisation yet. This fixes that.

Here we add a .gitattributes file at the top level with the policy `* text=auto` which roughly translates as files that git considers "text" are stored internally in history as LF and checked out according to local user preference (which might be CRLF on a certain person's Win32 box, for example).

In addition to the new .gitattributes, files that had CRLF line-endings in the history have been renormalized with `git add --renormalize .`.

This should make cross-platform development easier while respecting local configs.

When in Rome...


kbenzie (approved):

ldrumm:
Since this is functionally NFC, but has the possibility to need rebasing several times would it be possible to please merge soon?

kbenzie:
> Since this is functionally NFC, but has the possibility to need rebasing several times would it be possible to please merge soon?

Yes, as soon as @oneapi-src/unified-runtime-level-zero-write review it.

PR #1656, opened 2024-05-23
fix leaks on level-zero interop program and kernel handles

pbalcer:


pbalcer:
@nrspruit having looked at this code, device -> handle maps for both kernel and programs may have introduced quite a few subtle issues that are hard to detect. This stems from state duplication. We should probably refactor this code so that the native handles are ideally stored only in one place.

test/adapters/level_zero/level_zero_memcheck.match
```diff
@@ -0,0 +1,2 @@
+{{IGNORE}}
+{{.*}} ERROR SUMMARY: {{[03]}} errors from {{[03]}} contexts {{.*}}
```
pbalcer, line 2: I was forced to allow 3 leaks. They originate in libigc, so there's not much we can do here. This should go away once we update all our runners to a newer driver version.

pbalcer:
SYCL E2E failure is unrelated. It's failing on other PRs as well (e.g., https://github.com/intel/llvm/actions/runs/9206096279/job/25325394817?pr=13892).

igchor (approved):

igchor:
> @nrspruit having looked at this code, device -> handle maps for both kernel and programs may have introduced quite a few subtle issues that are hard to detect. This stems from state duplication. We should probably refactor this code so that the native handles are ideally stored only in one place.

I started doing some refactor here: https://github.com/oneapi-src/unified-runtime/pull/1607/files

There's also a problem with handling multiple zeKernels that I want to address. It would be good to have a machine where we can test this with multiple devices, tough.

pbalcer:
All the test failures are unrelated.

pbalcer:
> It would be good to have a machine where we can test this with multiple devices

No promises, but hopefully we can add a runner with multiple devices next week.

PR #1661, opened 2024-05-23
[HIP] Add support for querying program binary type

frasercrmck:


source/adapters/hip/program.hpp
```diff
@@ -27,6 +27,12 @@ struct ur_program_handle_t_ {
   ur_device_handle_t Device;
   std::string ExecutableCache;
 
+  /* The ur_program_binary_type_t property is defined individually for every
```
frasercrmck, line 30: Note - this comment was copied from CUDA but appears out of date. I've left it as-is for now.

kbenzie:
I think it might be worth trying to remove `{{OPT}}` from most/all of the matches now that the segfault matches have been removed.

frasercrmck:
> I think it might be worth trying to remove `{{OPT}}` from most/all of the matches now that the segfault matches have been removed.
> 
> Edit: This can be a separate PR.

Good idea - see #1662. I left in the `{{OPT}}` matches from `program_adapter_hip.match` as there's still a segfault match there.

source/adapters/hip/program.cpp
```diff
@@ -283,7 +283,11 @@ urProgramCreateWithIL(ur_context_handle_t, const void *, size_t,
 UR_APIEXPORT ur_result_t UR_APICALL
 urProgramCompile(ur_context_handle_t hContext, ur_program_handle_t hProgram,
                  const char *pOptions) {
-  return urProgramBuild(hContext, hProgram, pOptions);
+  UR_CHECK_ERROR(urProgramBuild(hContext, hProgram, pOptions));
+  // urProgramBuild sets the BinaryType to UR_PROGRAM_BINARY_TYPE_EXECUTABLE, so
+  // set it to the correct value for urProgramCompile post-hoc.
+  hProgram->BinaryType = UR_PROGRAM_BINARY_TYPE_COMPILED_OBJECT;
```
hdelan, line 289: I'm not sure if this entry point should be supported at all by AMD adapter tbh. What is the philosophy of UR here, should `urProgramCompile` be something that you do in UR regardless of the backend, or is it backend dependent on whether or not your backend supports JIT?

I especially think that it is not useful to have two entry points (`urProgramBuild` and `urProgramCompile`) do the exact same thing, but then somehow differentiating the `BinaryType`, as if they have undergone different processes.

kbenzie: It would need investigation to validate this but my hunch is that SYCL RT expects these entry points to be functional.

kbenzie: As of today, these are required entry points. From an API consistency point of view, having only the HIP adapter not support these would be quite confusing.

hdelan: OK fair enough. As such I guess it's still a good idea to identify this with a different `BinaryType` than a program built with `urProgramBuild` even though the binary type is the same across both

frasercrmck: I agree it's a bit odd, but it's worth remembering that the CUDA adapter also does this. Not to say we should revisit it, but as @kbenzie says it's what UR currently requires.

hdelan: Yeah that's OK with me

hdelan (approved):

PR #1662, opened 2024-05-23
[HIP] Remove {{OPT}} from most matches

frasercrmck:


aarongreig (approved):

EwanC (commented):
What's the semantics of `{{OPT}}`? Is it an optional prefix that isn't required?

pbalcer:
> What's the semantics of `{{OPT}}`? Is it an optional prefix that isn't required?
https://github.com/oneapi-src/unified-runtime/blob/main/cmake/match.py#L13

In the context of ctest output match files, this is essentially a marker stating that the test is flaky. We initially added it for tests that worked on gcc and crashed on clang.
If all lines in a match file are OPT, it means that the entire test program randomly crashes.

EwanC (approved):
LGTM, thanks for the context @pbalcer

PR #1665, opened 2024-05-24
Prefer value over reference for `isPowerOf2`

ldrumm:
It's questionable whether we should ever have taken a reference instead of a value type here, but since 02bb6290 `isPowerOf2` is only available for integral types. Thus, take a value instead of a reference since we will never need reference semantics, and don't ever expect to be passing large arguments.


kbenzie (approved):

GeorgeWeb (approved):
Yes, looks good. Thanks, @ldrumm!

PR #1667, opened 2024-05-24
[UR] Fix Multi Device Event Cache for shared Root Device

nrspruit:
- Fix the creation of the multi device event to occur only when the root devices of the event and the executing queue do not match.

### DIFF
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -1500,7 +1500,20 @@ ur_result_t _ur_ze_event_list_t::createAndRetainUrZeEventList(
 
         std::shared_lock<ur_shared_mutex> Lock(EventList[I]->Mutex);
 
-        if (Queue && Queue->Device != CurQueueDevice &&
+        ur_device_handle_t QueueRootDevice;
+        ur_device_handle_t CurrentQueueRootDevice;
+        if (Queue) {
+          QueueRootDevice = Queue->Device;
+          CurrentQueueRootDevice = CurQueueDevice;
+          if (Queue->Device->isSubDevice()) {
+            QueueRootDevice = Queue->Device->RootDevice;
+          }
+          if (CurQueueDevice->isSubDevice()) {
+            CurrentQueueRootDevice = CurQueueDevice->RootDevice;
+          }
+        }
+
+        if (Queue && QueueRootDevice != CurrentQueueRootDevice &&
             !EventList[I]->IsMultiDevice) {
           ze_event_handle_t MultiDeviceZeEvent = nullptr;
           ur_event_handle_t MultiDeviceEvent;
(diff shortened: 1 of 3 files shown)


raiyanla (approved):

test/adapters/level_zero/CMakeLists.txt
```diff
@@ -44,3 +44,22 @@ if(LINUX)
 
     target_link_libraries(test-adapter-level_zero PRIVATE zeCallMap)
 endif()
+
+if(NOT WIN32)
+    # Make L0 use CallMap from a seprate shared lib so that we can access the map
+    # from the tests. This only seems to work on linux
+    add_library(zeCallMap SHARED zeCallMap.cpp)
+    target_compile_definitions(ur_adapter_level_zero PRIVATE UR_L0_CALL_COUNT_IN_TESTS)
```
igchor, line 52: I think we should only call this add_library, target_compile_definitions and target_link_libraries once. Can we remove it from here or from under the if above?

source/adapters/level_zero/event.cpp
```diff
@@ -1500,7 +1500,20 @@ ur_result_t _ur_ze_event_list_t::createAndRetainUrZeEventList(
 
         std::shared_lock<ur_shared_mutex> Lock(EventList[I]->Mutex);
 
-        if (Queue && Queue->Device != CurQueueDevice &&
+        ur_device_handle_t QueueRootDevice;
+        ur_device_handle_t CurrentQueueRootDevice;
+        if (Queue) {
+          QueueRootDevice = Queue->Device;
+          CurrentQueueRootDevice = CurQueue->Device;
```
pbalcer, line 1507: This section of the code may not be holding the lock on `CurQueue`. Please use the `CurQueueDevice` variable.

source/adapters/level_zero/event.cpp
```diff
@@ -1500,7 +1500,20 @@ ur_result_t _ur_ze_event_list_t::createAndRetainUrZeEventList(
...
+          if (Queue->Device->isSubDevice()) {
+            QueueRootDevice = Queue->Device->RootDevice;
+          }
+          if (CurQueue->Device->isSubDevice()) {
```
pbalcer, line 1511: same as above.

test/adapters/level_zero/multi_device_event_cache_tests.cpp
```diff
@@ -0,0 +1,107 @@
...
+// See LICENSE.TXT
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "ur_print.hpp"
+#include "uur/fixtures.h"
+#include "uur/raii.h"
+
+#include <map>
+#include <string>
+
```
pbalcer, line 12: This is just testing whether the implementation calls the L0 calls we expect it to. Those sort of tests tend to be fragile. Would it be possible to instead create a test that verifies the expected outcome (i.e., the commands are ordered correctly on multi-device systems)?

test/adapters/level_zero/CMakeLists.txt
igchor: Right, but once my fix is merged this might stop working since we will call add_library(zeCallMap) twice. If this PR is merged first then I can rebase and extract this in my PR but otherwise this would need to be fixed here.

PR #1674, opened 2024-05-28
[CUDA] wire up cooperative kernel launch

JackAKirk:
The DPC++ `root_group` test already passes (and still does with this patch), but it doesn't actually enable `root_group` sync in reality. However I think the only thing missing is this small patch. With this patch the the custom kernel launch is called with the cooperative attribute. 

I've tested this locally, but will link up a dpc++ ci test for completeness.


konradkusiak97 (approved):
LGTM if all tests pass

JackAKirk:
dpc++ test CI here: https://github.com/intel/llvm/pull/13933

JackAKirk:
Marking it as ready to merge again. The query is blocked by https://github.com/oneapi-src/unified-runtime/issues/1687
We may as well merge this part I think.

PR #1675, opened 2024-05-28
Fix duplicate targets for `test-adapter-level-zero`

omarahmed1111:
If we configured our build with `UR_DPCXX` specified and on `LINUX` platform, we will have a duplicate target for `test-adapter-level-zero` which will fail with an error of multiple targets with the same name. This PR should fix that.

This also changes `if(LINUX)` to `if(CMAKE_SYSTEM_NAME STREQUAL Linux)`.


omarahmed1111:
This is also fixed in this [PR](https://github.com/oneapi-src/unified-runtime/pull/1666) so closing this.

PR #1676, opened 2024-05-28
[DeviceSanitizer] Support out-of-bounds on private memory

AllanZyne:
- Add runtime support for private memory, and use "detect_privates" option to control it
- When the shadow memory for private or local memory is allocated failed, the checking will be disabled automatically
- Fix WA for local work group size since "urKernelGetSuggestedLocalWorkSize" is added
- Set the default level of sanitizer logger to "warning"

LLVM: https://github.com/intel/llvm/pull/13935


AllanZyne:
Hi @oneapi-src/unified-runtime-maintain, @zhaomaosu, @yingcong-wu and @wenju-he, can you please review this PR?
Thank you very much!

zhaomaosu (approved):
lgtm

source/loader/layers/sanitizer/asan_libdevice.hpp
```diff
@@ -82,6 +82,10 @@ struct LaunchInfo {
 constexpr unsigned ASAN_SHADOW_SCALE = 3;
 constexpr unsigned ASAN_SHADOW_GRANULARITY = 1ULL << ASAN_SHADOW_SCALE;
 
+// Based on the observation, only the last 24 bits of the address of the private
+// variable have changed, we use 31 bits(2G) to be safe.
+constexpr std::size_t ASAN_PRIVATE_SIZE = 0x7fffffffULL + 1;
```
wenju-he, line 87: 2G is probably too big, how about just use 24 bits?

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -202,6 +202,12 @@ SanitizerInterceptor::SanitizerInterceptor() {
         auto Value = KV->second.front();
         cl_DetectLocals = Value == "1" || Value == "true" ? true : false;
     }
+
+    KV = Options->find("detect_privates");
+    if (KV != Options->end()) {
+        auto Value = KV->second.front();
+        cl_DetectPrivates = Value == "1" || Value == "true" ? true : false;
```
pbalcer, line 209: nit: you have now at least two ways to check for true. Maybe create a helper?

source/loader/layers/sanitizer/asan_libdevice.hpp
AllanZyne: We should set ASAN_PRIVATE_SIZE as large as possible, if not, it will produce unpredictable wrong error report.

wenju-he: is there a reproducer that produces a wrong report if we use 24 bits?

AllanZyne: I tried a create a kernel that has large private memory but failed at "PI_OUT_OF_RESOURCES" when the private memory size is more than 8MB without asan.
Besides, if asan is enabled, the maximum size of private memory on user side should be small.

Therefore, I think it's fine to use 24 bits now.

wenju-he: Thank you for the test. Using 24bit lgtm. Though it will still cause allocating a huge memory, which may result in OOM, when number of workgroups is large. I guess we probably need to reduce ASAN_PRIVATE_SIZE further in the future.

AllanZyne: I think we can also prevent from OOM by limiting the detecting range of workgroup

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -760,12 +766,8 @@ ur_result_t SanitizerInterceptor::prepareLaunch(
...
                            sizeof(DeviceInfo->Type));
 
-        if (DeviceInfo->Type == DeviceType::CPU) {
-            break;
-        }
-
         if (LaunchInfo.LocalWorkSize.empty()) {
-            LaunchInfo.LocalWorkSize.reserve(3);
+            LaunchInfo.LocalWorkSize.resize(3);
             // FIXME: This is W/A until urKernelSuggestGroupSize is added
```
yingcong-wu, line 771: urKernelSuggestGroupSize has been added now, do you plan to address this FIXME?

AllanZyne: Nice catch! Added.

yingcong-wu (approved):
lgtm

wenju-he (approved):
LGTM

wenju-he (approved):
LGTM

AllanZyne:
Hi @oneapi-src/unified-runtime-maintain, @zhaomaosu, @yingcong-wu and @wenju-he, can you please review this PR again?
I added some detail changes since last review (also added description in PR).
Thank you very much!

yingcong-wu (approved):
lgtm

zhaomaosu (approved):
lgtm

callumfare:
@AllanZyne This has conflicts now that #1677 has merged. If you could fix them then I'll make sure this gets merged next. Thanks!

AllanZyne:
> @AllanZyne This has conflicts now that #1677 has merged. If you could fix them then I'll make sure this gets merged next. Thanks!

Fixed. Thanks very much!

callumfare (approved):

PR #1678, opened 2024-05-29
Fix return of component and composite device info queries

steffenlarsen:
The UR_DEVICE_INFO_COMPONENT_DEVICES and UR_DEVICE_INFO_COMPOSITE_DEVICE device info queries currently have confusing return values for all backends.

For backends that do not support these queries, they return 0, despite both queries being expected to return 1 or more devices, which may have a larger type than the integral 0. To address this, these backends now return UR_RESULT_ERROR_UNSUPPORTED_FEATURE to signal that the device info queries are not supported by the given backend.

For the L0 backend that supports this, the case where the component devices query has no devices, it also returns a 0 literal. Instead, this commit changes it to report a 0-sized return value and will not write to the return value if asked.


source/adapters/cuda/device.cpp
```diff
@@ -1078,7 +1078,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   case UR_DEVICE_INFO_COMPONENT_DEVICES:
   case UR_DEVICE_INFO_COMPOSITE_DEVICE:
     // These two are exclusive of L0.
-    return ReturnValue(0);
+    return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
```
aarongreig, line 1081: can you make the return `UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION` instead? that's what we have in the spec for unsupported queries and it's what the cts expects if a query is allowed to be unsupported

steffenlarsen: Sure, we can do that. It looks a little weird in the SYCL runtime because it currently maps to `PI_ERROR_INVALID_VALUE`, but I suppose "invalid" means "unsupported" when it comes to info queries anyway.

aarongreig: yeah that's a bad mapping, could change it to map to `PI_ERROR_UNSUPPORTED_FEATURE` in the corresponding llvm pr or just leave it and wait for the UR port to sort it out

steffenlarsen: I figure in the vast majority of cases, "invalid" from an info query means the query is either not recognized or intentionally not supported. In both those cases, the feature isn't supported, so it's probably fine. 😄

aarongreig (approved):

aarongreig:
@oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-hip-write @oneapi-src/unified-runtime-cuda-write @oneapi-src/unified-runtime-level-zero-write this change is now important for the PI->UR port, please review when you can

pbalcer (approved):

PietroGhg (approved):
Native CPU LGTM, thank you

PR #1679, opened 2024-05-29
[HIP] Add support for querying USM allocation size

frasercrmck:


source/adapters/hip/usm.cpp
```diff
@@ -230,8 +230,12 @@ urUSMGetMemAllocInfo(ur_context_handle_t hContext, const void *pMem,
       }
       return ReturnValue(Pool);
     }
+    case UR_USM_ALLOC_INFO_SIZE: {
+      size_t RangeSize;
```
jchlanda, line 234: Not sure if `pedantic` would complain about uninitialised variable here (probably yes), but you might want to give `RangeSize` a default value. Like we do in [here](https://github.com/oneapi-src/unified-runtime/blob/0959c61a557127baeec9cdcdc807969d647d84b9/source/adapters/hip/device.cpp#L167) for example.

source/adapters/hip/usm.cpp
```diff
@@ -230,8 +230,12 @@ urUSMGetMemAllocInfo(ur_context_handle_t hContext, const void *pMem,
...
+      UR_CHECK_ERROR(hipMemPtrGetInfo(const_cast<void *>(pMem), &RangeSize));
```
jchlanda, line 235: :)

jchlanda (approved):

source/adapters/hip/usm.cpp
frasercrmck: Good shout!

source/adapters/hip/usm.cpp
frasercrmck: Yeah.. I don't feel too bad as it's done in quite a few other places.

jchlanda: It's on HIP, great API design.

PR #1682, opened 2024-05-29
[cuda] Add cuda 11.8 #ifdef for `cuEnqueueKernelEx`.

JackAKirk:
This was erroring out for cuda earlier than 11.8 due to `cuEnqueueKernelEx` and associated attributes not being introduced till cuda 11.8


frasercrmck (approved):
LGTM

PR #1690, opened 2024-05-30
[Test] Add UUID to device name

hdelan:
Without having a device index or UUID added to the device name, tests were failing in multi GPU systems as tests would instantiate the same testnames for separate GPUs of the same model. This makes each GPU name unique, which in turn makes test names unique.

### DIFF
--- a/test/conformance/testing/include/uur/utils.h
+++ b/test/conformance/testing/include/uur/utils.h
@@ -189,9 +189,13 @@ inline std::string GetPlatformName(ur_platform_handle_t hPlatform) {
 }
 
 inline std::string GetDeviceName(ur_device_handle_t device) {
-    std::string device_name;
+    std::string device_name, device_uuid;
     GetDeviceInfo<std::string>(device, UR_DEVICE_INFO_NAME, device_name);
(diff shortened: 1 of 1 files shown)


test/conformance/context/context_adapter_native_cpu.match
```diff
@@ -1,3 +1,3 @@
-urContextCreateWithNativeHandleTest.InvalidNullPointerDevices/SYCL_NATIVE_CPU___SYCL_Native_CPU_
-urContextCreateWithNativeHandleTest.InvalidNullPointerContext/SYCL_NATIVE_CPU___SYCL_Native_CPU_
-urContextSetExtendedDeleterTest.Success/SYCL_NATIVE_CPU___SYCL_Native_CPU_
+urContextCreateWithNativeHandleTest.InvalidNullPointerDevices/SYCL_NATIVE_CPU___SYCL_Native_CPU_____
```
RossBrunton, line 1: Should we use `{{.*}}` when matching the device name for native CPU? Just in case it changes or a UUID is added in future.

PR #1691, opened 2024-05-30
Add support for comments in match files

frasercrmck:
These have a variety of uses. For example, by documenting the reasons for known failures inside the match files themselves, we can keep all relevant information in one place and help direct developers towards known reproducers, areas of code to investigate, etc.

The comment character is '#' which is fairly common and intuitive, especially in conf-style file types. Also no GTest tests should ever be named with that character.

### DIFF
--- a/cmake/match.py
+++ b/cmake/match.py
@@ -63,6 +63,7 @@ def check_status(input_lines, match_lines):
 class Tag(Enum):
     OPT = "{{OPT}}"         # makes the line optional
     IGNORE = "{{IGNORE}}"   # ignores all input until next match or end of input file
+    COMMENT = "#"           # comment - line ignored
 
 
 ## @brief main function for the match file processing script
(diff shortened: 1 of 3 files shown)


cmake/match.py
```diff
@@ -76,7 +77,14 @@ def main():
 
     with open(input_file, 'r') as input, open(match_file, 'r') as match:
         input_lines = input.readlines()
-        match_lines = match.readlines()
+        # Filter out comments (lines beginning with the comment character,
+        # ignoring leading whitespace)
+        match_lines = list(
```
RossBrunton, line 82: I wonder, is it worth filtering out empty lines as well, just so we can group things more logically?

```
# Foo doesn't work
foo_test
foo2_test

# Bar doesn't work
bar_test
```

RossBrunton (approved):

cmake/match.py
frasercrmck: That's a good idea, yeah. Thanks!

kbenzie (approved):

Issue #1068, opened 2023-11-11
`urPlatformCreateWithNativeHandle` should accept adapter


alexbatashev:
Currently `urPlatformCreateWithNativeHandle` tries to extract dispatch table from native platform handle. That, however, can't work because the native handle may be anything, including nullptr, which may be a valid case for some adapters. So, instead the API should be something like:

```
UR_APIEXPORT ur_result_t UR_APICALL urPlatformCreateWithNativeHandle(ur_native_handle_t hNativePlatform, ur_adapter_handle_t adapter, const ur_platform_native_properties_t *pProperties, ur_platform_handle_t *phPlatform)
```

Which would be consistent with how SYCL handles the matter:
```
namespace sycl {

template <backend Backend>
platform make_platform(const backend_input_t<Backend, platform>& backendObject);

}
```
where backend is part of the API call.

This also reveals a big gap in CTS testing - CTS gets native object from the UR API and then feeds it back, which means that the creation of UR objects from external handles (e.g., from cuBLAS or something) is never tested.

alexbatashev:
Other APIs seem to be OK, as they accept mostly accept a context (or something else in rare cases), and we can use that to extract ddi table. The PR with making this change is here: https://github.com/oneapi-src/unified-runtime/pull/1067

alycm:
Thank you for the PR fixing the immediate issue.  Yes, it seems like we should also improve both the specification and CTS here.

aarongreig:
this will be fixed by https://github.com/oneapi-src/unified-runtime/pull/1692

PR #1692, opened 2024-05-30
Add hAdapter parameter to urPlatformCreateWithNativeHandle

aarongreig:
Other CreateWithNative entry points need a UR handle primarily so the loader has something to obtain a ddi table from. This new adapter parameter fulfils that purpose for PlatformCreateWithNative, allowing us to remove a weird hack that saw the loader getting (apparently succesfully somehow??) a loader object out of the native handle itself.

fixes https://github.com/oneapi-src/unified-runtime/issues/1068

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -1188,6 +1188,8 @@ typedef struct ur_platform_native_properties_t {
 ///     - ::UR_RESULT_ERROR_UNINITIALIZED
 ///     - ::UR_RESULT_ERROR_DEVICE_LOST
 ///     - ::UR_RESULT_ERROR_ADAPTER_SPECIFIC
+///     - ::UR_RESULT_ERROR_INVALID_NULL_HANDLE
+///         + `NULL == hAdapter`
 ///     - ::UR_RESULT_ERROR_INVALID_NULL_POINTER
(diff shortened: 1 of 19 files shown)


callumfare (approved):

JackAKirk:
I'll approve this. I made some related points about the relevance of the usage if this function in cuda/hip. IMO we should probably change the implementations of these functions as returning an error code saying something to the effect of "platform interop doesn't make sense in cuda/hip backends"
But I see that this is not the main issue of the changes made in this patch.

JackAKirk (approved):

aarongreig:
added a comment explaining the unsupported error codes

aarongreig:
LLVM PR https://github.com/intel/llvm/pull/14012

pbalcer (approved):
You can also modify the helper script to get rid of now redundant types in the loader:
https://github.com/oneapi-src/unified-runtime/pull/1067#issuecomment-1808721072

aarongreig:
@oneapi-src/level-zero-maintainers please review, this is important for the sycl UR port

pbalcer (approved):
The L0 adapter change is insignificant.

PR #1693, opened 2024-05-31
Fix CUDA tracing when UR is built standalone

callumfare:
Previously, CUDA tracing only worked when the UR sources were built as part of the CUDA PI plugin. FindCUDACupti.cmake is copied verbatim from intel/llvm.

### DIFF
--- a/cmake/FindCUDACupti.cmake
+++ b/cmake/FindCUDACupti.cmake
@@ -1,0 +1,26 @@
+# Copyright (C) 2024 Intel Corporation
+# Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+# See LICENSE.TXT
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This is lifted from intel-llvm's FindCUDACupti implementation
+# https://github.com/intel/llvm/blob/0cd04144d9ca83371c212e8e4709a59c968291b9/sycl/cmake/modules/FindCUDACupti.cmake
+
+macro(find_cuda_cupti_library)
+  find_library(CUDA_cupti_LIBRARY
+    NAMES cupti
+    HINTS ${CUDA_TOOLKIT_ROOT_DIR}
+          ENV CUDA_PATH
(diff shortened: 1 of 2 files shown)


GeorgeWeb (approved):

cmake/FindCUDACupti.cmake
```diff
@@ -0,0 +1,24 @@
+# Copyright (C) 2024 Intel Corporation
+# Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+# See LICENSE.TXT
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+macro(find_cuda_cupti_library)
+# The following if can be removed when FindCUDA -> FindCUDAToolkit
```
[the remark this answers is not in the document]
callumfare: The comment was definitely misplaced, I found a similar comment elsewhere (in sycl-trace) which made more sense, so I've just updated it here

cmake/FindCUDACupti.cmake
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2024 Intel Corporation
```
aarongreig, line 1: we should maybe have something like what's in [Assertions](https://github.com/oneapi-src/unified-runtime/blob/c259937a690afb0a2a64446aa06310f95e6c411a/cmake/Assertions.cmake#L1) if this is lifted from LLVM

callumfare: It's from the SYCL specific cmake folder rather than upstream LLVM so I think the copyright is fine, but I can add a note here anyway

kbenzie:
Looks like [this](https://github.com/oneapi-src/unified-runtime/actions/runs/9397744079/job/25883839063?pr=1693) failed due to action e2e test failures, I thought these were not supposed to cause check failures. I think they are both know issues.

pbalcer:
> I thought these were not supposed to cause check failures

They are not:

https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/e2e_core.yml#L79

PR #1694, opened 2024-05-31
Set minimum L0 device support to support UR kernel update

EwanC:
In order to support all the ways a kernel command in a command-buffer can be updated in Unified Runtime, the following capabilities need to be available from the device:
* `ZE_MUTABLE_COMMAND_EXP_FLAG_KERNEL_ARGUMENTS`
* `ZE_MUTABLE_COMMAND_EXP_FLAG_GROUP_COUNT`
* `ZE_MUTABLE_COMMAND_EXP_FLAG_GROUP_SIZE`
* `ZE_MUTABLE_COMMAND_EXP_FLAG_GLOBAL_OFFSET`

Only supporting kernel argument update is not sufficient, as is currently the case. There is also no way in the UR API currently to take advantage of the `ZE_MUTABLE_COMMAND_EXP_FLAG_SIGNAL_EVENT` and `ZE_MUTABLE_COMMAND_EXP_FLAG_WAIT_EVENTS` capabilities, so these can be omitted.

Also fix debug printing of `pGroupCount` in command update, as we want to show the values it contains rather than pointer value.

DPC++ PR https://github.com/intel/llvm/pull/13987

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -1295,7 +1295,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferUpdateKernelLaunchExp(
     MutableGroupCountDesc->pNext = NextDesc;
     DEBUG_LOG(MutableGroupCountDesc->pNext);
     MutableGroupCountDesc->pGroupCount = &ZeThreadGroupDimensions;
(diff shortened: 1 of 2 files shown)


Bensuo (approved):

nrspruit (approved):
LGTM

PR #1695, opened 2024-05-31
[HIP][CUDA] Migrate memory across devices in ctx at getNativeHandle

hdelan:
Make sure the NativeHandle that is requested gets the most up to date view of data, which may be on a different device.


npmiller:
Should we add a test for this in the UR CTS? I know it would only work on multi-GPU systems but it would be good to have

hdelan:
> Should we add a test for this in the UR CTS? I know it would only work on multi-GPU systems but it would be good to have

I'm adding a test for SYCL atm. Please hold

hdelan:
https://github.com/intel/llvm/pull/13988

PR #1699, opened 2024-06-03
[NATIVECPU] Implement urUSMGetMemAllocInfo and aligned alloc

PietroGhg:
Implements `urUSMGetMemAllocInfo` and handles alignment requests in allocation functions.
Corresponding intel/llvm PR: https://github.com/intel/llvm/pull/14010

### DIFF
--- a/source/adapters/native_cpu/context.hpp
+++ b/source/adapters/native_cpu/context.hpp
@@ -10,13 +10,49 @@
 
 #pragma once
 
+#include <unordered_map>
 #include <ur_api.h>
 
 #include "common.hpp"
 #include "device.hpp"
+#include "ur/ur.hpp"
+
+namespace native_cpu {
+struct usm_alloc_info {
+  const ur_usm_type_t type;
+  const void *base_ptr;
+  const size_t size;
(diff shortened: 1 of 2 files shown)


source/adapters/native_cpu/context.hpp
```diff
@@ -10,13 +10,49 @@
...
+      return native_cpu::usm_alloc_info(UR_USM_TYPE_UNKNOWN, ptr, 0, nullptr,
+                                        nullptr);
+    }
+    return it->second;
+  }
+
+  void remove_alloc_info_entry(void *ptr) { alloc_info.erase(ptr); }
+
+private:
+  std::unordered_map<const void *, native_cpu::usm_alloc_info> alloc_info;
```
pbalcer, line 57: If performance is any concern here, a potentially simpler solution is to have a metadata header to store the alloc_info:
```
struct host_allocation {
  native_cpu::usm_alloc_info info;
  uint8_t data[0];
}
```
Then you can simply subtract the size of the header to get to the info structure. Should be both quicker and more memory efficient.

Issue #1330, opened 2024-02-09
Proposal for UR new warnings mechanism


fabiomestre:
The warning mechanism that currently exists in Unified Runtime is ambiguous and not very intuitive.

This is the current state (for adapters that use this functionality): 

  - The entrypoint that wants to send a warning sets an error message for `urAdapterGetLastError` and  returns  `UR_RESULT_ERROR_ADAPTER_SPECIFIC` 
  - The call to `urAdapterGetLastError` returns  `UR_RESULT_SUCCESS`  to signal that it is a warning. 
 - If it is an actual error,  The call to `urAdapterGetLastError`returns `UR_RESULT_ERROR_ADAPTER_SPECIFIC` instead.

This is not intuitive because `urAdapterGetLastError` is the only entrypoint that can return an error code in a success state. In addition, setting `UR_RESULT_SUCCESS` to signal a warning doesn't make much sense.

After some internal discussion, we think that the following changes could improve the current workflow:
 - Create a new loader entrypoint that accepts a callback function. This callback function will be called everytime a UR adapter emits a warning.
 - The application would be able to opt out of warnings by simply not setting the callback function.
 - At runtime, we can provide an environment variable to allow the user to enable/disable warnings.
 - `urAdapterGetLastError` should be simplified to always return `UR_RESULT_SUCCESS` unless something is wrong with its parameters. This would make it behave in a way that is more in line with other UR entrypoints.

Risks:
 - This would require some changes on the way that SYCL Runtime handles warnings and adapter specific errors from UR.

Related issue: https://github.com/oneapi-src/unified-runtime/issues/762

fabiomestre:
This was discussed on the WG meeting of 14/02/2024. There was no objections to this approach. So we will go ahead with adding the new entrypoint to the specification.

PR #1707, opened 2024-06-04
[HIP][CUDA] Allow customization of the default logging level

frasercrmck:
The HIP and CUDA adapaters would like a more verbose logging level by default. This patch allows customization per logger.

If the default log level is unsupplied, the old default continues to be used. Users can also continue to override the logging level via the various environment variables.

### DIFF
--- a/source/adapters/cuda/adapter.cpp
+++ b/source/adapters/cuda/adapter.cpp
@@ -36,8 +36,12 @@ public:
 
   ~ur_legacy_sink() = default;
 };
+
+// FIXME: Remove the default log level when querying logging info is supported
+// through UR entry points. See #1330.
 ur_adapter_handle_t_::ur_adapter_handle_t_()
-    : logger(logger::get_logger("cuda")) {
+    : logger(logger::get_logger("cuda",
+                                /*default_log_level*/ logger::Level::ERR)) {
 
   if (std::getenv("UR_LOG_CUDA") != nullptr)
     return;
--- a/source/common/logger/ur_logger.hpp
+++ b/source/common/logger/ur_logger.hpp
@@ -15,10 +15,15 @@
 namespace logger {
 
 Logger create_logger(std::string logger_name, bool skip_prefix = false,
-                     bool skip_linebreak = false);
+                     bool skip_linebreak = false,
+                     logger::Level default_log_level = logger::Level::QUIET);
 
-inline Logger &get_logger(std::string name = "common") {
-    static Logger logger = create_logger(std::move(name));
+inline Logger &
+get_logger(std::string name = "common",
+           logger::Level default_log_level = logger::Level::QUIET) {
+    static Logger logger =
+        create_logger(std::move(name), /*skip_prefix*/ false,
+                      /*slip_linebreak*/ false, default_log_level);
     return logger;
 }
 
@@ -99,6 +104,8 @@ template <typename T> inline std::string toHex(T t) {
 ///             UR_LOG_LOADER="level:info;flush:warning;output:file,out.log"
 /// @param logger_name name that should be appended to the `UR_LOG_` prefix to
 ///        get the proper environment variable, ie. "loader"
+/// @param default_log_level provides the default logging configuration when the environment
+///        variable is not provided or cannot be parsed
 /// @return an instance of a logger::Logger. In case of failure in the parsing of
 ///         the environment variable, returns a default logger with the following
 ///         options:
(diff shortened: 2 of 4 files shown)


pbalcer:
Libraries shouldn't output logs unless explicitly asked to by the application.

source/adapters/cuda/adapter.cpp
```diff
@@ -37,7 +37,7 @@ class ur_legacy_sink : public logger::Sink {
   ~ur_legacy_sink() = default;
 };
 ur_adapter_handle_t_::ur_adapter_handle_t_()
-    : logger(logger::get_logger("cuda")) {
+    : logger(logger::get_logger("cuda", /*default_log_level*/ "level:info")) {
```
hdelan, line 40: I think `level:error` gives the same amount of info as the old logger used to by default

hdelan:
> Libraries shouldn't output logs unless explicitly asked to by the application.

In general I think this is true, but `level:error` I think should be default as, at least within SYCL, when an error occurs the SYCL runtime calls `std::terminate`. Without having `level:error` by default, users get no useful information when their code crashes.

For a sample code dereferencing a nullptr on device:

```
$ ./a.out
terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  Native API failed. Native API returns: -999 (Unknown PI error) -999 (Unknown PI error)
[1]    1158548 IOT instruction (core dumped)  ./a.out
$ UR_LOG_CUDA="level:error" ./a.out
<CUDA>[ERROR]: 
UR CUDA ERROR:
	Value:           700
	Name:            CUDA_ERROR_ILLEGAL_ADDRESS
	Description:     an illegal memory access was encountered
	Function:        wait
	Source Location: /home/hugh/unified-runtime/source/adapters/cuda/event.cpp:135

terminate called after throwing an instance of 'sycl::_V1::runtime_error'
  what():  Native API failed. Native API returns: -999 (Unknown PI error) -999 (Unknown PI error)
[1]    1158563 IOT instruction (core dumped)  UR_LOG_CUDA="level:error" ./a.out
```

UR error codes are completely insufficient to give any information about what has happened. For the average user, having the second case by default is far better than having the first case.

hdelan:
I am not sure what the philosophy of UR is wrt errors vs warnings, but in general it is my understanding that a program should be aborted if an error occurs. In this case I think it is always good to have some information about the error that happened, and the native error code is the best way of transmitting this information.

pbalcer:
> but in general it is my understanding that a program should be aborted if an error occurs.

My opinion (and I don't think this is controversial :-)) is that application should be aborted if we detect a violation of invariants that we always expect to hold (e.g., there's a bug in the code that is hitting an assert, or some memory corruption). Otherwise the error should be propagated to the layer above, with sufficient information for the user to make an informed decision (or simply include the error message as part of the exception as in your example).


I don't feel super strongly about this, if that was the behavior before, and there's a good reason for it, then we should restore it. Just wanted to give some reasoning why the default is off.

hdelan:
I agree with your explanation, that it should pass information to the level above. Does this mean that in the SYCL runtime we should be able to set the default log level of UR? 

Our main concern is restoring the previous level of error reporting, so perhaps this should be configured in SYCL RT (if possible) and if not in UR as a default log level of the CUDA and HIP adapters.

pbalcer:
> Does this mean that in the SYCL runtime we should be able to set the default log level of UR?

Right now the only way to do that would be to use setenv in SYCL, that has its own problems. I'd like to eventually expose all knobs through a programmatic interface (https://github.com/oneapi-src/unified-runtime/issues/1454).

One solution might be for `urAdapterGetLastError` to return the last message written through `logger:error()`.

kbenzie:
FYI for emitting warnings in a way that doesn't output to `stdout`/`stderr` we are working on https://github.com/oneapi-src/unified-runtime/issues/1330 to expose those to SYCL RT via a callback to do with as it pleases. We could make this accept a warning level, then implement it as a sink for the logger.

hdelan:
OK great thanks @kbenzie . In that case, @frasercrmck will you add a FIXME to the code here and then I think this should be good to go

source/adapters/cuda/adapter.cpp
frasercrmck: Yes I think you're right. I had `level:error` but then got confused by the wording in internal ticket. Fixed.

frasercrmck:
> OK great thanks @kbenzie . In that case, @frasercrmck will you add a FIXME to the code here and then I think this should be good to go

FIXME and link to ticket added, thanks :+1:

source/common/logger/ur_logger.hpp
```diff
@@ -107,7 +112,8 @@ template <typename T> inline std::string toHex(T t) {
 ///                            to be printed immediately as they occur
 ///             - output: stderr
 inline Logger create_logger(std::string logger_name, bool skip_prefix,
-                            bool skip_linebreak) {
+                            bool skip_linebreak,
+                            std::string default_log_level) {
```
pbalcer, line 116: It might be simpler to take `logger::Level default_level = logger::Level::QUIET` as a parameter instead of changing how the env variables are parsed.

frasercrmck: It might be, yeah. I was originally hoping that being able to more generically provide default values for environment variables would be useful across more of the project. But so far it's just coming down to the default logging configuration (well, level) and the device selector.

I don't have a strong preference either way.

pbalcer: Up to you. I think I'd slightly prefer passing variables instead of default env string. Less error prone and simpler.

frasercrmck: Yep, agreed. I've pushed that change now. I had to do a bit of plumbing to allow users to set `quiet` (i.e., back to the default) via the environment variable, as it isn't currently a valid value.

konradkusiak97 (approved):

pbalcer (approved):

PR #1708, opened 2024-06-04
Implement callback layer naive approach

aarongreig:
We shouldn't do it this way, but this does demonstrate that it's possible.


aarongreig:
closing for now so I don't needlessly spam CI

PR #1711, opened 2024-06-04
[HIP][CUDA] Several changes to kernel launch

hdelan:
Fixes some bad things that multi dev ctx was doing:

- Don't automatically initialize event vector unless extra events needed for mem args. This avoids dynamic allocation by default at each `urEnqueueKernelLaunch`.
- Make migrateMemoryToDeviceIfNeeded into an async enqueue. This is needed since all the previous work that the memory migration depends on is synchronized with on a stream. So we must dispatch the memory transfers on the same stream. If we don't do this, and naively dispatch synchronous work, this will not depend on the previous work that was enqueued to the stream.
- Add helper func to check if a list contains an elem.
- Make sure that the enqueue write operations call `setLastQueueWritingToMemObj`
- Remove dependency building for buffers and images. All necessary dependencies should be passed to the ur enqueue entry points by the user.


source/ur/ur.hpp
```diff
@@ -375,3 +375,12 @@ static inline void roundToHighestFactorOfGlobalSizeIn3d(
                MaxBlockDim[2]));
   roundToHighestFactorOfGlobalSize(ThreadsPerBlock[2], GlobalSize[2]);
 }
+
+template <typename T>
+inline bool listContainsElem(uint32_t numEventsInList, const T *phList,
```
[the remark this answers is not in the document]
hdelan: The previous `std::find` solution requires that use a vector, whereas this patch tries to avoid unnecessary allocations of vectors at each `urEnqueueKernelLaunch`

source/adapters/hip/enqueue.cpp
```diff
@@ -313,19 +312,15 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
 
   // Early exit for zero size range kernel
   if (*pGlobalWorkSize == 0) {
-    if (DepEvents.size()) {
-      return urEnqueueEventsWaitWithBarrier(hQueue, DepEvents.size(),
-                                            phEventWaitList, phEvent);
-    }
-    return UR_RESULT_SUCCESS;
+    return urEnqueueEventsWaitWithBarrier(hQueue, numEventsInWaitList,
+                                          phEventWaitList, phEvent);
```
npmiller, line 316: This will enqueue a barrier all the time not just when there's `DepEvents`, the `enqueueEventsWait` helper will skip if there's no events but I don't think this will.

hdelan: This is only true if `*pGlobalSize` is zero. This is the correct behaviour. Execution should not proceed past this point if there is no global size. Checking for size previously was a bug

npmiller: Oh I see that's what the CUDA adapter is doing, and I guess without it you might end up with return event not set

npmiller:
Is this covered by the E2E tests in intel/llvm? It could be good to have some tests for this in UR CTS as well

hdelan:
> Is this covered by the E2E tests in intel/llvm? It could be good to have some tests for this in UR CTS as well

Yeah I think I should add a more local test in UR. Will add tomorrow

hdelan:
@npmiller Have added a test for multi dev ctx. Passing on my multi NVIDIA GPU workstation.

test/conformance/memory/urMemBufferMigrateAcrossDevices.cpp
```diff
@@ -0,0 +1,259 @@
...
+using T = uint32_t;
+
+struct urMultiDeviceContextTest : uur::urPlatformTest {
+    void SetUp() {
+        uur::urPlatformTest::SetUp();
+        ASSERT_SUCCESS(urDeviceGet(platform, UR_DEVICE_TYPE_ALL, 0, nullptr,
+                                   &num_devices));
+        ASSERT_NE(num_devices, 0);
+        if (num_devices == 1) {
+            return;
```
npmiller, line 20: Isn't there a `GTEST_SKIP()` or something we could use here instead so the test doesn't show passing when it doesn't actually run?

npmiller (approved):

test/conformance/memory/urMemBufferMigrateAcrossDevices.cpp
hdelan: Good question. Will check

hdelan: Yep looks like `GTEST_SKIP` works. Have changed.

PR #1712, opened 2024-06-05
[CI] Run E2E on PR

lukaszstolarczuk:
We don't want to treat these workflows as blockers though, so we enable continue-on-error for step with tests.
Add comment with actual test status to the PR.

// I've tested the flow on my fork: https://github.com/lukaszstolarczuk/unified-runtime/pull/7
// comment is added; flow only works if prev. (dependent) jobs are done (e.g. "cuda" required for "e2e_cuda")

### DIFF
--- a/.github/workflows/e2e_core.yml
+++ b/.github/workflows/e2e_core.yml
@@ -30,10 +30,6 @@ on:
         description: Tag defifned for the runner
         type: string
         required: true
-      trigger:
-        description: Type of workflow trigger
-        type: string
-        required: true
       xfail:
         description: Allow test failures
         type: string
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e_cuda.yml
```diff
@@ -4,21 +4,18 @@ on:
   schedule:
```
lukaszstolarczuk, line 4: should we still care about schedule?

.github/workflows/e2e_core.yml
```diff
@@ -184,8 +173,14 @@ jobs:
...
 
-    - name: Run e2e tests
+    - name: Run e2e tests (on schedule)
+      if: inputs.trigger == 'schedule'
+      run: ninja -C build-e2e check-sycl-e2e
+
+    - name: Run e2e tests (on PR; allow failures)
+      if: inputs.trigger != 'schedule'
       id: tests
+      continue-on-error: true
```
lukaszstolarczuk, line 183: should we make just this step "continue-on-error" or the whole job...?

.github/workflows/e2e_cuda.yml
pbalcer: No, we should just get rid of the nightly.

.github/workflows/e2e_core.yml
pbalcer: The whole job probably. There might be scenarios (e.g., API changes in UR) where SYCL won't even compile.

.github/workflows/e2e_cuda.yml
lukaszstolarczuk: Done.

.github/workflows/e2e_core.yml
lukaszstolarczuk: Done.

pbalcer (approved):

PR #1713, opened 2024-06-05
[HIP] Add support for querying USM base pointer

frasercrmck:
Also refactor some common code into helper methods.

### DIFF
--- a/source/adapters/hip/common.hpp
+++ b/source/adapters/hip/common.hpp
@@ -204,3 +204,38 @@ public:
   /// UR object.
   void dismiss() { Captive = nullptr; }
 };
+
+// Helper method to return a (non-null) pointer's attributes, or std::nullopt in
+// the case that the pointer is unknown to the HIP subsystem.
+inline static std::optional<hipPointerAttribute_t>
+getPointerAttributes(const void *pMem) {
+  // do not throw if hipPointerGetAttributes returns hipErrorInvalidValue
+  hipPointerAttribute_t hipPointerAttributes;
+  hipError_t Ret = hipPointerGetAttributes(&hipPointerAttributes, pMem);
+  if (Ret == hipErrorInvalidValue && pMem) {
+    // pointer non-null but not known to the HIP subsystem
+    return std::nullopt;
+  }
+  // Direct usage of the function, instead of UR_CHECK_ERROR, so we can get
+  // the line offset.
+  checkErrorUR(Ret, __func__, __LINE__ - 7, __FILE__);
+  // ROCm 6.0.0 introduces hipMemoryTypeUnregistered in the hipMemoryType
+  // enum to mark unregistered allocations (i.e., via system allocators).
+#if HIP_VERSION_MAJOR >= 6
+  if (hipPointerAttributes.type == hipMemoryTypeUnregistered) {
+    // pointer not known to the HIP subsystem
+    return std::nullopt;
+  }
+#endif
+  return hipPointerAttributes;
+}
+
+// Helper method to abstract away the fact that retrieving a pointer's memory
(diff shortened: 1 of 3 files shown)


source/adapters/hip/usm.cpp
```diff
@@ -148,45 +148,56 @@ ur_result_t USMHostAllocImpl(void **ResultPtr,
   return UR_RESULT_SUCCESS;
 }
 
+static std::optional<hipPointerAttribute_t>
```
hdelan, line 151: Nit: I think in general we use anonymous namespace rather than static, although this is sort of matter of style

frasercrmck: I'll go with that, thanks

GeorgeWeb (commented):
Can we move both helpers to adapters/hip/common.hpp - I'd like to make use of them in adapters/hip/enqueue.cpp as well where we workaround getting the memory type to decide the copy direction for the `urEnqueueUSMMemcpy2D` entry point in order to figure out if any of the pointers from an unregistered/os-allocation. Otherwise, I am glad it finally came to refactoring that mess that HIP created. Thanks!

---

You don't need to refactor the the `urEnqueueUSMMemcpy2D` entry point in the same PR, can be done separately, but the helpers will be much appreciated in common. Can be use as well in other places where we enqueue commands on a usm pointer to check if it is a valid unregistered one but can be used in the driver call because system shared usm / pageable memory access is supported on the system.

frasercrmck:
> Can we move both helpers to adapters/hip/common.hpp - I'd like to make use of them in adapters/hip/enqueue.cpp as well where we workaround getting the memory type to decide the copy direction for the `urEnqueueUSMMemcpy2D` entry point in order to figure out if any of the pointers from an unregistered/os-allocation. Otherwise, I am glad it finally came to refactoring that mess that HIP created. Thanks!
> 
> You don't need to refactor the the `urEnqueueUSMMemcpy2D` entry point in the same PR, can be done separately, but the helpers will be much appreciated in common. Can be use as well in other places where we enqueue commands on a usm pointer to check if it is a valid unregistered one but can be used in the driver call because system shared usm / pageable memory access is supported on the system.

Good idea, thanks. I've pushed that change now :+1:

aarongreig (approved):

hdelan (approved):
LGTM

source/adapters/hip/common.hpp
```diff
@@ -204,3 +204,38 @@ template <typename T> class ReleaseGuard {
...
 };
+
+// Helper method to return a pointer's attributes, or std::nullopt in the case
+// that the pointer is unknown to the HIP subsystem.
+inline static std::optional<hipPointerAttribute_t>
+getPointerAttributes(const void *pMem) {
+  // do not throw if hipPointerGetAttributes returns hipErrorInvalidValue
+  hipPointerAttribute_t hipPointerAttributes;
+  hipError_t Ret = hipPointerGetAttributes(&hipPointerAttributes, pMem);
+  if (Ret == hipErrorInvalidValue) {
```
GeorgeWeb, line 215: Last one, I just caught that - `hipErrorInvalidValue` can also be returned very early, before even checking the pointer base type, if either the ptr or the attributes passed are nullptr. In this case I think it we need to change the if-statement to
`if (Ret == hipErrorInvalidValue && pMem)` ensuring that the we are not going to skip erroring out on actual null memory.

GeorgeWeb (changes requested):

source/adapters/hip/common.hpp
frasercrmck: Thanks! I've also updated the function's comments to try and clarify that a null pointer is an error, rather than returning `std::nullopt`

PR #1717, opened 2024-06-05
[CUDA] use advise ACCESSED_BY for all devs in `ur_context_handle_t` for managed memory

JackAKirk:
This change improves the performance of managed memory in the cuda backend for all devices tested, by resulting in better default on device caching of the managed memory. We have tested this using a range of use cases. The performance benefit increases with increased number of devices, but can still be considerable when using only a single device if the managed memory is accessed frequently.
There appears to be no drawbacks to doing this by default (no use case was found where this led to a drop in performance), and the observed behaviour with respect to copies to and from the host now matches to what we observe to happen in the cuda runtime api. If a device in the picontext does not ever access the shared memory this use case also does not lead to a drop in performance even though we mark all devices in the picontext as ACCESSED_BY for this managed memory.


source/adapters/cuda/usm.cpp
```diff
@@ -159,6 +159,12 @@ ur_result_t USMSharedAllocImpl(void **ResultPtr, ur_context_handle_t,
     ScopedContext Active(Device);
     UR_CHECK_ERROR(cuMemAllocManaged((CUdeviceptr *)ResultPtr, Size,
                                      CU_MEM_ATTACH_GLOBAL));
+    for (const auto &device : Context->getDevices()) {
```
frasercrmck, line 162: nit: I don't claim to know anything about the coding style used here, but we have `Context`, `Device`, `Size,` and `Err` in this function, and we're introducing lower-case `device` here. Is it not inconsistent, and isn't it confusing to have two uses of "device" in the same function?

JackAKirk: > Agreed. I think in general good to go for `hDevice` for the device arg and then `Dev` for the local device. But I don't think the device arg is actually needed? Maybe better to do:
> 
> ```
> ScopedContext Active(Context->getDevices()[0]);
> ```
> 
> And just remove the `Device` arg altogether

OK I'll switch the naming. But I don't think any device param can be removed. 

What is happening here is that a stream within CUcontext of the Device that was associated with a runtime (sycl) queue is allocating managed memory. Probably in many cases we could ignore this prescription provided by the programmer and just use the first device from the `Context->getDevices()[0]` and this wouldn't matter to a user. However I think it could do. 
Say for example an application from User A is on a node but sees all devices in the platform on that node (in perlmutter there are 4 gpus). If the Context corresponds to a sycl default context then all four gpus will be in the context. The user provides device 1 as the device whos allocated CUcontext they prescribe to allocate the managed memory via. User A only wants to use device 0. We mark accessed_by for all devices but this isn't used - not a big deal, it is cheap
User B is using device 0 for a different application on the same node. I imagine that if user B is saturating the gpu, it might be a bad idea for us to break the prescription from user A and use device 0 instead.
Probably it doesn't make a big difference in practice, but hopefully the above details there is a potential difference. If the interface used provides a specific device, then I think this should be used as the "command device"

hdelan: In that case we should also be calling `setAccessedBy` for the `hDevice` param, or asserting that the device is within the context

JackAKirk: OK I'll add an assert.

source/adapters/cuda/usm.cpp
```diff
@@ -151,18 +151,22 @@ ur_result_t USMDeviceAllocImpl(void **ResultPtr, ur_context_handle_t,
...
     UR_CHECK_ERROR(cuMemAllocManaged((CUdeviceptr *)ResultPtr, Size,
                                      CU_MEM_ATTACH_GLOBAL));
-    for (const auto &device : Context->getDevices()) {
-      if (getAttribute(device, CU_DEVICE_ATTRIBUTE_CONCURRENT_MANAGED_ACCESS)) {
+    if (getAttribute(CommandDevice, CU_DEVICE_ATTRIBUTE_CONCURRENT_MANAGED_ACCESS)) {
+        UR_CHECK_ERROR(cuMemAdvise((CUdeviceptr)*ResultPtr, Size,
+                                   CU_MEM_ADVISE_SET_ACCESSED_BY, CommandDevice->get()));
+      }
+    for (const auto &Dev : Context->getDevices()) {
+      if (getAttribute(Dev, CU_DEVICE_ATTRIBUTE_CONCURRENT_MANAGED_ACCESS)) {
```
GeorgeWeb, line 167: It feelks like we should cache the result for this property on device initialisation to save us from querying it at this point regardless of how cheap it likely is. It is also used is some other USM managed memory entry points too.
That's a just a note, can totally be a separate PR.

JackAKirk: I'm not sure it needs to be used again? Where are you thinking it would be reused?

GeorgeWeb: Yeh pretty much my thinking and set in the `ur_device_handle_t_` constructor.

source/adapters/cuda/usm.cpp
JackAKirk: I've just ended up calling the advise also via the commandDevice. This will lead to a duplication in normal circumstances, but I've checked you can advise twice without an error.

source/adapters/cuda/usm.cpp
JackAKirk: But yeh probs to just merge this asap, so people can actually use malloc_shared in sycl cuda backend and the performance not being way worse than cuda runtime.

source/adapters/cuda/usm.cpp
hdelan: Yeah that sounds good to me.

hdelan (approved):

JackAKirk:
Closing in favour of https://github.com/oneapi-src/unified-runtime/pull/1774

PR #1718, opened 2024-06-05
[L0] fix report of runtime version as api version_driverVersion

nrspruit:


ph0b:
how can developers/end-users associate a number such as 1.3.4294967295 with an actual driver ? human readable versions like what's reported by the OpenCL backend (ex. 31.0.101.5534 on Windows defined here https://www.intel.com/content/www/us/en/support/articles/000005654/graphics.html, 24.17.29377.6 on Linux which is compute-runtime version) would be much better.
This is basic requirements for ISVs to understand application debug logs and establish driver version filters to avoid crashes/incompatibilities.

nrspruit:
not moving ahead with this change, closing.

PR #1720, opened 2024-06-06
Add [release] tag to parameters where appropriate.

aarongreig:


aarongreig:
@pbalcer I found that a part of the loader template is expecting these [release] tags, and adding them generates additional code that releases loader handles. We've been fine without this so far, I only noticed because having [release] is also useful for the mock replacement layer. Do you know if the omission of these was deliberate for some reason?

pbalcer:
> Do you know if the omission of these was deliberate for some reason?

No, probably an accidental omission. LGTM.

EwanC (approved):
command-buffer LGTM

Is there a list of all the possible tags somewhere? Didn't know this one existed

DBDuncan:
`[Release]` tags seem to be missing from line 410 and line 621 of https://github.com/oneapi-src/unified-runtime/pull/1720/files#diff-a02a3dd544128a2e5c5b34b481f592108bd0c8cabc9d961b46ad46346bf94e72

aarongreig:
> Is there a list of all the possible tags somewhere? Didn't know this one existed

not that I could find, created an issue https://github.com/oneapi-src/unified-runtime/issues/1742

pbalcer:
> > Is there a list of all the possible tags somewhere? Didn't know this one existed
> 
> not that I could find, created an issue #1742

I think they are supposed to be documented here: https://github.com/oneapi-src/unified-runtime/blob/main/scripts/YaML.md. The `release` that is listed there.

aarongreig:
> > > Is there a list of all the possible tags somewhere? Didn't know this one existed
> > 
> > 
> > not that I could find, created an issue #1742
> 
> I think they are supposed to be documented here: https://github.com/oneapi-src/unified-runtime/blob/main/scripts/YaML.md. The `release` that is listed there.

ah nice, was looking in `scripts/core`

source/loader/ur_ldrddi.cpp
```diff
@@ -102,6 +102,13 @@ __urdlllocal ur_result_t UR_APICALL urAdapterRelease(
     // forward to device-platform
     result = pfnAdapterRelease(hAdapter);
 
+    if (UR_RESULT_SUCCESS != result) {
+        return result;
+    }
+
+    // release loader handle
+    ur_adapter_factory.release(hAdapter);
```
aarongreig, line 110: eh.. it occurred to me as a result of adding an equivalent [retain] tag in #1759 that this isn't reference counted, it's just unconditionally releasing on the first `Release` call that comes in, might need to add a reference counting mechanism to the factory objects

isaacault (commented):
tl;dr question at the bottom.

---

I was about to request some additions, but then I found the link above with the definition of the [release] tag as

> Used for params that are handles or pointers to handles where the function will destroy any backing memory associated with the handle(s).

We have a somewhat grey area here with the bindless images API as we rely on 2 types of objects:

1. A handle to the underlying memory i.e. `ur_exp_image_mem_handle_t`, `ur_exp_interop_mem_handle_t`
2. A handle to the object in the memory i.e. `ur_exp_image_handle_t`.

Before reading the definition of the [release] tag, I thought we should add the tag to any params which are destroyed and whereby use of the param after the function call would result in undefined behaviour. This would then require adding the tag to:

* `urBindlessImagesUnsampledImageHandleDestroyExp`
* `urBindlessImagesSampledImageHandleDestroyExp`
* `urBindlessImagesDestroyExternalSemaphoreExp`.

However, if we continue with the strict definition above, the only bindless images function which actually "destroys the backing memory" would be 

* `urBindlessImagesImageFreeExp`

as all others are destroying the objects within the memory, and not the memory itself. Note this includes the interop functions as the interop API only creates object handles and does not take ownership of the underlying memory.  

---

So I'm asking for clarification: is the [release] tag meant to be limited to destroying memory, or should the definition be expanded for destroyed handles for which the underlying memory remains intact?

aarongreig:
I think we need to re-write the definition in terms of UR handles, really this is to provide lifetime information about handles to the loader and layers rather than anything to do with underlying memory. For example the only mechanism this currently affects is our loader handle/platform handle mapping:

output handles are returned in a loader wrapper e.g. [here](https://github.com/oneapi-src/unified-runtime/blob/06cb1b907eb5cbcee4c88a5c1bc01cf4fb53c8c8/source/loader/ur_ldrddi.cpp#L6052)
```
    try {
        // convert platform handle to loader handle
        *phImage = reinterpret_cast<ur_exp_image_handle_t>(
            ur_exp_image_factory.getInstance(*phImage, dditable));
    } catch (std::bad_alloc &) {
        result = UR_RESULT_ERROR_OUT_OF_HOST_MEMORY;
    }
```
the factory objects are [instantiations](https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_ldrddi.hpp#L67) of [singleton_factory_t](https://github.com/oneapi-src/unified-runtime/blob/main/source/common/ur_singleton.hpp#L20), the immediate effect of adding in the `[release]` tags is that we now generate calls to [release](https://github.com/oneapi-src/unified-runtime/blob/main/source/common/ur_singleton.hpp#L70) which removes dead handles from the platform/loader handle map

(as mentioned in another comment I'll need to add some kind of reference counting in there - we inherited this mechanism from level zero which doesn't have `Retain` type entry points so it made sense there to immediately erase a released handle)

isaacault (approved):
Bindless images LGTM

PR #1725, opened 2024-06-06
[CUDA][HIP] Fix build info log reporting

frasercrmck:
We were always returning the maximum log size (of 8192) even when the actual log was much shorter. Strictly speaking, we should be returning the log only up to the first null terminator. Doing otherwise could cause strange behaviour, and may account for the flaky fails we've seen in this test on both adapters.


mmoadeli (approved):
LGTM

PR #1722, opened 2024-06-06
fix permissions in e2e workflow

pbalcer:
Ref: #1719

PR #1732, opened 2024-06-07
[GHA] Fix e2e workflow for forks

kbenzie:
This follows on from #1722 which set the `pull-request: write`
permission on for the e2e workflows. However, this did not fully resolve
the 403 issue, see [1], [2], and [3]. Those jobs have the following
token permissions:

```
GITHUB_TOKEN Permissions
  Contents: read
  Metadata: read
  PullRequests: read
```

After some digging, I found a [post](https://stackoverflow.com/a/78444521)
which notes that using the `pull_request` workflow trigger will never
grant write permissions when the pull request comes from a public fork.

The solution is to instead use the `pull_request_target` trigger, I
found this [post](https://stackoverflow.com/a/74959635) helpful reading
on the trade-offs of using this trigger. The main difference, other than
allowing write permissions, is that the workflow config used comes from
the pull request's base branch so does not include any workflow changes
from the pull request.

As such, in this patch I've split out the e2e workflows from our main
build & test workflow to restrict the scope of jobs which may be granted
write permissions.

[1]: https://github.com/oneapi-src/unified-runtime/actions/runs/9417159892/job/25942801242?pr=1639
[2]: https://github.com/oneapi-src/unified-runtime/actions/runs/9417159892/job/25942802400?pr=1639
[3]: https://github.com/oneapi-src/unified-runtime/actions/runs/9417159892/job/25942803189?pr=1639


kbenzie:
I do not expect the e2e workflows to run in the pull request due to the changes I've made to use the `pull_request_target` workflow trigger. They will only begin to take effect post-merge.

pbalcer:
From one of your links:
> event runs against the workflow and code from the base of the pull request

I think this means that we need to add a step to fetch the code from the actual PR:
https://github.com/oneapi-src/unified-runtime/commit/3f60e4e0b3eb2ea55d4984e9371c264ee79139a6#diff-c68dec5b230f8243fe4144d52e61bbad25a1aecad217e3a1efab02195e8e8713L109-L118

kbenzie:
> From one of your links:
> 
> > event runs against the workflow and code from the base of the pull request
> 
> I think this means that we need to add a step to fetch the code from the actual PR: [3f60e4e#diff-c68dec5b230f8243fe4144d52e61bbad25a1aecad217e3a1efab02195e8e8713L109-L118](https://github.com/oneapi-src/unified-runtime/commit/3f60e4e0b3eb2ea55d4984e9371c264ee79139a6#diff-c68dec5b230f8243fe4144d52e61bbad25a1aecad217e3a1efab02195e8e8713L109-L118)

My reading of that was only the workflow config is read from the base branch. Since the checkout action is responsible for actually checking out the code I believe this rule from the [docs](https://github.com/actions/checkout) has the same behaviour either `pull_request` or `pull_request_target` are used:

```yaml
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
```

kbenzie:
The annoying thing about that is errors like [this](https://github.com/oneapi-src/unified-runtime/actions/runs/9445948022) don't show up until you merge. I'll push a fix to the oneapi-src to actually test it this time.

lukaszstolarczuk:
ech, these permissions :/ The other solution would be to just drop writing comments. I added them, because we have "continue-on-error" enabled and GHA does not provide us any info if it indeed was "continued on error".
We could drop "continue-on-error", have red statuses from E2E workflows and ignore them "manually" when merging...?

kbenzie:
> edit: och, it seems I'm too late 😄

sorry 😬 

> ech, these permissions :/ The other solution would be to just drop writing comments

yeah, I don't love that though. 

> We could drop "continue-on-error", have red statuses from E2E workflows and ignore them "manually" when merging...?

This would make managing the merge queue much harder, so I'd like to avoid this if possible.

PR #1734, opened 2024-06-10
[GHA] Revert e2e workflow & disable comment

kbenzie:
- **Revert "Merge pull request #1732 from kbenzie/benie/gha-fix-e2e-for-forks"**
- **[GHA] Disable the e2e comment step**


pbalcer (approved):

lukaszstolarczuk (approved):
if we go with this change we should make sure anyone from the team is aware of potentially hidden issues within E2E logs

pbalcer:
I found something like this: https://github.com/marketplace/actions/wait-on-check It polls PRs jobs for status. I wonder whether it would be possible to do that, and then extract E2E test status from the job logs.

kbenzie:
> if we go with this change we should make sure anyone from the team is aware of potentially hidden issues within E2E logs

My understanding is the checks will still show as red when there are e2e failures due to the checks UI not being great when combined with `continue-on-error: true`.

> I found something like this: https://github.com/marketplace/actions/wait-on-check It polls PRs jobs for status. I wonder whether it would be possible to do that, and then extract E2E test status from the job logs.

That could certainly be worth investigating.

lukaszstolarczuk:
> I found something like this: https://github.com/marketplace/actions/wait-on-check It polls PRs jobs for status. I wonder whether it would be possible to do that, and then extract E2E test status from the job logs.

you could wait up, read the log, but there would be still the same problem - reporting it back to PR requires `write` permission, which we don't have here... or am I missing something in this `wait-on-check` action?


> > if we go with this change we should make sure anyone from the team is aware of potentially hidden issues within E2E logs
> 
> My understanding is the checks will still show as red when there are e2e failures due to the checks UI not being great when combined with `continue-on-error: true`.

I'm not sure if I understand correctly - setting `continue-on-error: true` for the job makes it always green (but hides the real problem). If we want the job to be red we should just remove this param (currently set [in e2e_core.yml](https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/e2e_core.yml#L79)).

kbenzie:
> I'm not sure if I understand correctly - setting `continue-on-error: true` for the job makes it always green (but hides the real problem). If we want the job to be red we should just remove this param (currently set [in e2e_core.yml](https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/e2e_core.yml#L79)).

I've maybe misunderstood this [discussion](https://github.com/orgs/community/discussions/15452), then. Lets see how things look now this is merged.

kbenzie:
> > I'm not sure if I understand correctly - setting `continue-on-error: true` for the job makes it always green (but hides the real problem). If we want the job to be red we should just remove this param (currently set [in e2e_core.yml](https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/e2e_core.yml#L79)).
> 
> I've maybe misunderstood this [discussion](https://github.com/orgs/community/discussions/15452), then. Lets see how things look now this is merged.

So it does look like it shows up as red [here](https://github.com/oneapi-src/unified-runtime/pull/1731).

image

lukaszstolarczuk:
So I don't understand what this `continue-on-error` does, after all 😕
We set this up in [core_e2e.yml#79](https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/e2e_core.yml#L79)
[Docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) says it should `Prevents a workflow run from failing when a job fails.`

lukaszstolarczuk:
@kbenzie, is the current situation acceptable for you?

pbalcer:
> So I don't understand what this `continue-on-error` does, after all 😕
Hm, maybe we can try setting this for steps, not the whole job?

kbenzie:
> So I don't understand what this `continue-on-error` does, after all 😕

So my understanding is it means the steps following the failing step still run and are not cancelled. This is why I was talking about `|| true` yesterday, to mark the step as passed/green, even when it fails.

> @kbenzie, is the current situation acceptable for you?

This is workable for now, not ideal but I'm also not wanting to spend a lot more time on this.

lukaszstolarczuk:
> > So I don't understand what this `continue-on-error` does, after all 😕
> 
> Hm, maybe we can try setting this for steps, not the whole job?

yeah, let's try this out, it might behave slightly differently:
https://github.com/oneapi-src/unified-runtime/pull/1745

PR #1738, opened 2024-06-10
[CMDBUF][L0] Use copy engine to optimize cmd-buffer usage

Bensuo:
- Out-of-order command-buffers now use copy-engine in some cases where available
- Also respect existing L0 env vars for using copy-engine with fill and d2d copies

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -41,17 +41,30 @@ bool IsDriverVersionNewerOrSimilar(ur_context_handle_t Context,
           (DriverVersionMinor >= VersionMinor) &&
           (DriverVersionBuild >= VersionBuild));
 }
+
+// Default to using compute engine for fill operation, but allow to
+// override this with an environment variable.
+bool PreferCopyEngineForFill = [] {
+  const char *UrRet = std::getenv("UR_L0_USE_COPY_ENGINE_FOR_FILL");
+  const char *PiRet =
+      std::getenv("SYCL_PI_LEVEL_ZERO_USE_COPY_ENGINE_FOR_FILL");
+  return (UrRet ? std::stoi(UrRet) : (PiRet ? std::stoi(PiRet) : 0));
+}();
+
 }; // namespace
 
 ur_exp_command_buffer_handle_t_::ur_exp_command_buffer_handle_t_(
     ur_context_handle_t Context, ur_device_handle_t Device,
     ze_command_list_handle_t CommandList,
     ze_command_list_handle_t CommandListResetEvents,
+    ze_command_list_handle_t CopyCommandList,
     ZeStruct<ze_command_list_desc_t> ZeDesc,
+    ZeStruct<ze_command_list_desc_t> ZeCopyDesc,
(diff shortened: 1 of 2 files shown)


Bensuo:
DPC++ PR: https://github.com/intel/llvm/pull/13051

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -41,17 +41,30 @@ bool IsDriverVersionNewerOrSimilar(ur_context_handle_t Context,
           (DriverVersionMinor >= VersionMinor) &&
           (DriverVersionBuild >= VersionBuild));
 }
+
+// Default to using compute engine for fill operation, but allow to
+// override this with an environment variable.
+static bool PreferCopyEngineForFill = [] {
```
EwanC, line 47: Don't think `static` adds anything here when we're in an anonymous namespace.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -74,6 +87,9 @@ ur_exp_command_buffer_handle_t_::~ur_exp_command_buffer_handle_t_() {
   if (ZeCommandList) {
```
EwanC, line 87: I'm not 100% about this, but wondering if we should rename `ZeCommandList` to `ZeComputeCommandList` everywhere

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -427,20 +463,51 @@ static ur_result_t enqueueCommandBufferMemCopyRectHelper(
...
-              ->QueueGroup[ur_device_handle_t_::queue_group_info_t::Compute]
-              .ZeProperties.maxMemoryFillPatternSize,
-      UR_RESULT_ERROR_INVALID_VALUE);
+  ze_command_list_handle_t ZeCommandList;
+  // If the copy engine available, the command is enqueued in the
+  // ZeCopyCommandList.
+  if (PreferCopyEngine && CommandBuffer->UseCopyEngine()) {
+    // Check engine fill capabilty (as this is done for non-graph cmd
+    // submission)
+    if (CommandBuffer->Device->hasMainCopyEngine() &&
```
EwanC, line 479: When we create the copy command-list, we already check `Device->hasMainCopyEngine() `, so I think we could remove it from this check as we're preconditioned on `UseCopyEngine()`.

Ideally we could also refactor this `if` so that's it's to conditions are copy engine or compute. Rather than having this nested if

source/adapters/level_zero/command_buffer.cpp
Bensuo: Agreed, have removed `static` here.

source/adapters/level_zero/command_buffer.cpp
Bensuo: I think this makes sense so I'e made the change.

source/adapters/level_zero/command_buffer.cpp
Bensuo: Good points both, I've removed the extra check and refactored this to be simpler.

EwanC (approved):
LGTM

reble (approved):

PR #1739, opened 2024-06-10
[UR] Add some information about buffers in spec

hdelan:
Add a short section for Memory in UR, which says how ur_mem_handle_ts can be expected to work.


RossBrunton (approved):

hdelan:
@kbenzie looks like this is good to go

PR #1741, opened 2024-06-11
[UR] Test that an in order queue returns same backend q

hdelan:
Test that an in order queue will return the same backend queue when calling urQueueGetNativeHandle.

Ping @JackAKirk

### DIFF
--- a/test/conformance/queue/urQueueGetNativeHandle.cpp
+++ b/test/conformance/queue/urQueueGetNativeHandle.cpp
@@ -24,3 +24,17 @@ TEST_P(urQueueGetNativeHandleTest, InvalidNullPointerNativeHandle) {
     ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_NULL_POINTER,
                      urQueueGetNativeHandle(queue, nullptr, nullptr));
 }
+
+TEST_P(urQueueGetNativeHandleTest, InOrderQueueSameNativeHandle) {
+    ur_queue_handle_t in_order_queue;
+    ur_native_handle_t native_handle1 = nullptr, native_handle2 = nullptr;
+    ASSERT_SUCCESS(urQueueCreate(context, device, nullptr, &in_order_queue));
+    if (auto error =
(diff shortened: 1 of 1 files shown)


test/conformance/queue/urQueueGetNativeHandle.cpp
```diff
@@ -24,3 +24,17 @@ TEST_P(urQueueGetNativeHandleTest, InvalidNullPointerNativeHandle) {
...
+    ur_queue_handle_t in_order_queue;
+    ur_native_handle_t native_handle1 = nullptr, native_handle2 = nullptr;
+    ASSERT_SUCCESS(urQueueCreate(context, device, nullptr, &in_order_queue));
+    if (auto error =
+            urQueueGetNativeHandle(in_order_queue, nullptr, &native_handle1)) {
+        ASSERT_EQ_RESULT(UR_RESULT_ERROR_UNSUPPORTED_FEATURE, error);
+        return;
+    }
+    ASSERT_SUCCESS(
+        urQueueGetNativeHandle(in_order_queue, nullptr, &native_handle2));
```
aarongreig, line 38: is it correct that we're doing this twice on the same queue and making sure the output doesn't change? I'm curious about the circumstances that could make this fail

hdelan: SYCL interop with in order queues may rely on `ih.get_native_queue` returning the same `CUstream` over multiple invocations. This `ih.get_native_queue` ends up calling this UR entry point so I think it's best to check in UR

hdelan: With an out of order queue, each call will likely return different native handles, since an out of order queue maps to multiple `CUstream`s

hdelan: Does that answer your question @aarongreig ?

aarongreig: ok makes sense, thanks for explaining

aarongreig (approved):

PR #1748, opened 2024-06-12
Logger callback function sink

martygrant:
This PR implements https://github.com/oneapi-src/unified-runtime/issues/1330 through a new logger sink: a user configurable callback. It introduces some spec additions:

- `typedef void (*ur_logger_output_callback_t)(ur_logger_level_t level, const char *pLoggerMsg, void *pUserData)`
- `urSetLoggerCallback(ur_adapter_handle_t hAdapter, ur_logger_output_callback_t pfnLoggerCallback, void *pUserData, ur_logger_level_t level`
)`
- `urSetLoggerCallbackLevel(ur_adapter_handle_t hAdapter, ur_logger_level_t level)`
- `typedef enum ur_logger_level_t` (moved the logger::level enum into the spec)

This new logger sink will only be constructed once a user makes a call to `urSetLoggerCallback`, supplying their own callback function. They can set the minimum logging level through `urSetLoggerCallbackLevel`. Any subsequent logging calls will additionally make a call to the supplied callback where the log level, message and user data will be sent.

A new test suite `LoggerWithCallbackSink` has been added to test this new functionality.

intel/llvm PR: https://github.com/intel/llvm/pull/14422

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -381,54 +381,58 @@ typedef enum ur_function_t {
   UR_FUNCTION_COMMAND_BUFFER_GET_INFO_EXP = 221,
   /// Enumerator for ::urEnqueueTimestampRecordingExp
   UR_FUNCTION_ENQUEUE_TIMESTAMP_RECORDING_EXP = 223,
-  /// Enumerator for ::urEnqueueKernelLaunchCustomExp
-  UR_FUNCTION_ENQUEUE_KERNEL_LAUNCH_CUSTOM_EXP = 224,
-  /// Enumerator for ::urKernelGetSuggestedLocalWorkSize
-  UR_FUNCTION_KERNEL_GET_SUGGESTED_LOCAL_WORK_SIZE = 225,
-  /// Enumerator for ::urBindlessImagesImportExternalMemoryExp
-  UR_FUNCTION_BINDLESS_IMAGES_IMPORT_EXTERNAL_MEMORY_EXP = 226,
-  /// Enumerator for ::urBindlessImagesImportExternalSemaphoreExp
-  UR_FUNCTION_BINDLESS_IMAGES_IMPORT_EXTERNAL_SEMAPHORE_EXP = 227,
-  /// Enumerator for ::urEnqueueNativeCommandExp
-  UR_FUNCTION_ENQUEUE_NATIVE_COMMAND_EXP = 228,
   /// Enumerator for ::urLoaderConfigSetMockingEnabled
-  UR_FUNCTION_LOADER_CONFIG_SET_MOCKING_ENABLED = 229,
+  UR_FUNCTION_LOADER_CONFIG_SET_MOCKING_ENABLED = 224,
+  /// Enumerator for ::urAdapterSetLoggerCallback
+  UR_FUNCTION_ADAPTER_SET_LOGGER_CALLBACK = 225,
+  /// Enumerator for ::urAdapterSetLoggerCallbackLevel
+  UR_FUNCTION_ADAPTER_SET_LOGGER_CALLBACK_LEVEL = 226,
+  /// Enumerator for ::urKernelGetSuggestedLocalWorkSize
+  UR_FUNCTION_KERNEL_GET_SUGGESTED_LOCAL_WORK_SIZE = 227,
+  /// Enumerator for ::urBindlessImagesImportExternalMemoryExp
+  UR_FUNCTION_BINDLESS_IMAGES_IMPORT_EXTERNAL_MEMORY_EXP = 228,
+  /// Enumerator for ::urBindlessImagesMapExternalLinearMemoryExp
+  UR_FUNCTION_BINDLESS_IMAGES_MAP_EXTERNAL_LINEAR_MEMORY_EXP = 229,
   /// Enumerator for ::urBindlessImagesReleaseExternalMemoryExp
   UR_FUNCTION_BINDLESS_IMAGES_RELEASE_EXTERNAL_MEMORY_EXP = 230,
+  /// Enumerator for ::urBindlessImagesImportExternalSemaphoreExp
+  UR_FUNCTION_BINDLESS_IMAGES_IMPORT_EXTERNAL_SEMAPHORE_EXP = 231,
   /// Enumerator for ::urCommandBufferAppendUSMMemcpyExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_USM_MEMCPY_EXP = 231,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_USM_MEMCPY_EXP = 232,
   /// Enumerator for ::urCommandBufferAppendUSMFillExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_USM_FILL_EXP = 232,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_USM_FILL_EXP = 233,
   /// Enumerator for ::urCommandBufferAppendMemBufferCopyExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_COPY_EXP = 233,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_COPY_EXP = 234,
   /// Enumerator for ::urCommandBufferAppendMemBufferWriteExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_WRITE_EXP = 234,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_WRITE_EXP = 235,
   /// Enumerator for ::urCommandBufferAppendMemBufferReadExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_READ_EXP = 235,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_READ_EXP = 236,
   /// Enumerator for ::urCommandBufferAppendMemBufferCopyRectExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_COPY_RECT_EXP = 236,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_COPY_RECT_EXP = 237,
   /// Enumerator for ::urCommandBufferAppendMemBufferWriteRectExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_WRITE_RECT_EXP = 237,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_WRITE_RECT_EXP = 238,
   /// Enumerator for ::urCommandBufferAppendMemBufferReadRectExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_READ_RECT_EXP = 238,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_READ_RECT_EXP = 239,
   /// Enumerator for ::urCommandBufferAppendMemBufferFillExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_FILL_EXP = 239,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_MEM_BUFFER_FILL_EXP = 240,
   /// Enumerator for ::urCommandBufferAppendUSMPrefetchExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_USM_PREFETCH_EXP = 240,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_USM_PREFETCH_EXP = 241,
   /// Enumerator for ::urCommandBufferAppendUSMAdviseExp
-  UR_FUNCTION_COMMAND_BUFFER_APPEND_USM_ADVISE_EXP = 241,
+  UR_FUNCTION_COMMAND_BUFFER_APPEND_USM_ADVISE_EXP = 242,
   /// Enumerator for ::urCommandBufferEnqueueExp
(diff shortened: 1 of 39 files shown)


scripts/core/loader.yml
```diff
@@ -225,3 +225,23 @@ ordinal: "1"
...
+type: function
+desc: "Set a callback function for use by the logger to retrieve logging output."
+class: $xLoaderConfig
+loader_only: True
+name: SetLoggerCallback
+decl: static
+params:
+    - type: $x_logger_output_callback_t
+      name: pfnLoggerCallback
+      desc: "[in] Function pointer to callback from the logger."
```
kbenzie, line 247: There should always be a `void * user_data` argument which gets passed into the callback to let the user access any context they need.

pbalcer:
If the idea is for the SYCL to hoist UR logs and print them using its own facility, wouldn't it make sense for `urLoaderConfigSetLoggerCallback` to override (or be in addition to) the default sink so that the users only ever need to interact with SYCL interfaces?

scripts/core/loader.yml
martygrant: I've added this now. It's passed all the way through to the logger sink not sure if that's correct?

test/unit/logger/logger.cpp
```diff
@@ -207,3 +208,15 @@ TEST_P(CommonLoggerWithMultipleThreads, StdoutMultithreaded) {
         thread.join();
     }
 }
+
+//////////////////////////////////////////////////////////////////////////////
+void printLoggerMessages(const char *msg) { std::cout << msg << std::endl; }
```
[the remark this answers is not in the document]
martygrant: Well spotted, I was using it to debug the callback was working but didn't add any actual testing functionality. None of the other tests here actually do any tests either it seems, just printing the messages. I can add an issue for this and get round to it myself.

source/loader/ur_lib.cpp
```diff
@@ -215,6 +215,17 @@ urLoaderConfigSetCodeLocationCallback(ur_loader_config_handle_t hLoaderConfig,
     return UR_RESULT_SUCCESS;
 }
 
+ur_result_t
+urLoaderConfigSetLoggerCallback(ur_logger_output_callback_t pfnLoggerCallback) {
+    if (!pfnLoggerCallback) {
+        return UR_RESULT_ERROR_INVALID_NULL_POINTER;
+    }
+
+    logger::get_logger().setCallbackSinkFunction(pfnLoggerCallback);
```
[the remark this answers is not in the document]
martygrant: I meant to highlight this as a question because I wasn't sure about how to get the callback function pointer into the logger as `get_logger` takes a name parameter. I can't see any loader references in DDI tables, would I need to add one for the loader or should the entry point be part of another group maybe?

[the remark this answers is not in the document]
pbalcer: This is not really a loader API, so I'd call this something like `urSetLoggerCallback` and remove `loader_only` from the spec. This will generate a DDI function where you can implement this for each adapter. And each adapter will then need to have the `logger::get_logger().setSink()` call in their implementation. Because the instance returned by `logger::get_logger()` is different for each adapter.

Additionally, if you want the application to be able to also get callbacks on loader-related messages, you'd need to make an exception in the template file so that do the `logger::get_logger().setSink()` at the loader layer, before the function is forwarded to the adapter-specific implementation.

pbalcer:
> @pbalcer thanks for your comments! I'm not sure about this one maybe you can clarify?
> 
> > If the idea is for the SYCL to hoist UR logs and print them using its own facility, wouldn't it make sense for `urLoaderConfigSetLoggerCallback` to override (or be in addition to) the default sink so that the users only ever need to interact with SYCL interfaces?
> 
> I think you mean so the user at the SYCL level doesn't need to interact with UR directly right? I'm not sure if that would be possible for this case as the callback would need to be passed through programmatically?

The way you implemented it right now, for a SYCL end-user (I don't mean developer of SYCL runtime) to receive any sort of messages from the UR layer, they'd have to set `UR_LOG_FOO=callback` env variable. If the plan for the callback is to pass the responsibility for emitting logs to SYCL, it would make more sense to me if SYCL was also responsible for controlling the verbosity of those logs, using whatever methods that exist today. This way, end-users interact only with the highest level of abstraction (SYCL), not having to be aware of the underlying implementation (UR).

But for that to be possible, the set callback function needs to always work, independent of the UR-level log setting.

scripts/core/loader.yml
```diff
@@ -225,3 +225,29 @@ ordinal: "1"
...
+    - type: void*
+      name: pUserData
+      desc: "[in][out][optional] pointer to data to be passed to callback"
```
pbalcer, line 253: The way this is implemented right now is not thread safe with respect to any other API call. This should be stated clearly in the docs.
(Ideally we'd figure out how to make it thread-safe though, but I don't see a simple way other than having a mutex around the logger, and that's just iffy).

include/ur_api.h
```diff
@@ -226,6 +226,8 @@ typedef enum ur_function_t {
     UR_FUNCTION_BINDLESS_IMAGES_IMPORT_EXTERNAL_MEMORY_EXP = 226,         ///< Enumerator for ::urBindlessImagesImportExternalMemoryExp
     UR_FUNCTION_BINDLESS_IMAGES_IMPORT_EXTERNAL_SEMAPHORE_EXP = 227,      ///< Enumerator for ::urBindlessImagesImportExternalSemaphoreExp
     UR_FUNCTION_ENQUEUE_NATIVE_COMMAND_EXP = 228,                         ///< Enumerator for ::urEnqueueNativeCommandExp
+    UR_FUNCTION_SET_LOGGER_CALLBACK = 230,                                ///< Enumerator for ::urSetLoggerCallback
```
frasercrmck, line 229: Is this an intentional jump from `228` to `230`?

martygrant: Nope! I must've made a mistake there when rebasing, I'll get that corrected. Thanks :)

scripts/core/loader.yml
martygrant: I've updated the description here to say the creator of the callback function should be responsible for making it thread-safe.

scripts/core/adapter.yml
```diff
@@ -200,3 +200,76 @@ etors:
...
+      Set a callback function for use by the logger to retrieve logging output.
+      It is a requirement that the callback function is thread safe and the
+      creator of the function will be responsible for this.
+class: $x
+name: SetLoggerCallback
+decl: static
+params:
+    - type: $x_adapter_handle_t
+      name: hAdapter
+      desc: "[in] handle of the adapter"
```
kbenzie, line 252: Taking an adapter here means the user needs to loop over all adapters to be able to set the logger globally. It also means we don't have the ability to emit logs for loader entry points which can means we can't report some useful errors, espeically around loading adapter libraries.

Perhaps this can be extended later, though.

scripts/core/adapter.yml
```diff
@@ -200,3 +200,76 @@ etors:
     - name: NATIVE_CPU
       value: "5"
       desc: "The backend is Native CPU"
+--- #--------------------------------------------------------------------------
+type: enum
+desc: "Minimum level of messages to be processed by the logger."
+class: $x
```
kbenzie, line 206: These should all be in the `$xAdapter` class.


```suggestion
class: $xAdapter
```

scripts/core/adapter.yml
```diff
@@ -200,3 +200,76 @@ etors:
...
+      desc: "Used to warn users about potential problems."
+    - name: ERR
+      value: "3"
+      desc: "Used when an error has occurred."
+    - name: QUIET
+      value: "4"
+      desc: "Restrict logger processing any messages."
+--- #--------------------------------------------------------------------------
+type: fptr_typedef
+desc: "Callback function to retrieve output from the logger."
```
kbenzie, line 226:
```suggestion
desc: "Callback function to retrieve output from the logger."
class: $xAdapter
```

scripts/core/adapter.yml
```diff
@@ -200,3 +200,76 @@ etors:
...
+    - type: void*
+      name: pUserData
+      desc: "[in][out] pointer to data to be passed to callback"
+--- #--------------------------------------------------------------------------
+type: function
+desc: |
+      Set a callback function for use by the logger to retrieve logging output.
+      It is a requirement that the callback function is thread safe and the
+      creator of the function will be responsible for this.
+class: $x
```
kbenzie, line 246: This will change the entry point name to `urAdapterSetLoggerCallback`:

```suggestion
class: $xAdapter
```

scripts/core/adapter.yml
```diff
@@ -200,3 +200,76 @@ etors:
...
+      name: pUserData
+      desc: "[in][out][optional] pointer to data to be passed to callback"
+    - type: $x_logger_level_t
+      name: level
+      desc: "[in] logging level"
+      init: 0
+--- #--------------------------------------------------------------------------
+type: function
+desc: "Set the minimum logging level for the logger Callback function."
+class: $x
```
kbenzie, line 266: This will change the entry point name to `urAdapterSetLoggerCallbackLevel`:

```suggestion
class: $xAdapter
```

scripts/core/adapter.yml
martygrant: Will revisit this later as discussed offline.

scripts/core/adapter.yml
kbenzie: This once got missed.

scripts/core/registry.yml
```diff
@@ -592,6 +592,12 @@ etors:
 - name: ENQUEUE_NATIVE_COMMAND_EXP
   desc: Enumerator for $xEnqueueNativeCommandExp
   value: '228'
+- name: ADAPTER_SET_LOGGER_CALLBACK
+  desc: Enumerator for $xAdapterSetLoggerCallback
+  value: '231'
+- name: ADAPTER_SET_LOGGER_CALLBACK_LEVEL
+  desc: Enumerator for $xAdapterSetLoggerCallbackLevel
+  value: '232'
```
kbenzie, line 600: This need reset, changing the entry point names will have done this.

scripts/core/adapter.yml
martygrant: Not sure about this one? Ends up making it being called `urAdapterur_logger_output_callback_t`. Looking at other function pointer examples none of them seem to start with a `$xClassName`

scripts/core/registry.yml
martygrant: Oh strange why would it increment the value rather than just updating the name?

kbenzie: The script doesn't know the difference between a new entry point having a name change and an existing entry point being removed. The ID's are intended to be globally unqiue for use in external tooling.

omarahmed1111 (approved):

omarahmed1111:
There is a workaround for warnings for now in this [PR](https://github.com/oneapi-src/unified-runtime/pull/1828). and it is probably going to be merged first so could you delete it in this PR once merged as this should fix it properly @martygrant? Thanks!

source/common/logger/ur_sinks.hpp
```diff
@@ -195,6 +197,35 @@ class FileSink : public Sink {
     std::ofstream ofstream;
 };
 
+class CallbackSink : public Sink {
+  public:
+    CallbackSink(std::string logger_name, bool skip_prefix = false,
```
pbalcer, line 202: I can't find where the `logger_name` is used for. I realize it's part of the `Sink` interface, but it seems to be unused?

pbalcer (approved):

include/ur_api.h
```diff
@@ -987,6 +989,70 @@ typedef enum ur_adapter_backend_t {
...
+    ur_logger_level_t level, ///< [out] Minimum level of messages to be processed by the logger.
+    const char *pLoggerMsg,  ///< [in][out] pointer to data to be passed to callback
+    void *pUserData          ///< [in][out] pointer to data to be passed to callback
+);
+
+///////////////////////////////////////////////////////////////////////////////
+/// @brief Set a callback function for use by the logger to retrieve logging output.
+///        It is a requirement that the callback function is thread safe and the
+///        creator of the function will be responsible for this.
+///
```
pbalcer, line 1018: One more thing to document is that calling this, `urAdapterSetLoggerCallback`, and `urAdapterSetLoggerCallbackLevel` functions is not thread-safe with respect to any other function in the API. The reason is that they are modifying global state without using atomics or grabbing a lock. In other words, when something is writing a log, you may invertedly change the logger callback pointer.

source/common/logger/ur_sinks.hpp
martygrant: it gets used here in the `log` method from the generic `Sink` https://github.com/oneapi-src/unified-runtime/blob/main/source/common/logger/ur_sinks.hpp#L26 but that doesn't get used in this new class

include/ur_api.h
martygrant: Okay no problem, I reworded the description for `urAdapterSetLoggerCallback` but will add it to `urAdapterSetLoggerCallbackLevel` as well. It should specifically say not "thread safe with respect to any other function in the API"? Should the explanation be added here or is that enough?

pbalcer: Most people understand thread-safe functions as safe to be invoked concurrently by multiple threads. And that is indeed how POSIX defines it. In this case, the situation is a bit different, since `urAdapterSetLoggerCallback` is not safe to call even with respect to other UR API functions (that are otherwise thread-safe). Small caveat is that implementations may choose to limit this unsafety to be per-adapter.

In my opinion this break [the principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), and should be thoroughly documented. Saying "not thread-safe" is not enough.

Ideally we'd fix this - one approach would be to only allow setting the callback as part of the loader config, and then change the logger level variable to be an atomic.

pbalcer (approved):

include/ur_api.h
martygrant: Thanks Piotr. @kbenzie do you have any thoughts on this? Would we want to modify the design now or just get this PR in as it is now and modify it later? If the latter then I'll move these comments to a new issue to track this.

martygrant:
## Information

## PR Migration

---

martygrant:

---

PR #1749, opened 2024-06-12
[L0] Maintain Lock of Queue while syncing the Last Command Event and update Last Command Event only if matching

nrspruit:
- To avoid usecases where the LastcommandEvent could be updated in another thread and a thread can be expecting the Queue Wait to include this event, then hold the queue lock such that the last command event remains unchanged during synchronization.

### DIFF
--- a/source/adapters/level_zero/queue.cpp
+++ b/source/adapters/level_zero/queue.cpp
@@ -1696,6 +1696,7 @@ ur_result_t ur_queue_handle_t_::synchronize() {
     // If event is discarded then it can be in reset state or underlying level
     // zero handle can have device scope, so we can't synchronize the last
     // event.
+    auto savedLastCommandEvent = LastCommandEvent;
     if (isInOrderQueue() && !LastCommandEvent->IsDiscarded) {
       ZE2UR_CALL(zeHostSynchronize, (LastCommandEvent->ZeEvent));
 
@@ -1737,7 +1738,12 @@ ur_result_t ur_queue_handle_t_::synchronize() {
         }
       }
     }
-    LastCommandEvent = nullptr;
+    // If the current version of the LastCommandEvent == savedLastCommandEvent,
+    // then LastCommandEvent = nullptr; Otherwise, if LastCommandEvent !=
+    // savedLastCommandEvent, then LastCommandEvent is unchanged.
+    if (LastCommandEvent == savedLastCommandEvent) {
+      LastCommandEvent = nullptr;
+    }
   }
 
   // Since all timestamp recordings should have finished with the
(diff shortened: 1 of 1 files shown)


source/adapters/level_zero/queue.cpp
```diff
@@ -1697,14 +1697,7 @@ ur_result_t ur_queue_handle_t_::synchronize() {
...
     if (isInOrderQueue() && !LastCommandEvent->IsDiscarded) {
-      if (UrL0QueueSyncNonBlocking) {
-        auto SyncZeEvent = LastCommandEvent->ZeEvent;
-        this->Mutex.unlock();
-        ZE2UR_CALL(zeHostSynchronize, (SyncZeEvent));
-        this->Mutex.lock();
-      } else {
-        ZE2UR_CALL(zeHostSynchronize, (LastCommandEvent->ZeEvent));
-      }
+      ZE2UR_CALL(zeHostSynchronize, (LastCommandEvent->ZeEvent));
```
pbalcer, line 1700: Below this, there's `LastCommandEvent = nullptr;` assignment, which, if it was changed from another thread, will be overwritten. It may need to changed to something like:
```
auto savedLastEvent = LastCommandEvent
...
compare_and_swap(LastCommandEvent, savedLastEvent, null);
```
This way, this thread will only update the LastCommandEvent if nothing else has changed it.

pbalcer (commented):
can we add a multi-threaded test that exercises this path? If merging this is urgent, we should at least create an issue to add one.

source/adapters/level_zero/queue.cpp
nrspruit: Should we be doing that in a different patch?  I can incorporate that into this patch, but that was not the original target. The purpose of this patch was to maintain the lock and call the host sync in this case.

pbalcer: Up to you, I think this patch will fix in order queues, may still have a race in the out-of-order scenario. And it's possible that using CAS addresses both at the same time.

nrspruit: Good idea, let me add the CAS and retest to ensure the original issue is still addressed.

pbalcer: I just realized that we might not need a CAS here. This variable is always modified and read under a lock. 
So a simple `if (LastCommandEvent == savedLastEvent) LastCommandEvent = nullptr;` (with `savedLastEvent` assigned [here](https://github.com/oneapi-src/unified-runtime/pull/1749/files#diff-2c1af7ac7b1b3f44932c1e685b141784c2d14a66126ea7a383f9f30d919678f5L1695)) should do the trick.

pbalcer (approved):

PR #1750, opened 2024-06-13
[CTS] Fix queue-releated nullptr issues in tests

frasercrmck:
These queue handles (i.e., pointers) weren't being initialized to nullptr in the constructor and were in an indeterminate state. If they weren't initialized as intended in the tests' SetUp routine (e.g., something else aborting the test beforehand), the queues were still in an indeterminate state and possibly non-null, in which case the test would try to free the garbage queues in TearDown.

This was responsible for one segfault in the Native CPU tests (which is why we need to add a lot of new {{OPT}} failures, now that it may run to completion) and for some HIP tests in an unrelated PR.


test/conformance/enqueue/enqueue_adapter_native_cpu.match
```diff
@@ -46,4 +46,328 @@
 {{OPT}}urEnqueueKernelLaunchTestWithParam.Success/SYCL_NATIVE_CPU___SYCL_Native_CPU___3D_256_79_8
 {{OPT}}urEnqueueKernelLaunchWithVirtualMemory.Success/SYCL_NATIVE_CPU___SYCL_Native_CPU_
 {{OPT}}urEnqueueKernelLaunchMultiDeviceTest.KernelLaunchReadDifferentQueues/SYCL_NATIVE_CPU___SYCL_Native_CPU_
-{{Segmentation fault|Aborted}}
+{{OPT}}urEnqueueKernelLaunchUSMLinkedList.Success/SYCL_NATIVE_CPU___SYCL_Native_CPU___UsePoolEnabled
```
pbalcer, line 49: if these tests are deterministically failing now, you shouldn't need `{{OPT}}`.

frasercrmck: True, but I can't guarantee they deterministically fail. While they don't segfault for me locally, I still see them segfault in [a job](https://github.com/oneapi-src/unified-runtime/actions/runs/9495222693/job/26167050849?pr=1689) which is HIP-only but contains the same changes to the test fixtures.

Unfortunately it's very difficult to see what's happening when the job passes, as in this PR, as it obscures the real results.

frasercrmck: Specifically, it seems like Native CPU, or the test, has non-deterministic problems:

```
/home/test-user/actions-runners/nativecpu-1/_work/unified-runtime/unified-runtime/test/conformance/enqueue/urEnqueueTimestampRecording.cpp:15: Skipped

Timestamp recording is not supported

corrupted double-linked list
```

pbalcer: Hm, ok, that looks like heap/memory corruption. Probably needs to be check with memsan or valgrind. Thanks.

pbalcer (approved):

test/conformance/enqueue/enqueue_adapter_native_cpu.match
frasercrmck: Yep. It's possible there's a fairly reliable point at which it hits this problem, and we could remove `{{OPT}}` from all tests before it, but unless we could reproduce it locally, I think that'd take a fair bit of CI time to ensure we're not introducing flakiness to the CI or other developers' local machines. As you say it would probably be better to actually just investigate and fix the problem with Native CPU, rather than playing around with match files.

PR #1753, opened 2024-06-13
[CUDA][HIP] Always wait on interop events

hdelan:
We can't make assumptions about interop events being on the same stream as other interop events, since in UR we don't track the streams that native events are recorded on.


hdelan:
Friendly ping @mmoadeli

mmoadeli (approved):
LGTM

PR #1755, opened 2024-06-14
[test] Move mem migrate to its own dir

hdelan:
Not having its own dir was breaking some builds.


test/conformance/CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ add_subdirectory(platform)
 add_subdirectory(device)
 add_subdirectory(context)
 add_subdirectory(memory)
+add_subdirectory(memory-migrate)
```
kbenzie, line 112: I think this should move down to line 140 along with all the other test suites that depend on kernel execution.

test/conformance/memory-migrate/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+if (UR_DPCXX)
```
kbenzie, line 1: With the move of `add_subdirectory(memory-migrate)` this check will no longer be necessary.

kbenzie, line 1: This is missing the license comment.

hdelan: Thanks have added.

hdelan: Thanks change made

test/conformance/CMakeLists.txt
hdelan: OK sure, have moved.

kbenzie (approved):

PR #1759, opened 2024-06-14
Implement layer to allow adapter mocking.

aarongreig:


source/loader/layers/mock/ur_mock_layer.hpp
```diff
@@ -0,0 +1,79 @@
...
+#pragma once
+#include "ur_ddi.h"
+#include "ur_proxy_layer.hpp"
+#include "ur_util.hpp"
+
+#include <atomic>
+
+namespace ur_mock_layer {
+
+struct dummy_handle_t_ {
```
aarongreig, line 21: I haven't hooked up the storage and data members of this yet, I'm kind of hoping that when it comes to porting the particular sycl unittests that were using them I can achieve the same effect with callbacks defined in the test rather than default behavior in our layer. I've left them in for now in case that isn't so.

aarongreig:
I'm going to leave this in draft until I've ported most/all of the PI unittests just to make sure our initial version satisfies all the requirements there, but this is a complete PR with documentation and testing for the new layer. It isn't urgent since it'll take me a bit to port those tests I mentioned but I'd appreciate some feedback on the interface, scripting and docs @pbalcer

pbalcer (commented):
A general comment: if we focus on the mocking use-case instead of creating a generic callback mechanism, it might be easier to instead create a mock adapter, with its own include header that the application would use to modify its behavior. We could even modify the existing null adapter for this purpose.

A side note - did you considering just using something existing like using `-Wl,--wrap` for the mocked tests?

aarongreig:
> A general comment: if we focus on the mocking use-case instead of creating a generic callback mechanism, it might be easier to instead create a mock adapter, with its own include header that the application would use to modify its behavior. We could even modify the existing null adapter for this purpose.
> 
> A side note - did you considering just using something existing like using `-Wl,--wrap` for the mocked tests?

I actually didn't know about `--wrap`, I think it could work with some clever cmake but at this point I don't really want to spend the time to try and get a whole new approach working.

I have hit some stuff while porting tests that looks like it's going to necessitate a mock header from the UR side being included there.. the only reason this is still done in a layer is that it started as the generic callback layer before I realised we'd need to focus in on the mock element so I might need to look at moving the mock stuff into the null adapter (I think the only thing we use that for in its current state is layer testing right?)

aarongreig:
closed in favour of https://github.com/oneapi-src/unified-runtime/pull/1780

PR #1765, opened 2024-06-17
[EXP][CUDA] Enable Large cluster sizes and fix cluster dimensions being set for dimensions less than 3

AD2605:
Fix the ordering of cluster dimension in accordance to the grid Dimensions.

Also adds a call to `CU_FUNC_ATTRIBUTE_NON_PORTABLE_CLUSTER_SIZE_ALLOWED` to allow 
cluster sizes greater than 8


source/adapters/cuda/enqueue.cpp
```diff
@@ -530,22 +530,41 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunchCustomExp(
...
       launch_attribute[i].id = CU_LAUNCH_ATTRIBUTE_CLUSTER_DIMENSION;
       // Note that cuda orders from right to left wrt SYCL dimensional order.
-      launch_attribute[i].value.clusterDim.x =
-          launchPropList[i].value.clusterDim[2];
-      launch_attribute[i].value.clusterDim.y =
-          launchPropList[i].value.clusterDim[1];
-      launch_attribute[i].value.clusterDim.z =
-          launchPropList[i].value.clusterDim[0];
+      if (workDim == 3) {
+        launch_attribute[i].value.clusterDim.x =
```
AD2605, line 547: Could use some help Here - 

I was not able to figure out where this flipping of order happens, 
I see it's being set in `setKernelParams` but how it flips it I was not able to understand it

source/adapters/cuda/enqueue.cpp
```diff
@@ -530,22 +530,41 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunchCustomExp(
   }
 
   std::vector<CUlaunchAttribute> launch_attribute(numPropsInLaunchPropList);
+  bool has_property_cluster_launch = false;
+
   for (uint32_t i = 0; i < numPropsInLaunchPropList; i++) {
```
JackAKirk, line 535: You may as well move this whole for loop down past line 594 where cuFunc is defined, then you can move
```
UR_CHECK_ERROR(cuFuncSetAttribute(
          CuFunc, CU_FUNC_ATTRIBUTE_NON_PORTABLE_CLUSTER_SIZE_ALLOWED, 1));
```
inside `UR_EXP_LAUNCH_PROPERTY_ID_CLUSTER_DIMENSION` case, then you don't need `has_property_cluster_launch` at all.

AD2605: Done

JackAKirk (approved):

AD2605:
@oneapi-src/unified-runtime-cuda-write  gentle ping r.e. this PR.
 E2E Cuda fails because It was unable to find a cuda GPU, seems unrelated to my changes - 

```cpp
lit.py: /home/test-user/actions-runners/01/_work/unified-runtime/unified-runtime/sycl-repo/sycl/test-e2e/lit.cfg.py:718: error: Cannot detect device aspect for cuda:gpu
stdout:

Platforms: 0
default_selector()      : No device of requested type available. -1 (PI_ERRO...
accelerator_selector()  : No device of requested type available. -1 (PI_ERRO...
cpu_selector()          : No device of requested type available. -1 (PI_ERRO...
gpu_selector()          : No device of requested type available. -1 (PI_ERRO...
custom_selector(gpu)    : No device of requested type available. -1 (PI_ERRO...
custom_selector(cpu)    : No device of requested type available. -1 (PI_ERRO...
custom_selector(acc)    : No device of requested type available. -1 (PI_ERRO...
```

Maybe re-triggering the job might help ?

JackAKirk:
yeh CI seems to be broken currently, same here: https://github.com/oneapi-src/unified-runtime/actions/runs/9584385509/job/26430113727?pr=1774

MartinWehking:
Hi, thanks for your patch.
Are there any test cases that fail without your fix and if not, could you provide one?

AD2605:
> Hi, thanks for your patch. Are there any test cases that fail without your fix and if not, could you provide one?

So the test cases that are being added, are in this PR here - https://github.com/intel/llvm/pull/14113, which would test this PR fully. I do not suppose I can add a test which will check the ordering of the cluster dimensions in this UR PR, however I can change increase the cluster size in the tests added in #1643 and increase the cluster size, such that it tests `CU_FUNC_ATTRIBUTE_NON_PORTABLE_CLUSTER_SIZE_ALLOWED` flag being added

callumfare:
I've removed the _ready-to-merge_ label since https://github.com/intel/llvm/pull/14113 isn't passing CI and also has the _abi-break_ label - it will need to wait for the ABI breaking window to open before it can be merged.

MartinWehking:
> > Hi, thanks for your patch. Are there any test cases that fail without your fix and if not, could you provide one?
> 
> So the test cases that are being added, are in this PR here - [intel/llvm#14113](https://github.com/intel/llvm/pull/14113), which would test this PR fully. I do not suppose I can add a test which will check the ordering of the cluster dimensions in this UR PR, however I can change increase the cluster size in the tests added in #1643 and increase the cluster size, such that it tests `CU_FUNC_ATTRIBUTE_NON_PORTABLE_CLUSTER_SIZE_ALLOWED` flag being added
> 
> However, note that this runs on SM_90 only, which I do not suppose is on the CI. If you prefer, I can add like a log of the test run on H100 ?

Thanks, that would be great!

MartinWehking (approved):

callumfare:

PR #1768, opened 2024-06-18
[L0] Enable default support for L0 in-order lists

raiyanla:
- Checks if compatible L0 driver for default enable
- Allow for force on/off toggle using env variable

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -28,31 +28,6 @@ template <typename T> T *getPointerFromVector(std::vector<T> &V) {
 }
 
 /**
- * Checks the version of the level-zero driver.
- * @param[in] Context Execution context
- * @param[in] VersionMajor Major version number to compare to.
- * @param[in] VersionMinor Minor version number to compare to.
- * @param[in] VersionBuild Build version number to compare to.
- * @return true if the version of the driver is higher than or equal to the
- * compared version.
- */
-bool isDriverVersionNewerOrSimilar(ur_context_handle_t Context,
-                                   uint32_t VersionMajor, uint32_t VersionMinor,
-                                   uint32_t VersionBuild) {
-  ZeStruct<ze_driver_properties_t> ZeDriverProperties;
-  ZE2UR_CALL(zeDriverGetProperties,
-             (Context->getPlatform()->ZeDriver, &ZeDriverProperties));
-  uint32_t DriverVersion = ZeDriverProperties.driverVersion;
-  auto DriverVersionMajor = (DriverVersion & 0xFF000000) >> 24;
-  auto DriverVersionMinor = (DriverVersion & 0x00FF0000) >> 16;
-  auto DriverVersionBuild = DriverVersion & 0x0000FFFF;
-
-  return ((DriverVersionMajor >= VersionMajor) &&
-          (DriverVersionMinor >= VersionMinor) &&
-          (DriverVersionBuild >= VersionBuild));
-}
-
-/**
  * Default to using compute engine for fill operation, but allow to override
  * this with an environment variable. Disable the copy engine if the pattern
  * size is larger than the maximum supported.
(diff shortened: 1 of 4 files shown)


nrspruit:
This looks good to me,  but you will need to fix the formatting, please address the code format issues and update the patch, thanks!

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -27,31 +27,6 @@ template <typename T> T *getPointerFromVector(std::vector<T> &V) {
...
-/**
- * Checks the version of the level-zero driver.
- * @param[in] Context Execution context
- * @param[in] VersionMajor Major version number to compare to.
- * @param[in] VersionMinor Minor version number to compare to.
- * @param[in] VersionBuild Build version number to compare to.
- * @return true if the version of the driver is higher than or equal to the
- * compared version.
- */
-bool isDriverVersionNewerOrSimilar(ur_context_handle_t Context,
```
EwanC, line 39: Think the callsite in command_buffer.cpp will need updated since we changed the first character to uppercase `I`

source/adapters/level_zero/device.cpp
```diff
@@ -1107,10 +1107,15 @@ bool ur_device_handle_t_::useRelaxedAllocationLimits() {
 bool ur_device_handle_t_::useDriverInOrderLists() {
   // Use in-order lists implementation from L0 driver instead
   // of adapter's implementation.
-  static const bool UseDriverInOrderLists = [] {
+
+  ze_driver_handle_t ZeDriver = this->Platform->ZeDriver;
+
+  static const bool UseDriverInOrderLists = [&] {
     const char *UrRet = std::getenv("UR_L0_USE_DRIVER_INORDER_LISTS");
+    bool compatibleDriver =
```
EwanC, line 1115: nitpick 
```suggestion
    bool CompatibleDriver =
```

EwanC (approved):
LGTM once the CI is happy

source/adapters/level_zero/command_buffer.cpp
raiyanla: Thanks, will update. Just noticed that failure now after rebasing around the conflicts.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -637,7 +612,7 @@ ur_result_t createMainCommandList(ur_context_handle_t Context,
 bool canBeInOrder(ur_context_handle_t Context,
                   const ur_exp_command_buffer_desc_t *CommandBufferDesc) {
   // In-order command-lists are not available in old driver version.
-  bool CompatibleDriver = isDriverVersionNewerOrSimilar(Context, 1, 3, 28454);
+  bool CompatibleDriver = IsDriverVersionNewerOrSimilar(Context, 1, 3, 28454);
```
nrspruit, line 615: The new api takes the driver handle ie ze_driver_handle_t not the context, you need to change the variable to be passed in.

source/adapters/level_zero/device.cpp
```diff
@@ -1107,10 +1107,15 @@ bool ur_device_handle_t_::useRelaxedAllocationLimits() {
...
   // Use in-order lists implementation from L0 driver instead
   // of adapter's implementation.
-  static const bool UseDriverInOrderLists = [] {
+
+  ze_driver_handle_t ZeDriver = this->Platform->ZeDriver;
+
+  static const bool UseDriverInOrderLists = [&] {
     const char *UrRet = std::getenv("UR_L0_USE_DRIVER_INORDER_LISTS");
+    bool CompatibleDriver =
+        isDriverVersionNewerOrSimilar(ZeDriver, 1, 3, 29534);
```
pbalcer, line 1116: The two driver versions you used in this patch are not identical. Why?
If they are supposed to be, please create a `#define` or some `static const` variable that defines the correct version to support this feature.

raiyanla: Yes these should be identical, since NEO had some bugs in the in-order implementation before this particular version. I'll update it to use a common #define then.

raiyanla:
I see one SYCL e2e test failing (minmax) in CI, haven't seen this one fail before for enabling in-order by default so need to look into this. Not sure if some new regression.

nrspruit (approved):
LGTM

pbalcer (approved):

pbalcer:
> I see one SYCL e2e test failing (minmax) in CI, haven't seen this one fail before for enabling in-order by default so need to look into this. Not sure if some new regression.
> 
> Edit: Irrelevant error, it is showing on every PR. Not related to these changes.

Unfortunately our HW platform is different than what intel/llvm CI uses, so we sometimes see failures that they don't.
https://github.com/oneapi-src/unified-runtime/pull/1870

PR #1770, opened 2024-06-18
[L0] implement command list cache for queue

igchor:


source/adapters/level_zero/v2/command_list_cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+  ~command_list_cache_t();
+
+  ze_command_list_handle_t
+  getCommandList(const command_list_descriptor_t &desc);
+  void addCommandList(const command_list_descriptor_t &desc,
+                      ze_command_list_handle_t cmdList);
+
+private:
+  std::unordered_map<command_list_descriptor_t,
+                     std::stack<ze_command_list_handle_t>,
```
pbalcer, line 57: maybe wrap `ze_command_list_handle_t` in a struct/class with a destructor? will eliminate the function pointer you are storing.

source/adapters/level_zero/v2/command_list_cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <ur/ur.hpp>
+#include <ur_api.h>
+#include <ze_api.h>
+
+#include "common.hpp"
+
+namespace v2 {
+struct immediate_command_list_descriptor_t {
+  ze_device_handle_t Device;
+  ZeStruct<ze_command_queue_desc_t> QueueDesc;
```
pbalcer, line 23: all of QueueDesc? Won't that be a little annoying to use?

source/adapters/level_zero/v2/command_list_cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+struct command_list_cache_t {
+  command_list_cache_t(
+      std::function<void(ze_command_list_handle_t)> Destructor =
+          [](ze_command_list_handle_t cmdList) {
+            ZE_CALL_NOCHECK(zeCommandListDestroy, (cmdList));
+          });
+  ~command_list_cache_t();
+
+  ze_command_list_handle_t
+  getCommandList(const command_list_descriptor_t &desc);
```
pbalcer, line 51: tbh, I think I'd simplify this and do `getImmCommandList` and `getRegularCommandList`, but this also works I guess :P

source/adapters/level_zero/common.cpp
```diff
@@ -66,9 +66,6 @@ ur_result_t ze2urResult(ze_result_t ZeResult) {
   }
 }
 
-usm::DisjointPoolAllConfigs DisjointPoolConfigInstance =
```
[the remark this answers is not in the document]
igchor: I included this because it's not possible to build unit tests without this

test/adapters/level_zero/v2/command_list_cache_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+        cache.addCommandList(immediateDesc1,
+                             reinterpret_cast<ze_command_list_handle_t>(i));
+        cache.addCommandList(
+            regularDesc1,
+            reinterpret_cast<ze_command_list_handle_t>(i + numListsPerType));
+    }
+
+    for (int i = 0; i < numListsPerType; ++i) {
+        auto immCmdList =
+            reinterpret_cast<int64_t>(cache.getCommandList(immediateDesc1));
```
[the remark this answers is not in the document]
igchor: Sure, we can do that but I was thinking that we might want to do some preallocation of command lists (e.g. during context creation) and the decision about what exact command list we want should be made outside of the cache (in context in this case)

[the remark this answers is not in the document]
pbalcer: I still think keeping related things in one place is better. If you really want to populate the cache in advance, maybe have a dedicated function to do that (or just do it in a constructor)?

source/adapters/level_zero/v2/command_list_cache.hpp
igchor: I think we need all the fields from ze_command_queue_desc_t anyway (expect for index), no?

test/adapters/level_zero/v2/command_list_cache_test.cpp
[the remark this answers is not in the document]
igchor: Ok, I don't have a strong opinion on that. Done. I will implement the preallocation function if needed once I start the integration with context.

source/adapters/level_zero/v2/command_list_cache.hpp
igchor: Done. The callback was only introduced so that I could easily test the cache without actually creating proper command lists but this is now outdated.

source/adapters/level_zero/common.cpp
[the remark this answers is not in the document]
igchor: In general, I think we might want to do some refactoring in CMakeFiles to add unittests more easily.

source/adapters/level_zero/v2/command_list_cache.hpp
pbalcer: This is false equivalence. This struct contains stuff we don't need (like pNext) that this will likely never use. I think we should specify each variable we need as part of the struct (or, which would be my preference, have a function like `getImmediateCommandList(params for immedaite stuff, all spelled out)`.

igchor: Ok, I've changed it as you requested but that argument can be made for most places where we use/store L0 desc structures. Do we want to only create desc structure as late as possible (just before calling the L0 API call) and just have custom structures before that?

pbalcer: > Do we want to only create desc structure as late as possible (just before calling the L0 API call) and just have custom structures before that?

If the intent is for the desc structure to be passed as-is to the L0 call, and if it more convenient, then I think its fine for it to be passed as an argument. However, if the structure is then somehow interpreted/manipulated by the adapter function, we should prefer to have our own structures/arguments, where we can define all the behavior for all the cases.
My concern here is that, when using L0 structs for non-L0 API calls, we potentially leave the API underdefined. For example, let's imagine a scenario where L0 has some sort of enum that lists supported usage models, but the L0 Adapter only supports a select number of those modes/behaviors. In this case, accepting the L0 type would make it seem as-if the API supports everything that the L0 specifies, which is not the case. In such scenarios, my preference would be for adapter to define its own types, that are aligned with that it supports.

pbalcer (approved):

nrspruit (approved):
LGTM

PR #1771, opened 2024-06-18
[L0] Initial version of queue dispatcher 

igchor:


source/adapters/level_zero/queue.cpp
```diff
@@ -593,11 +603,17 @@ UR_APIEXPORT ur_result_t UR_APICALL urQueueRetain(
   return UR_RESULT_SUCCESS;
 }
 
-UR_APIEXPORT ur_result_t UR_APICALL urQueueRelease(
-    ur_queue_handle_t UrQueue ///< [in] handle of the queue object to release
+UR_APIEXPORT ur_result_t UR_APICALL urQueueRetain(
```
pbalcer, line 606: Aliasing API function names like this is probably going to be confusing.

source/adapters/level_zero/queue.cpp
```diff
@@ -504,6 +504,16 @@ UR_APIEXPORT ur_result_t UR_APICALL urQueueCreate(
 
   UR_ASSERT(Context->isValidDevice(Device), UR_RESULT_ERROR_INVALID_DEVICE);
 
+  // optimized path for immediate, in-order command lists
+  if (v2::shouldUseQueueV2() && (Flags & UR_QUEUE_FLAG_SUBMISSION_IMMEDIATE) &&
+      Device->useImmediateCommandLists() &&
+      (Flags & UR_QUEUE_FLAG_OUT_OF_ORDER_EXEC_MODE_ENABLE) == 0) {
+    *Queue = new ur_queue_handle_t_(
+        std::in_place_type<v2::ur_queue_immediate_in_order_t>, Context, Device,
```
pbalcer, line 512: I have to say, this is an incredibly clunky way of defining a variant of an enum. Can we e.g., create a factory function that would be take a templated variant argument?

source/adapters/level_zero/queue.cpp
```diff
@@ -687,17 +703,31 @@ UR_APIEXPORT ur_result_t UR_APICALL urQueueRelease(
...
   return UR_RESULT_SUCCESS;
 }
 
-UR_APIEXPORT ur_result_t UR_APICALL urQueueGetNativeHandle(
-    ur_queue_handle_t UrQueue, ///< [in] handle of the queue.
+UR_APIEXPORT ur_result_t UR_APICALL urQueueRelease(
+    ur_queue_handle_t UrQueue ///< [in] handle of the queue object to release
+) {
+  return UrQueue->Dispatcher.dispatch([&](auto Queue) {
+    if constexpr (std::is_same_v<decltype(Queue), ur_queue_handle_legacy_t>) {
```
pbalcer, line 714: why just this function is implemented in the closure for the dispatcher?

source/adapters/level_zero/device.hpp
```diff
@@ -201,4 +201,7 @@ struct ur_device_handle_t_ : _ur_object {
   ZeCache<struct ze_global_memsize> ZeGlobalMemSize;
   ZeCache<ZeStruct<ze_mutable_command_list_exp_properties_t>>
       ZeDeviceMutableCmdListsProperties;
+
+  // Global (per-adapter) Id of the device
+  size_t Id;
```
pbalcer, line 206: this is unrelated.

source/adapters/level_zero/queue.cpp
```diff
@@ -593,11 +603,17 @@ UR_APIEXPORT ur_result_t UR_APICALL urQueueRetain(
   return UR_RESULT_SUCCESS;
 }
 
-UR_APIEXPORT ur_result_t UR_APICALL urQueueRelease(
-    ur_queue_handle_t UrQueue ///< [in] handle of the queue object to release
+UR_APIEXPORT ur_result_t UR_APICALL urQueueRetain(
+    ur_queue_handle_t UrQueue ///< [in] handle of the queue object to get access
 ) {
-  auto Queue = Legacy(UrQueue);
+  return UrQueue->Dispatcher.dispatch(
```
pbalcer, line 609: I know we are trying to minimize changes and use static dispatch where possible, but looking at this right now, I think simple inheritance from a base abstract class might be much simpler.

source/adapters/level_zero/queue.cpp
igchor: Alternatives are:
- add prefix to the function, e.g. this would become urQueueRetainImpl or something like that
- make those functions member functions of the proper handle type, i.e. we would call UrQueue->Retain()

source/adapters/level_zero/device.hpp
igchor: I've put this in a separate commit. Do you want me to move it to a separate PR?

source/adapters/level_zero/queue.cpp
igchor: It might be true, if we want to implement each function for each queue variant then there are not many benefits of using a variant anyway (like template functions), Approach with inheritance would look like this: https://github.com/igchor/unified-runtime/tree/queue_v2_inheritance

source/adapters/level_zero/device.hpp
pbalcer: I don't think we should have this at all yet, since it's dead code.

source/adapters/level_zero/queue.cpp
pbalcer: Hm, both approaches have merit, but I think my preference would be inheritance just because its simpler to grok.

igchor: Ok, I've just pushed a version with inheritance. The base class and API functions that operate on queue are now auto-generated. The script and templates for generation can be merged or not, but in either case, if we'd like to go back to the variant-approach for some reason, it should be pretty easy to do.

isaacault (approved):
`source/adapters/level_zero/image.cpp` LGTM.

pbalcer (approved):

nrspruit (approved):

PR #1773, opened 2024-06-19
[HIP] Improve urEnqueueUSMAdvise return code for non-shared mem

frasercrmck:
We can only call hipMemAdvise on managed pointers allocated via hipMallocManaged, so can only support the UR entry point on pointers allocated via urUSMSharedAlloc.

When faced with other pointers, we're supposed to ignore them and result "success", but instead this PR continues the tradition of returning UR_RESULT_ERROR_ADAPTER_SPECIFIC with a warning message providing the user more information. This should all be fixed up later when there's a better warning mechanism.


GeorgeWeb (approved):
Thank you! LGTM

aarongreig (approved):

callumfare:

PR #1778, opened 2024-06-20
[L0] Fix Handle used in calls to L0 Driver zex apis given multi drivers

nrspruit:


pbalcer (approved):
There are more uses of `zeDriverGetExtensionFunctionAddress` that are probably similarly broken right now, but are never tested. Its `zeMemGetPitchFor2dImage` and a few functions related to mutable command lists [here](https://github.com/oneapi-src/unified-runtime/blob/41c6290ed6802a2d49a12be0bddcb924ab0829bb/source/adapters/level_zero/platform.cpp#L222). 

Also, we should probably consider adding test passes that set `ZE_ENABLE_LOADER_INTERCEPT=1`. This would have found the problem.

nrspruit:
> There are more uses of `zeDriverGetExtensionFunctionAddress` that are probably similarly broken right now, but are never tested. Its `zeMemGetPitchFor2dImage` and a few functions related to mutable command lists [here](https://github.com/oneapi-src/unified-runtime/blob/41c6290ed6802a2d49a12be0bddcb924ab0829bb/source/adapters/level_zero/platform.cpp#L222).
> 
> Also, we should probably consider adding test runs that set `ZE_ENABLE_LOADER_INTERCEPT=1`. This would have found the problem.

I will see if they apply. If so, I will update this patch to cover all those cases.

nrspruit:
> > There are more uses of `zeDriverGetExtensionFunctionAddress` that are probably similarly broken right now, but are never tested. Its `zeMemGetPitchFor2dImage` and a few functions related to mutable command lists [here](https://github.com/oneapi-src/unified-runtime/blob/41c6290ed6802a2d49a12be0bddcb924ab0829bb/source/adapters/level_zero/platform.cpp#L222).
> > Also, we should probably consider adding test runs that set `ZE_ENABLE_LOADER_INTERCEPT=1`. This would have found the problem.
> 
> I will see if they apply. If so, I will update this patch to cover all those cases.

Fixed in the latest version. I went thru and addressed all the experimental usecases. If there is a single driver, this call is a noop so there is no issue.

pbalcer (approved):

raiyanla (approved):

nrspruit:
I have rebased this change on the latest in an effort to address the strange failures in https://github.com/intel/llvm/pull/14250 .

Those errors do not reproduce locally thus far.

EwanC (commented):
> Also, we should probably consider adding test runs that set ZE_ENABLE_LOADER_INTERCEPT=1. This would have found the problem.

The change looks fine to me, but I'm curious without this how this functionality is expected to continue to work going forward with testing. Could you point me to where the testing is for this?

pbalcer:
> The best way to catch these problems is to actually run all the tests with ZE_ENABLE_LOADER_INTERCEPT=1 enabled.

That'd be my preference. The way we can do that is either by modifying the tests themselves (adding something like `if level_zero { env +=  "ZE_ENABLE_LOADER_INTERCEPT=1" }` [here](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/CMakeLists.txt#L40) or by adding a new CI job that simply sets the variable externally (to do that you'd need to modify the [hw reusable workflow](https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/build-hw-reusable.yml#L74) to accept an optional env variable, and then use that [here](https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/cmake.yml#L164)).

For now, I think we should just create an issue to do so, and go ahead with merging this PR as-is.

callumfare:
@oneapi-src/unified-runtime-bindless-images-write This PR is high priority so could you review this ASAP? Thanks

isaacault (approved):
`source/adapters/level_zero/image.cpp` LGTM.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -56,12 +56,14 @@ bool PreferCopyEngineForFill = [] {
...
     ur_context_handle_t Context, ur_device_handle_t Device,
     ze_command_list_handle_t CommandList,
+    ze_command_list_handle_t CommandListTranslated,
     ze_command_list_handle_t CommandListResetEvents,
     ze_command_list_handle_t CopyCommandList,
     ZeStruct<ze_command_list_desc_t> ZeDesc,
     ZeStruct<ze_command_list_desc_t> ZeCopyDesc,
     const ur_exp_command_buffer_desc_t *Desc, const bool IsInOrderCmdList)
     : Context(Context), Device(Device), ZeComputeCommandList(CommandList),
+      ZeComputeCommandListTranslated(CommandListTranslated),
```
zzdanowicz, line 66: for the mutation of wait and signal events, I think you would need to add translation of their respective handles too

PR #1780, opened 2024-06-24
Rework null adapter into mock adapter

aarongreig:
This also comes with a new accompanying helper header which allows tests to override entry points and create their own reference counted dummy handles.


aarongreig:
@pbalcer This is a working (insofar as I have one of the sycl unittest suites passing with it) prototype of the mock adapter + header approach. The interesting stuff is in `scripts/core/loader.yml` where there's a new loader entry point to force load the mock adapter (and the implementation for that in `source/loader/ur_lib.cpp`) and in `source/mock` where the mock interface header lives. If the overall approach looks ok to you I'll come back to polish and add docs/testing

source/loader/ur_callbacks.hpp
```diff
@@ -0,0 +1,91 @@
...
+                               callback_properties->pCallback);
+            break;
+        default:
+            return UR_RESULT_ERROR_INVALID_ENUMERATION;
+        }
+        return UR_RESULT_SUCCESS;
+    }
+
+  private:
+    std::unordered_map<std::string, ur_mock_callback_t> beforeCallbacks;
```
pbalcer, line 86: is this the same thing as in ur_mock_helpers?

source/mock/ur_mock_helpers.hpp
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+// Increment reference counter for the handle
+template <class T> inline static void retainDummyHandle(T Handle) {
+    auto DummyHandlePtr = reinterpret_cast<dummy_handle_t>(Handle);
+    ++DummyHandlePtr->refCounter;
+}
+
+struct callbacks {
```
pbalcer, line 53: I think I'd prefer for this to be a singleton instead of every individual variable being static.

scripts/core/loader.yml
```diff
@@ -187,6 +186,31 @@ params:
       name: pUserData
       desc: "[in][out][optional] pointer to data to be passed to callback."
 --- #--------------------------------------------------------------------------
+type: fptr_typedef
```
pbalcer, line 189: This is unused?

pbalcer (commented):
Looks good. Much cleaner than trying to shoehorn this functionality into a layer :)

source/loader/ur_callbacks.hpp
aarongreig: yeah this was left over from the layer implementation, have removed

scripts/core/loader.yml
aarongreig: I'm still using the callback function pointer definition in the [callback helpers](https://github.com/oneapi-src/unified-runtime/pull/1780/files#diff-9355e740a21703116ed14e0e168aabc2c999b47a27f6dd934a65b658a95a4981R67)

pbalcer: In which case you should be able to simply define that type there. Or do you want that type to be available without the callback helper header?

aarongreig:
this is now complete with docs and testing, it depends on https://github.com/oneapi-src/unified-runtime/pull/1720 but since that's a NFC I'll be able to merge it pretty soon

source/mock/ur_mock_helpers.hpp
```diff
@@ -0,0 +1,129 @@
...
+        afterCallbacks.clear();
+    }
+
+  private:
+    std::unordered_map<std::string, ur_mock_callback_t> beforeCallbacks;
+    std::unordered_map<std::string, ur_mock_callback_t> replaceCallbacks;
+    std::unordered_map<std::string, ur_mock_callback_t> afterCallbacks;
+};
+
+extern callbacks_t callbacks;
```
pbalcer, line 127: nit: I think I'd prefer an exported function with implementation like:
```
getCallbacks() {
  static callbacks_t callbacks;
  return callbacks;
}
```

pbalcer (approved):
lgtm

source/mock/ur_mock_helpers.hpp
aarongreig: I don't mind making the change but what's the advantage? Something to do with how it's linked?

scripts/core/exp-command-buffer.yml
```diff
@@ -893,7 +893,7 @@ name: ReleaseCommandExp
 params:
     - type: $x_exp_command_buffer_command_handle_t
       name: hCommand
-      desc: "[in] Handle of the command-buffer command."
+      desc: "[in][release] Handle of the command-buffer command."
```
EwanC, line 896: Need a matching  `[retain]` for this on line 883?

source/mock/ur_mock_helpers.hpp
pbalcer: Well, personally I don't like variables as part of public interfaces, a function seems better. Using a static inside of a function also abstracts the global nature of this type little better. AFAIR, on Windows there are different rules on how public variables and public functions are exported (see [docs](https://learn.microsoft.com/en-us/cpp/build/reference/exports?view=msvc-170)).

And finally, what I suggested is inline with [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i3-avoid-singletons).

aarongreig: made the change

EwanC (approved):

PR #1781, opened 2024-06-24
Make device param in urMemGetNativeHandle optional

callumfare:
#1226 added a device handle parameter to `urMemGetNativeHandle`, but some uses in SYCL pass a null device handle, which causes unexpected failures when the validation layer is enabled, particularly for adapters where the device isn't required (OpenCL and L0).

This makes the device optional, with adapters allowed to fail when given a null device if they require it to get the native handle. This should preserve the intended implementation of all the adapters, and just moves the null handle check out of the validation layer.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -2820,6 +2820,8 @@ urMemBufferPartition(
 ///     - The application may call this function from simultaneous threads for
 ///       the same context.
 ///     - The implementation of this function should be thread-safe.
+///     - The implementation may require a valid device handle to return the
+///       native mem handle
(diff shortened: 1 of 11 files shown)


hdelan (approved):
Sorry I should have done this in the original patch. LGTM

konradkusiak97 (approved):

aarongreig (approved):

callumfare:
intel-llvm PR: https://github.com/intel/llvm/pull/14299

callumfare:

PR #1782, opened 2024-06-25
[CUDA] Get scopedcontext device from UR context

hdelan:
An event created with interop doesn't have an associated queue. So get an active device through the context.


JackAKirk (approved):

callumfare:

PR #1791, opened 2024-06-26
Make urKernelCreateWithNativeHandle's program param optional.

aarongreig:
Not all backends requires this parameter so sycl doesn't always provide it. This prevents those cases from getting caught in the validation layer.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -5276,6 +5276,8 @@ typedef struct ur_kernel_native_properties_t {
 ///     - The application may call this function from simultaneous threads for
 ///       the same context.
 ///     - The implementation of this function should be thread-safe.
+///     - The implementation may require a valid program handle to return the
+///       native kernel handle
 ///
 /// @returns
 ///     - ::UR_RESULT_SUCCESS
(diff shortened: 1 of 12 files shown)


martygrant (approved):

include/ur_api.h
```diff
@@ -5234,7 +5236,7 @@ typedef struct ur_kernel_native_properties_t {
 ///     - ::UR_RESULT_ERROR_ADAPTER_SPECIFIC
 ///     - ::UR_RESULT_ERROR_INVALID_NULL_HANDLE
 ///         + `NULL == hContext`
-///         + `NULL == hProgram`
+///         + If `hProgram == NULL` and the implementation requires a valid program.
```
RossBrunton, line 5239: Do we have a test for this with a backend that does return `INVALID_NULL_HANDLE`?

omarahmed1111: I think the current backends could query the program associated with the kernel via getInfo. So, I don't think the current backends have a hard requirement for the program being passed. However, I am not sure of `native_cpu` position on this.

aarongreig: it is a hard requirement for level zero as far as I can tell, I've added a check and an adapter specific test

aarongreig:
LLVM PR since this touches adapter code now https://github.com/intel/llvm/pull/14710

omarahmed1111 (approved):

RossBrunton:
This seems to fix https://github.com/intel/llvm/issues/14714 as well, could the `UNSUPPORTED` in that test be removed here if that turns out to be the case?

aarongreig:
> This seems to fix [intel/llvm#14714](https://github.com/intel/llvm/issues/14714) as well, could the `UNSUPPORTED` in that test be removed here if that turns out to be the case?

we'll need to remove it on PI2UR-main once this merges and gets pulled in via a `sycl` merge

RossBrunton (approved):

PR #1794, opened 2024-06-27
Add separate CODEOWNERS entry for L0 v2 dispatcher

pbalcer:


pbalcer:
The @oneapi-src/unified-runtime-level-zero-v2-write team is a superset of @oneapi-src/unified-runtime-level-zero-write that includes @igchor and me.

aarongreig:
is there a problem with modifying the original team?

pbalcer:
> is there a problem with modifying the original team?

I didn't want to presume :-) That could also work, assuming @nrspruit is ok with that solution.

PR #1800, opened 2024-06-27
[L0] skip checks in test if L0 function is not supported

igchor:
older driver versions might not implement zeCommandListGet* functions.


pbalcer (commented):
LGTM, but I think there's an existing assert macro that does this all in one.

igchor:
> LGTM, but I think there's an existing assert macro that does this all in one.

There is, but only for checking against UR results (not ZE results). Also, that macro skips the entire test if the feature is not supported - here, we might have only some of the checks unavailable.

pbalcer (approved):

PR #1801, opened 2024-06-27
[L0 v2] Implement command list initialization

igchor:


source/adapters/level_zero/v2/queue_immediate_in_order.cpp
```diff
@@ -11,8 +11,51 @@
 #include "queue_immediate_in_order.hpp"
 
 namespace v2 {
+
+static int32_t getZeOrdinal(ur_device_handle_t Device, queue_group_type type) {
+  if (type == queue_group_type::MainCopy && Device->hasMainCopyEngine()) {
+    return Device->QueueGroup[queue_group_type::MainCopy].ZeOrdinal;
+  }
+  return Device->QueueGroup[queue_group_type::Compute].ZeOrdinal;
```
igchor, line 19: If there's no copy engine available, we could have a special path (just a single CommandList and a single LastEvent in the queue) but I don't know if it;s worth the extra complexity - I assume that would be a rare case.

pbalcer: nah, not worth the complication.

source/adapters/level_zero/v2/command_list_cache.cpp
```diff
@@ -10,8 +10,8 @@
 
 #include "command_list_cache.hpp"
 
-#include "context.hpp"
-#include "device.hpp"
+#include "../context.hpp"
```
pbalcer, line 13: hm, I'd prefer if you added add_include_directories to point to the adapters' dir.

source/adapters/level_zero/v2/command_list_cache.cpp
```diff
@@ -87,41 +88,29 @@ raii::ze_command_list_t command_list_cache_t::getImmediateCommandList(
   Desc.Mode = Mode;
   Desc.Priority = Priority;
   Desc.Index = Index;
-  return getCommandList(Desc);
+
+  auto CommandList = getCommandList(Desc).release();
+  return raii::cache_borrowed_command_list_t(
+      CommandList, [Cache = this, Desc](ze_command_list_handle_t CmdList) {
+        Cache->addCommandList(
```
pbalcer, line 95: is this thread-safe?

pbalcer (commented):
lgtm, but I think this is not mt-safe. If you agree, please fix and add mt tests if possible.

source/adapters/level_zero/v2/command_list_cache.cpp
igchor: I thought about this, but then we have a name collision (level_zero/context.hpp and level_zero/v2/context.hpp), and we would need to rename this header to v2_context.hpp or something like this.

igchor:
> lgtm, ~but I think this is not mt-safe. If you agree, please fix and~ add mt tests if possible.

Done.

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
```diff
@@ -11,8 +11,51 @@
...
+}
+
+static std::optional<int32_t> getZeIndex(const ur_queue_properties_t *Props) {
+  if (Props) {
+    if (Props->pNext) {
```
EwanC, line 24: This would be more readable as `if (Props && Props->pNext)`

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
```diff
@@ -11,8 +11,51 @@
...
+      const ur_base_properties_t *extendedDesc =
```
EwanC, line 25: I noticed that the v2 code doesn't follow a consistent variable naming convention, e.g. we start with a lowercase letter here but a few lines on there's a variable defined starting with an upper case letter.

There's quite a few cases of this so maybe not something for this PR, but could be good to decide on a consistent format while v2 is still relatively new.

test/adapters/level_zero/v2/command_list_cache_test.cpp
```diff
@@ -167,3 +162,97 @@ TEST_P(CommandListCacheTest, ImmediateCommandListsHaveProperAttributes) {
         }
     }
 }
+
+TEST_P(CommandListCacheTest, CommandListsAreReusedByQueues) {
+    static constexpr int NumQueuesPerType = 5;
```
EwanC, line 167: What's the benefit of using `static` with `constexpr` ?

EwanC (approved):
Graphs change LGTM, minor questions about the v2 adapter code that I don't own

test/conformance/kernel/kernel_adapter_level_zero-v2.match
```diff
@@ -1,173 +1,17 @@
...
 urKernelGetSuggestedLocalWorkSizeTest.Success2D/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}_
-urKernelGetSuggestedLocalWorkSizeTest.Success3D/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}_
-urKernelGetSuggestedLocalWorkSizeTest.InvalidNullHandleKernel/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}_
-urKernelGetSuggestedLocalWorkSizeTest.InvalidNullHandleQueue/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}_
-urKernelGetSuggestedLocalWorkSizeTest.InvalidWorkDimension/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}_
-urKernelGetSuggestedLocalWorkSizeTest.InvalidGlobalOffset/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}_
-urKernelGetSuggestedLocalWorkSizeTest.InvalidGlobalSize/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}_
-urKernelGetSuggestedLocalWorkSizeTest.InvalidSuggestedLocalWorkSize/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}_
\ No newline at end of file
+urKernelGetSuggestedLocalWorkSizeTest.Success3D/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}_
```
kbenzie, line 17: missing new line at the end of the file

source/adapters/level_zero/context.cpp
```diff
@@ -36,7 +38,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urContextCreate(
   ZE2UR_CALL(zeContextCreate, (Platform->ZeDriver, &ContextDesc, &ZeContext));
   try {
     ur_context_handle_t_ *Context =
-        new ur_context_handle_t_(ZeContext, DeviceCount, Devices, true);
+        new v2::ur_context_handle_t_(ZeContext, DeviceCount, Devices, true);
```
kbenzie, line 41: This still defaults to non v2 somehow?

kbenzie (approved):
Testing LGTM

source/adapters/level_zero/v2/command_list_cache.cpp
pbalcer: ok, makes sense. let's leave it as is then.

source/adapters/level_zero/context.cpp
pbalcer: See the relevant [commit message](https://github.com/oneapi-src/unified-runtime/pull/1801/commits/58ddeb2aa8098adb06dcf6c653476f3cd38afb59):
```
that inherits from the old context and extends it with v2 functionality.
urContextCreate will now always create v2::context. All legacy
functions will continue to operate as normal and the optimized
code path can levarage additional features implemented in v2.
```

kbenzie: Cool, thanks.

pbalcer:
E2E is failing, but I think its unrelated:
```

********************
Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90.. 
********************
Failed Tests (1):
  SYCL :: ESIMD/regression/minmax.cpp
```

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
igchor: Sure, done.

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
igchor: Yeah, good point. I copied part of this from different places from L0 adapter but I agree it would be better to have a consistent style. I updated all variable names (that I introduce in this PR) to start with lowercase letter. I will also do a follow-up PR to make naming in command_list_cache.*pp consistent.

test/adapters/level_zero/v2/command_list_cache_test.cpp
igchor: I just kind of use `static constexpr` automatically when defining a const. Here, it doesn't matter much but in general, `static constexpr` allows you to use it in compile-time expressions (as template parameter, etc.)

test/conformance/kernel/kernel_adapter_level_zero-v2.match
igchor: Done.

pbalcer (approved):

PR #1802, opened 2024-06-28
[L0] Fix immediate command list use in Command Queues

nrspruit:
- Keep track of all immediate command lists ensuring that they are not used in regular queues after they are placed back into the command list map.


pbalcer (approved):

kbenzie:
There's a [failure](https://github.com/intel/llvm/actions/runs/9705904572/job/26794155114?pr=14341) in the `Regression/commandlist/gpu.cpp` test. I'll update the branches to see if it goes away.

PR #1805, opened 2024-06-28
[Testing] extend test names to include device id

igchor:
(position of the device handle inside a vector of handles). This avoids duplicate test names (disallowed by gtests) and makes it possible to run the tests in multi-device environment.


pbalcer (approved):

igchor:
Turns out NATIVE_CPU tests had to be updated to include possible device id

PR #1806, opened 2024-06-28
[UR][L0] Fix undefined behavior caused by shifting more than bits count

againull:
Shifting more than number of bits in the type is undefined behavior. Current code results in incorrect TimestampMaxValue calculation when valid bits >= 64 for the device. This PR is fixing this.

intel/llvm PR: https://github.com/intel/llvm/pull/14360

### DIFF
--- a/source/adapters/level_zero/device.cpp
+++ b/source/adapters/level_zero/device.cpp
@@ -1545,8 +1545,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetGlobalTimestamps(
 ) {
   const uint64_t &ZeTimerResolution =
       Device->ZeDeviceProperties->timerResolution;
-  const uint64_t TimestampMaxCount =
-      ((1ULL << Device->ZeDeviceProperties->kernelTimestampValidBits) - 1ULL);
+  const uint64_t TimestampMaxCount = Device->getTimestampMask();
   uint64_t DeviceClockCount, Dummy;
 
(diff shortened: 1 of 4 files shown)


nrspruit (approved):
LGTM

source/adapters/level_zero/device.cpp
```diff
@@ -1545,8 +1545,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetGlobalTimestamps(
 ) {
   const uint64_t &ZeTimerResolution =
       Device->ZeDeviceProperties->timerResolution;
+  auto ValidBits = Device->ZeDeviceProperties->kernelTimestampValidBits;
```
pbalcer, line 1548: when is it possible for this value to be > 64? Isn't that a bug in L0?
Shouldn't this be value capped at 64 when it exceeds it?

againull: Main intention of the PR was to fix == 64 case, I didn't get the case >64.  This implementation still caps it at 64 bit, right?  even though I believe you are right that L0 runtime caps it itself as well. Do you want me to change this it to "assert" for >64 bit case?

pbalcer: in that case, I think this should become `max(ValidBits, 63)`. Maybe create a function in device that returns `TimestampMask` instead of repeating this 3x ?

againull: Thanks for suggestions, fixed.

pbalcer (approved):

PR #1807, opened 2024-07-02
[L0] Add missing check for WorkDim to KernelLaunch funcs

igchor:
This fixes urEnqueueKernelLaunchTest.InvalidWorkDimension.

There might be more failing tests in enqueue. Right now this test seems to crash (in the driver) on Arcs so I'm not sure if we even run all the tests from enqueue.cpp I found this on PVC where the error was reported differently and didn't match with the matchfile.

I'm not sure what the best approach is here, but maybe we should just filter out the tests we know fail in the driver so that we can run all of them?

### DIFF
--- a/source/adapters/level_zero/kernel.cpp
+++ b/source/adapters/level_zero/kernel.cpp
@@ -130,8 +130,10 @@ ur_result_t ur_queue_handle_legacy_t_::enqueueKernelLaunch(
         *OutEvent ///< [in,out][optional] return an event object that identifies
                   ///< this particular kernel execution instance.
 ) {
-  auto Queue = this;
+  UR_ASSERT(WorkDim > 0, UR_RESULT_ERROR_INVALID_WORK_DIMENSION);
+  UR_ASSERT(WorkDim < 4, UR_RESULT_ERROR_INVALID_WORK_DIMENSION);
 
+  auto Queue = this;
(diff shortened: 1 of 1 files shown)


pbalcer (approved):
> I'm not sure what the best approach is here, but maybe we should just filter out the tests we know fail in the driver so that we can run all of them?

Match files in CTS were always supposed to be a temporary measure until all the problems are addressed. I wouldn't want to do a list of crashing tests, since you'd have to maintain that (and it would be easy to forget to update). We contemplated simply running each test as a standalone process.
My opinion? We should just fix the crashes. It probably wouldn't take that much more time than modifying the testing framework to account for them.

nrspruit (approved):

igchor:
> > I'm not sure what the best approach is here, but maybe we should just filter out the tests we know fail in the driver so that we can run all of them?
> 
> Match files in CTS were always supposed to be a temporary measure until all the problems are addressed. I wouldn't want to do a list of crashing tests, since you'd have to maintain that (and it would be easy to forget to update). We contemplated simply running each test as a standalone process. My opinion? We should just fix the crashes. It probably wouldn't take that much more time than modifying the testing framework to account for them.

Sure, but here the crash comes from the driver so I'm not sure there is much we can do (other than waiting for a fixed version).

PR #1816, opened 2024-07-03
[L0] Use Intel Level Zero Driver String extension

nrspruit:
- Given zeIntelGetDriverVersionString is available, then use the new driver extension to read the driver runtime version string.

### DIFF
--- a/source/adapters/level_zero/platform.cpp
+++ b/source/adapters/level_zero/platform.cpp
@@ -159,18 +159,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urPlatformCreateWithNativeHandle(
 }
 
 ur_result_t ur_platform_handle_t_::initialize() {
-  // Cache driver properties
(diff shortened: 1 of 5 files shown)


source/adapters/level_zero/platform.cpp
```diff
@@ -211,6 +199,33 @@ ur_result_t ur_platform_handle_t_::initialize() {
     zeDriverExtensionMap[extension.name] = extension.version;
   }
 
+  ZE2UR_CALL(zelLoaderTranslateHandle, (ZEL_HANDLE_DRIVER, ZeDriver,
+                                        (void **)&ZeDriverHandleExpTranslated));
+
+  // Check if intel Driver Version Extension is supported.
+  ZeDriverVersionString.setZeDriverVersionString(this);
```
pbalcer, line 206: writing to the global makes this non thread-safe. I think you should move this object into the platform class.

pbalcer (approved):

PR #1823, opened 2024-07-04
Add event update to command-buffers

EwanC:
Expand the command-buffer experimental feature API so that it can be used to implement [SYCL-Graph dynamic
events](https://github.com/reble/llvm/pull/372).

This involves extending each command append entry-point to include the following extra parameters:
* An output `ur_exp_command_buffer_command_handle_t`.
* An Input `ur_event_handle_t` event wait-list of dependent events.
* An output `ur_event_handle_t` event that is signaled when the command completes its next execution.

TODO - Update APIs

For this commit the implementation of these new APIs is stubbed out. Implementing the APIs for at least one adapter and adding new UR CTS tests will be follow-on work.


scripts/core/EXP-COMMAND-BUFFER.rst
EwanC, line 1: I should put a blank line in before line 81 so that the bullet list renders correctly

test/conformance/exp_command_buffer/event_sync_kernel_command.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+        if (second_cmd_buf_handle) {
+            EXPECT_SUCCESS(urCommandBufferReleaseExp(second_cmd_buf_handle));
+        }
+
+        UUR_RETURN_ON_FATAL_FAILURE(
+            urCommandBufferExpExecutionTest::TearDown());
+    }
+
+    // First two shared points are inputs to be tested, last is an output
```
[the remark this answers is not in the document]
EwanC: "points" is a typo on "pointers". Will clean up these test comments.

aarongreig (approved):

source/adapters/cuda/command_buffer.cpp
```diff
@@ -96,6 +122,28 @@ ur_exp_command_buffer_command_handle_t_::
...
+
+ur_exp_command_buffer_command_handle_t_::
+    ~ur_exp_command_buffer_command_handle_t_() {
+  // We create the ur_event_t returned to the user for a signal node using
+  // `makeWithNative` which sets `HasOwnership` to false. Therefore destruction
+  // of the `ur_event_t` object doesn't free the underlying CuEvent_t object and
+  // we need to do it manually ourselves.
+  if (SignalNode) {
+    CUevent SignalEvent;
+    cuGraphEventRecordNodeGetEvent(SignalNode, &SignalEvent);
```
hdelan, line 144: Should we be checking for an error here?

source/adapters/cuda/command_buffer.cpp
```diff
@@ -261,11 +319,28 @@ static ur_result_t enqueueCommandBufferFillHelper(
...
-    auto SyncPoint = CommandBuffer->addSyncPoint(GraphNode);
+    CUgraphNode SyncPointNode = SignalNode ? SignalNode : GraphNode;
+    auto SyncPoint = CommandBuffer->addSyncPoint(SyncPointNode);
     if (RetSyncPoint) {
       *RetSyncPoint = SyncPoint;
     }
+
+    std::vector<CUgraphNode> WaitNodes =
+        NumEventsInWaitList ? DepsList : std::vector<CUgraphNode>();
+    auto NewCommand = new T(CommandBuffer, GraphNode, SignalNode, WaitNodes);
```
hdelan, line 337: Would it be possible to use a `unique_ptr` instead of calling `new`?

source/adapters/cuda/command_buffer.cpp
```diff
@@ -71,16 +71,42 @@ ur_exp_command_buffer_handle_t_::~ur_exp_command_buffer_handle_t_() {
...
-        const size_t *GlobalWorkOffsetPtr, const size_t *GlobalWorkSizePtr,
-        const size_t *LocalWorkSizePtr)
-    : CommandBuffer(CommandBuffer), Kernel(Kernel), Node(Node), Params(Params),
-      WorkDim(WorkDim), RefCountInternal(1), RefCountExternal(1) {
-  CommandBuffer->incrementInternalReferenceCount();
+std::unique_ptr<ur_event_handle_t_>
+ur_exp_command_buffer_handle_t_::addSignalNode(CUgraphNode DepNode,
+                                               CUgraphNode &SignalNode) {
+  CUevent Event;
+  UR_CHECK_ERROR(cuEventCreate(&Event, CU_EVENT_DEFAULT));
```
hdelan, line 78: In the normal non graph API, we only record events at the beginning of an event if profiling is enabled. Events recorded afterwards for dependency analysis are always recorded if profiling is not being used. Should we have some sort of check here that checks if profiling is enabled? We can skip recording an extra event before each node this way

source/adapters/cuda/command_buffer.cpp
```diff
@@ -495,11 +639,29 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferAppendMemBufferCopyExp(
...
-    auto SyncPoint = hCommandBuffer->addSyncPoint(GraphNode);
+    CUgraphNode SyncPointNode = SignalNode ? SignalNode : GraphNode;
+    auto SyncPoint = hCommandBuffer->addSyncPoint(SyncPointNode);
     if (pSyncPoint) {
       *pSyncPoint = SyncPoint;
     }
+
+    std::vector<CUgraphNode> WaitNodes =
+        numEventsInWaitList ? DepsList : std::vector<CUgraphNode>();
+    auto NewCommand = new buffer_copy_command_handle(hCommandBuffer, GraphNode,
```
hdelan, line 658: More use of `new`. `unique_ptr`?

source/adapters/cuda/command_buffer.cpp
```diff
@@ -603,11 +824,30 @@ ur_result_t UR_APICALL urCommandBufferAppendMemBufferReadExp(
...
-    auto SyncPoint = hCommandBuffer->addSyncPoint(GraphNode);
+    CUgraphNode SyncPointNode = SignalNode ? SignalNode : GraphNode;
+    auto SyncPoint = hCommandBuffer->addSyncPoint(SyncPointNode);
     if (pSyncPoint) {
       *pSyncPoint = SyncPoint;
     }
+
+    std::vector<CUgraphNode> WaitNodes =
+        numEventsInWaitList ? DepsList : std::vector<CUgraphNode>();
+    auto NewCommand = new buffer_read_command_handle(hCommandBuffer, GraphNode,
```
hdelan, line 843: Also here

source/adapters/cuda/command_buffer.cpp
```diff
@@ -642,11 +889,30 @@ ur_result_t UR_APICALL urCommandBufferAppendMemBufferWriteRectExp(
...
-    auto SyncPoint = hCommandBuffer->addSyncPoint(GraphNode);
+    CUgraphNode SyncPointNode = SignalNode ? SignalNode : GraphNode;
+    auto SyncPoint = hCommandBuffer->addSyncPoint(SyncPointNode);
     if (pSyncPoint) {
       *pSyncPoint = SyncPoint;
     }
+
+    std::vector<CUgraphNode> WaitNodes =
+        numEventsInWaitList ? DepsList : std::vector<CUgraphNode>();
+    auto NewCommand = new buffer_write_rect_command_handle(
```
hdelan, line 908: And here

hdelan (commented):
Mostly LGTM but a few small questions. Most importantly about checking whether profiling is enabled or not.

source/adapters/cuda/command_buffer.cpp
hdelan: There are some more other uses below but I haven't commented on all of them

source/adapters/cuda/command_buffer.cpp
EwanC: These CUDA calls were inside a destructor, so throwing exception with `UR_CHECK_ERROR` would be bad form. So i've moved the CUDA calls just outside of the destructor so that we can do this checking with `UR_CHECK_ERROR`

source/adapters/cuda/command_buffer.cpp
EwanC: So this heap allocated object is returned to the user as an output parameter as handle typedefed to a raw pointer, which they can refcount with `release`/`retain` counts. So I don't think `unique_ptr` is the right abstraction because it is not singly owned by the runtime.

While ruminating about this, I think we don't actually need these objects to be user refcounted by the user, and can just have the same lifetime as the parent UR command-buffer (OpenCL does this already). At which point the command-buffer could store unique ptrs to these heap object. Making this change is out of scope for this PR, but I propose making it separately as a standalone PR? I can create an issue to track it

source/adapters/cuda/command_buffer.cpp
EwanC: SYCL-Graphs doesn't support profiling on the node level (i.e. inside a graph rather than whole graph submission) yet, so even if profiling was enabled we don't have a way to express back up the stack any of the counters results from per node events.

One of the intents of this PR is to provide a SYCL functional equivalent of `cudaGraphAddEventRecordNode`, we decided not to create a separate UR entry-point (e.g. `urCommandBufferAppendEventSignal`) for this, nor a separate SYCL-Graph node type (e.g `graph.add_signal_node()`) and instead add an extra parameter to existing UR/SYCL-Graph command entry-points, which when non-null would result in the creation of the extra `cudaGraphAddEventRecordNode` node.

source/adapters/cuda/command_buffer.cpp
hdelan: Nice thanks

source/adapters/cuda/command_buffer.cpp
hdelan: That sounds like a good solution to me. 

FWIW in other UR entry points we tend to use `unique_ptr`s for heap allocations within entry points, just so resources are freed if something goes wrong at some stage within the UR entry point. At the end of the UR call we usually call `out_param = unique_ptr.release()` and give ownership of the mem to the user.

source/adapters/cuda/command_buffer.cpp
hdelan: Fair enough. That sounds reasonable to me.

hdelan (approved):
LGTM

source/adapters/cuda/command_buffer.cpp
```diff
@@ -1045,16 +1394,82 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferUpdateKernelLaunchExp(
...
+  if (!CommandBuffer->CudaGraphExec) {
+    return UR_RESULT_ERROR_INVALID_OPERATION;
+  }
+
+  // Update requires command-buffer to be created with update enabled
+  if (!CommandBuffer->IsUpdatable) {
+    return UR_RESULT_ERROR_INVALID_OPERATION;
+  }
+
+  // Error to try update the signal event, when a signal event wasn't set on
```
fabiomestre, line 1422:
```suggestion
  // Error to try to update the signal event, when a signal event wasn't set on
```

test/conformance/exp_command_buffer/fixtures.h
```diff
@@ -304,6 +301,128 @@ struct urCommandBufferMultipleKernelUpdateTest
...
+        for (auto &buffer : buffers) {
+            ASSERT_SUCCESS(urMemBufferCreate(context, UR_MEM_FLAG_READ_WRITE,
+                                             allocation_size, nullptr,
+                                             &buffer));
+            ASSERT_NE(buffer, nullptr);
+        }
+
+        // Create a command-buffer with update enabled.
+        ur_exp_command_buffer_desc_t desc{
+            UR_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_DESC, nullptr, true};
```
fabiomestre, line 340: It is a bit confusing that this fixture is does not mention updates and creates an updatable command buffe. Does it have to be marked as updatable?

fabiomestre (approved):

test/conformance/exp_command_buffer/fixtures.h
EwanC: Good catch! This was just an oversight. Updated to fully initalize the struct rather than only some of the members, and make updatable false, i.e.
```cpp
         ur_exp_command_buffer_desc_t desc{
            /*.stype=*/UR_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_DESC,
            /*.pNext =*/nullptr,
            /*.isUpdatable =*/false,
            /*.isInOrder =*/false,
            /*.enableProfiling =*/false,
        };
```

EwanC:
ping @oneapi-src/unified-runtime-level-zero-write for review

pbalcer (approved):

PR #1826, opened 2024-07-05
refactor loader lifetime management

pbalcer:
This patch implements an atomic singleton class for managing the lifecycle of the context objects inside of the loader. This class ensures that the contexts always exist and lets the loader manually destroy them on user request (during teardown).

Thanks to this change, the loader no longer relies on the order of library constructors and destructors. It also gets us 90% towards allowing the loader to be statically linked with the application (which is why I did this :P).


kbenzie (approved):
I've been investigating tear down regression in https://github.com/intel/llvm/pull/14145 and I believe this will solve those.

I'll be adding some logging on top of this once it merges for loading/unloader apdater libs to satisfy some e2e testing.

PR #1027, opened 2023-11-01
[SYCL][HIP] Implement mem_advise for HIP

GeorgeWeb:
Adds initial implementation for mem_advise to the HIP adapter.

PR #1828, opened 2024-07-08
Add a workaround for urEnqueue loader problems with ADAPTER_SPECIFIC

omarahmed1111:
This is a temporarily fix for a problem with `urEnqueue` entries in `ldrddi` loader. The problem is that these entries could return a valid `Event` handle while in the same time the entry is returning `UR_RESULT_ADAPTER_SPECIFIC_ERROR` like in [here](https://github.com/oneapi-src/unified-runtime/blob/01af5602345ac3289cc084eb19bf266c054da075/source/adapters/hip/enqueue.cpp#L1395C14-L1395C46). This would cause a seg-fault where the loader wrapper would not be called on this event handle like in [here](https://github.com/oneapi-src/unified-runtime/blob/01af5602345ac3289cc084eb19bf266c054da075/source/loader/ur_ldrddi.cpp#L5298) and later when `urEventRelease` is called on this valid event handle, it will pass by the loader to extract the dditable from the handle like in [here](https://github.com/oneapi-src/unified-runtime/blob/01af5602345ac3289cc084eb19bf266c054da075/source/loader/ur_ldrddi.cpp#L3972) and that will be invalid as that is not a loader handle but the original event handle.

### DIFF
--- a/scripts/templates/ldrddi.cpp.mako
+++ b/scripts/templates/ldrddi.cpp.mako
@@ -272,9 +272,16 @@ namespace ur_loader
         %>
         %for i, item in enumerate(epilogue):
         %if 0 == i and not item['release'] and not item['retain'] and not th.always_wrap_outputs(obj):
-        if( ${X}_RESULT_SUCCESS != result )
+        ## TODO: Remove once we have a concrete way for submitting warnings in place.
+        %if re.match(r"urEnqueue\w+", th.make_func_name(n, tags, obj)):
+        // In the event of ERROR_ADAPTER_SPECIFIC we should still attempt to wrap any output handles below.
+        if( ${X}_RESULT_SUCCESS != result && ${X}_RESULT_ERROR_ADAPTER_SPECIFIC != result )
+            return result;
+        %else:
+        if( ${X}_RESULT_SUCCESS != result)
             return result;
 
+        %endif
         %endif
         ## Before we can re-enable the releases we will need ref-counted object_t.
         ## See unified-runtime github issue #1784
--- a/source/loader/ur_ldrddi.cpp
+++ b/source/loader/ur_ldrddi.cpp
@@ -4144,10 +4144,11 @@ __urdlllocal ur_result_t UR_APICALL urEnqueueKernelLaunch(
                         pGlobalWorkSize, pLocalWorkSize, numEventsInWaitList,
                         phEventWaitListLocal.data(), phEvent);
 
-    if (UR_RESULT_SUCCESS != result) {
+    // In the event of ERROR_ADAPTER_SPECIFIC we should still attempt to wrap any output handles below.
+    if (UR_RESULT_SUCCESS != result &&
+        UR_RESULT_ERROR_ADAPTER_SPECIFIC != result) {
         return result;
     }
-
     try {
         // convert platform handle to loader handle
         if (nullptr != phEvent) {
@@ -4200,10 +4201,11 @@ __urdlllocal ur_result_t UR_APICALL urEnqueueEventsWait(
     result = pfnEventsWait(hQueue, numEventsInWaitList,
                            phEventWaitListLocal.data(), phEvent);
 
-    if (UR_RESULT_SUCCESS != result) {
+    // In the event of ERROR_ADAPTER_SPECIFIC we should still attempt to wrap any output handles below.
+    if (UR_RESULT_SUCCESS != result &&
+        UR_RESULT_ERROR_ADAPTER_SPECIFIC != result) {
         return result;
     }
-
     try {
         // convert platform handle to loader handle
         if (nullptr != phEvent) {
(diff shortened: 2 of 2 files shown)


scripts/templates/ldrddi.cpp.mako
```diff
@@ -272,9 +272,14 @@ namespace ur_loader
         %>
         %for i, item in enumerate(epilogue):
         %if 0 == i and not item['release']:
-        if( ${X}_RESULT_SUCCESS != result )
+        %if re.match(r"urEnqueue\w+", th.make_func_name(n, tags, obj)):
+        if( ${X}_RESULT_SUCCESS != result && ${X}_RESULT_ERRPR_ADAPTER_SPECIFIC != result )
```
aarongreig, line 276: my typo from the chat
```suggestion
        if( ${X}_RESULT_SUCCESS != result && ${X}_RESULT_ERROR_ADAPTER_SPECIFIC != result )
```

also I think it's worth adding a comment here to the effect of "In the event of ERROR_ADAPTER_SPECIFIC we should still attempt to wrap any output handles below"

pbalcer:
> The problem is that these entries could return a valid `Event` handle while in the same time the entry is returning `UR_RESULT_ADAPTER_SPECIFIC_ERROR` like in [here](https://github.com/oneapi-src/unified-runtime/blob/01af5602345ac3289cc084eb19bf266c054da075/source/adapters/hip/enqueue.cpp#L1395C14-L1395C46).

Is that correct though? Is the user expected to check for this error, and then handle the output handles if they were created? If so, is this documented?

I'd expect the proper fix here to be mixing the adapters so that they never return valid handles on error.

omarahmed1111:
> > The problem is that these entries could return a valid `Event` handle while in the same time the entry is returning `UR_RESULT_ADAPTER_SPECIFIC_ERROR` like in [here](https://github.com/oneapi-src/unified-runtime/blob/01af5602345ac3289cc084eb19bf266c054da075/source/adapters/hip/enqueue.cpp#L1395C14-L1395C46).
> 
> Is that correct though? Is the user expected to check for this error, and then do something with the output handles if they were created? That would mean that all input handles to these entry points need to be NULL-initialized so that the user can distinguish during error handling between a properly uninitialized handle and a handle created by the adapter. If so, is this documented?
> 
> I'd expect the proper fix here to be fixing the adapters so that they never return valid handles on error. Otherwise error handling becomes complex:
> 
> ```
> ur_event_handle_t e = nullptr;
> ur_result_t res = urEnqueueKernelLaunch(..., &e);
> if (res != UR_RESULT_SUCCESS) {
>    if (e != nullptr) {
>      urEventRelease(e);
>    }
> }
> ```

I am not sure if that should be the correct behaviour, it make sense to me that entries shouldn't return a valid handle in case of errors. I don't see in the spec that this is a documented behaviour But I noticed this [PR](https://github.com/oneapi-src/unified-runtime/pull/1027) which seems that this was an intentional behaviour. I guess @GeorgeWeb you might have a better idea on this or thoughts?

omarahmed1111:
> > The problem is that these entries could return a valid `Event` handle while in the same time the entry is returning `UR_RESULT_ADAPTER_SPECIFIC_ERROR` like in [here](https://github.com/oneapi-src/unified-runtime/blob/01af5602345ac3289cc084eb19bf266c054da075/source/adapters/hip/enqueue.cpp#L1395C14-L1395C46).
> 
> Is that correct though? Is the user expected to check for this error, and then do something with the output handles if they were created? That would mean that all input handles to these entry points need to be NULL-initialized so that the user can distinguish during error handling between a properly uninitialized handle and a handle created by the adapter. If so, is this documented?
> 
> I'd expect the proper fix here to be fixing the adapters so that they never return valid handles on error. Otherwise error handling becomes complex:
> 
> ```
> ur_event_handle_t e = nullptr;
> ur_result_t res = urEnqueueKernelLaunch(..., &e);
> if (res != UR_RESULT_SUCCESS) {
>    if (e != nullptr) {
>      urEventRelease(e);
>    }
> }
> ```

After some other investigations, I think this was more meant as a warning, looking at this [code](https://github.com/intel/llvm/blob/1fcc1cf48e016a199131ab13d3d1bb201c495d8a/sycl/source/detail/plugin.hpp#L158) and how [`UR_RESULT_ERROR_ADAPTER_SPECIFIC`](https://oneapi-src.github.io/unified-runtime/core/api.html#ur__api_8h_1a70996548e8016a2761523966dd1a470ba0618033f629dd69b011f6e69b5d92fff) in the spec, I think the solution in this PR makes more sense to me now as a temporary solution as `UR_RESULT_ERROR_ADAPTER_SPECIFIC` should not make an exception and should in case of warning be treated as `UR_RESULT_SUCCESS` so the loader should also align with that temporarily and treat it as `UR_RESULT_SUCCESS`. As the current situation is that sycl-rt and UR adapters is treating `UR_RESULT_ERROR_ADAPTER_SPECIFIC` as `UR_RESULT_SUCCESS` but UR loaders are treating it as errors. So, I think the 3 layers should align for now (sycl-rt, UR loaders, UR adapters) and later we could change that but that would need to be changed on this 3 layers together.

pbalcer:
Makes sense. Please add a TODO in the script to remove the hack once the proper solution is in place.

aarongreig (approved):

PR #1829, opened 2024-07-08
Treat bindless images handle types as native handles

callumfare:
The bindless image extension added new handle types, however some of them are used in the SYCL Runtime in ways that contradict how handles are expected to be implemented, by casting them directly between other types. This breaks when UR is used with the UR loader, which transparently wraps handles to allow dispatching to specific adapters. This usage implies they are more like native handles, so they should be implemented like native handles.

Rename the image and image_mem types - if they end in `native_handle_t`, they are treated as native handles. Also update ImageGetInfoExp to take a context handle so it still possible for the loader to dispatch to the adapters.

Also fix incorrect return condition for `UR_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR` in bindless image functions

### DIFF
--- a/scripts/core/exp-bindless-images.yml
+++ b/scripts/core/exp-bindless-images.yml
@@ -15,12 +15,12 @@ ordinal: "99"
 type: handle
 desc: "Handle of bindless image"
 class: $xBindlessImages
-name: "$x_exp_image_handle_t"
+name: "$x_exp_image_native_handle_t"
 --- #--------------------------------------------------------------------------
 type: handle
 desc: "Handle of bindless image memory"
 class: $xBindlessImages
-name: "$x_exp_image_mem_handle_t"
+name: "$x_exp_image_mem_native_handle_t"
 --- #--------------------------------------------------------------------------
 type: handle
 desc: "Handle of interop memory"
@@ -355,7 +355,7 @@ params:
     - type: $x_device_handle_t
       name: hDevice
       desc: "[in] handle of the device object"
-    - type: $x_exp_image_handle_t
+    - type: $x_exp_image_native_handle_t
       name: hImage
       desc: "[in][release] pointer to handle of image object to destroy"
 returns:
@@ -376,7 +376,7 @@ params:
     - type: $x_device_handle_t
       name: hDevice
       desc: "[in] handle of the device object"
-    - type: $x_exp_image_handle_t
+    - type: $x_exp_image_native_handle_t
       name: hImage
       desc: "[in][release] pointer to handle of image object to destroy"
 returns:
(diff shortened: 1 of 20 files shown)


scripts/core/exp-bindless-images.yml
```diff
@@ -578,7 +578,10 @@ class: $xBindlessImages
 name: ImageGetInfoExp
 ordinal: "0"
 params:
-    - type: $x_exp_image_mem_handle_t
+    - type: $x_context_handle_t
+      name: hContext
+      desc: "[in] handle of the context object"
```
przemektmalon, line 583: Why is the context parameter being added here? No implementations use the context, and the extra parameter doesn't seem to be reflected in all of the adapters code. Additionally, if this change is necessary in any way, it will require changes on the DPC++ side too.

callumfare: The UR loader wraps every `_handle_t` type to insert a pointer to the adapter that created the handle. This way the loader is able to dispatch UR function calls to the correct adapter. This breaks DPC++'s usage of these image handles because it reinterpret casts them straight to pointers etc. This PR makes them native handles to disable the handle wrapping in the loader. However that means that `ImageGetInfoExp` no longer has any 'real' handles so the loader cannot dispatch the function, which is why the context handle has been added. It could alternatively be some other handle we know is available like the platform or adapter.

I realise this is a slightly awkward and disruptive change but it's the easiest fix I can come up with to fix bindless images for https://github.com/intel/llvm/pull/14145 - happy to take other suggestions but this is probably easiest in the short term

I'll put a intel/llvm PR up shortly.

przemektmalon: I see.

Does the context handle have to be valid? In which case we need to change the SYCL API to receive a context handle, and any code using this API. Or can we just pass a `nullptr` to UR since the value is never used by any backend?

kbenzie: The handle does need to be valid because `nullptr` doesn't have any reference to an adapters dispatch table containing function pointers to the implementation of any UR entry point.

callumfare:
intel/llvm PR - https://github.com/intel/llvm/pull/14516

scripts/core/exp-bindless-images.yml
callumfare: @przemektmalon I've just put up a PR for DPC++ - https://github.com/intel/llvm/pull/14516 - the SYCL API doesn't need changed because the implementation already has access to the context any time ImageGetInfo is used

przemektmalon: > the SYCL API doesn't need changed because the implementation already has access to the context any time ImageGetInfo is used

Ah, you're right. I left two minor comments on the DPC++ PR. Everything LGTM otherwise.

przemektmalon (approved):

kbenzie (approved):

DBDuncan (approved):

PR #1831, opened 2024-07-08
Always use translated command-lists

fabiomestre:
Updates the implementation of command-buffers to use translated command-list handles in every API call instead of only using them for experimental L0 API's. This avoids having to keep track of 2 separate command-list handles which can be confusing.


EwanC (approved):

EwanC:
@nrspruit You originally authored https://github.com/oneapi-src/unified-runtime/pull/1778, does this change make sense to you?

EwanC (approved):

pbalcer (commented):
please test this with `ZE_ENABLE_TRACING_LAYER=1` set. I have a suspicion that you can't just use a translated handles in regular APIs.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -1754,7 +1757,7 @@ ur_result_t updateKernelCommand(
   auto Platform = CommandBuffer->Context->getPlatform();
   ZE2UR_CALL(
       Platform->ZeMutableCmdListExt.zexCommandListUpdateMutableCommandsExp,
-      (CommandBuffer->ZeComputeCommandListTranslated, &MutableCommandDesc));
```
nrspruit, line 1757: You still need the translated compute command list handle. Any call to a n extension function pointer requires all handles used are translated.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -862,9 +865,9 @@ createCommandHandle(ur_exp_command_buffer_handle_t CommandBuffer,
                                ZE_MUTABLE_COMMAND_EXP_FLAG_GLOBAL_OFFSET;
 
   auto Platform = CommandBuffer->Context->getPlatform();
-  ZE2UR_CALL(Platform->ZeMutableCmdListExt.zexCommandListGetNextCommandIdExp,
-             (CommandBuffer->ZeComputeCommandListTranslated,
-              &ZeMutableCommandDesc, &CommandId));
+  ZE2UR_CALL(
+      Platform->ZeMutableCmdListExt.zexCommandListGetNextCommandIdExp,
+      (CommandBuffer->ZeComputeCommandList, &ZeMutableCommandDesc, &CommandId));
```
nrspruit, line 870: This will fail, you need to translate both handles, not just the copy above. This will break the code.

fabiomestre:
Closing due to review feedback mentioning that this won't work.

PR #1833, opened 2024-07-08
[OpenCL] Add OpenCL version check for independent forward progress query

ayylol:
As per [OpenCL docs](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clGetDeviceInfo.html), `CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS` is missing in OpenCL versions prior to 2.1. This change adds a check to the OpenCL version to ensure it is newer than 2.1, otherwise returning a default value for independent forward progress.


aarongreig (approved):

aarongreig:
LLVM PR https://github.com/intel/llvm/pull/15872

PR #1840, opened 2024-07-09
Refactor L0 command-buffer cleanup.

EwanC:
Coverity has reported that there is the possibility of the command-buffer destructor throwing an exception from its calls
to `CleanupCompletedEvent`.

This is from an underlying `throw` in [GetQueue](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/queue.hpp#L890) when a `dynamic_cast` doesn't behave as expected. I initially tried
changing this to an assert but the V2 Level Zero adapter is triggering this case while in it's developmental state, so an assert brings down the whole test suite rather than just being reported as a fail (the UR loader catches exceptions and returns them as error codes).

Therefore this PR addresses the Coverity report by moving the calls to `CleanupCompletedEvent()` outside of the command-buffer destructor. This allows us to cleanup prior to calling the destructor, without the risk of throwing an exception.

DPC++ PR https://github.com/intel/llvm/pull/14503

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -424,9 +424,7 @@ ur_exp_command_buffer_handle_t_::ur_exp_command_buffer_handle_t_(
   urDeviceRetain(Device);
 }
 
-// The ur_exp_command_buffer_handle_t_ destructor releases all the memory
-// objects allocated for command_buffer management.
-ur_exp_command_buffer_handle_t_::~ur_exp_command_buffer_handle_t_() {
+void ur_exp_command_buffer_handle_t_::cleanupCommandBufferResources() {
   // Release the memory allocated to the Context stored in the command_buffer
   urContextRelease(Context);
 
@@ -703,6 +701,7 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t CommandBuffer) {
   if (!CommandBuffer->RefCount.decrementAndTest())
     return UR_RESULT_SUCCESS;
 
+  CommandBuffer->cleanupCommandBufferResources();
   delete CommandBuffer;
   return UR_RESULT_SUCCESS;
 }
--- a/source/adapters/level_zero/command_buffer.hpp
+++ b/source/adapters/level_zero/command_buffer.hpp
@@ -34,8 +34,6 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
       ur_event_handle_t WaitEvent, ur_event_handle_t AllResetEvent,
       const ur_exp_command_buffer_desc_t *Desc, const bool IsInOrderCmdList);
 
-  ~ur_exp_command_buffer_handle_t_();
-
   void registerSyncPoint(ur_exp_command_buffer_sync_point_t SyncPoint,
                          ur_event_handle_t Event);
 
@@ -64,6 +62,10 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
    * @return The chosen command list.
    */
   ze_command_list_handle_t chooseCommandList(bool PreferCopyEngine);
+
+  // Releases the resources associated with the command-buffer before the
+  // command-buffer object is destroyed.
+  void cleanupCommandBufferResources();
 
   // UR context associated with this command-buffer
   ur_context_handle_t Context;


source/adapters/level_zero/command_buffer.hpp
```diff
@@ -65,6 +65,13 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
    */
   ze_command_list_handle_t chooseCommandList(bool PreferCopyEngine);
 
+  /**
+   * Frees the resources associated with the command-buffer.
+   * @return UR_RESULT_SUCCESS or an error if a resources wasn't freed
```
fabiomestre, line 70:
```suggestion
   * @return UR_RESULT_SUCCESS or an error if a resource wasn't freed
```

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -449,6 +449,59 @@ ur_exp_command_buffer_handle_t_::ur_exp_command_buffer_handle_t_(
   urDeviceRetain(Device);
 }
 
+ur_result_t ur_exp_command_buffer_handle_t_::cleanupCommandBufferResources() {
```
fabiomestre, line 452: Since we are still not doing a try / catch in this function, I'm struggling to understand how exactly it fixes the coverity issue?

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -475,22 +528,18 @@ ur_exp_command_buffer_handle_t_::~ur_exp_command_buffer_handle_t_() {
 
   // Release additional signal and wait events used by command_buffer
   if (SignalEvent) {
-    CleanupCompletedEvent(SignalEvent, false);
     urEventReleaseInternal(SignalEvent);
```
fabiomestre, line 531: We are still ignoring the error codes returned by urEventReleaseInternal, zeFenceDestroy and urKernelRelease. Since we already have to move some of the code, I wonder if it would be better to just move everything from the destructor so that we can do proper error checking?

source/adapters/level_zero/command_buffer.cpp
EwanC: The Coverity issue is that an exception can be thrown in `~ur_exp_command_buffer_handle_t_()` specifically, not that a exception is thrown in general. By doing the cleanup work that could throw an exception prior to calling a C++ destructor, I believe that we address the problem.

pbalcer:
E2E L0 failed, but it doesn't seem related:
```
Failed Tests (1):
  SYCL :: ESIMD/regression/minmax.cpp
```

pbalcer (approved):
lgtm, but this is quite a lot of changes to handle a fairly trivial problem.

@igchor it may make sense to create a nothrow variant of the getQueue method that returns something like an std::optional if succesful. Thoughts?

EwanC:
> @igchor it may make sense to create a nothrow variant of the getQueue method that returns something like an std::optional if succesful. Thoughts?

ping @igchor, I could try make this change instead if you think it's a better way to go

igchor:
> > @igchor it may make sense to create a nothrow variant of the getQueue method that returns something like an std::optional if succesful. Thoughts?
> 
> ping @igchor, I could try make this change instead if you think it's a better way to go

Sorry, I missed the notification.

I think it would be best to just move all the logic from the destructor to a function like you did but **not** using `UR_CALL_STORE_ERR` inside - leaving all the logic as is. Also, the function can just return `void`.

If an exception from GetQueue is thrown, (.e.g from `urDeviceRelease`) the rest of the cleanup functions won't be called anyway because the exception will abort the `cleanupCommandBufferResources` function and propagate up. To avoid that, you would have to wrap each function in a try/catch but I don't think that's a good idea (you might end up with some inconsistent state - it's better to just have a leak). Also, as far as I understand this should not happen during execution anyway, it is just something coverity thinks is possible, right?

EwanC:
Thanks for your reply @igchor, I've updated the PR with your suggestion.

It is indeed only the possibility of an exception being thrown in a C++ object destructor that Coverity has detected. Avoiding leaks and returning a reliable error when things go wrong when freeing resources is over an above that, so can be considered out of scope for the purposes of this PR.

igchor (approved):

Bensuo (approved):

fabiomestre (approved):

Issue #1837, opened 2024-07-09
Use of urEnqueueUSMFill on OpenCL enforces requirement on Host Mem Allocs for odd sizes of pattern


coldav:
On oneapi-construction-kit/refsi we support USM but not Host Allocs. The workaround for odd sizes of pattern_size calls clHostMemAllocINTEL_fn  which refsi does not support, and likely many implementations won't. refsi does support device allocations however.

This can be reproduced by building cloning https://github.com/codeplaysoftware/oneapi-construction-kit.git and building with:
```
cmake -GNinja -DCA_MUX_TARGETS_TO_ENABLE="riscv" -DCA_RISCV_ENABLED=ON  -DCA_ENABLE_API=cl       -DCA_LLVM_INSTALL_DIR=<llvm_install>  -DCA_CL_ENABLE_ICD_LOADER=ON   -DOCL_EXTENSION_cl_khr_command_buffer=ON -DOCL_EXTENSION_cl_khr_command_buffer_mutable_dispatch=ON
```
LLVM Install requires RISCV and lld to be enabled.

This is shown up in the SYCL CTS test:
test_queue/queue shortcuts unified shared memory core

where it looks to cover a fill of pattern_size 12 after commit https://github.com/intel/llvm/pull/13788

PR #1845, opened 2024-07-10
Remove the power of 2 pattern size requirement from USMFill

aarongreig:
All adapters have sufficiently robust workarounds that this is no longer necessary, and the sycl runtime is now using USMFill in such a way that if we enable the validation layer this case will get hit and break stuff.

### DIFF
--- a/scripts/core/enqueue.yml
+++ b/scripts/core/enqueue.yml
@@ -1061,7 +1061,6 @@ returns:
     - $X_RESULT_ERROR_INVALID_SIZE:
         - "`patternSize == 0 || size == 0`"
         - "`patternSize > size`"
-        - "`(patternSize & (patternSize - 1)) != 0`"
         - "`size % patternSize != 0`"
         - "If `size` is higher than the allocation size of `ptr`"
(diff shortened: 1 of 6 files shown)


scripts/core/enqueue.yml
```diff
@@ -1061,7 +1061,6 @@ returns:
     - $X_RESULT_ERROR_INVALID_SIZE:
         - "`patternSize == 0 || size == 0`"
         - "`patternSize > size`"
-        - "`(patternSize & (patternSize - 1)) != 0`"
```
EwanC, line 1064: I think this change should be reflected in the equivalent command-buffer entry-point too https://github.com/oneapi-src/unified-runtime/blob/daa3ee1a8922030e655cd7a4a71ed7727f7b0bfc/scripts/core/exp-command-buffer.yml#L411

kbenzie:
Does this fix #1837?

aarongreig:
> Does this fix #1837?

no it fixes some new e2e regression on the port branch

kbenzie (approved):

EwanC (approved):

Issue #888, opened 2023-09-25
Generate scripts are broken on windows runners following recent image update


aarongreig:
Since this update https://github.com/actions/runner-images/releases/tag/win22%2F20230918.1 the check-generated target fails on windows 2022 and 2019 runners like [this](https://github.com/oneapi-src/unified-runtime/actions/runs/6247903093/job/16961414508#step:6:91). A bit of debugging showed that this is coming from this line https://github.com/oneapi-src/unified-runtime/blob/main/scripts/generate_docs.py#L246 (see https://github.com/oneapi-src/unified-runtime/actions/runs/6263499058/job/17008038672?pr=883#step:6:18). Simply adding a step to install doxygen didn't fix the issue so it isn't clear if we really are failing to find doxygen on the PATH or if something weirder is going on. This issue does appear to only affect the generate scripts so we could consider disabling that step on windows to recover some CI functionality while we figure this out

PR #1847, opened 2024-07-10
[CMake] Enable clang-format command on Windows

lukaszstolarczuk:
It seems like list of args for Windows was too long. Split cppformat commands into two, each formatting half of passed files. Re-enable 'check-generated' check on Windows CI again.

ref. #888


scripts/templates/stype_map_helpers.hpp.mako
```diff
@@ -7,7 +8,7 @@ from templates import helper as th
     x=tags['$x']
     X=x.upper()
 %>
-// This file is autogenerated from the template at ${self.template.filename}
+// This file is autogenerated from the template at ${os.path.dirname(self.template.filename)}/${os.path.basename(self.template.filename)}
```
pbalcer, line 11: is this related?

pbalcer (approved):

scripts/templates/stype_map_helpers.hpp.mako
lukaszstolarczuk: yeah, generated file contains a comment:
```
// This file is autogenerated from the template at templates/stype_map_helpers.hpp.mako
```

on Windows, there's different separator (-> `\`) and it's displayed as a diff in the generated files (causing failure)

lukaszstolarczuk: alternatively we could just drop this comment 😉

pbalcer: ok, lets leave it. thanks.

lukaszstolarczuk:
Rebased with new `source checks` workflow.

lukaszstolarczuk:
ref. #2158

cmake/helpers.cmake
```diff
@@ -28,16 +28,22 @@ function(add_cppformat name)
...
+        add_custom_target(cppformat-${name}-1
+            COMMAND ${CLANG_FORMAT} --style=file --i ${selected_files}
+            COMMENT "Format CXX source files (pt 1)"
+        )
+
+        list(SUBLIST ARGN 251 -1 selected_files)
+        add_custom_target(cppformat-${name}-2
+            COMMAND ${CLANG_FORMAT} --style=file --i ${selected_files}
+            COMMENT "Format CXX source files (pt 2)"
+        )
```
kbenzie, line 42: A custom target can contain multiple commands, there's no need for multiple `cppformat-${name}` targets. e.g.

```cmake
add_custom_target(example
  COMMAND echo first
  COMMAND echo second
  COMMENT Multiple commands example
)
```

Has this output when invoking the `example` target:

```
first
second
```

```suggestion
        # Split args into 2 parts (in Windows the list is probably too long)
        list(SUBLIST ARGN 0 250 selected_files_1)
        list(SUBLIST ARGN 251 -1 selected_files_2)
        add_custom_target(cppformat-${name}
            COMMAND ${CLANG_FORMAT} --style=file --i ${selected_files_1}
            COMMAND ${CLANG_FORMAT} --style=file --i ${selected_files_2}
            COMMENT "Format CXX source files (pt 1)"
        )
```

lukaszstolarczuk: thx! that's super nice, applied.

kbenzie (approved):

lukaszstolarczuk:
soo.. can we merge this? I'd like to set up some more changes around CI workflows - it would be nice to not have to re-base it again :)

PR #1849, opened 2024-07-10
[OpenCL] Add extension check for sub group sizes query

ayylol:
`cl_intel_required_subgroup_size` is needed to be able to check for `CL_DEVICE_SUB_GROUP_SIZES_INTEL` (See [here](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_required_subgroup_size.html)). This patch adds a check for this extension, if it is not present a default sub group sizes list is returned.


source/adapters/opencl/device.cpp
```diff
@@ -938,17 +938,30 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
...
-                                                     SubGroupSizes.size());
+    /* CL_DEVICE_SUB_GROUP_SIZES_INTEL is only supported if the device has the
+     * cl_intel_required_subgroup_size extension, if it does not have it we
+     * return a default subgroups sizes list of {1}
+     * */
+    size_t ExtSize = 0;
+    urDeviceGetInfo(hDevice, UR_DEVICE_INFO_EXTENSIONS, 0, nullptr, &ExtSize);
+    std::string ExtStr(ExtSize, 0);
+    urDeviceGetInfo(hDevice, UR_DEVICE_INFO_EXTENSIONS, ExtSize, ExtStr.data(),
+                    nullptr);
```
kbenzie, line 949: Using `urDeviceGetInfo` here results in unnecessary indirection. We have the `cl_adapter::checkDeviceExtensions()` helper function which calls OpenCL directly which should be preferred.

source/adapters/opencl/device.cpp
```diff
@@ -938,17 +938,30 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
...
+          clGetDeviceInfo(cl_adapter::cast<cl_device_id>(hDevice), CLPropName,
+                          0, nullptr, &SubGroupSizesSize));
+      std::vector<size_t> SubGroupSizes(SubGroupSizesSize / sizeof(size_t));
+      CL_RETURN_ON_FAILURE(
+          clGetDeviceInfo(cl_adapter::cast<cl_device_id>(hDevice), CLPropName,
+                          SubGroupSizesSize, SubGroupSizes.data(), nullptr));
+      return ReturnValue.template operator()<uint32_t>(SubGroupSizes.data(),
+                                                       SubGroupSizes.size());
+    } else {
+      return ReturnValue.template operator()<uint32_t>(std::data({1}), 1);
```
kbenzie, line 963: If the query is not supported we should not be returning a value but only and erro code `return UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION`.

kbenzie (approved):

kbenzie:
@ayylol as noted in #1863 I'm going to have to revert this due to `check-sycl-e2e` regressions.

PR #1852, opened 2024-07-10
[L0] Use all devices from a context in urProgramBuild

igchor:
Using just the first device makes creating a kernel from that program for any other devices impossible. urEnqueueKernelLaunch fails with UR_RESULT_ERROR_INVALID_QUEUE.

I don't know if using only the first device was intentional or not (perhaps an optimization?) but the test that I added passes for opencl so I think it would make sense to make L0 behave similarly.

### DIFF
--- a/source/adapters/level_zero/program.cpp
+++ b/source/adapters/level_zero/program.cpp
@@ -284,8 +284,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramLink(
     ur_program_handle_t
         *Program ///< [out] pointer to handle of program object created.
 ) {
-  return urProgramLinkExp(Context, Count, Context->Devices.data(), 1, Programs,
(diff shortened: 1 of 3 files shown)


nrspruit:
you copied my review none looked at :) https://github.com/oneapi-src/unified-runtime/pull/1276/files , thanks for making the test change as well.

nrspruit:
Could you add the LinkExp change as well?

nrspruit (approved):
LGTM, optional to add linkexp

igchor:
> Could you add the LinkExp change as well? or that can wait for a follow-up review.

Sure, done. Actually, there was one other problem with urProgramLink, arguments were passed in wrong order (number od devices and number of program handles)

pbalcer (approved):

PR #1856, opened 2024-07-11
Make urContextCreateWithNativeHandle work for SYCL use with loader

aarongreig:
The public sycl interface allows you to create a context from a native context handle without passing in anything else, to match this we need the device parameter to be optional. Allowing the device to be null has the potential to break the loader since it no longer has a loader handle to get function pointers from, so we also need to add a handle parameter that the sycl RT is guaranteed to have access to when it needs to call `ContextCreateWithNative`.

This PR also makes the device param in `QueueCreateWithNative` optional for similar reasons.

### DIFF
--- a/scripts/core/context.yml
+++ b/scripts/core/context.yml
@@ -228,12 +228,15 @@ params:
       name: hNativeContext
       desc: |
             [in][nocheck] the native handle of the context.
+    - type: $x_adapter_handle_t
+      name: hAdapter
+      desc: "[in] handle of the adapter that owns the native handle"
     - type: uint32_t
       name: numDevices
       desc: "[in] number of devices associated with the context"
     - type: "const $x_device_handle_t*"
       name: phDevices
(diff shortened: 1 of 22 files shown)


kbenzie (approved):

scripts/core/context.yml
```diff
@@ -228,12 +228,15 @@ params:
       name: hNativeContext
       desc: |
             [in][nocheck] the native handle of the context.
+    - type: $x_platform_handle_t
+      name: hPlatform
```
aarongreig, line 232: making this an adapter instead, it makes the sycl changes cleaner and only needs a slight modification to the conformance test fixtures

pbalcer (approved):

jchlanda (approved):

omarahmed1111 (approved):

aarongreig:
ping @oneapi-src/unified-runtime-native-cpu-write

PR #1857, opened 2024-07-12
[Testing] Handle multi-device and PVC test failures

igchor:
To handle output matching on multi-device systems, parse input file and remove duplicate lines (line which differ only by device id).

Also, skip sampler and usm tests if required functions are not supported to enable running them on PVC.


cmake/match.py
```diff
@@ -87,6 +88,12 @@ def main():
             )
         )
 
+    # Every fail has a single corresponding match line but if there are multiple
+    # devices being tested there will be multiple input lines with the same failure
+    # message. To avoid matching mismatch, remove lines that differ only by device ID.
+    input_lines = [re.sub(r'ID[0-9]ID', 'X', line) for line in input_lines]
```
pbalcer, line 94: we use this for more than just test match files. Maybe you can remove the duplicate lines earlier, like in run_cts.py?

igchor: That makes sense. Done.

igchor:
> lgtm, but please add a workflow that runs these tests on PVC (`L0_E2E`).

I've added a run on `L0_2T` - this additionally enables multi-device tests.

pbalcer (approved):

PR #1862, opened 2024-07-15
[ASAN] fix recursive initialization of sanitizer context

pbalcer:
The AsanOptions class, a singleton, was being instantiated inside of the sanitizer context constructor, but, inside of its contructor, it was using the logger from inside of the context. This only worked before because the context variable was a global and the logger just so happened to be initialized prior to the AsanOptions.

Now that the layer contexts are not globals, this was causing a deadlock on trying to retrieve a context while it was being initialized. The fix is to pass the logger manually to the AsanOptions class.

A better fix might be to refactor AsanOptions to no longer be a singleton, but I was trying to minimize changes since this is blocking intel/llvm update.

### DIFF
--- a/source/loader/layers/sanitizer/asan_interceptor.cpp
+++ b/source/loader/layers/sanitizer/asan_interceptor.cpp
@@ -146,10 +146,12 @@ ur_result_t enqueueMemSetShadow(ur_context_handle_t Context,
 
 } // namespace
 
-SanitizerInterceptor::SanitizerInterceptor() {
-    if (Options().MaxQuarantineSizeMB) {
+SanitizerInterceptor::SanitizerInterceptor(logger::Logger &logger)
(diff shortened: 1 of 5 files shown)


kbenzie (approved):

pbalcer:
` 2/18 Test  #8: platform-adapter_native_cpu ................***Timeout 180.15 sec`

This is unrelated. I've seen this failing before a few times.

pbalcer:
@AllanZyne This was failing on intel/llvm CI after #1826, and we needed to quickly fix it. So we made a decision to merge this without waiting for your review. The changes are fairly small, but feel free to make a follow-up patch if you want to refactor this code.

yingcong-wu:
Thank you @pbalcer , We will make a follow-up PR for this later.

PR #1865, opened 2024-07-16
[DeviceSanitizer] refactor options

yingcong-wu:
A follow-up of https://github.com/oneapi-src/unified-runtime/pull/1862, refactor AsanOptions to avoid initialization deadlock.

SYCLOS PR: https://github.com/intel/llvm/pull/14579

### DIFF
--- a/source/loader/layers/sanitizer/asan_interceptor.cpp
+++ b/source/loader/layers/sanitizer/asan_interceptor.cpp
@@ -146,12 +146,13 @@ ur_result_t enqueueMemSetShadow(ur_context_handle_t Context,
 
 } // namespace
 
-SanitizerInterceptor::SanitizerInterceptor(logger::Logger &logger)
-    : logger(logger) {
-    if (Options(logger).MaxQuarantineSizeMB) {
+// Context is not ready because this constructor is part of the context
+// initialization, so we pass AsanOptions to this constructor and this
+// constructor alone.
+SanitizerInterceptor::SanitizerInterceptor(AsanOptions &Options) {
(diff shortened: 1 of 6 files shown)


pbalcer (approved):

yingcong-wu:
I disagree, AsanOptions class is nothing but a helper that parses and validates the options, it does not have any implementation details. And move it to context would allow easy access to it just like the logger.

AllanZyne:
> I disagree, AsanOptions class is nothing but a helper that parses and validates the options, it does not have any implementation details. And move it to context would allow easy access to it just like the logger.

"options" is too general in context, since we will support other sanitizers in the future. 
I suggest you can rename it to "asanOptions", or move it under interceptor.

AllanZyne:
I prefer to move it under interceptor, and create a new method "getOption()" for outside code.

AllanZyne:
Maybe "getContext()->interceptor" will be rename to "asanInterceptor" in the future. 
Or we'll create a base class for interceptor, since sanitizers can not be enabled at the same time.

yingcong-wu:
> Maybe "getContext()->interceptor" will be rename to "asanInterceptor" in the future.
> Or we'll create a base class for interceptor, since sanitizers can not be enabled at the same time.

That sounds reasonable. If we want to put many sanitizers in the same context, then there are many refactoring to be done. And since we use the name `UR_LAYER_ASAN_OPTIONS` which is specified for ASAN, then it is more reasonable to put in under a ASAN interceptor than put in in a general context.

I would close this PR and this refactor can be in the same change of the ASAN interceptor refactoring.

PR #1867, opened 2024-07-16
initial v2 event pool implementation

pbalcer:


source/adapters/level_zero/v2/event_provider_normal.hpp
```diff
@@ -0,0 +1,67 @@
...
+public:
+  provider_pool(ur_context_handle_t, ur_device_handle_t, event_type,
+                queue_type);
+  ~provider_pool();
+
+  event_borrowed allocate();
+  size_t nfree() const;
+
+private:
+  ze_event_pool_handle_t Pool;
```
igchor, line 43: I would prefer to have some raii wrappers around this (any L0 handle tbh) so that you don't have to manually call destroy in the dtor. Something like this (similar to what we have in tests): https://github.com/igchor/unified-runtime/commit/e775e85da47d29a13be81d1dc63aa3a4019371b3

source/adapters/level_zero/v2/event_pool.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+  ~event_pool(){};
+
+  DeviceId Id() { return Provider->device()->Id; };
+
+  ur_event *allocate();
+  void free(ur_event *event);
+
+private:
+  std::deque<ur_event> Events;
```
igchor, line 43: Hm, do we really need to keep cache of ur_events as well? They are pretty cheap to construct (do you think this will change?) so why not just create a new one for each allocation and only reuse ze_event_handles?

source/adapters/level_zero/v2/event_pool.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace v2 {
+
+class event_pool {
+public:
+  event_pool(std::unique_ptr<event_provider> Provider)
+      : Provider(std::move(Provider)){};
+  event_pool(event_pool &&other) : Provider(std::move(other.Provider)) {}
+
+  ~event_pool(){};
```
igchor, line 35: nit: not needed

source/adapters/level_zero/v2/event.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <ur/ur.hpp>
+#include <ur_api.h>
+#include <ze_api.h>
+
+#include "event_provider.hpp"
+
+namespace v2 {
+
+class ur_event {
```
igchor, line 22: Will this be equivalent to ur_event_handle_t_?

source/adapters/level_zero/v2/event_pool.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+  DeviceId Id() { return Provider->device()->Id; };
+
+  ur_event *allocate();
```
igchor, line 39: Shouldn;'t we make this method thread safe?

source/adapters/level_zero/v2/event_pool_cache.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+  event_pool_cache(ProviderCreateFunc);
+  ~event_pool_cache();
+
+  event_pool_borrowed borrow(DeviceId);
+
+private:
+  ur_mutex Mutex;
+  ProviderCreateFunc ProviderCreate;
+  std::unordered_map<DeviceId, std::vector<std::unique_ptr<event_pool>>> Pools;
```
igchor, line 44: I don;t think you need to use unique_ptr here, you should be able to just store event_pool directly.

igchor: Also, do we actually need unordered_map here? Vector should work too I think

source/adapters/level_zero/v2/event.hpp
```diff
@@ -0,0 +1,37 @@
...
+public:
+  ur_event();
```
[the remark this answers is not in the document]
pbalcer: I just created them for everything when writing out all interfaces. I removed the unused ones.

source/adapters/level_zero/v2/event_provider.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace v2 {
+
+enum event_type { EVENT_REGULAR, EVENT_COUNTER };
+
+using event_borrowed =
+    std::unique_ptr<_ze_event_handle_t, std::function<void(ze_event_handle_t)>>;
+
+struct event_allocation {
+  event_type Type;
```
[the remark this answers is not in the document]
pbalcer: I don't care either way, I just thought this was the coding style :) Its inconsistent everywhere.
I renamed everything to lower-case.

source/adapters/level_zero/v2/event_provider_normal.hpp
pbalcer: Let's refactor this once your PR comes in and we can use the wrapper. I added a TODO.

source/adapters/level_zero/v2/event_pool.hpp
pbalcer: The existing implementation caches the event structure allocations. I think it makes sense. While a fast-path malloc is indeed very quick, the worst-case is not (mmap + page faults). Having a collection of pre-allocated (and pre-faulted) events in a contiguous array may help us avoid latency spikes.

Whether that's going to be measurable it's hard to say, but I say we keep it for now and we can reevaluate once more of the implementation is done and we can run benchmarks.

source/adapters/level_zero/v2/event_pool.hpp
pbalcer: removed.

source/adapters/level_zero/v2/event.hpp
pbalcer: Yes. I don't think there's a point in reusing the existing event structure. We can just allocate this one for all events and cast it to `ur_event_handle_t *`.

source/adapters/level_zero/v2/event_pool.hpp
pbalcer: No, because `event_pool` is meant to be stored in the queue, under its lock. And the queue **must** outlive all the events, so theoretically there should be no problem in waiting with event deallocation until the queue is finished.

source/adapters/level_zero/v2/event_pool_cache.hpp
pbalcer: > I don;t think you need to use unique_ptr here, you should be able to just store event_pool directly.

I sort this vector and wanted to avoid moving possibly large structs.

> Also, do we actually need unordered_map here? Vector should work too I think

Done, but this will require us to specify the max number of devices when constructing the cache.

pbalcer:
The test failure is odd, it's passing on my system. Maybe a driver version? I'll need to login on the CI system and debug.

igchor (approved):

source/adapters/level_zero/v2/event_provider_normal.cpp
```diff
@@ -0,0 +1,119 @@
...
+provider_pool::provider_pool(ur_context_handle_t context,
+                             ur_device_handle_t device, event_type events,
+                             queue_type queue) {
+  ZeStruct<ze_event_pool_desc_t> desc;
+  desc.count = EVENTS_BURST;
+  desc.flags = 0;
+
+  if (events == event_type::EVENT_COUNTER) {
+    ze_event_pool_counter_based_exp_desc_t counterBasedExt = {
+        ZE_STRUCTURE_TYPE_COUNTER_BASED_EVENT_POOL_EXP_DESC, nullptr};
```
al42and, line 30: Shouldn't the declaration be moved up to avoid storing a pointer to an out-of-scope object in `desc.pNext`?

igchor: https://github.com/oneapi-src/unified-runtime/pull/1976

PR #1872, opened 2024-07-17
Revert use of ACCESSED_BY for managed memory

npmiller:
This reverts commits:
* d49cc0f5ba2cce990a30a8ea42387b8935c683a8
* 76c9653835910942a78d757d7f435f00db0fbd68
* 11bd10373300b65ff1b3694e39378bb1c3e9d282
* e18cf578b46c713e2ac89ae642ed8909ce9c57fc

While this is beneficial in some cases it causes very important slow downs when using atomcis in some cases, so reverting this until we can figure out what's going on.


hdelan (approved):

mmoadeli (approved):

hdelan:
Is there a DPC++ PR for this as well?

npmiller:
No it's an adapter only change, it should be good to merge, thanks for adding the label!

PR #1875, opened 2024-07-18
Add option to build loader as a static library

kswiecicki:


CMakeLists.txt
```diff
@@ -134,6 +135,13 @@ if(UR_ENABLE_TRACING)
         set_target_properties(xptifw PROPERTIES
             LIBRARY_OUTPUT_DIRECTORY ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
         )
+
+        install(TARGETS xpti xptifw
```
pbalcer, line 139: hm, I don't know how I feel about adding xpti and xptifw here. could this conflict with anything if a project that uses ur (e.g., SYCL) also has xpti/xptifw targets?
I think we should only do this (and the ur-common) if static linking is enabled.

Also, it may make sense  to do an alias:
```
add_library(ur_xpti ALIAS xpti)
add_library(ur_xptifw ALIAS xptifw)
```
so the names are different. But I'm not sure on that one.

kswiecicki: It seems that it's not possible to install targets from alias.
`CMake Error at CMakeLists.txt:143 (install): install TARGETS given target "ur_xpti" which is an alias.`

pbalcer (approved):

.github/workflows/build-hw-reusable.yml
```diff
@@ -29,14 +29,18 @@ jobs:
...
+          {name: "${{inputs.name}}", platform: "${{inputs.platform}}", static_Loader: OFF},
         ]
         build_type: [Debug, Release]
         compiler: [{c: gcc, cxx: g++}, {c: clang, cxx: clang++}]
         # TODO: The latest L0 loader segfaults when built with clang.
         exclude:
          - adapter: {name: L0, platform: ""}
            compiler: {c: clang, cxx: clang++}
+        include:
+         - adapter: {name: L0, platform: "", static_Loader: ON}
```
pbalcer, line 41: did you figure out how we can avoid having this for every adapter? I think that's the only thing blocking us from merging this.

kswiecicki: I've added a `static_loader` param to the `build-hw-reusable.yml` and a `level-zero-static` job that sets this param. Take a look.

igchor (approved):

CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ option(UR_USE_ASAN "enable AddressSanitizer" OFF)
 option(UR_USE_UBSAN "enable UndefinedBehaviorSanitizer" OFF)
 option(UR_USE_MSAN "enable MemorySanitizer" OFF)
 option(UR_USE_TSAN "enable ThreadSanitizer" OFF)
-option(UR_ENABLE_TRACING "enable api tracing through xpti" OFF)
+option(UR_ENABLE_TRACING "enable api tracing through xpti" ON)
```
pbalcer, line 38: intentional?

CMakeLists.txt
```diff
@@ -48,7 +48,8 @@ option(UR_BUILD_ADAPTER_ALL "Build all currently supported adapters" OFF)
 option(UR_BUILD_EXAMPLE_CODEGEN "Build the codegen example." OFF)
 option(VAL_USE_LIBBACKTRACE_BACKTRACE "enable libbacktrace validation backtrace for linux" OFF)
 option(UR_ENABLE_ASSERTIONS "Enable assertions for all build types" OFF)
-option(UR_BUILD_XPTI_LIBS "Build the XPTI libraries when tracing is enabled" ON)
+option(UR_BUILD_XPTI_LIBS "Build the XPTI libraries when tracing is enabled" OFF)
```
pbalcer, line 51: ?

CMakeLists.txt
kswiecicki: No, I had it turned on for testing, left it by mistake.

CMakeLists.txt
kswiecicki: All should be set back to default now.

pbalcer (approved):

PR #1876, opened 2024-07-18
generalize benchmark script, add more benchmarks

pbalcer:
This patch creates an abstraction for defining benchmarks and
then uses it to add 6 more benchmarks from Velocity Bench.

The build steps of the individual benchmarks have been moved
into the script itself. This is to make the definition of
a benchmark self-contained.


pbalcer:
Example: https://github.com/pbalcer/unified-runtime/pull/14#issuecomment-2243183846

igchor:
Looks good to me. Could you just paste a sample output?

igchor (approved):

pbalcer:
> Just looking at the example, shouldn't baseline be at 100%? (Right now, it is baseline | 101.58%)

The way this works is that I measure "This PR", and then, you can add `--compare label` to compare to some previously saved result (using `--save label`). This will then give you the results. So I can only rely on having data for "This PR", hence why it's 100%. `baseline` is just a label for a previously stored result.
We could add something like `--baseline some_label`, which I could then use for normalization of the results.

PR #1881, opened 2024-07-22
[Benchmark] Add a simple framework and port api_overhead_benchmark

igchor:
from compute-benchmarks.


test/performance/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+        ${CMAKE_CURRENT_SOURCE_DIR}
+        ${CMAKE_CURRENT_SOURCE_DIR}/utility
+        ${CMAKE_CURRENT_SOURCE_DIR}/../conformance/testing/include) # for raii helpers
+
+    target_link_libraries(${name} PRIVATE
+        ${PROJECT_NAME}::loader
+        ${PROJECT_NAME}::headers)
+endfunction()
+
+add_benchmark(submit_kernel benchmarks/api_overhead/submit_kernel.cpp)
```
pbalcer, line 26: We can add this as a scenario in the benchmark scripts ;) But we can do that later.

igchor:
Already implemented in compute-benchmarks: https://github.com/intel/compute-benchmarks/blob/master/source/benchmarks/api_overhead_benchmark/implementations/ur/submit_kernel_ur.cpp

I'll create a separate PR to run this on our CI.

PR #1879, opened 2024-07-22
[DeviceSanitizer] Fix intercepter destruction order

AllanZyne:

PR #1878, opened 2024-07-19
[DeviceSanitizer] Extend a variable lifetime to fix a flaky failure

yingcong-wu:
`EnqueueWriteGlobal` is a async write, so `&Debug`'s content is undetermined at the time of writing. Therefore we need to extend its lifetime to make it work.

PR #1883, opened 2024-07-23
Don't retain device handle references in sanitizer layer.

aarongreig:
Also use uint32_t for Debug value.

### DIFF
--- a/source/loader/layers/sanitizer/asan_interceptor.hpp
+++ b/source/loader/layers/sanitizer/asan_interceptor.hpp
@@ -45,17 +45,7 @@ struct DeviceInfo {
     std::queue<std::shared_ptr<AllocInfo>> Quarantine;
     size_t QuarantineSize = 0;
 
-    explicit DeviceInfo(ur_device_handle_t Device) : Handle(Device) {
-        [[maybe_unused]] auto Result =
-            getContext()->urDdiTable.Device.pfnRetain(Device);
-        assert(Result == UR_RESULT_SUCCESS);
-    }
-
-    ~DeviceInfo() {
-        [[maybe_unused]] auto Result =
-            getContext()->urDdiTable.Device.pfnRelease(Handle);
-        assert(Result == UR_RESULT_SUCCESS);
-    }
+    explicit DeviceInfo(ur_device_handle_t Device) : Handle(Device) {}
 
     ur_result_t allocShadowMemory(ur_context_handle_t Context);
 };
--- a/source/loader/layers/sanitizer/asan_interceptor.hpp
+++ b/source/loader/layers/sanitizer/asan_interceptor.hpp
@@ -45,6 +45,9 @@ struct DeviceInfo {
     std::queue<std::shared_ptr<AllocInfo>> Quarantine;
     size_t QuarantineSize = 0;
 
+    // TODO: re-enable retaining and releasing device handles in DeviceInfo
+    // constructor/destructor. See PR
+    // https://github.com/oneapi-src/unified-runtime/pull/1883
     explicit DeviceInfo(ur_device_handle_t Device) : Handle(Device) {}
 
     ur_result_t allocShadowMemory(ur_context_handle_t Context);


aarongreig:
This fixes https://github.com/intel/llvm/issues/14658 and all the other asan fails.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -708,8 +708,7 @@ ur_result_t SanitizerInterceptor::prepareLaunch(
         };
 
         // Write debug
-        // We use "uint64_t" here because EnqueueWriteGlobal will fail when it's "uint32_t"
```
aarongreig, line 711: I didn't observe this fail when running locally, and I discovered that changing this to `uint32_t` fixes a test for the Debug option. Was this a fix for a specific platform @AllanZyne @yingcong-wu?

aarongreig:
LLVM PR https://github.com/intel/llvm/pull/14719

source/loader/layers/sanitizer/asan_interceptor.cpp
yingcong-wu: A fix related to this var is purposed here: https://github.com/oneapi-src/unified-runtime/pull/1878

yingcong-wu:
Could you please explain why is this change needed?

source/loader/layers/sanitizer/asan_interceptor.cpp
AllanZyne: > I didn't observe this fail when running locally, and I discovered that changing this to `uint32_t` fixes a test for the Debug option. Was this a fix for a specific platform @AllanZyne @yingcong-wu?

I remember correctly, `uint32_t` was not worked on PVC. 
If you tested this and it worked, I also think it'd be better to use `uint32_t`.

source/loader/layers/sanitizer/asan_interceptor.cpp
```diff
@@ -708,8 +708,7 @@ ur_result_t SanitizerInterceptor::prepareLaunch(
...
-        uint64_t Debug = Options(logger).Debug ? 1 : 0;
+        uint32_t Debug = Options(logger).Debug ? 1 : 0;
         EnqueueWriteGlobal(kSPIR_AsanDebug, &Debug, sizeof(Debug));
```
AllanZyne, line 712: I think we can use Options directly, no need to use a temporary variable
```suggestion
        EnqueueWriteGlobal(kSPIR_AsanDebug, &Option(logger).Debug, sizeof(Option(logger).Debug));
```

AllanZyne: If the type of `Option::Debug` is not correct, we can change it at `Option`

AllanZyne:
Could you please explain why this change is needed?

source/loader/layers/sanitizer/asan_interceptor.cpp
yingcong-wu: PR #1878 is merged recently, you can see if that fix the problem you are trying to fix with this change.

aarongreig:
> Could you please explain why not retain device handle references in sanitizer layer?

Following the PI->UR port we started seeing fails in a handful of AddressSanitizer e2e tests that were kind of similar to the issue https://github.com/oneapi-src/unified-runtime/pull/1879 is trying to solve: the sycl RT unloads the adapter and tears down the loader before the `SanitizerInterceptor` destructor is called, so by the time device references were released [here](https://github.com/oneapi-src/unified-runtime/blob/907f1542c7498b2115c3eff8286dbd09e50c9457/source/loader/layers/sanitizer/asan_interceptor.hpp#L56) the adapter was no longer around to properly release the device. My initial approach to this was to add reference counting for all the devices saved in [m_DeviceMap](https://github.com/oneapi-src/unified-runtime/blob/907f1542c7498b2115c3eff8286dbd09e50c9457/source/loader/layers/sanitizer/asan_interceptor.hpp#L248) but that quickly got complicated. I figured that since the sanitizer layer only uses devices while the application still considers them valid (e.g. during [memory allocations](https://github.com/oneapi-src/unified-runtime/blob/907f1542c7498b2115c3eff8286dbd09e50c9457/source/loader/layers/sanitizer/asan_interceptor.cpp#L177), [kernel launches](https://github.com/oneapi-src/unified-runtime/blob/907f1542c7498b2115c3eff8286dbd09e50c9457/source/loader/layers/sanitizer/asan_interceptor.cpp#L337) etc.) we can get away without artificially extending the lifetime of these handles and solve the SYCL RT/UR destruction order problems in a simpler fashion.

aarongreig:
it's possible that https://github.com/oneapi-src/unified-runtime/pull/1879 could also solve this, although in its current state that PR actually causes all the AddressSanitizer e2e tests to fail on the PI2UR-main branch

pbalcer:
> the sycl RT unloads the adapter and tears down the loader before the SanitizerInterceptor destructor is called

This should not be happening. The layer objects are teardown'ed inside of [`urLoaderTeardown`](https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_lib.cpp#L210), prior to the loader's own context (which includes handle maps). Unless the sanitizer state is stored somewhere in a global variable. In which case the fix is to get rid of the global state.

The loader does have one hack that causes `getContext()` to actually re-initialize all the state if its called after the loader has been already destroyed. This is to prevent misbehaving applications from crashing, but leaking data is expected in such cases. But in that case, the SanitizerInterceptor should never have been called.

pbalcer (approved):
I think I get it now. The adapter is destroyed before all device handles are released...

We need to start running tests with full validation layer enabled. It would have caught it.

aarongreig:
yeah I double checked and it's literally just that these release calls happen when the layers are (correctly) torn down in `urLoaderTearDown`, which necessarily happens after the `urAdapterRelease` call renders them invalid

aarongreig:
the multi device fail is unrelated

AllanZyne (approved):
I think this fix can be treated as a hotfix, but not final solution.
Can you add a comment in "struct DeviceInfo" to remind me about this?
Thanks.

aarongreig:
> I made an issue for this, so please add
> 
> "Fixes: [intel/llvm#14825](https://github.com/intel/llvm/issues/14825)" to the commit description, thanks

I've added it here https://github.com/intel/llvm/pull/14719 because that change will re-enable the tests and actually land the fix

AllanZyne:
Hi @pbalcer, please merge this PR. 
This bugfix is needed for 2025.0.

zhaomaosu (approved):
lgtm

PR #1886, opened 2024-07-24
[L0] Use zesInit for SysMan API usage

nrspruit:
- Change to using zesInit and zes data structures for accessing L0
  SysMan functionality.
- Updated Platform & Devices to store zes handles if sysman support is
  available.
- Given Legacy Environment Variable from user, then fallback to
  old functionality.
- Fixed Return code on error to be consistently unsupported enumeration.

### DIFF
--- a/source/adapters/level_zero/adapter.cpp
+++ b/source/adapters/level_zero/adapter.cpp
@@ -35,19 +35,41 @@ public:
   ~ur_legacy_sink() = default;
 };
 
-ur_result_t initPlatforms(PlatformVec &platforms) noexcept try {
+ur_result_t initPlatforms(PlatformVec &platforms,
+                          ze_result_t ZesResult) noexcept try {
   uint32_t ZeDriverCount = 0;
   ZE2UR_CALL(zeDriverGet, (&ZeDriverCount, nullptr));
   if (ZeDriverCount == 0) {
     return UR_RESULT_SUCCESS;
   }
 
+  uint32_t ZesDriverCount = 0;
+  ze_result_t zesDriverGetResult = ZE_RESULT_ERROR_UNINITIALIZED;
+  if (ZesResult == ZE_RESULT_SUCCESS) {
+    zesDriverGetResult =
+        ZE_CALL_NOCHECK(zesDriverGet, (&ZesDriverCount, nullptr));
+  }
+
   std::vector<ze_driver_handle_t> ZeDrivers;
   ZeDrivers.resize(ZeDriverCount);
 
+  std::vector<zes_driver_handle_t> ZesDrivers;
+  ZesDrivers.resize(ZesDriverCount);
+
   ZE2UR_CALL(zeDriverGet, (&ZeDriverCount, ZeDrivers.data()));
+  if (zesDriverGetResult == ZE_RESULT_SUCCESS && ZesDriverCount > 0) {
+    ZE_CALL_NOCHECK(zesDriverGet, (&ZesDriverCount, ZesDrivers.data()));
+  }
   for (uint32_t I = 0; I < ZeDriverCount; ++I) {
+    zes_driver_handle_t zesDriverHandle = nullptr;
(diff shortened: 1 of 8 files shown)


source/adapters/level_zero/adapter.cpp
```diff
@@ -35,19 +35,41 @@ class ur_legacy_sink : public logger::Sink {
...
+ur_result_t initPlatforms(PlatformVec &platforms,
+                          ze_result_t ZesResult) noexcept try {
   uint32_t ZeDriverCount = 0;
   ZE2UR_CALL(zeDriverGet, (&ZeDriverCount, nullptr));
   if (ZeDriverCount == 0) {
     return UR_RESULT_SUCCESS;
   }
 
+  uint32_t ZesDriverCount = 0;
+  ze_result_t zesDriverGetResult = ZE_RESULT_ERROR_UNINITIALIZED;
```
aravindksg, line 47: zesinit/zesDriverGet was introduced from v1.5 onwards IIRC, so do we need a way to check version here before call zesDriverGet?

source/adapters/level_zero/adapter.cpp
```diff
@@ -35,19 +35,41 @@ class ur_legacy_sink : public logger::Sink {
...
+  if (zesDriverGetResult == ZE_RESULT_SUCCESS && ZesDriverCount > 0) {
+    ZE_CALL_NOCHECK(zesDriverGet, (&ZesDriverCount, ZesDrivers.data()));
+  }
   for (uint32_t I = 0; I < ZeDriverCount; ++I) {
+    zes_driver_handle_t zesDriverHandle = nullptr;
+    // If the number of ZE Drivers is greater than the number of ZES Drivers,
+    // then a null sysman driver is associated. The index of the ZE and ZES
+    // driver should match.
+    if (I < ZesDrivers.size()) {
+      zesDriverHandle = ZesDrivers[I];
```
aravindksg, line 69: hmm how can we be sure the drivers are enumerated in same order? but having said that i see core driver_properties at least provide a a driver_uuid but Sysman does not have that.

source/adapters/level_zero/platform.cpp
```diff
@@ -304,14 +304,30 @@ ur_result_t ur_platform_handle_t_::populateDeviceCacheIfNeeded() {
 
   uint32_t ZeDeviceCount = 0;
   ZE2UR_CALL(zeDeviceGet, (ZeDriver, &ZeDeviceCount, nullptr));
+  uint32_t ZesDeviceCount = 0;
+  if (ZesDriver) {
+    ZE2UR_CALL(zesDeviceGet, (ZesDriver, &ZesDeviceCount, nullptr));
```
aravindksg, line 309: we might again need version check before calling zesDeviceGet to make sure we have support?

source/adapters/level_zero/platform.cpp
```diff
@@ -304,14 +304,30 @@ ur_result_t ur_platform_handle_t_::populateDeviceCacheIfNeeded() {
...
+      ZE2UR_CALL(zesDeviceGet, (ZesDriver, &ZesDeviceCount, ZesDevices.data()));
+    }
 
     for (uint32_t I = 0; I < ZeDeviceCount; ++I) {
+      zes_device_handle_t zesDeviceHandle = nullptr;
+      // If the number of ZE Devices is greater than the number of ZES Devices,
+      // then a null sysman device is associated. The index of the ZE and ZES
+      // devices should match.
+      if (I < ZesDevices.size()) {
+        zesDeviceHandle = ZesDevices[I];
```
aravindksg, line 326: do we need to match uuid here between core and sysman devices before association here? or you are not doing that for now since it was a very recent change in UMD?

nrspruit:
replaced by https://github.com/oneapi-src/unified-runtime/pull/2017 which itself is awaiting a patch for dynamic dependency loading.

PR #1893, opened 2024-07-24
[L0 v2] event improvements

igchor:


pbalcer (commented):
@igchor you made more changes to this code in #1894. do you still want this to be merged separately?

igchor:
> @igchor you made more changes to this code in #1894. do you still want this to be merged separately?

I think we could merge it separately, extra changes in the other PR are due to renaming and implementation of release() and retain() but that should be in a separate commit anywy

source/adapters/level_zero/v2/event_provider_normal.cpp
```diff
@@ -37,7 +42,7 @@ provider_pool::provider_pool(ur_context_handle_t context,
   }
 
   ZE2UR_CALL_THROWS(zeEventPoolCreate,
-                    (context->ZeContext, &desc, 1,
+                    (context->hContext, &desc, 1,
```
pbalcer, line 45: I think we should have getters for ze handles. But we can do that later.

source/adapters/level_zero/v2/event.cpp
```diff
@@ -7,28 +7,49 @@
...
+    return UR_RESULT_SUCCESS;
 
-} // namespace v2
+  pool->free(this);
+  RefCount.increment();
+
+  return UR_RESULT_SUCCESS;
+}
+
+UR_APIEXPORT ur_result_t UR_APICALL urEventRetain(ur_event_handle_t hEvent) {
```
pbalcer, line 49: can you add a simple MT test that creates and destroys events in a loop? Ideally with at least one scenario where events are created in one thread and then released in another.

pbalcer: nvm, I realized we don't have any entry point that would allow you create an event. let's do this later.

igchor: Sure, I think once we add a MT submit_kernel benchmark we can cover this.

source/adapters/level_zero/v2/event_provider_normal.cpp
igchor: Yes, agree. I actually have the getters implemented already on a branch with enqueue implementation because we need them to be able to reuse some code from legacy adapter (like program.cpp)

source/adapters/level_zero/v2/event.cpp
```diff
@@ -7,28 +7,49 @@
...
+  return hEvent->retain();
+}
+
+UR_APIEXPORT ur_result_t UR_APICALL urEventRelease(ur_event_handle_t hEvent) {
+  return hEvent->release();
```
pbalcer, line 54: This is not MT-safe. The hEvent needs to store a queue, not the pool, and then grab the queue lock to access its pool. In debug we could add a check to make sure that the queue's pool is exactly the same as the event's pool.

I know we can't implement that right now, so let's just leave this as TODO (no need to push another revision of PR :P I don't want to waste time on CI).

igchor: Right, event_pool is not thread safe. I'll make a note

.github/workflows/cmake.yml
```diff
@@ -140,6 +140,12 @@ jobs:
     with:
       name: L0
 
+  level-zero-v2:
+    name: Level Zero V2
+    uses: ./.github/workflows/build-hw-reusable.yml
+    with:
+      name: L0_V2
```
pbalcer, line 147: this will not run, there are no such runners.

igchor: Right, done.

EwanC (commented):
It looks like there are two changes in this PR, an update to GitHub workflows to run the V2 tests, and a change to the v2 event implementation. Should these be separate PRs?

EwanC (approved):
command-buffer match file changes LGTM

PR #1896, opened 2024-07-25
Map ur_bool_t to cl_bool in sampler getinfo for opencl adapter

omarahmed1111:
[INTEL/llvm](https://github.com/intel/llvm/pull/14769)

### DIFF
--- a/source/adapters/opencl/sampler.cpp
+++ b/source/adapters/opencl/sampler.cpp
@@ -158,16 +158,38 @@ urSamplerGetInfo(ur_sampler_handle_t hSampler, ur_sampler_info_t propName,
   static_assert(sizeof(cl_addressing_mode) ==
                 sizeof(ur_sampler_addressing_mode_t));
 
-  size_t CheckPropSize = 0;
-  ur_result_t Err = mapCLErrorToUR(
-      clGetSamplerInfo(cl_adapter::cast<cl_sampler>(hSampler), SamplerInfo,
-                       propSize, pPropValue, &CheckPropSize));
-  if (pPropValue && CheckPropSize != propSize) {
-    return UR_RESULT_ERROR_INVALID_SIZE;
-  }
-  UR_RETURN_ON_FAILURE(Err);
-  if (pPropSizeRet) {
-    *pPropSizeRet = CheckPropSize;
+  ur_result_t Err = UR_RESULT_SUCCESS;
+  // ur_bool_t have a size of uint8_t, but cl_bool size have the size of
+  // uint32_t so this adjust UR_SAMPLER_INFO_NORMALIZED_COORDS info to map
+  // between them.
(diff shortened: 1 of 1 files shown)


source/adapters/opencl/sampler.cpp
```diff
@@ -159,6 +159,12 @@ urSamplerGetInfo(ur_sampler_handle_t hSampler, ur_sampler_info_t propName,
                 sizeof(ur_sampler_addressing_mode_t));
 
   size_t CheckPropSize = 0;
+  // ur_bool_t have a size of uint8_t, but cl_bool size have the size of
+  // uint32_t so this makes a mapping between them.
+  if (propName == UR_SAMPLER_INFO_NORMALIZED_COORDS &&
+      propSize == sizeof(ur_bool_t)) {
+    propSize = sizeof(cl_bool);
```
aarongreig, line 166: I think we want an entirely different path for `INFO_NORMALIZED_COORDS`, this is passing `pPropValue` directly into `clGetSamplerInfo` and returning the property size we get from CL directly in `pPropSizeRet`. We should really be passing a `cl_bool*` to the CL api and converting it to `ur_bool_t` to be returned, as well as returning `sizeof(ur_bool_t)` for `pPropSizeRet`

omarahmed1111: Good point, I changed that. however, not sure if I handled it in the most efficient way (redundant code I mean) so let me know if I could optimize this code more.

source/adapters/opencl/sampler.cpp
```diff
@@ -158,16 +158,37 @@ urSamplerGetInfo(ur_sampler_handle_t hSampler, ur_sampler_info_t propName,
...
+    cl_bool normalized_coords = false;
+    Err = mapCLErrorToUR(
+        clGetSamplerInfo(cl_adapter::cast<cl_sampler>(hSampler), SamplerInfo,
+                         sizeof(cl_bool), &normalized_coords, nullptr));
+    if (pPropValue && propSize != sizeof(ur_bool_t)) {
+      return UR_RESULT_ERROR_INVALID_SIZE;
+    }
+    UR_RETURN_ON_FAILURE(Err);
+    if (pPropValue) {
+      *static_cast<ur_bool_t *>(pPropValue) = (ur_bool_t)normalized_coords;
```
aarongreig, line 175:
```suggestion
      *static_cast<ur_bool_t *>(pPropValue) = static_cast<ur_bool_t>(normalized_coords);
```

source/adapters/opencl/sampler.cpp
aarongreig: nah looks good, we have similar code to deal with edge cases elsewhere in the CL adapter

source/adapters/opencl/sampler.cpp
omarahmed1111: Fixed

aarongreig (approved):

PR #1898, opened 2024-07-25
[RFC] Fixing dangling pointers caused by urCommandBufferRelease

ianayl:
Some functions in the UR leaves dangling pointers, [such as `commandBufferReleaseInternal`, which is called by `urCommandBufferReleaseExp`](https://github.com/oneapi-src/unified-runtime/blob/3d8fe8d298cec8db624fc230fa5c0e19865aa6f1/source/adapters/opencl/command_buffer.cpp#L21). Coverity has detected that on intel/llvm, when [`urCommandBufferReleaseExp` is called from the pi](https://github.com/intel/llvm/blob/4329b0c5f0f72f13aa72753f25578aed27525136/sycl/plugins/unified_runtime/pi2ur.hpp#L4631) with tracing enabled, dangling pointers left by `urCommandBufferReleaseExp` are used again [here](https://github.com/intel/llvm/blob/4329b0c5f0f72f13aa72753f25578aed27525136/sycl/source/detail/plugin.hpp#L222) (specifically, the `pi_ext_command_buffer`/`ur_exp_command_buffer_handle_t` fed into `Args`, which was already freed in `urCommandBufferRelease`).

The best course of action to fix this Coverity hit, to the best of my knowledge, is to fix the dangling pointers left by e.g. `urCommandBufferReleaseExp`. This PR contains a potential fix by taking references of `ur_exp_command_buffer_handle_t` when calling `urCommandBufferReleases` instead, and setting the pointers to `nullptr` after they are deleted. This would allow code using `urCommandBufferReleaseExp` to actually catch that the command buffer has been freed.

Unfortunately, this requires the API to be changed, as I've changed the function signature of `urCommandBufferReleaseExp`. I'm also aware that changing only `urCommandBufferRelease` may result in asymmetry within the API. Thus, if this change was to be adopted, more functions may also have to be changed in a similar manner. Thus, I am opening this PR as an RFC to see what other people think of this, and to see if there are any potential issues with this approach.

Additionally, if anyone has better ideas for fixing/handling this issue/Coverity hit without changing the UR API, please feel free to let me know. From what I can tell, this is the course of action that'll result in the least amount of problems in the future.

The respective draft DPC++ testing PR is here: https://github.com/intel/llvm/pull/14782

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -8314,7 +8314,7 @@ urCommandBufferRetainExp(
 ///     - ::UR_RESULT_ERROR_OUT_OF_HOST_MEMORY
 UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferReleaseExp(
-    ur_exp_command_buffer_handle_t hCommandBuffer ///< [in][release] Handle of the command-buffer object.
+    ur_exp_command_buffer_handle_t &hCommandBuffer ///< [in][release] Handle of the command-buffer object.
 );
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -699,11 +699,12 @@ urCommandBufferRetainExp(ur_exp_command_buffer_handle_t CommandBuffer) {
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL
-urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t CommandBuffer) {
+urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t &CommandBuffer) {
   if (!CommandBuffer->RefCount.decrementAndTest())
     return UR_RESULT_SUCCESS;
 
   delete CommandBuffer;
+  CommandBuffer = nullptr;
   return UR_RESULT_SUCCESS;
 }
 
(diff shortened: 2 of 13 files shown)


include/ur_api.h
```diff
@@ -8314,7 +8314,7 @@ urCommandBufferRetainExp(
 ///     - ::UR_RESULT_ERROR_OUT_OF_HOST_MEMORY
 UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferReleaseExp(
-    ur_exp_command_buffer_handle_t hCommandBuffer ///< [in][release] Handle of the command-buffer object.
+    ur_exp_command_buffer_handle_t &hCommandBuffer ///< [in][release] Handle of the command-buffer object.
```
kbenzie, line 8317: This header is a C API, there can not be use of C++ references in it. Additionally, this header is generated from a spec description in yaml files which reside in `scripts/core/*.yml`.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -699,11 +699,12 @@ urCommandBufferRetainExp(ur_exp_command_buffer_handle_t CommandBuffer) {
 }
 
 UR_APIEXPORT ur_result_t UR_APICALL
-urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t CommandBuffer) {
+urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t &CommandBuffer) {
   if (!CommandBuffer->RefCount.decrementAndTest())
     return UR_RESULT_SUCCESS;
 
   delete CommandBuffer;
+  CommandBuffer = nullptr;
```
kbenzie, line 707: It is not the responsibility of the level zero adapter to zero out the handle pointer as part of the `urCommandBufferReleaseExp` entry point. The proper fix for this should be in the layer above which is calling this entry point.

ianayl:
With the removal of the PI layer in https://github.com/intel/llvm/pull/14145, there would no longer be a need for such mechanism, as it should now be actually possible for intel/llvm to check for the result of `urCommandBufferReleaseExp` directly. Thus, I am closing this. Thanks for your time and patience regardless!

PR #1906, opened 2024-07-26
[L0] Add check for Intel Flex/Arc for disabling use of copy engines.

nrspruit:


source/adapters/level_zero/command_buffer.cpp
```diff
@@ -922,6 +922,13 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferAppendUSMMemcpyExp(
...
   bool PreferCopyEngine = !IsDevicePointer(CommandBuffer->Context, Src) ||
                           !IsDevicePointer(CommandBuffer->Context, Dst);
+  // For better performance, Copy Engines are not preferred given Shared
+  // pointers on DG2.
+  if ((IsSharedPointer(CommandBuffer->Context, Src) ||
+       IsSharedPointer(CommandBuffer->Context, Dst)) &&
+      CommandBuffer->Device->isDG2()) {
+    PreferCopyEngine = false;
+  }
   PreferCopyEngine |= UseCopyEngineForD2DCopy;
```
johntfalk, line 933: So this "UseCopyEngineForD2DCopy" will not get set if logic for IsSharedPointer, above is set?   Also is the variable "isDG2" also true for ATS-M?

nrspruit: UseCopyEngineForD2DCopy is only set from an environment variable. It is a "debug" variable that can be used to override the defaults we are setting above. We need to allow that env to set the value if it is set since it is for "testing" this type of use case for performance measurement. It is default false. 

isDG2 is for all Arc and ATS-M products.

johntfalk: Thanks.   Everything looks good.

johntfalk (approved):

EwanC (approved):

source/adapters/level_zero/memory.cpp
```diff
@@ -1390,6 +1410,13 @@ ur_result_t ur_queue_handle_legacy_t_::enqueueUSMMemcpy2D(
   // (versus compute engine).
   bool PreferCopyEngine = !IsDevicePointer(Queue->Context, Src) ||
                           !IsDevicePointer(Queue->Context, Dst);
+  // For better performance, Copy Engines are not preferred given Shared
+  // pointers on DG2.
+  if ((IsSharedPointer(Queue->Context, Src) ||
+       IsSharedPointer(Queue->Context, Dst)) &&
+      Queue->Device->isDG2()) {
```
pbalcer, line 1417: The order of these conditions should be reversed. `Queue->Device->isDG2()` is just reading a variable (quick), whereas `IsSharedPointer` needs to call into the driver (how expensive is this operation). I'd be worried about impacting performance on non-DG2 systems.

nrspruit: it is a very small operation, but I see your point with regards to the check for DG2 first, then the pointer type. We already incur that cost in the previous line for isDevicePointer, but we can reorder to get some time back on other platforms.

pbalcer: Hm, ok, if you think it's OK the way it is, we can leave it as is for now (to save time on re-spinning CI), and revisit it later.

nrspruit: Updated in the recent commit.

pbalcer (approved):

PR #1909, opened 2024-07-30
[L0] create abstraction for ur_event_handle_t

igchor:
to allow switching the implementation to v2::ur_event_handle_t. This is similar to how ur_queue_handle_t behaves right now.


pbalcer (approved):
not the biggest fan of having to do this, but I don't see a better option...

igchor:
Alternative approach (not fully working yet): https://github.com/oneapi-src/unified-runtime/pull/1913

igchor:
Replaced by https://github.com/oneapi-src/unified-runtime/pull/1913

PR #1910, opened 2024-07-31
[CUDA][HIP] Improve command-buffer sync points

EwanC:
Several improvements to sync-point implementation in HIP and CUDA command-buffer adapters with additional CTS coverage to back it up.

* In the CUDA/HIP adapters we assume that there is always a return sync-point passed by the user. However, this is not required by the UR API, so we should check that the return value is non-null before dereferencing.
* The Fill helper function is can implement as fill as several commands for certain pattern sizes, we were creating a sync point for every internal command. This is not required, these commands from a linear dependency chain, so only the leaf command is required to be a sync point for future commands to depend on.
* Remove `shared_ptr` from `CUgraphNode` objects stored for sync-points. `CUgraphNode` is a pointer type, and is managed by the CUDA driver runtime rather than us.
* Simplify handling of return results. We don't always use the helper macro for returning the `ur_result_t` value no a function call fail, and also often unnecessarily use a variable to store return code.
* Use `hipMemcpyDefault` rather than ` hipMemcpyHostToHost` for [hipMemcpyKind](https://cgmb-hip.readthedocs.io/en/sphinx/api/index.html#c.hipMemcpyKind) parameter to `hipGraphAddMemcpyNode1D` used to implement `urCommandBufferAppendUSMMemcpyExp`. This is because the UR entry-point can take device or host parameters for the src/dst inputs. CTS test added in this PR fail on HIP without this change.
* Remove returned adapter specific error from prefetch & advise, since we are still doing the work of adding an empty node to the graph. An error implies that no work has been done. 

DPC++ PR https://github.com/intel/llvm/pull/14858


source/adapters/cuda/command_buffer.cpp
```diff
@@ -74,9 +74,9 @@ ur_exp_command_buffer_handle_t_::~ur_exp_command_buffer_handle_t_() {
 ur_exp_command_buffer_command_handle_t_::
     ur_exp_command_buffer_command_handle_t_(
         ur_exp_command_buffer_handle_t CommandBuffer, ur_kernel_handle_t Kernel,
-        std::shared_ptr<CUgraphNode> &&Node, CUDA_KERNEL_NODE_PARAMS Params,
-        uint32_t WorkDim, const size_t *GlobalWorkOffsetPtr,
-        const size_t *GlobalWorkSizePtr, const size_t *LocalWorkSizePtr)
+        CUgraphNode Node, CUDA_KERNEL_NODE_PARAMS Params, uint32_t WorkDim,
+        const size_t *GlobalWorkOffsetPtr, const size_t *GlobalWorkSizePtr,
+        const size_t *LocalWorkSizePtr)
     : CommandBuffer(CommandBuffer), Kernel(Kernel), Node{std::move(Node)},
```
EwanC, line 80:
```suggestion
    : CommandBuffer(CommandBuffer), Kernel(Kernel), Node(Node),
```

source/adapters/hip/command_buffer.cpp
```diff
@@ -76,9 +76,9 @@ ur_exp_command_buffer_handle_t_::~ur_exp_command_buffer_handle_t_() {
 ur_exp_command_buffer_command_handle_t_::
     ur_exp_command_buffer_command_handle_t_(
         ur_exp_command_buffer_handle_t CommandBuffer, ur_kernel_handle_t Kernel,
-        std::shared_ptr<hipGraphNode_t> &&Node, hipKernelNodeParams Params,
-        uint32_t WorkDim, const size_t *GlobalWorkOffsetPtr,
-        const size_t *GlobalWorkSizePtr, const size_t *LocalWorkSizePtr)
+        hipGraphNode_t Node, hipKernelNodeParams Params, uint32_t WorkDim,
+        const size_t *GlobalWorkOffsetPtr, const size_t *GlobalWorkSizePtr,
+        const size_t *LocalWorkSizePtr)
     : CommandBuffer(CommandBuffer), Kernel(Kernel), Node(std::move(Node)),
```
EwanC, line 82:
```suggestion
    : CommandBuffer(CommandBuffer), Kernel(Kernel), Node(Node),
```

PR #1911, opened 2024-07-31
[CUDA] Don't import XPTI symbols in the plugin library

ProGTX:
All other uses of XPTI also use the static library. This fixes compilation issues with `dllimport/dllexport` on Windows.


jchlanda (approved):
LGTM

ProGTX:
Note that I moved the macro definition into an existing `target_compile_definitions` instead of using a new one, and removed the leading `-D`, as CMake inserts that automatically.

PR #1894, opened 2024-07-24
[L0 v2] initial implementation of queueFinish and enqueueKernelLaunch

igchor:

PR #1912, opened 2024-07-31
[common] Histogram-based latency tracker

igchor:
This uses HdrHistrogram which is part of the public domain: https://github.com/HdrHistogram/HdrHistogram_c/blob/main/LICENSE.txt

The overhead of hdr_histogram seems to be acceptable: ~10ns for adding the value to histogram and few microseconds for creating the histogram (which happens once for every thread and for every measured scope). The creation overhead can be even smaller (a few hundred ns) if we tweak the configuration options for the histogram (at the cost of lower precision).

Sample output (run on https://github.com/oneapi-src/unified-runtime/pull/1894):

```

Latency histogram:
name,min,max,mean,stdev,p50,p90,p99,p99.9,p99.99,p99.999,p99.9999,unit
command_list_cache_t::getImmediateCommandList,88320,289791,189008,100656,88383,289791,289791,289791,289791,289791,289791,ns
event_pool::allocate,24,67711,28,214,28,29,31,36,50,2467,67711,ns
event_pool::free,25,2435,29,7,29,31,32,36,59,140,2435,ns
provider_normal::allocate,45,68287,1856,10381,48,55,52863,68287,68287,68287,68287,ns
ur_queue_immediate_in_order_t::enqueueKernelLaunch,1838,723455,2025,2297,2004,2107,2373,3647,9959,69759,723455,ns

```

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -20,6 +20,7 @@ jobs:
         compiler: [{c: gcc, cxx: g++}]
         libbacktrace: ['-DVAL_USE_LIBBACKTRACE_BACKTRACE=OFF']
         pool_tracking: ['-DUMF_ENABLE_POOL_TRACKING=ON', '-DUMF_ENABLE_POOL_TRACKING=OFF']
+        latency_tracking: ['-DUMF_ENABLE_LATENCY_TRACKING=OFF']
         include:
           - os: 'ubuntu-22.04'
             build_type: Release
@@ -36,6 +37,10 @@ jobs:
           - os: 'ubuntu-20.04'
             build_type: Release
             compiler: {c: gcc-7, cxx: g++-7}
+          - os: 'ubuntu-22.04'
+            build_type: Release
+            compiler: {c: clang, cxx: clang++}
+            latency_tracking: '-DUMF_ENABLE_LATENCY_TRACKING=ON'
     runs-on: ${{ (matrix.os == 'ubuntu-22.04' && github.repository_owner == 'oneapi-src') && 'intel-ubuntu-22.04' || matrix.os }}
 
     steps:
@@ -92,6 +97,7 @@ jobs:
         -DUR_DPCXX=${{github.workspace}}/dpcpp_compiler/bin/clang++
         ${{matrix.libbacktrace}}
         ${{matrix.pool_tracking}}
+        ${{matrix.latency_tracking}}
 
     - name: Configure CMake
       if: matrix.os == 'ubuntu-20.04'
@@ -106,6 +112,7 @@ jobs:
         -DUR_FORMAT_CPP_STYLE=ON
         ${{matrix.libbacktrace}}
         ${{matrix.pool_tracking}}
+        ${{matrix.latency_tracking}}
 
     - name: Generate source from spec, check for uncommitted diff
       if: matrix.os == 'ubuntu-22.04'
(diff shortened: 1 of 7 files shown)


source/adapters/level_zero/v2/latency_tracker.hpp
```diff
@@ -0,0 +1,212 @@
+//===--------- ur_latency_tracker.cpp - common ---------------------------===//
```
pbalcer, line 1: can we move this to `ur_common`? It might be useful for other adapters or even the loader.

source/adapters/level_zero/v2/latency_tracker.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+#include "logger/ur_logger.hpp"
+
+#if defined(UR_ENABLE_LATENCY_HISTOGRAM)
+
+#include <hdr/hdr_histogram.h>
+
+namespace v2 {
+
+static inline bool trackLatency = []() {
```
pbalcer, line 26: since we have the logger, can we control whether we actually collect based on whether the logger verbosity is set high enough (info)?
One less env variable to set :)

source/adapters/level_zero/v2/latency_tracker.hpp
```diff
@@ -0,0 +1,212 @@
...
+  std::chrono::time_point<std::chrono::steady_clock> begin;
+};
+
+// To resolve __COUNTER__
+#define CONCAT(a, b) a##b
+
+// Each tracker has it's own thread-local histogram.
+// At program exit, all histograms for the same scope are
+// aggregated.
+#define TRACK_SCOPE_LATENCY_CNT(name, cnt)                                     \
```
pbalcer, line 201: just a suggestion: you could make the name optional and use `__FUNCTION__`.

pbalcer (commented):
awesome, this is going to be very useful.

source/adapters/level_zero/v2/latency_tracker.hpp
igchor: Sure. Done.

source/adapters/level_zero/v2/latency_tracker.hpp
igchor: Good idea, done.

source/adapters/level_zero/v2/latency_tracker.hpp
igchor: Yeah, I thought about this, but when you're using this on member methods, __FUNCTION__ doesn't tell you the class name (and we might have the same methods' names) so I figures it would be better to just pass a name always

pbalcer: `__PRETTY_FUNCTION__` ? Google says it should include the class name, but we'd need a workaround for visual studio:
```
#if !defined(__PRETTY_FUNCTION__) && !defined(__GNUC__)
#define __PRETTY_FUNCTION__ __FUNCSIG__
#endif
```

igchor: Pretty function on the other hand prints too much imho ;d It includes the return type and all the arguments, so instead of `command_list_cache_t::getImmediateCommandList` we would have `raii::cache_borrowed_command_list_t
command_list_cache_t::getImmediateCommandList(
    ze_device_handle_t, bool, uint32_t,
    ze_command_queue_mode_t, ze_command_queue_priority_t,
    std::optional<uint32_t>)`

pbalcer: makes sense.

PR #1913, opened 2024-07-31
[L0 v2] Make L0 v2 implementation a seperate adapter

igchor:
Initially, L0 v2 adapter was supposed to reside in a separate namespace but be a part of legacy L0 adapter (with runtime option to switch between executing on legacy or v2). However, this turns out to require a lot of changes in the legacy code to allow for function dispatching to legacy/v2 implementations of queue, event, etc.

This approach allows us to keep the implementations separate while still resuing files when appropriate (e.g. for adapter.cpp or platform.cpp).


source/adapters/level_zero/v2/api.cpp
```diff
@@ -0,0 +1,1105 @@
...
+#include "../common.hpp"
+#include "logger/ur_logger.hpp"
+
+std::mutex ZeCall::GlobalLock;
+
+ur_result_t UR_APICALL
+urContextCreate(uint32_t DeviceCount, const ur_device_handle_t *phDevices,
+                const ur_context_properties_t *pProperties,
+                ur_context_handle_t *phContext) {
+  logger::error(logger::LegacyMessage("[UR][L0] {} function not implemented!"),
```
pbalcer, line 26: please remove the `logger::LegacyMessage`. This type only exists for compatibility with old PI-style logs.

source/adapters/level_zero/v2/queue_api.cpp
```diff
@@ -0,0 +1,321 @@
...
+ * Exceptions. See LICENSE.TXT
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ * @file queue_api.cpp
+ *
+ */
+
+#include "queue_api.hpp"
+
+UR_APIEXPORT ur_result_t UR_APICALL urQueueGetInfo(ur_queue_handle_t hQueue,
```
pbalcer, line 15: With no need to support the legacy path together with the new one, we might be able to significantly simplify the way we do dispatching, like we originally planned to. But we can revisit that later.

source/adapters/level_zero/queue.cpp
```diff
@@ -24,8 +24,6 @@
 #include "ur_util.hpp"
 #include "ze_api.h"
```
pbalcer, line 25: will this allow for all the `queue_legacy` to be reverted?

source/adapters/level_zero/v2/api.cpp
igchor: Done.

source/adapters/level_zero/queue.cpp
igchor: Yes, I have a revert commit ready but it might be better to merge it separately: https://github.com/oneapi-src/unified-runtime/pull/1919

CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ option(UR_BUILD_ADAPTER_CUDA "Build the CUDA adapter" OFF)
 option(UR_BUILD_ADAPTER_HIP "Build the HIP adapter" OFF)
 option(UR_BUILD_ADAPTER_NATIVE_CPU "Build the Native-CPU adapter" OFF)
 option(UR_BUILD_ADAPTER_ALL "Build all currently supported adapters" OFF)
+option(UR_BUILD_ADAPTER_L0_V2 "Build the (experimental) Level-Zero v2 adapter" OFF)
```
igchor, line 54: Not sure if we should build `UR_BUILD_ADAPTER_L0_V2` if `UR_BUILD_ADAPTER_ALL` is set - it is pretty quite far from fully functional at this moment.

igchor:
One consequence of this approach is that we will need to split some of the existing functions in legacy adapter (or at least move them to a separate file and possibly make them a templates) if we still want to reuse the logic in v2 adapter. For example, for kernel, there are 3 functions that would be nice to reuse as showed here: https://github.com/oneapi-src/unified-runtime/commit/0aac25597c99b88f9e441c7f9e892a5b760fbb60

CMakeLists.txt
pbalcer: No, let's leave it off.

EwanC (approved):

pbalcer (approved):

PR #1917, opened 2024-08-01
Initalize mock callbacks earlier

kbenzie:
Avoid use after static destruction in sycl unittests by moving the initialization of `mock::callbacks` from static function scope to static global scope.

### DIFF
--- a/source/mock/ur_mock_helpers.cpp
+++ b/source/mock/ur_mock_helpers.cpp
@@ -13,10 +13,8 @@
 #include "ur_mock_helpers.hpp"
 
 namespace mock {
+static callbacks_t callbacks = {};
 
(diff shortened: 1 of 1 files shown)


callumfare (approved):

pbalcer (approved):
if we can accept adding "callbacksTeardown" api or something like that, one solution to this problem might also be using the `AtomicSingleton` class. Otherwise lgtm.

kbenzie:
> if we can accept adding "callbacksTeardown" api or something like that, one solution to this problem might also be using the `AtomicSingleton` class. Otherwise lgtm.

I've been looking for a way to use a global environment in the LLVM/SYCL unittests but the all seem to be using the default one so I'm not sure where I'd invoke that.

PR #1923, opened 2024-08-02
[L0] Return the build log on compilation failure

sarnex:
Right now we never return the build log for L0 on build failure because we only add `ZeBuildLog` to `ZeBuildLogMap` if `ZeResult` is `ZE_RESULT_SUCCESS`.

I was running a SYCL program that didn't build and noticed the the log was printed for OCL but not L0, so I debugged into why L0 wasn't and found this.

I manually confirmed this fixes it.

### DIFF
--- a/source/adapters/level_zero/program.cpp
+++ b/source/adapters/level_zero/program.cpp
@@ -209,8 +209,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramBuildExp(
         }
       }
       hProgram->ZeModuleMap.insert(std::make_pair(ZeDevice, ZeModuleHandle));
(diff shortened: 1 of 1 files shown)


sarnex:
https://github.com/intel/llvm/pull/14934

sarnex:
@kbenzie Any idea what's going on with the CI here? There are failed/cancelled jobs that I don't understand.  If I'm support to restart it, I don't seem to have permission. Thanks

sarnex:
CI fails here all seem infra related and not related to this PR.

Testing in intel/llvm passed, so I think this is ready for review.

pbalcer (approved):

nrspruit (approved):

PR #1924, opened 2024-08-02
Add support for command-buffer kernel updates

fabiomestre:
- Updates the specification to add support for command-buffer kernel handle updates.
- Adds new UR tests for this feature.
- Adds an implementation for the Cuda and Hip adapters.
- Changes the `UR_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP` enum to a new  `UR_DEVICE_INFO_COMMAND_BUFFER_UPDATE_CAPABILITIES_EXP` which uses a bitfield instead of a boolean.
- Changes the spec for `urCommandBufferUpdateKernelLaunchExp` and `ur_exp_command_buffer_update_kernel_launch_desc_t` to make it more intuitive with less complicated errors codes:
  - Passing a non-nullptr local work-group will now always update the command to use that value.
  - Passing a nullptr local work-group will now always keep the current command behaviour (either generated by the implementation or user-defined).
  - Passing zero to newWorkDim is now an error.

intel/llvm PR: https://github.com/intel/llvm/pull/15287


scripts/core/exp-command-buffer.yml
```diff
@@ -307,6 +312,14 @@ params:
...
       name: pLocalWorkSize
       desc: "[in][optional] Local work size to use when executing kernel."
+    - type: uint32_t
+      name: "numKernelAlternatives"
+      desc: "[in] The number of kernel alternatives provided in pKernelAlternatives."
+    - type: $x_kernel_handle_t*
+      name: "phKernelAlternatives"
+      desc: |
+            [in][optional][range(0, numKernelAlternatives)] List of kernels handles that might be used to update the kernel in this 
+            command after the command-buffer is finalized.
```
EwanC, line 322: Rather than adding the list of kernels as a parameter, I think we should have a new descriptor parameter to the entry-point and then specify the kernel list as part of that. It gives more extensibility without API breakages.

To do that we'll need to first define something like:
```
name: $x_exp_command_buffer_kernel_command_desc_t
base: $x_base_desc_t
members:
    - type: uint32_t
      name: numKernelAlternatives
      desc: foo
    - type: $x_kernel_handle_t*
      name: phKernelAlternatives
      desc: "bar"
```

Then the new param would be like `const ur_exp_command_buffer_kernel_command_desc_t* pKernelDesc`

scripts/core/EXP-COMMAND-BUFFER.rst
EwanC, line 1: I think the "Updating Command-Buffer Commands" section should be updated with to document the new functionality, also add an entry to the changelog at the bottom and put yourself in "Contributors"

scripts/core/exp-command-buffer.yml
```diff
@@ -307,6 +312,14 @@ params:
     - type: "const size_t*"
       name: pLocalWorkSize
       desc: "[in][optional] Local work size to use when executing kernel."
+    - type: uint32_t
+      name: "numKernelAlternatives"
+      desc: "[in] The number of kernel alternatives provided in pKernelAlternatives."
```
EwanC, line 317:
```suggestion
      desc: "[in] The number of kernel alternatives provided in phKernelAlternatives."
```

scripts/core/EXP-COMMAND-BUFFER.rst
```diff
@@ -173,7 +173,14 @@ ${X}_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP query.
 Updating kernel commands is done by passing the new kernel configuration
 to ${x}CommandBufferUpdateKernelLaunchExp along with the command handle of
 the kernel command to update. Configurations that can be changed are the
-parameters to the kernel and the execution ND-Range.
+kernel handle, the parameters to the kernel and the execution ND-Range.
+
+To update the kernel handle, it is necessary to first register the new
+kernel handle before the command-buffer is finalized. This can be done
```
Bensuo, line 179: Nit: I think this could be clearer that they must be registered/provided in the call to add the kernel command to the command-buffer, it reads a bit ambiguously here to me saying it's just before finalization.

Bensuo (approved):

scripts/core/EXP-COMMAND-BUFFER.rst
fabiomestre: I updated the wording a bit and removed the finalize mention. Hopefully it reads less ambiguously :crossed_fingers:

test/conformance/exp_enqueue_native/CMakeLists.txt
EwanC, line 1: Would revert these changes to reduce scope of PR diff

scripts/core/exp-command-buffer.yml
```diff
@@ -170,6 +170,11 @@ desc: "Descriptor type for updating a kernel launch command."
 base: $x_base_desc_t
 name: $x_exp_command_buffer_update_kernel_launch_desc_t
 members:
+    - type: $x_kernel_handle_t
+      name: hNewKernel
+      desc: |
+        [in] The new kernel handle. If this value is equal to the current kernel handle associated
```
EwanC, line 176: It doesn't look like `hNewKernel` is optional. I'm not sure if we discussed whether it should be to allow devices that don't support kernel update to still be able to update other things.

The  `COMMAND_BUFFER_UPDATE_SUPPORT_EXP`   isn't fine-grained enough to represent individual aspects of mutability. Technically in this PR every adapter other than CUDA could report false for this now as they don't support kernel binary update, but I'm not sure if we actually want to do that as it will reduce test coverage for other aspects of update.

source/adapters/cuda/command_buffer.cpp
```diff
@@ -862,11 +879,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferUpdateKernelLaunchExp(
     return UR_RESULT_ERROR_INVALID_OPERATION;
   }
 
-  if (auto NewWorkDim = pUpdateKernelLaunch->newWorkDim) {
-    // Error if work dim changes
-    if (NewWorkDim != hCommand->WorkDim) {
-      return UR_RESULT_ERROR_INVALID_OPERATION;
```
EwanC, line 868: we don't need to check for this error anymore?

source/adapters/hip/command_buffer.cpp
EwanC, line 1: The CUDA adapter changes seem *relatively* straightforward, think it's worth having a quick go at trying to do the equivalents changes in HIP? If it turns out anything unexpected comes up then we can just abandon that scope creep.

test/conformance/exp_command_buffer/fixtures.h
```diff
@@ -139,19 +102,65 @@ struct urCommandBufferExpExecutionTest : uur::urKernelExecutionTest {
...
+        ur_exp_command_buffer_desc_t desc{
+            UR_STRUCTURE_TYPE_EXP_COMMAND_BUFFER_DESC, nullptr, true};
+
+        ASSERT_SUCCESS(urCommandBufferCreateExp(context, device, &desc,
+                                                &updatable_cmd_buf_handle));
+        ASSERT_NE(updatable_cmd_buf_handle, nullptr);
+
+        // Currently there are synchronization issue with immediate submission when used for command buffers.
+        // So, create queue with batched submission for this test suite if the backend is Level Zero.
+        if (backend == UR_PLATFORM_BACKEND_LEVEL_ZERO) {
```
EwanC, line 124: Noting https://github.com/oneapi-src/unified-runtime/pull/2046 that I don't think we need this

test/conformance/exp_command_buffer/update/kernel_handle_update.cpp
```diff
@@ -0,0 +1,347 @@
...
+        // Index 1 is A
+        ASSERT_SUCCESS(urKernelSetArgValue(Kernel, 1, sizeof(A), nullptr, &A));
+        // Index 2 is X
+        ASSERT_SUCCESS(urKernelSetArgPointer(Kernel, 2, nullptr, Memory[1]));
+        // Index 3 is Y
+        ASSERT_SUCCESS(urKernelSetArgPointer(Kernel, 3, nullptr, Memory[2]));
+    }
+
+    void destroyKernel() override {
+        for (auto &shared_ptr : Memory) {
```
EwanC, line 97: nitpick - there's a few places for here on where the variable naming convention changes
```suggestion
        for (auto &SharedPtr : Memory) {
```

scripts/core/EXP-COMMAND-BUFFER.rst
```diff
@@ -173,7 +173,14 @@ ${X}_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP query.
 Updating kernel commands is done by passing the new kernel configuration
 to ${x}CommandBufferUpdateKernelLaunchExp along with the command handle of
 the kernel command to update. Configurations that can be changed are the
-parameters to the kernel and the execution ND-Range.
+kernel handle, the parameters to the kernel and the execution ND-Range.
+
+Kernel handles that might be used to update the kernel of a command, need
```
EwanC, line 178: Is the original `hkernel` kernel an implicit alternative. i.e. if you updated back from the original kernel to an alternative, then back to original, would that still work? Might be worth adding a sentence to the spec on as well as a test

test/conformance/exp_enqueue_native/CMakeLists.txt
fabiomestre: Yeah, will do. My intention was to remove all the changes to this file (previously it was also fixing a bug that has since been reverted in UR)

scripts/core/exp-command-buffer.yml
fabiomestre: I think the reduction in test coverage would be a problem. I think that if we are going to have an implementation for other adapters soon, it would make sense to not change the aspect. 

I'm curious about why OpenCL is hardcoded to false? There seems to be an implementation of urCommandBufferUpdateKernelLaunchExp in the opencl adapter

source/adapters/hip/command_buffer.cpp
fabiomestre: Makes sense, will give it a try on hip :+1:

scripts/core/exp-command-buffer.yml
fabiomestre: Regarding `hNewKernel` not being optional: I wasn't sure whether to make it optional or not when writing the spec.

 In the end I decided to make it non-optional and require the user to explicitly pass the kernel handle that is currenty active if they only want to change the parameters. To me passing the kernel handle instead of nullptr seems a bit more consistent but happy to discuss this.

scripts/core/EXP-COMMAND-BUFFER.rst
fabiomestre: Yes, should still work. Will update the spec to make this behavior more clear :+1:

scripts/core/exp-command-buffer.yml
```diff
@@ -199,7 +226,7 @@ members:
       desc: "[in][optional][range(0, newWorkDim)] Array of newWorkDim unsigned values that describe the number of global work-items."
     - type: "size_t*"
       name: pNewLocalWorkSize
-      desc: "[in][optional][range(0, newWorkDim)] Array of newWorkDim unsigned values that describe the number of work-items that make up a work-group. If newWorkDim is non-zero and pNewLocalWorkSize is nullptr, then runtime implementation will choose the work-group size. If newWorkDim is zero and pNewLocalWorkSize is nullptr, then the local work size is unchanged."
+      desc: "[in][optional][range(0, newWorkDim)] Array of newWorkDim unsigned values that describe the number of work-items that make up a work-group. If pNewLocalWorkSize is nullptr, then the local work size is unchanged."
```
fabiomestre, line 229: As discussed on the standup, I'm unsure about what is the most intuitive behaviour for local work-size. It used to be that:

- `Work-dim == 0 && pNewLocalWorkSize == nullptr` - No update
- `Work-dim > 0 && pNewLocalWorkSize == nullptr` - Implementation defined local work-dim
- `Work-dim > 0 && pNewLocalWorkSize != nullptr` - User defined local work-dim

To make the api less confusing after the kernl binary update changes, I made it an error to have work-dim == 0. 
So the new behaviour is:

- `Work-dim == 0 && pNewLocalWorkSize == nullptr` - Error
- `Work-dim > 0 && pNewGlobalWorkSize != nullptr && pNewLocalWorkSize == nullptr` - Implementation defined local work-dim
- `Work-dim > 0 &&  && pNewGlobalWorkSize == nullptr  && pNewLocalWorkSize == nullptr` - No update
- `Work-dim > 0 && pNewLocalWorkSize != nullptr` - User defined local work-dim

We are losing the possibility of not updating the global work size and allowing the implementation to generate a local worksize. This seems like a very niche usecase. So maybe that's okay?

To me, this function has too many optional behaviours which makes it confusing. Anyway, would welcome feedback about which behaviour is more intuitive. Can go back to allowing dim == 0 if we think that's for the best. However that will add a few more verbose errors in the spec.

Bensuo (approved):

scripts/core/exp-command-buffer.yml
fabiomestre: Another option that was discussed is to allow all parameters (including newWorkDim) to be optional.

Essentially the implementation wouldn't do as many sanity checks on what the user provides. For example, if the user wants to change the work dimensions of a kernel node, that would be allowed. 

This approach could have the benefit of allowing the specification to be more consistent and easier to interpret due to less edge cases that need to return error codes.

- A point that still needs to be discussed is whether it can make sense to update the global work size without updating the local work-size (and vice-versa).
- Also unsure if allowing implementation defined local-work sizes on updates has a benefit.

EwanC: Thinking in terms of the sycl usage, there is `update_nd_range` where there is both local & global sizes, and `update_range` which is only global size. There is no way to set the local size without global size too.

If a user does `update_range()` I think it would be good to get a new implementation defined local work size. The user doesn't have the means to change this themselves. i.e. prefer option

```quote 
    newWorkDim == 0 && pNewLocalWorkSize == nullptr - Error
    newWorkDim > 0 && pNewGlobalWorkSize != nullptr && pNewLocalWorkSize == nullptr - Implementation defined local work-dim
    newWorkDim > 0 &&  pNewGlobalWorkSize == nullptr  && pNewLocalWorkSize == nullptr - No update
    newWorkDim > 0 && pNewLocalWorkSize != nullptr - User defined local work-dim
```


> Another option that was discussed is to allow all parameters (including newWorkDim) to be optional.

I'm not sure there is less edge cases, just edge cases in different places. e.g. what would the new value of nd-range be if you increased of decreased the work dim without updating that. How many elements would there be in the nd-range arrays if you didn't have to pass a work dim. I think these are solvable, but I think the least confusing API is one where everything is mandatory (apart from local work size, which can be nullptr for sycl::range kernels and stays null for all updates).

```quote 
    newWorkDim == 0 - Error
   newWorkDim > 0 &&  pNewGlobalWorkSize == nullptr - Error (global size not explicitly set)
   newWorkDim > 0 && pNewGlobalWorkSize != nullptr && pNewLocalWorkSize == nullptr - Implementation defined local work-dim
   newWorkDim > 0 && pNewGlobalWorkSize != nullptr  && pNewLocalWorkSize != nullptr - User defined local work-dim
```

fabiomestre: As discussed offline, I updated the spec to go mostly for the option that Ewan suggested with the exception that updating dimensions is also allowed as long as the global work size and the global work offset are updated at the same time

scripts/core/exp-command-buffer.yml
```diff
@@ -909,25 +948,24 @@ params:
...
+        - "`pUpdateKernelLaunch->pNewLocalWorkSize != NULL && pUpdateKernelLaunch->pNewGlobalWorkSize == NULL`"
+        - "If `pUpdateKernellaunch->hNewKernel` is equal to the currently active kernel in `hCommand`, and `pUpdateKernellaunch->newWorkDim` is different from the work-dim currently associated with `hCommand`."
     - $X_RESULT_ERROR_INVALID_COMMAND_BUFFER_COMMAND_HANDLE_EXP
     - $X_RESULT_ERROR_INVALID_MEM_OBJECT
     - $X_RESULT_ERROR_INVALID_KERNEL_ARGUMENT_INDEX
     - $X_RESULT_ERROR_INVALID_KERNEL_ARGUMENT_SIZE
     - $X_RESULT_ERROR_INVALID_ENUMERATION
-    - $X_RESULT_ERROR_INVALID_WORK_DIMENSION
+    - $X_RESULT_ERROR_INVALID_WORK_DIMENSION:
+        - "`pUpdateKernelLaunch->newWorkDim < 0 || pUpdateKernelLaunch->newWorkDim > 3`"
```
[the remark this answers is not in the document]
fabiomestre: Updated so that newWorkDim == 0 is an error now

source/adapters/opencl/common.cpp
```diff
@@ -129,21 +131,37 @@ cl_int deviceSupportsURCommandBufferKernelUpdate(cl_device_id Dev,
...
-      CL_MUTABLE_DISPATCH_EXEC_INFO_KHR;
-  Result = (mutable_capabilities & required_caps) == required_caps;
+      sizeof(MutableCapabilities), &MutableCapabilities, nullptr));
+
+  if (!(MutableCapabilities & CL_MUTABLE_DISPATCH_EXEC_INFO_KHR)) {
+    UpdateCapabilities = 0;
+    return CL_SUCCESS;
+  }
+
+  UpdateCapabilities = 0;
```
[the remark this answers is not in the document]
fabiomestre: Done :+1:

source/adapters/cuda/command_buffer.cpp
fabiomestre: This has been changed due to the discussions in https://github.com/oneapi-src/unified-runtime/pull/1924#discussion_r1754031968 . The check is now for when the workdim is changed, the global work size and global work offset are also updated.

scripts/core/exp-command-buffer.yml
fabiomestre: I went back on my opinion about this and decided to keep `kNewKernel` but make it optional (as Ewan had suggested before). It just seemed more consistent after the recent change to make the api more flexible with less error conditions.

Note: `hNewKernel` is not a pointer to a handle since it seems to be a common pattern UR to have optional handles that are not marked as pointers. For example: `urMemGetNativeHandle` and `urUSMHostAlloc`

EwanC (approved):

source/adapters/level_zero/device.cpp
```diff
@@ -994,20 +994,30 @@ ur_result_t urDeviceGetInfo(
...
+    if (supportsFlags(ZE_MUTABLE_COMMAND_EXP_FLAG_KERNEL_ARGUMENTS)) {
+      UpdateCapabilities |=
+          UR_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_KERNEL_ARGUMENTS;
+    }
+    if (supportsFlags(ZE_MUTABLE_COMMAND_EXP_FLAG_GROUP_COUNT |
+                      ZE_MUTABLE_COMMAND_EXP_FLAG_GROUP_SIZE)) {
+      UpdateCapabilities |=
+          UR_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_GLOBAL_WORK_SIZE |
+          UR_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_LOCAL_WORK_SIZE;
+    }
```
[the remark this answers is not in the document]
fabiomestre: I think that having this properties separated might cause issues in certain scenarions. For example:
When updating the global work size, we might also want to generate a local work size. If I mapped GlobalWorkSize -> ZE_MUTABLE_COMMAND_EXP_FLAG_GROUP_COUNT and LocalWorkSize -> ZE_MUTABLE_COMMAND_EXP_FLAG_GROUP_SIZE  this behaviour wouldn't work.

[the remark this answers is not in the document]
EwanC: That makes sense, could you stick that rationale in a quick code comment.

Bensuo (approved):

fabiomestre:
This PR should be ready for reviewing:
@oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-opencl-write @oneapi-src/unified-runtime-hip-write @oneapi-src/unified-runtime-cuda-write @oneapi-src/unified-runtime-level-zero-write @oneapi-src/unified-runtime-maintain

PietroGhg (approved):
Native CPU lgtm, thank you

source/adapters/opencl/device.cpp
```diff
@@ -1085,12 +1085,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     return ReturnValue(ExtStr.find("cl_khr_command_buffer") !=
                        std::string::npos);
   }
-  case UR_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP: {
+  case UR_DEVICE_INFO_COMMAND_BUFFER_UPDATE_CAPABILITIES_EXP: {
     cl_device_id Dev = cl_adapter::cast<cl_device_id>(hDevice);
-    bool Supported = false;
+    ur_device_command_buffer_update_capability_flags_t UpdateCapabilities;
```
aarongreig, line 1090:
```suggestion
    ur_device_command_buffer_update_capability_flags_t UpdateCapabilities = 0;
```

aarongreig (approved):

JackAKirk (approved):
cuda/hip lgtm

nrspruit (approved):

PR #1880, opened 2024-07-22
[L0] L0 impl for enqueue native command

hdelan:
Implement urEnqueueNativeCommandExp for the Level Zero adapter.

PR #1929, opened 2024-08-05
Revert "[L0] L0 impl for enqueue native command"

omarahmed1111:
Reverts oneapi-src/unified-runtime#1880


aarongreig (approved):

omarahmed1111:
@oneapi-src/unified-runtime-level-zero-v2-write This changes caused event CTS test failures described [here](https://github.com/intel/llvm/pull/14701#issuecomment-2268851766) So, reverting that.

pbalcer (approved):
This is a very isolated change that shouldn't have any side-effects, so I'm surprised that something broke after this. The only thing that comes to mind is the thread-local mechanism for native queue handles. But I don't see any obvious way this might impact something that doesn't use the native command enqueue.
@hdelan any ideas?

omarahmed1111:
> This is a very isolated change that shouldn't have any side-effects, so I'm surprised that something broke after this. The only thing that comes to mind is the thread-local mechanism for native queue handles. But I don't see any obvious way this might impact something that doesn't use the native command enqueue. @hdelan any ideas?

I will recheck that maybe I missed something in my configs

PR #1931, opened 2024-08-05
[L0] Enable L0 Driver Init to init all device types with filtering

nrspruit:
- Change zeInit usage to optionally init all driver types with all flags set, then
  filter the drivers currently supported.
- Enables for L0 application to init all driver types which unblocks
  other libraries using L0 drivers and enables for future use of other
driver types.
- Added Environment Variable UR_L0_INIT_ALL_DRIVERS which will set all
  driver type flags when enabled.

### DIFF
--- a/source/adapters/level_zero/adapter.cpp
+++ b/source/adapters/level_zero/adapter.cpp
@@ -43,15 +43,31 @@ ur_result_t initPlatforms(PlatformVec &platforms) noexcept try {
   }
 
   std::vector<ze_driver_handle_t> ZeDrivers;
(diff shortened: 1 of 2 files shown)


raiyanla (approved):

winstonzhang-intel (approved):

nrspruit:
Pending if changes can be made in other L0 products first to ensure avoiding possible breakage.

rwmcguir (approved):
I totally approve, and really like the environ we can pass around for components to test their compatibility and future issues.

PR #1936, opened 2024-08-06
Fix unintended crash in catch blocks.

isaacault:
Ensure that handles are default initialised and checked before attempting to destroy.

DPC++ PR: [ [Bindless] Fix unintended crash in catch blocks #14966 ](https://github.com/intel/llvm/pull/14966)

### DIFF
--- a/source/adapters/cuda/image.cpp
+++ b/source/adapters/cuda/image.cpp
@@ -461,12 +461,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageAllocateExp(
       UR_CHECK_ERROR(cuArray3DCreate(&ImageArray, &array_desc));
       *phImageMem = (ur_exp_image_mem_native_handle_t)ImageArray;
     } catch (ur_result_t Err) {
-      if (ImageArray) {
+      if (ImageArray != CUarray{}) {
         UR_CHECK_ERROR(cuArrayDestroy(ImageArray));
       }
       return Err;
     } catch (...) {
-      if (ImageArray) {
+      if (ImageArray != CUarray{}) {
         UR_CHECK_ERROR(cuArrayDestroy(ImageArray));
       }
       return UR_RESULT_ERROR_UNKNOWN;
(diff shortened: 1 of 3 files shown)


source/adapters/cuda/image.cpp
```diff
@@ -455,32 +455,40 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageAllocateExp(
...
   if (pImageDesc->numMipLevel == 1) {
-    CUarray ImageArray;
+    CUarray ImageArray{};
 
     try {
       UR_CHECK_ERROR(cuArray3DCreate(&ImageArray, &array_desc));
       *phImageMem = (ur_exp_image_mem_native_handle_t)ImageArray;
     } catch (ur_result_t Err) {
-      cuArrayDestroy(ImageArray);
+      if (ImageArray) {
```
hdelan, line 464: I'm sure it's OK to do as you suggest but just in case `CUarray` is not aliased to a ptr type. I think this should be applied for all of the checks

```suggestion
      if (ImageArray != CUarray{}) {
```

isaacault: Good point! Applied. With the hip adapter the handle is declared as a pointer but I added it nonetheless for completeness if the code is ever copied for use where that safety isn't guaranteed.

source/adapters/cuda/image.cpp
```diff
@@ -455,32 +455,40 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageAllocateExp(
...
-      cuArrayDestroy(ImageArray);
+      if (ImageArray != CUarray{}) {
+        UR_CHECK_ERROR(cuArrayDestroy(ImageArray));
+      }
       return UR_RESULT_ERROR_UNKNOWN;
     }
   } else // Allocate a cuMipmappedArray
   {
-    CUmipmappedArray mip_array;
+    CUmipmappedArray mip_array{};
```
hdelan, line 476: Nit: can we use CamelCase here like the other local vars in this file?

isaacault: Although that would seem the obvious match to the above `ImageArray`, this entire file is lacking consistency and clarity on the convention used. This is being tracked internally by our team for future clean up and I'd rather leave this to be addressed then.

hdelan: Sure thing.

hdelan (approved):

DBDuncan (approved):

Seanst98 (approved):

cppchedy (approved):

PR #1940, opened 2024-08-06
[XPTI] Use `ur.call` rather than `ur` in XPTI

RossBrunton:
Closes: #14922


RossBrunton:
Requires DPCPP changes from https://github.com/intel/llvm/pull/14971

RossBrunton:
Should be ready to merge now - the failing E2E test is because there are required intel/llvm changes.

PR #1943, opened 2024-08-06
Fix various Coverity defects

kbenzie:
- Fix 1574354 Uninitialized scalar field
- Fix 1594027 Uncaught exception
- Fix 1595568, 1595570 Use of auto that causes a copy
- Fix 1595594 Copy instead of move
- Fix 1595785 Use of auto that causes a copy
- Fix 1594597 Dereference after null check
- Fix 1595225 Data race condition
- Fix 1598473 Resource leak
- Fix 14 instances of Resource leak
  - 1594026, 1594028, 1594029, 1594030, 1594031, 1594032, 1594033, 1594034,
    1594035, 1594036, 1594037, 1595372, 1595373, and 1598546

### DIFF
--- a/source/adapters/cuda/memory.hpp
+++ b/source/adapters/cuda/memory.hpp
@@ -409,10 +409,14 @@ struct ur_mem_handle_t_ {
   }
 
   ur_result_t clear() {
-    if (isBuffer()) {
-      return std::get<BufferMem>(Mem).clear();
+    try {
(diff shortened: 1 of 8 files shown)


steffenlarsen (approved):
Looks reasonable! 👍

pbalcer (approved):

pbalcer:
The e2e failures are expected.

PR #1948, opened 2024-08-07
[CMake] Add flag to disable fuzztesting

RossBrunton:
Default is enabled; this should not be a visible change without changing your cmake configuration.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -12,6 +12,7 @@ FetchContent_Declare(
 
 set(UR_TEST_DEVICES_COUNT 1 CACHE STRING "Count of devices on which conformance and adapters tests will be run")
 set(UR_TEST_PLATFORMS_COUNT 1 CACHE STRING "Count of platforms on which conformance and adapters tests will be run")
+set(UR_TEST_FUZZTESTS ON CACHE BOOL "Run fuzz tests if using clang and UR_DPCXX is specified")
 # For Windows: Prevent overriding the parent project's compiler/linker settings
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 FetchContent_MakeAvailable(googletest)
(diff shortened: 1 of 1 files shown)


kbenzie (approved):

pbalcer:
Why?

Also, please set this in workflow jobs (there are two - nightly and a normal one).

RossBrunton:
> Why?
> 
> Also, please set this in workflow jobs (there are two - nightly and a normal one).

The compiler I use to build LLVM is a clang without the fuzzing headers available, so I constantly have to comment out the include. Figured I'd just quickly add a variable and push it up in case it was useful to anyone else.

Given that it's enabled by default, does it need to be added to all the cmake invocations?

test/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ FetchContent_Declare(
 
 set(UR_TEST_DEVICES_COUNT 1 CACHE STRING "Count of devices on which conformance and adapters tests will be run")
 set(UR_TEST_PLATFORMS_COUNT 1 CACHE STRING "Count of platforms on which conformance and adapters tests will be run")
+set(UR_TEST_FUZZTESTS ON CACHE STRING "Run fuzz tests if using clang and UR_DPCXX is specified")
```
lukaszstolarczuk, line 15: I think it should be `BOOL` not `STRING`

pbalcer:
> > Why?
> > Also, please set this in workflow jobs (there are two - nightly and a normal one).
> 
> The compiler I use to build LLVM is a clang without the fuzzing headers available, so I constantly have to comment out the include. Figured I'd just quickly add a variable and push it up in case it was useful to anyone else.
> 
> Given that it's enabled by default, does it need to be added to all the cmake invocations?

Nvm, I thought you disabled it by default. lgtm.

PR #1952, opened 2024-08-08
Make USM parameter bounds checking configurable

kbenzie:
Bounds checking can be an expensive operation to perform, especially for applications with heavy use of commands operating on USM allocations.

This patch adds a new layer specifically for USM bounds checking called UR_LAYER_BOUNDS_CHECKING. This was previously part of UR_LAYER_PARAMETER_VALIDATION.

By removing USM bounds checking from the default parameter validation the overhead of using parameter validation greatly decreases allowing it to be enabled in more situations.

### DIFF
--- a/scripts/core/INTRO.rst
+++ b/scripts/core/INTRO.rst
@@ -295,6 +295,8 @@ Layers currently included with the runtime are as follows:
      - Description
    * - UR_LAYER_PARAMETER_VALIDATION
      - Enables non-adapter-specific parameter validation (e.g. checking for null values).
+   * - UR_LAYER_BOUNDS_CHECKING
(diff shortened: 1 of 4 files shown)


pbalcer (approved):
I've added some info about how to update the baseline: https://github.com/oneapi-src/unified-runtime/pull/1954

tl;dr: after this is merged, please run the compute benchmarks job with PR 0, and parameter `--save baseline`.

kbenzie:
> I've added some info about how to update the baseline: #1954
> 
> tl;dr: after this is merged, please run the compute benchmarks job with PR 0, and parameter `--save baseline`.

Done: https://github.com/oneapi-src/unified-runtime/actions/runs/10370029936

PR #1953, opened 2024-08-09
Change urDeviceCreateWithNativeHandle to take an adapter handle.

aarongreig:
It currently takes a platform handle, which is problematic for the sycl RT because its make_device api only takes a native handle, so to figure out the correct platform handle to pass at best we'd need to do some backend specific querying of the native object, but even then that isn't always possible as not all backends have a platform equivalent.

The platform handle does currently enable an optional (slightly) faster path to return the correct device in some adapter implementations but this isn't essential for them to work, so really its primary purpose is to serve as the wrapped UR handle for the loader to work. This purpose is equally well served by an adapter handle, which will also be a lot easier for the sycl rt to correctly provide.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -2062,7 +2062,7 @@ typedef struct ur_device_native_properties_t {
 ///     - ::UR_RESULT_ERROR_DEVICE_LOST
 ///     - ::UR_RESULT_ERROR_ADAPTER_SPECIFIC
 ///     - ::UR_RESULT_ERROR_INVALID_NULL_HANDLE
-///         + `NULL == hPlatform`
+///         + `NULL == hAdapter`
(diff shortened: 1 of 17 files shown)


aarongreig:
LLVM PR https://github.com/intel/llvm/pull/15023 sycl-e2e jobs here will fail due to missing accompanying rt change

aarongreig:
ping @oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-cuda-write @oneapi-src/unified-runtime-hip-write @oneapi-src/unified-runtime-level-zero-write @oneapi-src/unified-runtime-maintain this is necessary a fix for the release

PietroGhg (approved):

konradkusiak97 (approved):

kbenzie (approved):

PR #1955, opened 2024-08-09
[L0] fix Coverity issues for L0 Adapter

nrspruit:
- Fixed recast of Queue object vs usage of variable.
- fixed usm alignment check such that == 0 is correctly skipped from buffer overrun checks.
- fixed cast to show no need to check the return from decrement and test in queue finish.
- fixed release of resources for bindless import on error.

### DIFF
--- a/source/adapters/level_zero/queue.cpp
+++ b/source/adapters/level_zero/queue.cpp
@@ -599,7 +599,7 @@ ur_result_t ur_queue_handle_legacy_t_::queueRelease() {
       // internal reference count. When the External Reference count == 0, then
       // cleanup of the queue begins and the final decrement of the internal
       // reference count is completed.
-      Queue->RefCount.decrementAndTest();
+      static_cast<void>(Queue->RefCount.decrementAndTest());
       return UR_RESULT_SUCCESS;
     }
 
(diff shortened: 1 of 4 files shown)


source/adapters/level_zero/queue.cpp
```diff
@@ -599,7 +599,7 @@ ur_result_t ur_queue_handle_legacy_t_::queueRelease() {
       // internal reference count. When the External Reference count == 0, then
       // cleanup of the queue begins and the final decrement of the internal
       // reference count is completed.
-      Queue->RefCount.decrementAndTest();
+      static_cast<void>(Queue->RefCount.decrementAndTest());
```
igchor, line 602: Perhaps we should assert that the return value is 0 here?

igchor (approved):

source/adapters/level_zero/queue.cpp
nrspruit: No, because that would imply the refcnt "must" have been ==1 , which is not the case.

What occurs here is that the ExternalRef != 0 after being decremented. This means that the internal refcnt also needs to be decremented, but it may not be 0 so the refcnt being decremented, but not 0 is not a failure.

nrspruit: In this case we don't care that the refcnt is not 0 at this time, we just need to decrement the atomic refcnt.

isaacault (approved):

isaacault:
Tagging @wenju-he to test this.

cppchedy (approved):

wenju-he:
> Tagging @wenju-he to test this.

A subset of vulkan/dx12 interop tests are passing on DG2 Windows.
This PR has no regression.

PR #1956, opened 2024-08-09
[L0] Make destructor of the context virtual

igchor:
to make sure that v2::context will release all the resources on context destruction.

This patch is not applicable to main branch as we already removed the inheritance there.


nrspruit (approved):

omarahmed1111:
@igchor Could you please create an Intel/llvm PR for this? Thanks!

igchor:
> @igchor Could you please create an Intel/llvm PR for this? Thanks!

This PR only targets v0.10.0rc as the problem is already fixed on main (by making L0 v2 a separate adapter) and intel/llvm won't compile with the 0.10.0

kbenzie:
I've cherry-picked this onto v0.10.0rc

PR #1957, opened 2024-08-09
[L0] make device id an optional

igchor:
Since device id is initialized later (not in ctor), make it optional to avoid using uninitialized value.


nrspruit (approved):

omarahmed1111:
Could you please create a testing PR on intel/llvm for this to merge it? Thanks!

PR #1958, opened 2024-08-10
[L0] move kernel helper functions to a separate file

igchor:
to avoid code duplication between kernel.cpp and command_buffer.cpp and to enable code reuse with v2 adapter.


igchor:
I moved the function to a separate file instead of leaving them in kernel.cpp because L0 v2 adapter will have it's own implementation of the kernel and only those functions will be shared.

kswiecicki (approved):

igchor:
@oneapi-src/unified-runtime-command-buffer-write Could you please take a look at the changes?

EwanC (approved):

PR #1951, opened 2024-08-07
[L0] Enable Immediate Command List by default given Intel DG2

nrspruit:
- Enabled Immediate Command list usage per queue given Intel DG2 HW.
- Removed default setting of false on windows.

PR #1960, opened 2024-08-12
Command-buffer immediate command-list CTS fix

EwanC:
As discovered in https://github.com/oneapi-src/unified-runtime/pull/1951#issuecomment-2278519730 there is an issue running the command-buffer CTS tests on L0 with command-lists enabled.

After investigating, this was because the CTS tests we're not using the output event parameter to command-buffer enqueue. In the L0 adapter code a path for registering an event with the queue to waiting on the workload was guarded by this event being set. Therefore `urQueueFinish` was not working as expected, as the queue was not aware of this work to wait on.

Fixed by always registering the work to wait on with the queue, and miss out on propagation of the created event when user doesn't pass an output event, rather than not creating it at all.

DPC++ PR https://github.com/intel/llvm/pull/15035


nrspruit (approved):
LGTM, thank you for the fix!

PR #1964, opened 2024-08-12
Report correct extension string for devicelib assert in cuda + hip.

aarongreig:
It seems that the extension string SYCL RT checks for was changed by this commit https://github.com/intel/llvm/commit/0939f39818225ce3e469e08f6a45711b449a8ad4 in PI, but the change never made it to the UR adapters.


aarongreig:
Adapters reporting extension strings in general is something I think we should abolish but that's a larger change, this at least fixes the behaviour for now.

npmiller (approved):
LGTM

PR #1966, opened 2024-08-13
[L0] Fix incorrect profiling values for in-order queue barriers

raiyanla:
This change corresponds to the fix made in LLVM https://github.com/intel/llvm/pull/14123

With this change, the SYCL E2E test located at sycl/test-e2e/Regression/in_order_barrier_profiling.cpp will now pass.


nrspruit:
@raiyanla , please file a intel/llvm PR using this commit in UR to enable the pre-commit CI.

nrspruit (approved):

PR #1967, opened 2024-08-13
Improved wording of setKernelArgValue

coldav:
This identifies that the data pointed to needs to be copied, similar to clSetKerneLArg when it is by value.


aarongreig (approved):

coldav:
> you'll need to run `generate` to propogate the .yml change to the headers etc. @coldav

I don't see `generate` - is it `generate_docs.py`?

PR #1968, opened 2024-08-13
Add new launch property to support work_group_scratch_memory

Naghasan:
https://github.com/intel/llvm/pull/15061 introduce a new property `work_group_scratch_size` which allow the user to set a given amount of local memory to be used.

In order to pass this information to the adaptor, the patch adds a new launch property to `urEnqueueKernelLaunchCustomExp`.

The patch also changes the signature of `urEnqueueKernelLaunchCustomExp` to add `global offset` in order to maintain features when using this extension.

### DIFF
--- a/scripts/core/exp-launch-properties.yml
+++ b/scripts/core/exp-launch-properties.yml
@@ -29,6 +29,8 @@ etors:
       desc: "Whether to launch a cooperative kernel"
     - name: CLUSTER_DIMENSION
       desc: "work-group cluster dimensions"
+    - name: WORK_GROUP_MEMORY
+      desc: "Implicit work group memory allocation"
 --- #--------------------------------------------------------------------------
 type: union
 desc: "Specifies a launch property value"
@@ -45,6 +47,10 @@ members:
       name: cooperative
       desc: "[in] non-zero value indicates a cooperative kernel"
       tag: $X_EXP_LAUNCH_PROPERTY_ID_COOPERATIVE
+    - type: size_t
+      name: workgroup_mem_size
+      desc: "[in] non-zero value indicates the amount of work group memory to allocate in bytes"
+      tag: $X_EXP_LAUNCH_PROPERTY_ID_WORK_GROUP_MEMORY
 --- #--------------------------------------------------------------------------
 type: struct
 desc: "Kernel launch property"
@@ -82,6 +88,9 @@ params:
     - type: uint32_t
       name: workDim
       desc: "[in] number of dimensions, from 1 to 3, to specify the global and work-group work-items"
+    - type: "const size_t*"
+      name: pGlobalWorkOffset
+      desc: "[in] pointer to an array of workDim unsigned values that specify the offset used to calculate the global ID of a work-item"
     - type: const size_t*
       name: pGlobalWorkSize
       desc: "[in] pointer to an array of workDim unsigned values that specify the number of global work-items in workDim that will execute the kernel function"
@@ -97,10 +106,10 @@ params:
     - type: uint32_t
       name: numEventsInWaitList
       desc: "[in] size of the event wait list"
-    - type: const ur_event_handle_t*
+    - type: const $x_event_handle_t*
       name: phEventWaitList
       desc: "[in][optional][range(0, numEventsInWaitList)] pointer to a list of events that must be complete before the kernel execution. If nullptr, the numEventsInWaitList must be 0, indicating that no wait event. "
-    - type: ur_event_handle_t*
+    - type: $x_event_handle_t*
       name: phEvent
       desc: "[out][optional] return an event object that identifies this particular kernel execution instance. If phEventWaitList and phEvent are not NULL, phEvent must not refer to an element of the phEventWaitList array."
 returns:
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -422,11 +422,13 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueEventsWait(
                                         phEventWaitList, phEvent);
 }
 
-UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
-    ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel, uint32_t workDim,
-    const size_t *pGlobalWorkOffset, const size_t *pGlobalWorkSize,
-    const size_t *pLocalWorkSize, uint32_t numEventsInWaitList,
-    const ur_event_handle_t *phEventWaitList, ur_event_handle_t *phEvent) {
+static ur_result_t
+enqueueKernelLaunch(ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel,
+                    uint32_t workDim, const size_t *pGlobalWorkOffset,
+                    const size_t *pGlobalWorkSize, const size_t *pLocalWorkSize,
+                    uint32_t numEventsInWaitList,
+                    const ur_event_handle_t *phEventWaitList,
+                    ur_event_handle_t *phEvent, size_t WorkGroupMemory) {
   // Preconditions
   UR_ASSERT(hQueue->getDevice() == hKernel->getProgram()->getDevice(),
             UR_RESULT_ERROR_INVALID_KERNEL);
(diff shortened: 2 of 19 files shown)


source/adapters/level_zero/queue.cpp
```diff
@@ -896,8 +896,8 @@ ur_result_t urQueueFlush(
 
 ur_result_t urEnqueueKernelLaunchCustomExp(
     ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel, uint32_t workDim,
-    const size_t *pGlobalWorkSize, const size_t *pLocalWorkSize,
-    uint32_t numPropsInLaunchPropList,
+    const size_t *pGlobalWorkOffset, const size_t *pGlobalWorkSize,
```
kbenzie, line 899: You'll need to ignore `pGlobalWorkOffset` below.

kbenzie (approved):
We don't squash on merge so if you could squash the commits in this PR that would be appriciated.

source/adapters/level_zero/queue.cpp
kbenzie: This is still not ignored, its causing build failures

Naghasan: fixed

kbenzie:
@oneapi-src/unified-runtime-cuda-write @oneapi-src/unified-runtime-level-zero-write please review, this is intended for the next release.

scripts/core/exp-launch-properties.yml
```diff
@@ -45,6 +47,10 @@ members:
       name: cooperative
       desc: "[in] non-zero value indicates a cooperative kernel"
       tag: $X_EXP_LAUNCH_PROPERTY_ID_COOPERATIVE
+    - type: size_t
+      name: workgroup_mem_size
+      desc: "[in] non-zero value indicates the amount of work group memory to allocate"
```
frasercrmck, line 52: Is it worth being explicit and saying "in bytes"?

source/adapters/cuda/enqueue.cpp
```diff
@@ -506,6 +511,16 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
 }
 
+UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
+    ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel, uint32_t workDim,
+    const size_t *pGlobalWorkOffset, const size_t *pGlobalWorkSize,
+    const size_t *pLocalWorkSize, uint32_t numEventsInWaitList,
+    const ur_event_handle_t *phEventWaitList, ur_event_handle_t *phEvent) {
+  return enqueueKernelLaunch(hQueue, hKernel, workDim, pGlobalWorkOffset,
+                             pGlobalWorkSize, pLocalWorkSize,
+                             numEventsInWaitList, phEventWaitList, phEvent, 0);
```
frasercrmck, line 521:
```suggestion
                             numEventsInWaitList, phEventWaitList, phEvent, /*WorkGroupMemory=*/0);
```

(would probably require reformatting)

source/adapters/cuda/enqueue.cpp
```diff
@@ -561,17 +591,21 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunchCustomExp(
...
+  if (WorkGroupMemory)
+    hKernel->setWorkGroupMemory(WorkGroupMemory);
   uint32_t LocalSize = hKernel->getLocalSize();
   CUfunction CuFunc = hKernel->get();
 
   for (uint32_t i = 0; i < numPropsInLaunchPropList; i++) {
     switch (launchPropList[i].id) {
     case UR_EXP_LAUNCH_PROPERTY_ID_IGNORE: {
+      launch_attribute.push_back({});
       launch_attribute[i].id = CU_LAUNCH_ATTRIBUTE_IGNORE;
```
frasercrmck, line 604: If we hit the `UR_EXP_LAUNCH_PROPERTY_ID_WORK_GROUP_MEMORY` case, we don't push back a `CUlaunchAttribute`. At that point, the value of `i` goes out of sync with the number of elements in `launch_attribute`, and `launch_attribute[i]` is out of bounds. Maybe this is fine because `UR_EXP_LAUNCH_PROPERTY_ID_WORK_GROUP_MEMORY` only ever comes after the others?

I'd suggest we remove the reliance on `launch_attribute[i]` and use `launch_attribute.back()`.

Since we're C++17, we could do `auto &attr = launch_attribute.emplace_back({}); attr.id = ...`, right?

source/adapters/cuda/kernel.hpp
```diff
@@ -140,20 +148,40 @@ struct ur_kernel_handle_t_ {
...
     }
 
+    void setWorkGroupMemory(size_t memSize) {
+      assert(WorkGroupMemory == 0 &&
+             "Work Group Memory size can only be set once");
+      // Ensure first offset is MaxAlignment aligned
+      WorkGroupMemory = alignMemoryAllocation(MaxAlignment, memSize);
+
+      // Adjust local accessor setting
+      // the dynamic memory will start at offset 0 (allows us to keep access
```
frasercrmck, line 158:
```suggestion
      // the dynamic memory will start at offset 0 (allows us to keep accessing
```

source/adapters/cuda/kernel.hpp
```diff
@@ -140,20 +148,40 @@ struct ur_kernel_handle_t_ {
...
+      // local memory as a GV) and accessors will use the rest of the range
+      for (size_t i = 0; i < OffsetPerIndex.size(); i++) {
+        // if offset is 0, that's it is not a local accessor argument.
```
frasercrmck, line 161:
```suggestion
        // if offset is 0, it is not a local accessor argument.
```

source/adapters/cuda/kernel.hpp
```diff
@@ -140,20 +148,40 @@ struct ur_kernel_handle_t_ {
       MemObjArgs.push_back(arguments::mem_obj_arg{hMem, Index, Flags});
     }
 
+    void setWorkGroupMemory(size_t memSize) {
```
frasercrmck, line 151: `memSize` seems out of place with the naming conventions. Perhaps `MemSize`?

source/adapters/cuda/kernel.hpp
```diff
@@ -238,6 +266,7 @@ struct ur_kernel_handle_t_ {
     return Args.getIndices();
   }
 
+  void setWorkGroupMemory(size_t memSize) { Args.setWorkGroupMemory(memSize); }
```
frasercrmck, line 269:
```suggestion
  void setWorkGroupMemory(size_t MemSize) { Args.setWorkGroupMemory(MemSize); }
```

scripts/core/exp-launch-properties.yml
Naghasan: yes, changed

source/adapters/cuda/enqueue.cpp
Naghasan: done locally

source/adapters/cuda/kernel.hpp
Naghasan: 👍 fixed

source/adapters/cuda/kernel.hpp
Naghasan: fixed locally, thanks

source/adapters/cuda/kernel.hpp
Naghasan: fixed

source/adapters/cuda/kernel.hpp
Naghasan: fixed

source/adapters/cuda/enqueue.cpp
Naghasan: good catch, fixed. 

> Since we're C++17, we could do auto &attr = launch_attribute.emplace_back({}); attr.id = ..., right?

why not, cleaner than calling `launch_attribute.back()` over and over

frasercrmck (approved):
CUDA LGTM

JackAKirk:
> LGTM

Actually latest testing shows some related failures in the cuda build

Naghasan:
> Actually latest testing shows some related failures in the cuda build

as far as I can tell it is because needs the DPC++ change (function signature change). The related PR on the dpc++ side is green, same reason for the level 0 failure.

the other failures are unrelated to this patch (failure to fetch xpti)

pbalcer (approved):

Naghasan:
@callumfare the build failure was due to the fact you haven't bumped the tag, only updated the comment ;)

https://github.com/intel/llvm/pull/15061/commits/33ea3b0febb37a7470e748e22d45618bc9112d89

PR #1972, opened 2024-08-14
Fix CL adapter assuming incorrect type for sub group query.

aarongreig:
On a fallback path used when the clGetKernelSubGroupInfo query isn't supported the adapter tries the UR_DEVICE_INFO_SUB_GROUP_SIZES_INTEL device info query via urDeviceGetInfo. It currently assumes the type returned by that query is size_t, this patch currects that to uint32_t.


aarongreig:
LLVM PR https://github.com/intel/llvm/pull/15074

martygrant (approved):

omarahmed1111:
@aarongreig is the intel/llvm failures related, I don't think it does, but just confirming with you?

aarongreig:
No, they're level zero fails and this only touches the CL adapter. I'll push just now and hopefully they go away.

PR #1975, opened 2024-08-14
[Command-Buffers] Implement new command-list enqueue path

fabiomestre:
Adds a new path that submits command-buffers using zeCommandListImmediateAppendCommandListsExp() instead of zeCommandQueueExecuteCommandLists(). This allows:
 - Waiting for command-buffer submission dependencies without having to create a separate command-list.
 - Since the WaitEvent is no longer needed in this path, counter-based events can be used for helper events which  removes the need to create an event reset command-list.
- When the command-buffer is not in-order, the new path also moves the reset event command-list to execute after the command-buffer execution.

Intel/llvm PR: https://github.com/intel/llvm/pull/16096


source/adapters/level_zero/command_buffer.cpp
```diff
@@ -616,13 +624,79 @@ bool canBeInOrder(ur_context_handle_t Context,
...
              : false;
 }
-} // namespace
 
-UR_APIEXPORT ur_result_t UR_APICALL
-urCommandBufferCreateExp(ur_context_handle_t Context, ur_device_handle_t Device,
-                         const ur_exp_command_buffer_desc_t *CommandBufferDesc,
-                         ur_exp_command_buffer_handle_t *CommandBuffer) {
+ur_result_t
+commandBufferCreateNew(ur_context_handle_t Context, ur_device_handle_t Device,
```
fabiomestre, line 629: After the implementation is finished, will try to figure out if the "New" and "Old" entrypoints should be merged or not. For now keeping the implementations separate to make it less confusing

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -1346,12 +1515,174 @@ ur_result_t createUserEvent(ur_exp_command_buffer_handle_t CommandBuffer,
...
+ * @return
+ */
+ur_result_t submitComputeEngineCmdLists(
+    ur_queue_handle_legacy_t Queue,
+    ur_exp_command_buffer_handle_t CommandBuffer, uint32_t NumEventsInWaitList,
+    const ur_event_handle_t *EventWaitList, ur_event_handle_t *UserEvent) {
+
+  std::vector<ze_command_list_handle_t> ComputeCommandLists{
+      CommandBuffer->ZeCommandListResetEvents,
+      CommandBuffer->ZeComputeCommandList};
```
EwanC, line 1575: Wonder if this should this be `ZeComputeCommandListTranslated` if we're using an extension function, although it's a parameter not the command-list being executed on.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -1346,12 +1515,174 @@ ur_result_t createUserEvent(ur_exp_command_buffer_handle_t CommandBuffer,
...
+commandBufferEnqueueNew(ur_exp_command_buffer_handle_t CommandBuffer,
+                        ur_queue_handle_t UrQueue, uint32_t NumEventsInWaitList,
+                        const ur_event_handle_t *EventWaitList,
+                        ur_event_handle_t *Event) {
+  auto Queue = Legacy(UrQueue);
+  std::scoped_lock<ur_shared_mutex> Lock(Queue->Mutex);
+
+  if (!CommandBuffer->MCopyCommandListEmpty) {
+    submitCopyEngineCmdLists(CommandBuffer, Queue, NumEventsInWaitList,
+                             EventWaitList);
```
EwanC, line 1672: can be wrapped in UR_CALL

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -1761,6 +2108,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferUpdateKernelLaunchExp(
   UR_CALL(validateCommandDesc(Command, CommandDesc));
 
   // We must synchronize mutable command list execution before mutating.
+  // FIXME Is this needed for the new path?
```
EwanC, line 2111: I think we won't set `ZeActiveFence` in the new path since there won't be an output fence from the call

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -706,8 +793,78 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t CommandBuffer) {
...
+  ZE2UR_CALL(zeCommandListAppendSignalEvent,
+             (CommandBuffer->ZeCommandListResetEvents,
+              CommandBuffer->AllResetEvent->ZeEvent));
+
+  // Close the command lists and have them ready for dispatch.
+  ZE2UR_CALL(zeCommandListClose, (CommandBuffer->ZeComputeCommandList));
+  ZE2UR_CALL(zeCommandListClose, (CommandBuffer->ZeCommandListResetEvents));
+
+  if (CommandBuffer->useCopyEngine()) {
+    ZE2UR_CALL(zeCommandListAppendSignalEvent,
```
fabiomestre, line 855: I think this might need to be a barrier that waits on all the commands in the copy command-list before signaling.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -689,6 +763,19 @@ urCommandBufferCreateExp(ur_context_handle_t Context, ur_device_handle_t Device,
 
   return UR_RESULT_SUCCESS;
 }
+} // namespace
+
+UR_APIEXPORT ur_result_t UR_APICALL
+urCommandBufferCreateExp(ur_context_handle_t Context, ur_device_handle_t Device,
+                         const ur_exp_command_buffer_desc_t *CommandBufferDesc,
+                         ur_exp_command_buffer_handle_t *CommandBuffer) {
+  if (Device->isPVC() && Device->ImmCommandListUsed) {
```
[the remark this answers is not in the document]
fabiomestre: I'm checking the driver version in the latest commit. But this is definitely better. Will update the code to do it in this way.

source/adapters/level_zero/command_buffer.cpp
fabiomestre: Still unclear to me if the handles need to be translated for this specific entrypoint. Will do some more investigation into this.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -614,9 +799,29 @@ urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t CommandBuffer) {
...
+}
+
+UR_APIEXPORT ur_result_t UR_APICALL
+urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t CommandBuffer) {
+  UR_ASSERT(CommandBuffer, UR_RESULT_ERROR_INVALID_NULL_POINTER);
+
+  // It is not allowed to append to command list from multiple threads.
+  std::scoped_lock<ur_shared_mutex> Guard(CommandBuffer->Mutex);
+
+  if (checkImmediateAppendSupport(CommandBuffer->Context,
```
Bensuo, line 812: We already queried this when we created the command buffer, does it make sense to just store that in the command buffer itself or infer from existing info in the command buffer? SInce the device and context cannot change here.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -1168,67 +1373,127 @@ ur_result_t waitForDependencies(ur_exp_command_buffer_handle_t CommandBuffer,
...
-
-    ZE2UR_CALL(zeCommandListAppendQueryKernelTimestamps,
-               (SignalCommandList->first, CommandBuffer->ZeEventsList.size(),
-                CommandBuffer->ZeEventsList.data(),
-                (void *)Profiling->Timestamps, 0, RetEvent->ZeEvent, 1,
-                &(CommandBuffer->SignalEvent->ZeEvent)));
-
-    RetEvent->CommandData = static_cast<void *>(Profiling);
-  } else {
+ur_result_t doProfiling(ur_exp_command_buffer_handle_t CommandBuffer,
```
Bensuo, line 1387: Nit but `doProfiling` is not a very descriptive function name to me, would prefer something more like `appendProfilingQueries` or something like that.

source/adapters/level_zero/command_buffer.hpp
```diff
@@ -72,33 +74,48 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
   ur_context_handle_t Context;
   // Device associated with this command buffer
   ur_device_handle_t Device;
-  // Level Zero command list handle
+  // Level Zero command list handle that has the compute engine command for this
+  // command-buffer.
```
Bensuo, line 78:
```suggestion
  // Level Zero command list handle that has the compute engine commands for this
  // command-buffer.
```

source/adapters/level_zero/command_buffer.hpp
```diff
@@ -72,33 +74,48 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
...
   // Given a multi driver scenario, the driver handle must be translated to the
   // internal driver handle to allow calls to driver experimental apis.
   ze_command_list_handle_t ZeComputeCommandListTranslated;
-  // Level Zero command list handle
+  // Level Zero command list handle that is responsible for resetting
+  // the events after the compute and copy command-lists execute.
   ze_command_list_handle_t ZeCommandListResetEvents;
-  // Level Zero Copy command list handle
+  // Level Zero command list handle that has the copy engine command for this
+  // command-buffer.
```
Bensuo, line 87:
```suggestion
  // Level Zero command list handle that has the copy engine commands for this
  // command-buffer.
```

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -495,57 +543,121 @@ bool canBeInOrder(ur_context_handle_t Context,
...
+
+UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferCreateExp(ur_context_handle_t Context, ur_device_handle_t Device,
                          const ur_exp_command_buffer_desc_t *CommandBufferDesc,
                          ur_exp_command_buffer_handle_t *CommandBuffer) {
-
   bool IsInOrder = canBeInOrder(Context, CommandBufferDesc);
   bool EnableProfiling =
-      CommandBufferDesc && CommandBufferDesc->enableProfiling;
+      CommandBufferDesc && CommandBufferDesc->enableProfiling && !IsInOrder;
```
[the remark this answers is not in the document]
fabiomestre: In-order command-lists don't have sync-point events. So there is nothing to profile. This check already existed on the original implementation, but it was only checked during enqueue with:

```
const bool DoProfiling =
      (UrQueue->Properties & UR_QUEUE_FLAG_PROFILING_ENABLE) &&
      (!CommandBuffer->IsInOrderCmdList) &&
      (CommandBuffer->IsProfilingEnabled) && Event;
```

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -696,7 +901,7 @@ ur_result_t
 createCommandHandle(ur_exp_command_buffer_handle_t CommandBuffer,
                     ur_kernel_handle_t Kernel, uint32_t WorkDim,
                     const size_t *LocalWorkSize,
-                    ur_exp_command_buffer_command_handle_t &Command) {
+                    ur_exp_command_buffer_command_handle_t *Command) {
```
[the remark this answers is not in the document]
fabiomestre: I think those are leftovers from debugging. I will revert the changes :+1:

source/adapters/level_zero/command_buffer.cpp
fabiomestre: Confirmed that this is not needed since the entrypoint doesn't rely on zeDriverGetExtensionFunctionAddress

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -22,6 +22,71 @@
...
+                                                            MinDriverVersion);
+
+  // If this environment variable is:
+  //   - Set to 1: the immediate append path will always be enabled as long the
+  //   pre-requisites are met.
+  //   - Set to 0: the immediate append path will always be disabled.
+  //   - Not Defined: The default behaviour will be used which enables the
+  //   immediate append path only for some devices when the pre-requisites are
+  //   met.
+  const char *AppendEnvVarName = "UR_L0_CMD_BUFFER_USE_IMMEDIATE_APPEND_PATH";
```
EwanC, line 45: We should document this in https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/LEVEL_ZERO.rst

EwanC (commented):
Can you stick a link to the DPC++ PR in the description? (as part of that PR we might want to add an update to https://github.com/intel/llvm/blob/sycl/sycl/doc/design/CommandGraph.md#level-zero) to mention this new code path

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -22,6 +22,71 @@
 
 namespace {
 
+// Checks whether zeCommandListImmediateAppendCommandListsExp can be used for a
+// given Context and Device.
+bool checkImmediateAppendSupport(ur_context_handle_t Context,
+                                 ur_device_handle_t Device) {
+  // TODO The L0 driver is not reporting this extension yet. Once it does,
```
EwanC, line 29: Could you create a ticket in our internal tracker for this

aarongreig (approved):
doc change looks fine to me

EwanC (approved):

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -22,6 +22,71 @@
 
 namespace {
 
+// Checks whether zeCommandListImmediateAppendCommandListsExp can be used for a
+// given Context and Device.
+bool checkImmediateAppendSupport(ur_context_handle_t Context,
```
pbalcer, line 27: longer-term, is there anything stopping us from deprecating the old non-immediate path? I don't like how the two modes are intermingling together. It makes the code more complex.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -501,78 +588,132 @@ bool canBeInOrder(ur_context_handle_t Context,
...
+
+  std::vector<ze_event_handle_t> PrecondEvents;
+  if (CommandBuffer->ZeCommandListResetEvents) {
+    PrecondEvents.push_back(CommandBuffer->AllResetEvent->ZeEvent);
+  }
+  if (!CommandBuffer->UseImmediateAppendPath) {
+    PrecondEvents.push_back(CommandBuffer->WaitEvent->ZeEvent);
+  }
+
+  ZE2UR_CALL(zeCommandListAppendBarrier,
```
pbalcer, line 606: Does this have to be a barrier? Can you use `zeCommandListAppendWaitOnEvents` instead?

scripts/core/LEVEL_ZERO.rst
```diff
@@ -139,8 +139,15 @@ Environment Variables
 | UR_L0_DISABLE_USM_ALLOCATOR                 | Controls the use of the USM allocator.                       | "0": USM allocator is enabled.                               | "0"              |
 |                                             |                                                              | Any other value: USM allocator is disabled.                  |                  |
 +---------------------------------------------+--------------------------------------------------------------+--------------------------------------------------------------+------------------+
-
+| UR_L0_CMD_BUFFER_USE_IMMEDIATE_APPEND_PATH  | Controls which command-buffer implementation path is used.   | "1": the immediate append path will always be enabled as     | Unset            |
```
pbalcer, line 142: Which path is tested will depend on the CI runner. We should more explicit about it by testing twice - with and without the "immediate append path" enabled.

pbalcer (approved):
L0 changes lgtm, just a few questions.

source/adapters/level_zero/command_buffer.cpp
pbalcer: One consideration is the supported driver versions, we need to make sure this works with the LTS driver. 30898 is relatively new.
So no, we can't just remove it, at least not yet.

fabiomestre: We have an internal task to look into this in the future. At the moment it's not possible because it's still an experimental feature. It's also not clear if the implementation of zeCommandListImmediateAppendCommandListsExp has been properly tested on older GPU's and we still need to support those for Graphs.

source/adapters/level_zero/command_buffer.cpp
fabiomestre: The L0 documentation is not very clear. Could you clarify what are the advantages of zeCommandListAppendWaitOnEvents  over a barrier with a nullptr signal event?

Does zeCommandListAppendWaitOnEvents  provide the same guarantees as a barrier? This is, does it guarantee that any commands appended after the wait are not executed before the events are signalled?

scripts/core/LEVEL_ZERO.rst
fabiomestre: At the moment, the CI drivers are all older than 30898 and I would assume that that will be the case for the foreseeable future. So it will always use the old path. 

I think that, once this extension is not experimental anymore, if we are still using 2 paths in the code, we could add more testing to CI. But for now, I don't think there is much point in doing so.

source/adapters/level_zero/command_buffer.cpp
EwanC: Agree having 2 paths is definitely a maintenance burden, and I would very much like us to only have the new path in future. As has been said, in the short term we are blocked by:
1. LTS/public driver being recent enough to support this L0 feature
2. Verification on older GPUs for correctness/performance regressions.

source/adapters/level_zero/command_buffer.cpp
pbalcer: >zeCommandListAppendWaitOnEvents over a barrier with a nullptr signal event?

Barrier is coarse-grain and enforces global memory ordering.

>Does zeCommandListAppendWaitOnEvents provide the same guarantees as a barrier? This is, does it guarantee that any commands appended after the wait are not executed before the events are signalled

Not for out-of-order cmdlists AFAIK. That is, commands can be reordered past `AppendWaitOnEvents` if there's no explicit event synchronization point between these two operations. But you are using events anyway, so that's why I've asked.

@nrspruit ping, in case I got anything wrong.

scripts/core/LEVEL_ZERO.rst
pbalcer: AFAIK our PVC runners have a new enough driver (the newest available rolling is `24.39.31294.21`, but fair, we can expand CI later. Please create an issue.

EwanC: Created ticket https://github.com/oneapi-src/unified-runtime/issues/2373

PR #1987, opened 2024-08-16
Add dlerror output to loader logs

npmiller:
Failing to load an adapter is not an error in most cases the adapter is just not present.

However there's a lot of other reasons why an adapter may fail to load, such as a missing driver library or an unsupported driver. In these cases `dlerror` output is crucial to troubleshoot the issue.


RossBrunton (approved):

npmiller:
Tests: https://github.com/intel/llvm/pull/15110

igchor (approved):
The L0 v2 failures are unrelated. I will fix them today.

PR #1991, opened 2024-08-19
Clear OpenCL function pointer cache when context is released.

aarongreig:


aarongreig:
LLVM PR https://github.com/intel/llvm/pull/15135

PR #1994, opened 2024-08-20
Remove trailing semicolon. NFC

frasercrmck:


RossBrunton (approved):

omarahmed1111:
@frasercrmck As trivial that is, I would still be skeptical that we should create an Intel/llvm PR for it :smiley:

kbenzie:
> @frasercrmck As trivial that is, I would still be skeptical that we should create an Intel/llvm PR for it 😃

There's no need for an intel/llvm PR for this.

PR #1996, opened 2024-08-20
Add two new properties to ur_kernel_group_info_t

frasercrmck:
These two properties allow the program to specify a maximum work-group size in various ways.

They are intended to be targeted from languages such as SYCL (see https://github.com/intel/llvm/pull/14518).

This PR implements them for CUDA and Native CPU. It should also be able support them for HIP, in the same fashion. Other adapters using SPIR-V and/or Level Zero will need further changes to both of those specifications.


hdelan:
LGTM but I think it'd be a good idea to add a SYCL kernel [here](https://github.com/oneapi-src/unified-runtime/tree/main/test/conformance/device_code) that uses this SYCL property.

frasercrmck:
> LGTM but I think it'd be a good idea to add a SYCL kernel [here](https://github.com/oneapi-src/unified-runtime/tree/main/test/conformance/device_code) that uses this SYCL property.

I think that's a good idea but would depend on my DPC++ PR to be merged first, so that will have to be future work.

hdelan (approved):

source/adapters/native_cpu/kernel.cpp
```diff
@@ -31,14 +31,25 @@ urKernelCreate(ur_program_handle_t hProgram, const char *pKernelName,
...
-  } else {
-    kernel = new ur_kernel_handle_t_(hProgram, pKernelName, *f);
+  if (auto ReqdIt = ReqdMap.find(pKernelName); ReqdIt != ReqdMap.end()) {
+    ReqdWG = ReqdIt->second;
   }
 
+  std::optional<native_cpu::WGSize_t> MaxWG;
+  const auto &MaxMap = hProgram->KernelMaxWorkGroupSizeMD;
+  if (auto MaxIt = MaxMap.find(pKernelName); MaxIt != MaxMap.end()) {
+    ReqdWG = MaxIt->second;
```
PietroGhg, line 43: Don't we want `MaxWG` instead of `ReqdWG` here?

frasercrmck: Ooft good spot, thank you.

PietroGhg (approved):
Native CPU LGTM, thank you

frasercrmck:
Ping @oneapi-src/unified-runtime-opencl-write and @oneapi-src/level-zero-maintainers, thanks

aarongreig (approved):

frasercrmck:
Ping @oneapi-src/level-zero-maintainers, thanks

test/conformance/kernel/urKernelGetGroupInfo.cpp
```diff
@@ -85,11 +87,25 @@ TEST_P(urKernelGetGroupInfoWgSizeTest, CompileWorkGroupSize) {
...
+    // Returns 0 by default when there is no specific information
+    std::array<size_t, 3> read_dims{1, 1, 1};
+    std::array<size_t, 3> zero{0, 0, 0};
+    auto result = urKernelGetGroupInfo(
+        kernel, device, UR_KERNEL_GROUP_INFO_COMPILE_MAX_WORK_GROUP_SIZE,
+        sizeof(read_dims), read_dims.data(), nullptr);
+    if (result == UR_RESULT_SUCCESS) {
+        ASSERT_EQ(read_dims, zero);
+    } else {
+        ASSERT_EQ(result, UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION);
```
pbalcer, line 109: Please change this to a skip if the feature is unsupported. This will eliminate the need to include the test in the match files.

pbalcer (approved):
lgtm, just one nit.

Also, the windows build is failing - a rebase should fix that.

test/conformance/kernel/urKernelGetGroupInfo.cpp
kbenzie: This is the same pattern used elsewhere to handle the `UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION` is also success case.

pbalcer: Oh, right, I just noticed a match file change and didn't properly read the test before commenting :)

pbalcer (approved):

PR #1998, opened 2024-08-21
In build jobs, properly disable UR downloads

RossBrunton:
These jobs were not properly setting `SYCL_UR_USE_FETCH_CONTENT`, and so were fetching UR when they should have been using the UR under test.


kbenzie (approved):

RossBrunton:
This commit has no observable changes to intel/llvm, so merging without an associated MR there.

PR #2006, opened 2024-08-23
[DeviceSanitizer] Fix libstdc++ and libc++ mismatch problem

zhaomaosu:
UR will force link with libstdc++, but in in-tree build, if LLVM is built with libc++, this will cause mismatch problem. So, in such case, we need to build symbolizer.cpp with libc++ abi and link libc++ in.

### DIFF
--- a/source/loader/CMakeLists.txt
+++ b/source/loader/CMakeLists.txt
@@ -167,6 +167,26 @@ if(UR_ENABLE_SANITIZER)
         )
         target_include_directories(ur_loader PRIVATE ${LLVM_INCLUDE_DIRS})
         target_link_libraries(ur_loader PRIVATE LLVMSupport LLVMSymbolize)
+        # In in-tree build, if LLVM is built with libc++, we also need to build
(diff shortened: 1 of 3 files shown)


yingcong-wu (approved):

zhaomaosu:
Hi @oneapi-src/unified-runtime-maintain, could you please help review this PR? This is an important fix for sanitizer layer and we need this change to be included in v0.10.x release branch. Thanks.

zhaomaosu:
The risk of this change is low. It will not introduce any functionality of ABI change. Only fix a build failure and also will not impact syclos part.

zhaomaosu:
kindly ping @oneapi-src/unified-runtime-maintain

omarahmed1111 (approved):

omarahmed1111:
As I understand, this PR doesn't affect intel/llvm but I created a UR bump PR for it [here](https://github.com/intel/llvm/pull/15188). For syncing purposes as it will still need to land there and I didn't want to batch this change with another non `0.10.x` changes.

PR #2007, opened 2024-08-23
Add umfInit/umfTearDown to urAdapterGet/urAdapterRelease

omarahmed1111:
As windows order of unloading dlls is reversed from linux, windows will call umfTearDown before it could release umf objects in level_zero, so we call umfInit on urAdapterGet and umfAdapterTearDown to enforce the teardown of umf
after umf objects are destructed.

[Intel/llvm PR](https://github.com/intel/llvm/pull/15182)

### DIFF
--- a/source/adapters/level_zero/adapter.cpp
+++ b/source/adapters/level_zero/adapter.cpp
@@ -27,7 +27,12 @@ ur_adapter_handle_t_ *GlobalAdapter = new ur_adapter_handle_t_();
 #else
 ur_adapter_handle_t_ *GlobalAdapter;
 #endif
-
+// This is a temporary workaround on windows, where UR adapter is teardowned
+// before the UR loader, which will result in access violation when we use print
+// function as the overrided print function was already released with the UR
+// adapter.
+// TODO: Change adapters to use a common sink class in the loader instead of
+// using thier own sink class that inherit from logger::Sink.
 class ur_legacy_sink : public logger::Sink {
 public:
   ur_legacy_sink(std::string logger_name = "", bool skip_prefix = true)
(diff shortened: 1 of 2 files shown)


igchor (approved):

nrspruit (approved):

source/adapters/level_zero/adapter.cpp
```diff
@@ -12,14 +12,27 @@
...
 #else
 ur_adapter_handle_t_ *GlobalAdapter;
 #endif
-
+// This is a temporary workaround on windows, where UR adapter is teardowned
+// before the UR loader, which will result in access violation when we use print
+// function as the overrided print function was already released with the UR
+// adapter.
+// TODO: Change adapters to use a common sink class in the loader instead of
+// using thier own sink class that inherit from logger::Sink.
```
omarahmed1111, line 35: This is not ideal, but it should work for now to solve a release problem regarding the logger. @igchor @nrspruit If this is okay, I will merge this asap.

callumfare (approved):

nrspruit (approved):

source/adapters/level_zero/adapter.cpp
igchor: Yeah, I think that's fine for now. Eventually, we could do a similar thing like here: https://github.com/oneapi-src/unified-runtime/pull/2000. Just keep a global instance of shared_ptr<> to the logger (here https://github.com/oneapi-src/unified-runtime/blob/b766009add8dc41ad03e2a63be9f6ed40eea4c55/source/common/logger/ur_logger.hpp#L24) and every time get_logger is called, we would return shared_ptr instead of a reference.

igchor (approved):

PR #2008, opened 2024-08-23
Update CMake warning for tests hidden behind UR_DPCXX flag

martygrant:
Some conformance tests are disabled when UR_DPCXX is not set in CMake.
Only `test-program`, `test-kernel` and `test-enqueue` are displayed in the output warning when the flag is not set.
Added other tests to the output warning that also get disabled when the flag is not set.


test/conformance/CMakeLists.txt
```diff
@@ -143,18 +156,20 @@ if(UR_DPCXX)
...
-    add_subdirectory(exp_command_buffer)
-    add_subdirectory(exp_enqueue_native)
-    add_subdirectory(exp_usm_p2p)
-    add_subdirectory(exp_launch_properties)
-    add_subdirectory(memory-migrate)
+    foreach(dir ${TEST_SUBDIRECTORIES_DPCXX})
+        add_subdirectory(${dir})
+    endforeach()
 else()
+    set(DISABLED_TESTS)
```
omarahmed1111, line 163: NIT: I think it is better to set this as empty string
```suggestion
    set(DISABLED_TESTS "")
```

omarahmed1111 (approved):

PR #2012, opened 2024-08-26
[L0 v2] Use L0 provider from UMF and bump UMF tag

igchor:
Move L0 fetching to the CMakeLists.txt in common:
L0 needs to be fetched before UMF.


cmake/FetchLevelZero.cmake
```diff
@@ -0,0 +1,92 @@
+# Copyright (C) 2024 Intel Corporation
+# Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+# See LICENSE.TXT
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+set(UR_LEVEL_ZERO_LOADER_LIBRARY "" CACHE FILEPATH "Path of the Level Zero Loader library")
+set(UR_LEVEL_ZERO_INCLUDE_DIR "" CACHE FILEPATH "Directory containing the Level Zero Headers")
```
lukaszstolarczuk, line 7: nit: `PATH` not `FILEPATH` (it's just a small change in cmake GUI)

cmake/FetchLevelZero.cmake
```diff
@@ -0,0 +1,92 @@
...
+    INTERFACE "${LEVEL_ZERO_LIBRARY_SRC}"
+)
+target_link_libraries(LevelZeroLoader
+    INTERFACE "${LEVEL_ZERO_LIB_NAME}"
+)
+
+add_library (LevelZeroLoader-Headers INTERFACE)
+target_include_directories(LevelZeroLoader-Headers
+    INTERFACE "${LEVEL_ZERO_INCLUDE_DIR}"
+)
```
lukaszstolarczuk, line 92: missing EOF

source/common/CMakeLists.txt
```diff
@@ -20,12 +32,11 @@ target_include_directories(ur_common PUBLIC
 
 message(STATUS "Download Unified Memory Framework from github.com")
 if (NOT DEFINED UMF_REPO)
-    set(UMF_REPO "https://github.com/oneapi-src/unified-memory-framework.git")
+    set(UMF_REPO "https://github.com/igchor/unified-memory-framework.git")
```
lukaszstolarczuk, line 35: we merged https://github.com/oneapi-src/unified-memory-framework/pull/684 into main, but I guess we could stick with UMF commits from v0.9.x stable branch for now...?

I guess, you could only set UMF_TAG to main branch for L0_V2 adapter...? I'm not sure if this makes sense

igchor: Hm, why do we actually use v0.9.x on main? We should definitely use v0.9.x for UR 0.10.0 but for UR main we can use UMF main I think

cmake/FetchLevelZero.cmake
igchor: Done. I've done this in separate commit in case this breaks anything ;d

cmake/FetchLevelZero.cmake
igchor: Done.

source/common/CMakeLists.txt
lukaszstolarczuk: och, you're right, we already have v0.10.x stable branch in UR

lukaszstolarczuk (commented):
👍

pbalcer (approved):

lukaszstolarczuk:
I guess you should re-base, and BTW you could bump umf with your latest changes: https://github.com/oneapi-src/unified-memory-framework/pull/703

igchor:
> I guess you should re-base, and BTW you could bump umf with your latest changes: [oneapi-src/unified-memory-framework#703](https://github.com/oneapi-src/unified-memory-framework/pull/703)

Done.

PR #2016, opened 2024-08-27
[L0 v2] implement USM allocation functions using UMF 

igchor:
Based on: https://github.com/oneapi-src/unified-runtime/pull/2012


source/common/CMakeLists.txt
```diff
@@ -20,12 +26,11 @@ target_include_directories(ur_common PUBLIC
 
 message(STATUS "Download Unified Memory Framework from github.com")
 if (NOT DEFINED UMF_REPO)
-    set(UMF_REPO "https://github.com/oneapi-src/unified-memory-framework.git")
+    set(UMF_REPO "https://github.com/igchor/unified-memory-framework.git")
```
pbalcer, line 29: ?

test/conformance/enqueue/urEnqueueKernelLaunch.cpp
```diff
@@ -295,6 +295,68 @@ TEST_P(urEnqueueKernelLaunchTestWithParam, Success) {
     ValidateBuffer(buffer, buffer_size, val);
 }
 
+struct urEnqueueKernelLaunchWithUSM : uur::urKernelExecutionTest {
```
pbalcer, line 298: I'd like to see tests that allocate stuff in one thread, and use it in another, and then deallocate in yet another. But we can leave that as TODO.

pbalcer (approved):
mostly lgtm, but please rebase and remove the use of igchor umf fork :P

source/common/CMakeLists.txt
igchor: Done. This was for testing only.

test/conformance/enqueue/urEnqueueKernelLaunch.cpp
igchor: I'll create a task for that.

source/adapters/level_zero/v2/kernel.cpp
```diff
@@ -196,6 +196,48 @@ ur_program_handle_t ur_kernel_handle_t_::getProgramHandle() const {
   return hProgram;
 }
 
+ur_result_t ur_kernel_handle_t_::setExecInfo(ur_kernel_exec_info_t propName,
+                                             const void *pPropValue) {
+  std::scoped_lock<ur_shared_mutex> Guard(Mutex);
+
+  for (auto &kernel : deviceKernels) {
```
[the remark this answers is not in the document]
igchor: Yes, it's pretty much a copy-paste. Do you think it would be better to extract this logic from the legacy adapter to some common place and use it here? I didn't do this since the logic is quite simple but there is similar problem with urKernelGetGroupInfo

igchor:
@pbalcer can we bump the UMF version used by UMF to the latest main already?

source/adapters/level_zero/v2/kernel.cpp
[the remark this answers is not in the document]
igchor: I will share the logic in a separate PR.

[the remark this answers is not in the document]
igchor: https://github.com/oneapi-src/unified-runtime/pull/2145

PR #2017, opened 2024-08-28
[L0] Use zesInit for SysMan API usage

nrspruit:
- Change to using zesInit and zes data structures for accessing L0
SysMan functionality.
- Updated Platform & Devices to store zes handles if sysman support is
available.
- Given Legacy Environment Variable from user, then fallback to
old functionality.
- Fixed Return code on error to be consistently unsupported enumeration.

### DIFF
--- a/source/adapters/level_zero/adapter.cpp
+++ b/source/adapters/level_zero/adapter.cpp
@@ -219,25 +219,19 @@ ur_adapter_handle_t_::ur_adapter_handle_t_()
     // separately. This must be done to avoid attempting to use symbols that do
     // not exist in older loader runtimes.
 #ifdef _WIN32
-    std::string l0LoaderName = "ze_loader.dll";
+    HMODULE processHandle = GetModuleHandle(NULL);
 #else
-    std::string l0LoaderName = "libze_loader.so.1";
+    HMODULE processHandle = nullptr;
 #endif
-    GlobalAdapter->loaderHandle =
-        ur_loader::LibLoader::loadAdapterLibrary(l0LoaderName.c_str());
-    if (GlobalAdapter->loaderHandle.get() != nullptr) {
-      GlobalAdapter->getDeviceByUUIdFunctionPtr =
-          (zes_pfnDriverGetDeviceByUuidExp_t)
-              ur_loader::LibLoader::getFunctionPtr(
-                  GlobalAdapter->loaderHandle.get(),
-                  "zesDriverGetDeviceByUuidExp");
-      GlobalAdapter->getSysManDriversFunctionPtr =
-          (zes_pfnDriverGet_t)ur_loader::LibLoader::getFunctionPtr(
-              GlobalAdapter->loaderHandle.get(), "zesDriverGet");
-      GlobalAdapter->sysManInitFunctionPtr =
-          (zes_pfnInit_t)ur_loader::LibLoader::getFunctionPtr(
-              GlobalAdapter->loaderHandle.get(), "zesInit");
-    }
+    GlobalAdapter->getDeviceByUUIdFunctionPtr =
+        (zes_pfnDriverGetDeviceByUuidExp_t)ur_loader::LibLoader::getFunctionPtr(
+            processHandle, "zesDriverGetDeviceByUuidExp");
+    GlobalAdapter->getSysManDriversFunctionPtr =
+        (zes_pfnDriverGet_t)ur_loader::LibLoader::getFunctionPtr(
+            processHandle, "zesDriverGet");
+    GlobalAdapter->sysManInitFunctionPtr =
+        (zes_pfnInit_t)ur_loader::LibLoader::getFunctionPtr(processHandle,
+                                                            "zesInit");
     if (GlobalAdapter->getDeviceByUUIdFunctionPtr &&
         GlobalAdapter->getSysManDriversFunctionPtr &&
         GlobalAdapter->sysManInitFunctionPtr) {
--- a/source/adapters/level_zero/adapter.cpp
+++ b/source/adapters/level_zero/adapter.cpp
@@ -90,7 +90,8 @@ ur_result_t initPlatforms(PlatformVec &platforms,
 
   ZE2UR_CALL(zeDriverGet, (&ZeDriverCount, ZeDrivers.data()));
   for (uint32_t I = 0; I < ZeDriverCount; ++I) {
-    bool DriverInit = false;
+    // Keep track of the first platform init for this Driver
+    bool DriverPlatformInit = false;
     ze_device_properties_t device_properties{};
     device_properties.stype = ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES;
     uint32_t ZeDeviceCount = 0;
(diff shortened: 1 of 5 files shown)


nrspruit:
Hello @pbalcer , please take a look, this reuses the ur loader functions for the function pointer retrieval. See if you disagree with their reuse or if you are fine with this change.

source/adapters/level_zero/adapter.cpp
```diff
@@ -171,8 +209,40 @@ ur_adapter_handle_t_::ur_adapter_handle_t_()
...
+          (zes_pfnDriverGet_t)ur_loader::LibLoader::getFunctionPtr(
+              GlobalAdapter->loaderHandle.get(), "zesDriverGet");
+      GlobalAdapter->sysManInitFunctionPtr =
+          (zes_pfnInit_t)ur_loader::LibLoader::getFunctionPtr(
+              GlobalAdapter->loaderHandle.get(), "zesInit");
+    }
+    if (GlobalAdapter->getDeviceByUUIdFunctionPtr &&
+        GlobalAdapter->getSysManDriversFunctionPtr &&
+        GlobalAdapter->sysManInitFunctionPtr) {
+      logger::debug("\nzesInit with flags value of {}\n", static_cast<int>(0));
```
pbalcer, line 238: This will always print value of 0. Was that intentional?

source/adapters/level_zero/adapter.cpp
```diff
@@ -171,8 +209,40 @@ ur_adapter_handle_t_::ur_adapter_handle_t_()
 
       return;
     }
+    // Dynamically load the new L0 SysMan separate init and new EXP apis
+    // separately. This must be done to avoid attempting to use symbols that do
+    // not exist in older loader runtimes.
+#ifdef _WIN32
```
pbalcer, line 215: if these libraries are already part of the process, it should be possible to skip loading the library at all, and simply use:
```
dlsym(RTLD_DEFAULT, "zesSymbolName");
```
on Linux, and:
```
HMODULE hModule = GetModuleHandle(NULL);
FARPROC pFunc = GetProcAddress(hModule, "zesSymbolName");
````
on Windows.

But I won't insist, either solution works.

File: source/adapters/level_zero/device.cpp [hunk 3]
```diff
@@ -747,11 +747,13 @@ ur_result_t urDeviceGetInfo(
   }
 
   case UR_DEVICE_INFO_GLOBAL_MEM_FREE: {
-    if (getenv("ZES_ENABLE_SYSMAN") == nullptr) {
-      setErrorMessage("Set ZES_ENABLE_SYSMAN=1 to obtain free memory",
-                      UR_RESULT_ERROR_UNINITIALIZED,
+    bool SysManEnv = getenv_tobool("ZES_ENABLE_SYSMAN", false);
+    if ((Device->Platform->ZedeviceToZesDeviceMap.size() == 0) && !SysManEnv) {
+      setErrorMessage("SysMan support is unavailable on this system. Please "
```
pbalcer, line 752: `setErrorMessage` is only meant to be used together with `UR_RESULT_ERROR_ADAPTER_SPECIFIC`. Simply use the logger to print an error message.

source/adapters/level_zero/adapter.cpp
```diff
@@ -65,24 +90,37 @@ ur_result_t initPlatforms(PlatformVec &platforms) noexcept try {
...
+    auto platform = std::make_unique<ur_platform_handle_t_>(ZeDrivers[I]);
     // Check if this driver has GPU Devices
     for (uint32_t D = 0; D < ZeDeviceCount; ++D) {
       ZE2UR_CALL(zeDeviceGetProperties, (ZeDevices[D], &device_properties));
-
       if (ZE_DEVICE_TYPE_GPU == device_properties.type) {
-        // If this Driver is a GPU, save it as a usable platform.
-        auto platform = std::make_unique<ur_platform_handle_t_>(ZeDrivers[I]);
-        UR_CALL(platform->initialize());
+        if (!DriverInit) {
```
pbalcer, line 105: why this change? won't this mean that only the first platform will ever initialize?

source/adapters/level_zero/adapter.cpp
```diff
@@ -65,24 +90,37 @@ ur_result_t initPlatforms(PlatformVec &platforms) noexcept try {
...
+          DriverInit = true;
+        }
+        if (ZesResult == ZE_RESULT_SUCCESS) {
+          ur_zes_device_handle_data_t ZesDeviceData;
+          zes_uuid_t ZesUUID;
+          std::memcpy(&ZesUUID, &device_properties.uuid, sizeof(zes_uuid_t));
+          if (getZesDeviceHandle(
+                  ZesUUID, &ZesDeviceData.ZesDevice, &ZesDeviceData.SubDeviceId,
+                  &ZesDeviceData.SubDevice) == UR_RESULT_SUCCESS) {
+            platforms.back()->ZedeviceToZesDeviceMap.insert(
```
pbalcer, line 120: I don't think this is right. This code should move into the platform->initialize() function. Right now you will add all the mappings to the last platform in the container.

source/adapters/level_zero/adapter.cpp
nrspruit: This is because now we will walk thru all the devices vs only until the first device matches the GPU requirement.

This is to prevent more than one platform being created per Driver.

source/adapters/level_zero/adapter.cpp
nrspruit: This is adding the device mappings to the last platform that was added to the platforms container which matches the current driver whose devices are being examined.

The next time in the loop for the next driver, we would be adding the mappings to the next last platform added.

This is supposed to be adding the map to the last platform that has been added each time after then platform has been pushed back.

If you feel strongly about adding this to the init, I can see if that is possible, but it would require going back thru the zedevices again. The reason I have this here is because we are already going thru the zedevices, then this calls the function to further loop thru the zesdevices to find the match.

source/adapters/level_zero/adapter.cpp
nrspruit: It was, but let me create a variable holding the current flag value instead of making this fixed.

source/adapters/level_zero/device.cpp
nrspruit: ok, I will delete the set error message.

source/adapters/level_zero/adapter.cpp
nrspruit: Added a variable to allow this value to change.

source/adapters/level_zero/adapter.cpp
nrspruit: To prevent multiple loading that would not be a bad idea, let me see if I can add that style to the common functions.

source/adapters/level_zero/device.cpp
nrspruit: Deleted and replaced with the logger.

source/adapters/level_zero/adapter.cpp
nrspruit: I changed the name of the variable and added a comment to make this easier to understand.

source/adapters/level_zero/adapter.cpp
nrspruit: I updated the variable names used and added comments to explain the process. Please let me know if you still feel the logic is incorrect.

source/adapters/level_zero/adapter.cpp
nrspruit: I have changed this to use the current process handle vs additionally loading the loader.

source/adapters/level_zero/adapter.cpp
pbalcer: ah, ok, I see, the two outer loops threw me off :) I still think I'd be cleaner to have this inside of platform initialize, but that's less important.  No need to change it.

source/adapters/level_zero/adapter.cpp
pbalcer: I missed that this was in the inner per-device loop. Nevermind, thanks.

pbalcer (commented):
lgtm for the most part, but much of this is dead code, especially the zes initialization. Can you please add a test that exercises this code path to make sure the loading of the loader symbol works? (do we have a new enough driver in the CI?)

source/adapters/level_zero/adapter.cpp
```diff
@@ -171,8 +215,36 @@ ur_adapter_handle_t_::ur_adapter_handle_t_()
 
       return;
     }
+    // Dynamically load the new L0 SysMan separate init and new EXP apis
+    // separately. This must be done to avoid attempting to use symbols that do
+    // not exist in older loader runtimes.
+#ifdef _WIN32
+    HMODULE processHandle = GetModuleHandle(NULL);
+#else
+    HMODULE processHandle = nullptr;
```
pbalcer, line 224: nit: this will work, but should technically be `RTLD_DEFAULT` (which is NULL anyway, but... :P)

pbalcer (approved):

PR #2019, opened 2024-08-28
Remove superfluous namespace

ldrumm:
It only contained `static` functions so it just made uses within the same TU more cumbersome.


aarongreig:
looking at merging this soon, can you resolve the conflict @ldrumm

PR #2020, opened 2024-08-28
Fix unspecified alignment

ldrumm:
It's plausible that a user will pass a non-null `ur_usm_desc_t` argument that itself has a zero `align` member. We need to gracefully accept such a case.


ldrumm:
@@oneapi-src/unified-runtime-native-cpu-write ping

PietroGhg (approved):
I completely missed this one, sorry for the late review. LGTM, thank you

PR #2026, opened 2024-08-28
[Benchmark] use UR_ADAPTERS_FORCE_LOAD in benchmarks

igchor:
to enable testing L0 v2 adapter which cannot be selected by ONEAPI_SELECTOR

### DIFF
--- a/scripts/benchmarks/benches/compute.py
+++ b/scripts/benchmarks/benches/compute.py
@@ -15,7 +15,6 @@ class ComputeBench:
     def __init__(self, directory):
         self.directory = directory
         self.built = False
-        self.adapter_short_name = {'level_zero' : 'L0', "level_zero_v2" : 'L0_V2'}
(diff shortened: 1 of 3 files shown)


scripts/benchmarks/benches/compute.py
```diff
@@ -60,7 +60,7 @@ def bin_args(self) -> list[str]:
         return []
 
     def extra_env_vars(self) -> dict:
-        return {}
+        return {"UR_ADAPTERS_FORCE_LOAD" : os.path.join(self.bench.libs, f"libur_adapter_{options.ur_adapter_name}.so")}
```
pbalcer, line 63: hm, maybe we should move this to base benchmark and concatenate the returned values?

igchor: Yeah, we can do that. Done.

PR #2028, opened 2024-08-29
[DeviceSanitizer] Init tracing layer after sanitizer layer to achieve proper tracing

yingcong-wu:
Currently, the tracing layer is inited before sanitizer layer, which means if the sanitizer layer intercepts an API and does not call through layers. The tracing layer will not be able to print the API call.

We should always keep the tracing layer as the last one to be initialized to make sure it is the first layer to handle any API calls so that any API calls would be properly traced.

SYCLOS PR: https://github.com/intel/llvm/pull/15221


aarongreig (approved):

AllanZyne (approved):

yingcong-wu:
Hi @aarongreig , what should I do to move this PR forward? Thank you.

pbalcer (approved):

pbalcer:
@yingcong-wu CI needs to be green (or mostly green :-))

yingcong-wu:
@pbalcer Sure, thank you. But I just don't see how my PR relates to those failures. I will try to investigate those failures now.

pbalcer:
@yingcong-wu the remaining failures are unrelated. I'll mark this as ready to merge.

yingcong-wu:
Thank you!

PR #1869, opened 2024-07-17
[DeviceSanitizer] Memory overhead statistics

AllanZyne:
LLVM: https://github.com/intel/llvm/pull/14592

Statistic memory overhead (usm redzone + shadow memory) on each context, enabled by exporting asan flag `UR_LAYER_ASAN_OPTIONS=print_stats:1`.

Addtionally, move "AsanOptions" under "SanitizerInterceptor".

PR #2029, opened 2024-08-29
[DeviceSanitizer] Fix a heap use-after-free problem.

yingcong-wu:
`AllocInfo` is a shared_ptr, and we need to keep a copy of it to avoid possible early release of objects that cause heap use-aftre-free problem. 

SYCLOS PR: https://github.com/intel/llvm/pull/15222


AllanZyne (approved):

yingcong-wu:
Same fix included in https://github.com/oneapi-src/unified-runtime/pull/1869, closing this PR.

PR #2045, opened 2024-09-04
Generate .def file with all UR entry points

callumfare:
This was necessary for https://github.com/intel/llvm/pull/15262 - it allows defining macros over every function in the API


include/ur_api_funcs.def
```diff
@@ -0,0 +1,197 @@
+
```
EwanC, line 1: Could we get a comment at the top of this saying something like "this file is autogenerated, do not edit"

callumfare: Good shout, done

PR #2047, opened 2024-09-04
Add sycl-bench benchmarks

mateuszpn:
Add unisa-hpc/sycl-bench SYCL benchmarks


scripts/benchmarks/main.py
```diff
@@ -71,28 +115,31 @@ def main(directory, additional_env_vars, save_name, compare_names, filter):
...
+            for label in set([result.label for result in iteration_results]):
+                label_results = [result for result in iteration_results if result.label == label]
+                label_results.sort(key=lambda res: res.value)
+                median_index = len(label_results) // 2
+                median_result = label_results[median_index]
 
-            median_result.unit = benchmark.unit()
-            median_result.name = benchmark.name()
+                median_result.unit = benchmark.unit()
+                median_result.name = label
```
igchor, line 134: Why did you change this? Now, we have 'hashtable' instead of 'VelocityBench Hashtable'

igchor:
I think we should add a 'unit' to the results table - it's not clear when looking at it if the result is in milliseconds or something else.

Also, if it is in milliseconds, some of the benchmarks take very little time - can we increase number of iterations/size so that we can hopefully get more stable results?

.github/workflows/benchmarks_compute.yml
```diff
@@ -167,6 +167,14 @@ jobs:
       id: benchmarks
       run: numactl -N 0 ${{ github.workspace }}/ur-repo/scripts/benchmarks/main.py ~/bench_workdir ${{github.workspace}}/sycl_build ${{github.workspace}}/ur-repo ${{ matrix.adapter.str_name }} ${{ inputs.bench_script_params }}
 
+    - name: Upload results
```
pbalcer, line 170: This doesn't seem like a good idea from a security point of view. How is this git authenticated?
This should work similarly to how we publish documentation, see this existing workflow: https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/docs.yml

scripts/benchmarks/benches/compute.py
```diff
@@ -18,11 +18,13 @@ def __init__(self, directory):
         return
 
     def setup(self):
-        if self.built:
+        build_path = create_build_path(self.directory, 'compute-benchmarks-build')
+        self.bins = os.path.join(build_path, 'bin')
+
+        if self.built or not options.rebuild:
```
pbalcer, line 24: why?
self.built means that this was already built initialized by the previous invocation of setup. What rebuilt is preventing is removing the build directory, but we still want to pull the changes, call cmake and see whether there aren't any updates.

scripts/benchmarks/benches/compute.py
```diff
@@ -43,14 +45,14 @@ def setup(self):
...
         self.built = True
-        self.bins = os.path.join(build_path, 'bin')
 
 class ComputeBenchmark(Benchmark):
     def __init__(self, bench, name, test):
         self.bench = bench
         self.bench_name = name
         self.test = test
         super().__init__(bench.directory)
+        self.setup()
```
pbalcer, line 55: why?

scripts/benchmarks/benches/compute.py
```diff
@@ -78,7 +80,7 @@ def run(self, env_vars) -> Result:
 
         result = self.run_bench(command, env_vars)
         (label, mean) = self.parse_output(result)
-        return Result(label=label, value=mean, command=command, env=env_vars, stdout=result, lower_is_better=self.lower_is_better())
+        return [ Result(label=self.name(), value=mean, command=command, env=env_vars, stdout=result, lower_is_better=self.lower_is_better()) ]
```
pbalcer, line 83: why are you changing this into an array?
All the other benchmarks would have to change as well. You'd also need to change the return type of this function (and the base one as well).

scripts/benchmarks/benches/syclbench.py
```diff
@@ -0,0 +1,379 @@
...
+                if not row[0].startswith('#'):
+                    res_list.append(
+                        Result(label=row[0], 
+                               value=float(row[12]) * 1000, # convert to ms 
+                               command=command, 
+                               env=env_vars, 
+                               stdout=result))
+            
+            median_list = []
+            for label in set(result.label for result in res_list):
```
pbalcer, line 102: this is already done for all the benchmarks in main.py. why repeat it?

scripts/benchmarks/benches/syclbench.py
```diff
@@ -0,0 +1,379 @@
...
+        bin_dir = self.bench.bins
+
+        command += self.bin_args()
+        env_vars.update(self.extra_env_vars())
+
+        print(f"Command: {command}")
+        result = self.run_bench(command, env_vars)
+
+        with open(outputfile, 'r') as f:
+            reader = csv.reader(f)
```
pbalcer, line 90: I'm not sure if I follow. We open an "output" file that contains some benchmark definitions? Shouldn't that be input file then?

scripts/benchmarks/utils/utils.py
```diff
@@ -42,6 +42,11 @@ def run(command, env_vars={}, cwd=None, add_sycl=False):
 
 def git_clone(dir, name, repo, commit):
     repo_path = os.path.join(dir, name)
+    if not options.rebuild:
```
pbalcer, line 45: no, please remove this. The intention is for the benchmark git repos to still be pulled to check for updates.

scripts/benchmarks/benches/velocity.py
```diff
@@ -27,20 +28,20 @@ def download_deps(self):
...
-        configure_command = [
-            "cmake",
-            f"-B {build_path}",
-            f"-S {self.code_path}",
-            f"-DCMAKE_BUILD_TYPE=Release"
-        ]
-        run(configure_command, {'CC': 'clang', 'CXX':'clang++'}, add_sycl=True)
-        run(f"cmake --build {build_path} -j", add_sycl=True)
+        if options.rebuild:
+            self.download_deps()
```
pbalcer, line 35: this is not necessary, the underlying `download_untar` function will skip the file download if it exists.

.github/workflows/benchmarks_compute.yml
mateuszpn: In final solution there are no changes in .yml files

scripts/benchmarks/benches/compute.py
mateuszpn: Ok, if it is crucial. The change made it possible to run the scripts on cluster nodes with no access to external servers...

scripts/benchmarks/benches/compute.py
mateuszpn: Single benchmark in sycl-bench may return a set of results, because in single run we in fact have a group of benchmarks. Functions in other classes are adjusted

scripts/benchmarks/benches/syclbench.py
mateuszpn: This is an output file containing results - when put in file, they are structured as csv and easy to analyze. I must analyze the output, whether it is in stdout or file, to return results. Additionally, single benchmark has repetitions, so I have to find median of results

scripts/benchmarks/benches/compute.py
mateuszpn: Removed (after reverting the options.rebuild logic).

scripts/benchmarks/benches/syclbench.py
mateuszpn: Removed

scripts/benchmarks/utils/utils.py
mateuszpn: Got it. Removed all conditions checking options.rebuild, which are unnecessary with this approach.

mateuszpn:
Due to big number of script-generated comments, and changes which need to be withdrawn, new branch and new PR is created. Look at https://github.com/oneapi-src/unified-runtime/pull/2115 for current status.

PR #2050, opened 2024-09-04
[Cuda] Check for exceeding grid dimension limits (work-group count)

GeorgeWeb:
This change does the grid dimensions validation earlier as it was previously deferred to `cuLaunchKernel`. For now it returns `UR_RESULT_ERROR_INVALID_VALUE` as it is also handled this way in the DPC++ sycl runtime, but it allows better control to target this with a more suitable error code in the future.

Related Q&A discussion: https://github.com/intel/llvm/discussions/14160

intel/llvm PR: https://github.com/intel/llvm/pull/15390

### DIFF
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -250,9 +250,18 @@ setKernelParams([[maybe_unused]] const ur_context_handle_t Context,
       return UR_RESULT_ERROR_INVALID_WORK_GROUP_SIZE;
     }
 
+    size_t MaxGridDim[3] = {0, 0, 0};
+    urDeviceGetInfo(Device, UR_DEVICE_INFO_MAX_WORK_GROUPS_3D,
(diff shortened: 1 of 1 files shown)


JackAKirk:
I've got mixed feelings. On the one hand it allows you to give the user more information, that the invalid value is due to invalid grid size. On the other hand this is more information than you get compared to cuda runtime, and I'm not sure that it needs to be this specific at the expense of slowing down the kernel launch to do the query. It also sets a precedent to do this.

Personally I think we should just be preserving the native errors but emitting a sycl exception, which is what this achieves:
https://github.com/oneapi-src/unified-runtime/issues/500#issuecomment-1554854496

I had thought that we had already agreed this since we had a meeting about it and I thought that the above was the agreed solution.

JackAKirk (approved):

GeorgeWeb:
I see your point now. I'll close that PR because at the moment the same `INVALID_VALUE`  mapping is returned from the driver API and will use the above approach you already linked for errors in future instances where it is actually needed. Thanks!

PR #2054, opened 2024-09-05
[DeviceSanitizer] refine option parsing

yingcong-wu:
Handle `getenv_to_map()` exception and refine error message. 

SYCLOS PR: https://github.com/intel/llvm/pull/15293


AllanZyne (approved):

aarongreig (approved):

pbalcer (approved):

source/loader/layers/sanitizer/asan_options.hpp
```diff
@@ -42,7 +42,14 @@ struct AsanOptions {
 
   private:
     AsanOptions(logger::Logger &logger) {
-        auto OptionsEnvMap = getenv_to_map("UR_LAYER_ASAN_OPTIONS");
+        std::optional<EnvVarMap> OptionsEnvMap;
+        try {
+            OptionsEnvMap = getenv_to_map("UR_LAYER_ASAN_OPTIONS");
+        } catch (const std::invalid_argument &e) {
+            logger.error(e.what());
+            die("Sanitizer failed to parse options.\n");
```
pbalcer, line 50: should we really die here? this will crash the application.

zhaomaosu: We'd better to align the behavior with host sanitizer.

yingcong-wu: When fail to parse options, we can run with default options or simply die. I think if users set the `UR_LAYER_ASAN_OPTIONS`, they must have intentions to use options other than default. And if we cannot do that, then I think we better just die.

pbalcer: Hm, in that case I think we should use the `.always` function from the logger, so that the error message is always printed on the crash.

yingcong-wu: > We'd better to align the behavior with host sanitizer.

Address sanitizer will also die directly when having format error.

yingcong-wu: > Hm, in that case I think we should use the `.always` function from the logger, so that the error message is always printed on the crash.

Okay, I will test it and update the code later.

PR #2058, opened 2024-09-05
Return uninitialized on context destruction

omarahmed1111:


pbalcer:
It's possible we need to do change some calls to `getContext()` (skips refcount) with calls to `context_t::get();` (doesn't skip refcount and creates a context if doesn't exist).

omarahmed1111:
> It's possible we need to do change some calls to `getContext()` (skips refcount) with calls to `context_t::get();` (doesn't skip refcount and creates a context if doesn't exist).

This won't solve the problem unfortunately as there was another issue of l0 loader being also unloaded early (From further testing, this was the more frequent one).

PR #2062, opened 2024-09-05
[L0] Fix Command List Cache to correctly set In order List property

nrspruit:
- Fix to the emplace call in the context such that the in order list property is not lost from the cache.

### DIFF
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -767,9 +767,11 @@ ur_result_t ur_context_handle_t_::getAvailableCommandList(
         CommandList =
             Queue->CommandListMap
                 .emplace(ZeCommandList,
-                         ur_command_list_info_t(ZeFence, true, false,
-                                                ZeCommandQueue, ZeQueueDesc,
-                                                Queue->useCompletionBatching()))
+                         ur_command_list_info_t(
+                             ZeFence, true, false, ZeCommandQueue, ZeQueueDesc,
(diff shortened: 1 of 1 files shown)


source/adapters/level_zero/context.cpp
```diff
@@ -767,9 +767,11 @@ ur_result_t ur_context_handle_t_::getAvailableCommandList(
         CommandList =
             Queue->CommandListMap
                 .emplace(ZeCommandList,
-                         ur_command_list_info_t(ZeFence, true, false,
-                                                ZeCommandQueue, ZeQueueDesc,
-                                                Queue->useCompletionBatching()))
+                         ur_command_list_info_t(
```
igchor, line 770: Can we remove the default values for params in the ur_command_list_info_t constructor so that modes are always explicitly specified? This would be much less error-prone.

igchor (approved):

source/adapters/level_zero/context.cpp
nrspruit: I agree, I was planning on doing a follow-up patch to remove the defaults since this is error prone unless you know where to check.

igchor: Sounds, good. The only other place where defaults are used seems to be here: https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/queue.cpp#L2262 Everything else compiles once defaults are removed.

pbalcer (approved):

PR #2064, opened 2024-09-05
[L0 v2] add inital memory buffer support

igchor:
That is enought to run most of membuffer tests. This allows us to also test urEventsWait and urEnqueueEventsWait*

Based on: #2052


igchor:
The last remaining benchmark (VectorSum) requires profiling events support.

PR #2067, opened 2024-09-06
[OpenCL] Use older OpenCL-Headers commit to fix build

sarnex:
Build in DPCPP is broken because of the below error caused by the upstream headers change.

Just use the previous commit from the one that breaks the build until the adapter code can be updated.

```
 In file included from /__w/llvm/llvm/build/_deps/unified-runtime-src/source/adapters/opencl/adapter.cpp:11:
/__w/llvm/llvm/build/_deps/unified-runtime-src/source/adapters/opencl/common.hpp:317:29: error: 'cl_mutable_base_config_khr' does not name a type; did you mean 'cl_mutable_dispatch_config_khr'?
  317 |                       const cl_mutable_base_config_khr *mutable_config);
```

### DIFF
--- a/source/adapters/opencl/CMakeLists.txt
+++ b/source/adapters/opencl/CMakeLists.txt
@@ -54,7 +54,7 @@ if(UR_OPENCL_INCLUDE_DIR)
 else()
     FetchContent_Declare(OpenCL-Headers
         GIT_REPOSITORY  "https://github.com/KhronosGroup/OpenCL-Headers.git"
-        GIT_TAG         main
+        GIT_TAG         1e193332d02e27e15812d24ff2a3a7a908eb92a3
     )
     FetchContent_MakeAvailable(OpenCL-Headers)
     FetchContent_GetProperties(OpenCL-Headers


sarnex:
@oneapi-src/unified-runtime-opencl-write  @EwanC  

DPCPP build is broken because of this issue so a fast review is appreciated. After this I have to bump the DPCPP version of UR which hopefully wont break anything else

callumfare (approved):

callumfare:
@sarnex Thanks for the quick fix. We're already using the latest UR commit on intel/llvm so bumping it after this merges shouldn't be an issue

sarnex:
@callumfare Thanks, do you have permission to force merge this without CI? I want to fix DPCPP as fast as possible.

callumfare:
@sarnex I've merged, worst case we can revert it but it should be a very safe change

sarnex:
Thanks!

haonanya:
Do we have plan to Update adapters/opencl/common.hpp to align with latest change of opencl header? @callumfare, @martygrant, filed https://github.com/oneapi-src/unified-runtime/issues/2122, thanks.

EwanC:
> Do we have plan to Update adapters/opencl/common.hpp to align with latest change of opencl header? @callumfare, @martygrant, filed #2122, thanks.

Hi I missed that this issue has been filed so didn't reference it, but https://github.com/oneapi-src/unified-runtime/pull/2095 should do this

PR #2069, opened 2024-09-09
[DeviceSanitizer] Add a report flag to LaunchInfo

zhaomaosu:
This flag is used to control error report process.

SYCLOS Part: https://github.com/intel/llvm/pull/15326


source/loader/layers/sanitizer/asan_libdevice.hpp
```diff
@@ -79,6 +79,7 @@ struct LaunchInfo {
     uint32_t NumLocalArgs = 0;
     LocalArgsInfo *LocalArgs = nullptr; // Ordered by ArgIndex
 
+    int ReportFlag = 0;
```
pbalcer, line 82: Just a general comment: I really dislike the fact that we have an implicit ABI requirement of these structures, where they need to be kept in sync between the two repos.
Would it be possible to refactor this so that we move these structures into a separate header (in include/) that the `libdevice/include/asan_libdevice.hpp` uses?

AllanZyne: @pbalcer, can we make a common include folder in sycl runtime?

pbalcer: There's already a ur-related header in sycl runtime:
https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/ur.hpp
is that what you mean?

Why though? I was more thinking about putting the headers in UR and then including them in SYCL.

AllanZyne: Sorry, I changed my mind. please ignore this.

zhaomaosu: Done. I updated intel/llvm pr, then libdevice will re-use asan_libdevice.hpp in ur.

pbalcer (approved):

yingcong-wu (approved):

PR #2075, opened 2024-09-09
Print CTS invocation to stderr to ease reproduction

ldrumm:


ldrumm:
@oneapi-src/unified-runtime-maintain ping

test/conformance/cts_exe.py
```diff
@@ -2,48 +2,76 @@
 """
  Copyright (C) 2023 Intel Corporation
 
- Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
+ Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM
+ Exceptions.
```
kbenzie, line 6: I think the license string shouldn't change.

```suggestion
 Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
```

kbenzie (changes requested):
> Not sure if the reformatting was appropriate. Happy to roll back the surrounding whitespace changes if you're not into that

Certainly made this harder to grok than it could otherwise have. What formatting tool did you use?

test/conformance/cts_exe.py
ldrumm: I'm going to roll back the unrelated formatting, but in general: why not? The meaning of the words is still the same

ldrumm:
> What formatting tool did you use?

`black`. Think I assumed that since dpc++ used it...
Personally prefer formatting python by hand so happy to roll back

ldrumm:
That bandit failure looks like nonsense.

```
Test results:
>> Issue: [B603:subprocess_without_shell_equals_true] subprocess call - check for execution of untrusted input.
   Severity: Low   Confidence: High
   Location: .\test\conformance\cts_exe.py:49
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b603_subprocess_without_shell_equals_true.html
48	        invocation, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True,
49	        shell=False
50	    )
51	
52	    pat = re.compile(r'\[( )*FAILED( )*\]')

--------------------------------------------------
```
As far as I'm aware this warning is backwards since can't do command injection when `shell=False`

ldrumm:
> That bandit failure looks like nonsense.

Right. So there are two:
- [b602_subprocess_popen_with_shell_equals_true](https://bandit.readthedocs.io/en/latest/plugins/b602_subprocess_popen_with_shell_equals_true.html)
- [b603_subprocess_without_shell_equals_true](https://bandit.readthedocs.io/en/latest/plugins/b603_subprocess_without_shell_equals_true.html)

So basically they don't want us calling subprocesses?

This seems like an unwanted lint in testing infrastructure which relies on executing subprocesses. Can we disable this?

test/conformance/cts_exe.py
```diff
@@ -10,25 +10,44 @@
...
     args = parser.parse_args()
+    invocation = [
+        args.test_command,
+        "--gtest_brief=1",
+        f"--devices_count={args.devices_count}",
+        f"--platforms_count={args.platforms_count}",
+    ]
+    _print_cmdline(invocation, os.environ, os.getcwd())
 
-    result = subprocess.Popen([args.test_command, '--gtest_brief=1',        # nosec B603
```
kbenzie, line 28: > > That bandit failure looks like nonsense.
> 
> Right. So there are two:
> 
>     * [b602_subprocess_popen_with_shell_equals_true](https://bandit.readthedocs.io/en/latest/plugins/b602_subprocess_popen_with_shell_equals_true.html)
> 
>     * [b603_subprocess_without_shell_equals_true](https://bandit.readthedocs.io/en/latest/plugins/b603_subprocess_without_shell_equals_true.html)
> 
> 
> So basically they don't want us calling subprocesses?
> 
> This seems like an unwanted lint in testing infrastructure which relies on executing subprocesses. Can we disable this?

You've deleted the `# nosec B603` comment, I assume this is doing what you ask. As for B602, looks wrong to me. Since `shell=False` is the default, why not just delete it?

ldrumm: Thanks for the hint. Didn't realise that was related. Seems obvious in retrospect

kbenzie (approved):

PR #2080, opened 2024-09-11
fix urEnqueueEventsWaitWithBarrier returning events on failures

pbalcer:
SYCL RT may sometimes ignore errors and call release on returned handles from failing functions. This patch makes sure that the OutEvent remains unmodified when the urEnqueueEventsWaitWithBarrier function returns an error.


nrspruit (approved):
LGTM

Issue #2081, opened 2024-09-11
Add CI testing for builds with multiple adapters


RossBrunton:
Currently, we run a single build per adapter (e.g. one build for OpenCL, one build for Level Zero, etc.), which can miss failures in the multi-adapter dispatcher, ur_ldrddi.cpp (e.g. https://github.com/oneapi-src/unified-runtime/pull/2048 ).

We should add a CI workflow for the Build & Test HW and e2e jobs that use a build that has multiple adapters enabled (although it can use only one adapter for the test).

martygrant:
> (although it can use only one adapter for the test).

Could it be an entirely new job that does two test runs, one for each adapter?

RossBrunton:
@martygrant I could do that, but I'm concerned about how long the job will take to run. Once I have something working I'll have a look at how long it takes.

pbalcer:
Just keep in mind that not all systems have CUDA and HIP installed, and keeping the versions of those things up-to-date might be annoying.

I think one simple solution here would be to use `ONEAPI_DEVICE_SELECTOR` in the tests instead of `UR_ADAPTERS_FORCE_LOAD`: https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_libapi.cpp

The builds always have at least two adapters since the mock adapter is built unconditionally.

PR #2082, opened 2024-09-11
[CI] Add "loader" support to conformance testing

RossBrunton:
This expands our CI to test the loader; the dispatcher that is
used when multiple adapters are available.

The old "run on hardware" jobs should behave the same (they have
the loader tests disabled), but there are new "combined" jobs
with OpenCL/Level Zero + Native CPU.

Closes: #2081

### DIFF
--- a/.github/workflows/build-hw-reusable.yml
+++ b/.github/workflows/build-hw-reusable.yml
@@ -7,10 +7,15 @@ on:
       adapter_name:
         required: true
         type: string
+      other_adapter_name:
+        required: false
+        type: string
+        default: ""
       runner_name:
         required: true
         type: string
       platform:
+        description: "Platform string, `UR_CTS_ADAPTER_PLATFORM` will be set to this."
         required: false
         type: string
         default: ""
@@ -39,9 +44,13 @@ jobs:
     if: github.repository == 'oneapi-src/unified-runtime'  # run only on upstream; forks won't have the HW
     strategy:
       matrix:
-        adapter: [
-          {name: "${{inputs.adapter_name}}", platform: "${{inputs.platform}}", static_Loader: "${{inputs.static_loader}}", static_adapter: "${{inputs.static_loader}}"},
-        ]
+        adapter: [{
+          name: "${{inputs.adapter_name}}",
+          other_name: "${{inputs.other_adapter_name}}",
+          platform: "${{inputs.platform}}",
+          static_Loader: "${{inputs.static_loader}}",
+          static_adapter: "${{inputs.static_loader}}"
+        }]
         build_type: [Debug, Release]
         compiler: [{c: gcc, cxx: g++}, {c: clang, cxx: clang++}]
         # TODO: The latest L0 loader segfaults when built with clang.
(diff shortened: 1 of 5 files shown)


.github/workflows/build-hw-reusable.yml
```diff
@@ -7,13 +7,21 @@ on:
...
         required: true
         type: string
       platform:
         required: false
         type: string
         default: ""
+      other_platform:
+        required: false
+        type: string
+        default: ""
```
omarahmed1111, line 24: I don't see that this value was used in the added jobs, is there a need to have this?

.github/workflows/build-hw-reusable.yml
```diff
@@ -82,8 +95,9 @@ jobs:
         -DUR_ENABLE_TRACING=ON
         -DUR_DEVELOPER_MODE=ON
         -DUR_BUILD_TESTS=ON
-        -DUR_CONFORMANCE_TEST_LOADER=OFF
         -DUR_BUILD_ADAPTER_${{matrix.adapter.name}}=ON
+        -DUR_CONFORMANCE_TEST_LOADER=${{ matrix.adapter.other_name != '' && 'ON' || 'OFF' }}
```
omarahmed1111, line 99: I might be missing something but I don't get why we need to enable another adapter to test the loader? couldn't we just enable it for testing in a job if we need to test it?

.github/workflows/build-hw-reusable.yml
```diff
@@ -7,13 +7,21 @@ on:
       adapter_name:
         required: true
         type: string
+      other_adapter_name:
+        required: false
+        type: string
+        default: ""
```
omarahmed1111, line 13: Not sure if that would complicate things but it feels that if we make adapter name as a list of strings and use it might be more intuitive and expandable in the future than to limit it to only the main adapter and `other_adapter`.

.github/workflows/build-hw-reusable.yml
RossBrunton: UR will not use the loader if only one adapter is built. Multiple adapters in the build are required to make it use the loader.

test/conformance/event/event_adapter_opencl.match
```diff
@@ -0,0 +1,3 @@
+# Note: This file is only for use with cts_exe.py
```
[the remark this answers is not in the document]
RossBrunton: There are two programs now that handle match files; the older matching logic and cts_exe.py. When I updated the files, I did so in a way incompatible with the old match checker thing. Figured I'd add a note calling that out.

Although, happy to remove it now that we've transitioned to using cts_exe.py.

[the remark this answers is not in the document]
omarahmed1111: Ah okay make sense, I think it is better to keep it for now to avoid confusion then with the old matching logic. might be worth also calling out explicitly that this is incompatible with the old matching system. (doesn't have strong opinions about this)

.github/workflows/cmake.yml
```diff
@@ -187,6 +187,24 @@ jobs:
...
 
+  # Native CPU jobs are here to force the loader to be used
+  combined-opencl-native-cpu:
+    name: OpenCL + Native CPU
+    uses: ./.github/workflows/build-hw-reusable.yml
+    with:
+      adapter_name: OPENCL
+      other_adapter_name: NATIVE_CPU
+      runner_name: OPENCL
+      platform: "Intel(R) OpenCL"
```
omarahmed1111, line 198: [NIT] I feel it would be more clear to name the job a name that would relate that we are testing the loader here and explain in a comment that we are enabling multiple adapters to enable the loader to be used.

.github/workflows/build-hw-reusable.yml
omarahmed1111: Ah, got it. Thanks!

.github/workflows/build-hw-reusable.yml
RossBrunton: I've changed it so that "platform" is now directly the value for `UR_CTS_ADAPTER_PLATFORM` rather than rebuilding it.

.github/workflows/build-hw-reusable.yml
RossBrunton: I think that would require UR changes to allow a list of adapters to be specified in a single cmake variable, unless yaml has any fancy string mapping abilities.

omarahmed1111: Cool, I think lets keep it simple better then.

test/conformance/event/event_adapter_opencl.match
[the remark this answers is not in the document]
RossBrunton: https://github.com/oneapi-src/unified-runtime/pull/2285 Decided to go hard in the other direction. Drop support for the older matching logic and make our testing files look neater. Not sure how popular a change this will be though.

omarahmed1111 (approved):

RossBrunton:
Merging without LLVM change, as this only affects our testing.

PR #2076, opened 2024-09-10
Candidate for the v0.10.4 release tag

omarahmed1111:
Includes the following PR:
* https://github.com/oneapi-src/unified-runtime/pull/2009
* https://github.com/oneapi-src/unified-runtime/pull/2011
* https://github.com/oneapi-src/unified-runtime/pull/2083
* https://github.com/oneapi-src/unified-runtime/pull/2086 (0.10.x only PR, not exist in main)
* https://github.com/oneapi-src/unified-runtime/pull/2060

PR #2086, opened 2024-09-12
Bump UMF version to 0.9.0 release

lukaszstolarczuk:
This PR is targeting v0.10.x because it's using UMF v0.9.x stable branch - v0.9.0 release, to be precise.
On the main branch here we use UMF's main as well.

@omarahmed1111, please note this PR will conflict with https://github.com/oneapi-src/unified-runtime/pull/2076 which contains a little older UMF's commit. Please let me know how do you want to handle it.

### DIFF
--- a/source/common/CMakeLists.txt
+++ b/source/common/CMakeLists.txt
@@ -23,8 +23,8 @@ if (NOT DEFINED UMF_REPO)
 endif()
 
 if (NOT DEFINED UMF_TAG)
-    # v0.9.x 19.08.2024: Merge pull request #688 ...
-    set(UMF_TAG 59c4150b7120a7af5b3c8eb2d9b8bbb5d2e96aa3)
+    # v0.9.x 12.09.2024: 0.9.0 release
+    set(UMF_TAG v0.9.0)
 endif()
 
 message(STATUS "Will fetch Unified Memory Framework from ${UMF_REPO}")


bratpiorka (approved):

kbenzie:
> @omarahmed1111, please note this PR will conflict with https://github.com/oneapi-src/unified-runtime/pull/2076 which contains a little older UMF's commit. Please let me know how do you want to handle it.

I think it would make sense to include this #2076

source/common/CMakeLists.txt
```diff
@@ -23,8 +23,8 @@ if (NOT DEFINED UMF_REPO)
 endif()
 
 if (NOT DEFINED UMF_TAG)
-    # v0.9.x 19.08.2024: Merge pull request #688 ...
-    set(UMF_TAG 59c4150b7120a7af5b3c8eb2d9b8bbb5d2e96aa3)
+    # v0.9.x 12.09.2024: 0.9.0 release
+    set(UMF_TAG 57e819999a262b79bcceec47cca24f5015611ea1)
```
kbenzie, line 27: Do you have a tag we could use here instead of the commit hash?

lukaszstolarczuk:
> > @omarahmed1111, please note this PR will conflict with #2076 which contains a little older UMF's commit. Please let me know how do you want to handle it.
> 
> I think it would make sense to include this in #2076

Please note, that I want to target this PR only to v0.10.x branch - on main we use UMF's main.

source/common/CMakeLists.txt
lukaszstolarczuk: yes, of course we have a tag... but we tend to use commits, as tags can be force-pushed, and not all forks/clones contain tags - if you prefer to use a tag I'll do that :)

kbenzie: I think for the release branches using a release tag like [v0.9.0](https://github.com/oneapi-src/unified-memory-framework/releases/tag/v0.9.0) makes it much clearer to see the tag is pointing to the correct place.

I'd hope that release tags aren't getting changed after creation...

kbenzie:
> > > @omarahmed1111, please note this PR will conflict with #2076 which contains a little older UMF's commit. Please let me know how do you want to handle it.
> > 
> > 
> > I think it would make sense to include this in #2076
> 
> Please note, that I want to target this PR only to v0.10.x branch - on main we use UMF's main.

Makes sense 👍

source/common/CMakeLists.txt
lukaszstolarczuk: Yeah, we don't want to force-push tags, but it still possible.

Changed to tag.

omarahmed1111:
@lukaszstolarczuk I cherry-picked this commit to `v0.10.4` PR: https://github.com/oneapi-src/unified-runtime/pull/2076

kbenzie:
This commit has been cherry-picked to v0.10.4rc as 2dfc006e1aaeb03d5e48e9295eac49e8fa865a4e. Closing the PR.

PR #2087, opened 2024-09-12
[L0] Fix urEnqueueEventsWaitWithBarrier for driver in order lists

nrspruit:
- Fix urEnqueueEventsWaitWithBarrier to avoid appending a wait on events given a waitlist, but driver in order lists.
- Driver in order lists gurantees this implicitly and is not needed.


pbalcer:
I think this failed e2e test may be related:
`  SYCL :: Plugin/level_zero_barrier_optimization.cpp`

nrspruit:
> I think this failed e2e test may be related: ` SYCL :: Plugin/level_zero_barrier_optimization.cpp`

Looks like the test passed, but the e2e "file check" failed because I removed the wait on events given driver in order lists.

I will have to fix the e2e test to make that optional.

raiyanla (commented):
LGTM once the e2e test mentioned above is fixed

raiyanla (approved):

PR #2090, opened 2024-09-13
Hardening flags

RossBrunton:
UR_DEVELOPER_MODE has also been changed to only set Werror. The
previous other flags set by this variable either have no effect
(-fno-omit-frame-pointer) or are now always enabled
(-fstack-protector-strong).

Some smaller warnings have also been fixed.

### DIFF
--- a/cmake/helpers.cmake
+++ b/cmake/helpers.cmake
@@ -60,25 +60,39 @@ endmacro()
 
 function(add_ur_target_compile_options name)
     if(NOT MSVC)
+        target_compile_definitions(${name} PRIVATE -D_FORTIFY_SOURCE=2)
         target_compile_options(${name} PRIVATE
-            -fPIC
+            # Warning options
             -Wall
             -Wpedantic
             -Wempty-body
+            -Wformat
+            -Wformat-security
             -Wunused-parameter
+
+            # Hardening options
+            -fPIC
+            -fstack-protector-strong
+            -fvisibility=hidden # Required for -fsanitize=cfi
+            # -fsanitize=cfi requires -flto, which breaks a lot of things
+            # See: https://github.com/oneapi-src/unified-runtime/issues/2120
+            # -flto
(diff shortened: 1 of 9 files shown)


cmake/helpers.cmake
```diff
@@ -59,17 +59,32 @@ endmacro()
...
+                -fcf-protection=full
+                -fstack-protector-strong
+                -fvisibility=hidden # Required for -fsanitize=cfi
+                # -fsanitize=cfi requires -flto, which breaks a lot of things
+                # See: https://github.com/oneapi-src/unified-runtime/issues/2120
+                # -flto
+                # $<$<CXX_COMPILER_ID:Clang,AppleClang>:-fsanitize=cfi>
+            )
+            if (CMAKE_BUILD_TYPE STREQUAL "Release")
+                target_compile_options(${name} PRIVATE -Werror)
```
aarongreig, line 84: is this just for convenience if you're hacking around in debug? I'd vote for enabling it all the time if we can

source/adapters/level_zero/CMakeLists.txt
```diff
@@ -66,6 +66,21 @@ if(UR_BUILD_ADAPTER_L0)
         )
     endif()
 
+    # Ensure UR flags are propogated to level zero
```
aarongreig, line 69:
```suggestion
    # Ensure UR flags are propagated to level zero
```

kbenzie (approved):

RossBrunton:
@oneapi-src/unified-runtime-level-zero-write Are you okay with this change? It's adding a bunch of "hardening" flags to the loader targets.

pbalcer:
> @oneapi-src/unified-runtime-level-zero-write Are you okay with this change? It's adding a bunch of "hardening" flags to the loader targets.

We can do that, but it will only affect UR and intel/llvm builds. Normally the loader is shipped from the driver package.

My (very slight) preference would be that we don't set any flags that might affect the performance or cause difference in behavior compared to the unmodified loader.

pbalcer (approved):

PR #2096, opened 2024-09-16
[L0] Fix urEnqueueDeviceGlobalVariableWrite to use device specific mo…

nrspruit:
…dules

- Fixed urEnqueueDeviceGlobalVariableWrite to check the map for a device specific module for the device used by the queue if it exists, otherwise use the ZeModule in the Program handle.

### DIFF
--- a/source/adapters/level_zero/kernel.cpp
+++ b/source/adapters/level_zero/kernel.cpp
@@ -494,11 +494,19 @@ ur_result_t urEnqueueDeviceGlobalVariableWrite(
 ) {
   std::scoped_lock<ur_shared_mutex> lock(Queue->Mutex);
 
+  ze_module_handle_t ZeModule{};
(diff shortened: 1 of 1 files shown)


pbalcer (approved):

raiyanla (approved):

source/adapters/level_zero/kernel.cpp
```diff
@@ -494,11 +494,19 @@ ur_result_t urEnqueueDeviceGlobalVariableWrite(
 ) {
   std::scoped_lock<ur_shared_mutex> lock(Queue->Mutex);
 
+  ze_module_handle_t ZeModule{};
```
zhaomaosu, line 497: Do we need to apply this fix to urEnqueueDeviceGlobalVariableRead() as well?

PR #2100, opened 2024-09-18
Raise MSVC warning level from /W3 to /W4

kbenzie:
This patch increases the warning level when using the MSVC compiler from `/W3` to `/W4` and fixes the issues found. Four warnings introduced by `/W4` are disabled, all related to variable name shadowing, as they are overly prescriptive to valid code.

https://github.com/intel/llvm/pull/15745


source/common/ur_util.hpp
```diff
@@ -98,8 +98,9 @@ std::optional<std::string> ur_getenv(const char *name);
 
 inline bool getenv_tobool(const char *name, bool def = false) {
     if (auto env = ur_getenv(name); env) {
-        std::transform(env->begin(), env->end(), env->begin(),
-                       [](unsigned char c) { return std::tolower(c); });
+        std::transform(
+            env->begin(), env->end(), env->begin(),
+            [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
```
kbenzie, line 103: Should be `char` not `unsigned char`.

kbenzie:
> On GCC, `-WExtra -WError` has an issue where newer versions of the compiler might introduce new warnings (or older versions might have false positives), so you end up creating a hidden dependency on a specific version of GCC. Is this an issue with MSVC as well? And if so, is it worth having a separate CMake var to turn on `-WExtra` / `/W4`?

I don't think extra warnings being introduced by new compilers is actually an issue, by default warnings are not treated as errors, that's only enabled with the `UR_DEVELOPER_MODE=ON` option. We enable that in UR CI but not downstream.

test/conformance/exp_command_buffer/update/buffer_fill_kernel_update.cpp
```diff
@@ -336,7 +336,7 @@ TEST_P(BufferFillCommandTest, OverrideUpdate) {
     ASSERT_SUCCESS(urCommandBufferUpdateKernelLaunchExp(command_handle,
                                                         &first_update_desc));
 
-    uint32_t second_val = -99;
```
EwanC, line 339: I think the usage of a negative value here and in the next test was an oversight, and using `0xffffff9d; // -99` is just making the test less readable than a simple unsigned value. Could we change the `-99` to `99` (or any unsigned value other than `33` from `first_val` really)

kbenzie: These are both now using `99`.

EwanC (approved):

martygrant (approved):

kbenzie:
@oneapi-src/unified-runtime-level-zero-write do these changes look okay?

ProGTX (approved):

cppchedy (approved):

source/adapters/cuda/command_buffer.cpp
```diff
@@ -242,7 +242,7 @@ static ur_result_t enqueueCommandBufferFillHelper(
     if ((PatternSize == 1) || (PatternSize == 2) || (PatternSize == 4)) {
       CUDA_MEMSET_NODE_PARAMS NodeParams = {};
       NodeParams.dst = DstPtr;
-      NodeParams.elementSize = PatternSize;
+      NodeParams.elementSize = static_cast<unsigned int>(PatternSize);
```
jchlanda, line 245: I'm guessing that the variables (`PatternSize`, `gridDims`, `blockDims` and friends) are casted from `size_t`, that seems more suitable type than `unsigned int`. Perhaps, outside of the scope of this PR, but it feels like adjusting the types in node params structs would be a more elegant solution.

jchlanda (approved):

source/adapters/cuda/command_buffer.cpp
jchlanda: Actually, maybe not, native (cuda) calls seem to prefer [`unsigned int`](https://github.com/oneapi-src/unified-runtime/pull/2100/files#diff-9dcf8810f13f5d1a1049e13dc684ccc23bb4739a8c3efdb1ae46e64475f5cc18R484). It's probably not worth trying to align all the different apis, never mind.

kbenzie: Yeah its probably an impossible job having all the types align.

pbalcer (approved):

PR #2102, opened 2024-09-19
Improvements to align CTS and Spec for Loader

martygrant:
- Add a null pointer check for urLoaderConfigCreate
- Add a null pointer and invalid size check for urLoaderConfigGetinfo
- Add invalid size test for urLoaderConfigGetInfo
- Add reference count test for urLoaderConfigGetInfo
- Add new test for urLoaderConfigSetCodeLocationCallback
- Add new test for UR_LOADER_CONFIG_INFO_AVAILABLE_LAYERS to test the list is valid (empty or valid values)
- Updated doc wording to say no layers are enabled by default

### DIFF
--- a/test/loader/loader_config/urLoaderConfigGetInfo.cpp
+++ b/test/loader/loader_config/urLoaderConfigGetInfo.cpp
@@ -5,7 +5,7 @@
 
 #include "fixtures.hpp"
 
-struct urLoaderConfigGetInfoTest
+struct urLoaderConfigGetInfoWithParamTest
     : LoaderConfigTest,
       ::testing::WithParamInterface<ur_loader_config_info_t> {
     void SetUp() override {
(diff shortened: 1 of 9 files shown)


test/loader/loader_config/urLoaderConfigGetInfo.cpp
```diff
@@ -23,30 +23,115 @@ struct urLoaderConfigGetInfoTest
...
+    ASSERT_EQ(UR_RESULT_ERROR_INVALID_SIZE,
+              urLoaderConfigGetInfo(loaderConfig, infoType, 0,
+                                    infoAllocation.data(), &infoSize));
+
+    ASSERT_EQ(UR_RESULT_ERROR_INVALID_SIZE,
+              urLoaderConfigGetInfo(loaderConfig, infoType, 1,
+                                    infoAllocation.data(), &infoSize));
+}
+
+struct urLoaderConfigGetInfoTest : urLoaderConfigGetInfoWithParamTest {};
```
aarongreig, line 68: instead of this you can do `using urLoaderConfigGetInfoTest = LoaderConfigTest;` and then you don't need to do the `INSTANTIATE_TEST_SUITE` stuff, you can just define the tests like `TEST_F(urLoaderConfigGetInfoTest, nameOfTest) {`

test/loader/loader_config/urLoaderConfigGetInfo.cpp
```diff
@@ -23,30 +23,115 @@ struct urLoaderConfigGetInfoTest
...
                                     nullptr, &infoSize));
 }
+
+TEST_P(urLoaderConfigGetInfoWithParamTest, InvalidSize) {
+    ASSERT_EQ(UR_RESULT_ERROR_INVALID_SIZE,
+              urLoaderConfigGetInfo(loaderConfig, infoType, 0,
+                                    infoAllocation.data(), &infoSize));
+
+    ASSERT_EQ(UR_RESULT_ERROR_INVALID_SIZE,
+              urLoaderConfigGetInfo(loaderConfig, infoType, 1,
```
aarongreig, line 64: rather than `1` for these cases it's better to do `infoSize - 1` or something similar, to guarantee that the size is incorrect

test/loader/loader_config/urLoaderConfigGetInfo.cpp
martygrant: yeah googletest can be a bit awkward sometimes, think I done your suggestion in a later PR as well when I noticed this option. Updated it, cheers!

test/loader/loader_config/urLoaderConfigGetInfo.cpp
martygrant: makes sense, updated to this, thanks!

aarongreig (approved):

Issue #2103, opened 2024-09-19
Re-enable crashing level zero tests


RossBrunton:
A number of tests in the CTS cause crashes or memory corruption with the level zero backend. This causes the entire gtest process to die, which results in any tests that would run after the crashing test not doing so. As well as "hiding" results from tests, this causes the failures reported (for use in our CI match scripts) to change based on which order tests are ran in (this can happen if `GTEST_SHUFFLE` is specified, or the presence/absence of `-flto`).

These tests have been disabled, and this issue is a tracker for enabling them again once the crashing issues have been resolved:

- [ ] urEnqueueEventsWaitTest.Success
- [ ] urEnqueueEventsWaitWithBarrierTest.Success
- [ ] urEnqueueMemBufferCopyTestWithParam.Success
- [ ] urEnqueueMemBufferCopyRectTestWithParam.Success
- [ ] urEnqueueMemBufferFillTest.Success
- [ ] urEnqueueMemBufferFillTest.SuccessPartialFill
- [ ] urEnqueueMemBufferFillTest.SuccessOffset
- [ ] urEnqueueMemBufferWriteRectTestWithParam.Success
- [ ] urEnqueueMemImageCopyTest.Success
- [ ] urEnqueueMemImageCopyTest.SuccessPartialCopy
- [ ] urEnqueueMemImageCopyTest.SuccessPartialCopyWithSrcOffset
- [ ] urEnqueueMemImageCopyTest.SuccessPartialCopyWithDstOffset
- [ ] urEnqueueMemImageReadTest.InvalidRegion1D
- [ ] urEnqueueMemImageReadTest.InvalidRegion2D
- [ ] urEnqueueMemImageReadTest.InvalidRegion3D
- [ ] urEventCreateWithNativeHandleTest.Success
- [ ] urEventGetInfoTest.Success
- [ ] urMemImageCreateTest.InvalidSize
- [ ] urMemImageCreateWithHostPtrFlagsTest.Success
- [ ] urMemImageCreateTestWithImageFormatParam.Success

aleroldan156:
I think this will help you.
https://bit.ly/3TyLBMS
If you don't have the c compliator, install it.(gcc or clang)

RossBrunton:
This is out of date, and our testing now considers crashing tests as failures (and thus puts them in the match files).

PR #2104, opened 2024-09-19
Disable crashing l0 tests

RossBrunton:
See: #2103


RossBrunton:
Probably going to try and do this in a different way. Closing.

PR #2108, opened 2024-09-20
fix perf regression in tracing layer

pbalcer:
In the current tracing layer, the parameters of the function being called are always printed to a string stream that may be used for logging. However, if logs are disabled, this just unnecessarily wastes times.


kbenzie (approved):

pbalcer:
the build failure is unrelated, i'll restart once it all passes.

kbenzie:
> the build failure is unrelated, i'll restart once it all passes.

This type of failure where its specific to one runner/config is why I'd like to split up the workflows so its easier and faster to retry those jobs.

PR #2110, opened 2024-09-20
[Cuda] Simplify and optimise the `getMaxRegistersJitOptionValue` utility

GeorgeWeb:
This change simplifies the `getMaxRegistersJitOptionValue` function as it was previously unclear what the return bool value actually meant, because it doesn't really signify whether the value was valid or not but it was more of a whether the parsing was successful. Instead, we now just return the parsed value if it is successfully parsed, otherwise we return the default.
Additionally, we replace the use of `std::stoi` with `std::from_chars` which is much lighter weight (and works well with `std::string_view`s), in order to avoid extra dynamic allocation(s).

### DIFF
--- a/source/adapters/cuda/program.cpp
+++ b/source/adapters/cuda/program.cpp
@@ -13,47 +13,50 @@
 
 #include <algorithm>
 #include <array>
+#include <charconv>
 #include <cstdint>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


martygrant:
## Information

## PR Migration

---

PR #2117, opened 2024-09-23
Fix urEnqueueEventsWaitWithBarrier when used with interop events

pbalcer:
This patch makes sure that only events from the exact same
cmdlists are filtered out when appending events to wait.

### DIFF
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -171,48 +171,63 @@ ur_result_t urEnqueueEventsWaitWithBarrier(
   std::scoped_lock<ur_shared_mutex> lock(Queue->Mutex);
 
   // Helper function for appending a barrier to a command list.
-  auto insertBarrierIntoCmdList =
-      [&Queue](ur_command_list_ptr_t CmdList,
-               const _ur_ze_event_list_t &EventWaitList,
-               ur_event_handle_t &Event, bool IsInternal) {
-        UR_CALL(createEventAndAssociateQueue(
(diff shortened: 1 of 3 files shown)


nrspruit (approved):

source/adapters/level_zero/event.cpp
```diff
@@ -171,48 +171,63 @@ ur_result_t urEnqueueEventsWaitWithBarrier(
...
+    // TODO: this and other special handling of in-order queues to be
+    // updated when/if Level Zero adds native support for in-order queues.
+    //
+    if (Queue->isInOrderQueue() && InOrderBarrierBySignal &&
+        !Queue->isProfilingEnabled()) {
+      // If we are using driver in order lists, then append wait on events
+      // is unnecessary IF the cmdlists match.
+      if (EventWaitList.Length) {
+        if (CmdList->second.IsInOrderList) {
+          for (unsigned i = EventWaitList.Length; i-- < 0;) {
```
bader, line 203: /unified-runtime-src/source/adapters/level_zero/event.cpp:203:55: warning: comparison of unsigned expression in ‘< 0’ is always false [-Wtype-limits]
  203 |           for (unsigned i = EventWaitList.Length; i-- < 0;) {
      |                                                   ~~~~^~~

pbalcer: Yup, we are aware of that typo. Thanks. This patch will solve the problem: https://github.com/oneapi-src/unified-runtime/pull/2134

PR #2118, opened 2024-09-23
 Fix urEnqueueEventsWaitWithBarrier when used with interop events (simpler)

pbalcer:
This patch partially reverts changes made in https://github.com/oneapi-src/unified-runtime/pull/2087 and allows
all events to be appended for waiting.


pbalcer:
We went with #2117, closing.

PR #2121, opened 2024-09-24
[L0] Refcnt Parent Buffer on Sub Buffer Create and die on use of buffer after free

nrspruit:
- Refcnt the Parent Buffer on Sub Buffer Creation to avoid free until
  after all sub buffers are released.
- Avoid use after free of buffers by tracking if a buffer has already
  been called with free() and verify that the parent of a sub buffer has
not already been released before attempting to get ze handles from the
parent.
- In either use after free case, die aborts before the memory is accessed.

### DIFF
--- a/source/adapters/level_zero/memory.cpp
+++ b/source/adapters/level_zero/memory.cpp
@@ -2045,19 +2045,30 @@ ur_result_t _ur_buffer::getZeHandle(char *&ZeHandle, access_mode_t AccessMode,
 
   auto &Allocation = Allocations[Device];
 
+  if (this->isFreed) {
+    die("getZeHandle() buffer already released, no valid handles.");
+  }
+
   // Sub-buffers don't maintain own allocations but rely on parent buffer.
   if (SubBuffer) {
-    UR_CALL(SubBuffer->Parent->getZeHandle(ZeHandle, AccessMode, Device,
-                                           phWaitEvents, numWaitEvents));
-    ZeHandle += SubBuffer->Origin;
-    // Still store the allocation info in the PI sub-buffer for
-    // getZeHandlePtr to work. At least zeKernelSetArgumentValue needs to
-    // be given a pointer to the allocation handle rather than its value.
-    //
-    Allocation.ZeHandle = ZeHandle;
-    Allocation.ReleaseAction = allocation_t::keep;
-    LastDeviceWithValidAllocation = Device;
-    return UR_RESULT_SUCCESS;
+    // Verify that the Parent Buffer is still valid or if it has been freed.
+    if (SubBuffer->Parent && !SubBuffer->Parent->isFreed) {
+      UR_CALL(SubBuffer->Parent->getZeHandle(ZeHandle, AccessMode, Device,
+                                             phWaitEvents, numWaitEvents));
+      ZeHandle += SubBuffer->Origin;
+      // Still store the allocation info in the PI sub-buffer for
+      // getZeHandlePtr to work. At least zeKernelSetArgumentValue needs to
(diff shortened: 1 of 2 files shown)


nrspruit:
Marking this as ready to review since this is needed to properly throw a recoverable error given the parent of a sub buffer has been released.

source/adapters/level_zero/memory.cpp
```diff
@@ -1682,6 +1682,7 @@ ur_result_t urMemRelease(
     Buffer->free();
   }
   delete Mem;
+  Mem = nullptr;
```
pbalcer, line 1685: this is a local variable passed by value, not by reference, so this assignment has no effect.

source/adapters/level_zero/memory.cpp
```diff
@@ -2024,18 +2025,37 @@ ur_result_t _ur_buffer::getZeHandle(char *&ZeHandle, access_mode_t AccessMode,
 
   auto &Allocation = Allocations[Device];
 
+  if (this->isFreed) {
```
pbalcer, line 2028: Just to make sure I understand, this will only ever be set to true after the application calls `urMemRelease` on `this` handle, right? In other words, this will only ever be true in case of a use-after-free on this handle? So essentially, this is hardening against a situation that should not be occurring.

I'm not against adding this if here, but we should die() with "use-after-free" message, not return back to the application.
The reason is that if we return back to the application, it may never know that it's doing something incorrectly and come to expect this broken behavior from the adapter. All memory allocators that I know of (except for something like mmap) will fail fast with an assert instead of trying to continue in a broken state.

Additionally, UR_LAYER_LIFETIME_VALIDATION layer should already be able to tell the user that they are doing a use-after-free or using an invalid handle. So instead of trying to create workarounds, we should be encouraging users to enable the full validation layer in tests to find issues like this.

nrspruit: Thank you for the explanation, let me adjust the changes to assert that way this situation is caught sooner.

nrspruit:
Hello @pbalcer , please re-review, this addressed the segfault situation and instead asserts. See if you see a problem with my logic.

pbalcer (approved):

source/adapters/level_zero/memory.cpp
```diff
@@ -2045,19 +2045,37 @@ ur_result_t _ur_buffer::getZeHandle(char *&ZeHandle, access_mode_t AccessMode,
...
+    logger::debug("getZeHandle(ur_device_handle{}) buffer already released, no "
+                  "valid handles.",
+                  (void *)Device);
+    UR_ASSERT(!this->isFreed, UR_RESULT_ERROR_INVALID_NULL_HANDLE);
```
pbalcer, line 2052: `UR_ASSERT` is not a real assert: https://github.com/oneapi-src/unified-runtime/blob/d9608131d61159952db00bead44efffd3fab0ee3/source/adapters/level_zero/common.hpp#L306 We should get rid of that at some point, but in here I think we need to use [`die()`](https://github.com/oneapi-src/unified-runtime/blob/d9608131d61159952db00bead44efffd3fab0ee3/source/ur/ur.hpp#L63).

pbalcer (changes requested):

source/adapters/level_zero/memory.cpp
nrspruit: Hello @pbalcer , I have updated the code to die() in these cases and I have added a missing refcnt during sub buffer creation such that this issue should be avoided in most cases, meaning the die() is now simply a guard.

pbalcer (approved):

pbalcer (approved):

PR #2130, opened 2024-09-25
Fix `for` in level_zero EventWaitList handling

RossBrunton:


omarahmed1111 (approved):

nrspruit (approved):

pbalcer (commented):
why?

RossBrunton:
> why?

`i-- < 0` is always false, since i is unsigned. Even if it was signed, it would still be false anyway, since i is always negative. I think this was meant to be a greater than rather than a less than?

pbalcer (approved):

alycm:
> How is this passing tests?

Can a test be added to trigger this issue?  It seems like maybe this entire case is currently missed (or would indeed surely be failing).

pbalcer:
> > How is this passing tests?
> 
> Can a test be added to trigger this issue? It seems like maybe this entire case is currently missed (or would indeed surely be failing).

I added a test for this code path, but what that test verified is that the correct events are *not* removed from the list, not whether the intended events are removed. So not removing any events makes the test pass :)

This loop is intended to remove unneeded events from a waitlist. It's a performance optimization and may be needed to workaround (what I think is) a driver issue.

nrspruit:
preferred change:
https://github.com/oneapi-src/unified-runtime/pull/2134

PR #2131, opened 2024-09-25
[EXP][CMDBUF] Make command handle behaviour consistent

Bensuo:
- Attempting to obtain a command handle when the command buffer is not updatable is now an error across all adapters
- command-buffer spec and adapter code updated to reflect this
- invalid_update CTS test updated to reflect new behaviour

### DIFF
--- a/scripts/core/exp-command-buffer.yml
+++ b/scripts/core/exp-command-buffer.yml
@@ -111,6 +111,12 @@ etors:
             [uint32_t] Reference count of the command-buffer object.
             The reference count returned should be considered immediately stale.
             It is unsuitable for general use in applications. This feature is provided for identifying memory leaks.
+    - name: DESCRIPTOR
+      desc: |
+            [$x_exp_command_buffer_desc_t] Returns a $x_exp_command_buffer_desc_t
+            with the properties of the command-buffer. Returned values may differ
+            from those passed on construction if the property was ignored by the
+            adapter.
 --- #--------------------------------------------------------------------------
 type: enum
 desc: "Command-buffer command query information type"
@@ -380,7 +386,8 @@ params:
       desc: "[out][optional] return an event object that will be signaled by the completion of this command in the next execution of the command-buffer."
     - type: "$x_exp_command_buffer_command_handle_t*"
       name: phCommand
-      desc: "[out][optional] Handle to this command."
+      desc: "[out][optional] Handle to this command. Only available if the
+      command-buffer is updatable."
 returns:
     - $X_RESULT_ERROR_INVALID_COMMAND_BUFFER_EXP
     - $X_RESULT_ERROR_INVALID_KERNEL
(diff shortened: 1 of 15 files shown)


test/conformance/exp_command_buffer/invalid_update.cpp
```diff
@@ -151,11 +155,11 @@ TEST_P(InvalidUpdateTest, NotUpdatableCommandBuffer) {
...
     };
 
-    // Update command to command-buffer that doesn't have updatable set should
-    // be an error
+    // Since no command handle was returned Update command to command-buffer
+    // should also be an error.
     ur_result_t result =
         urCommandBufferUpdateKernelLaunchExp(test_command_handle, &update_desc);
-    EXPECT_EQ(UR_RESULT_ERROR_INVALID_OPERATION, result);
+    EXPECT_EQ(UR_RESULT_ERROR_INVALID_NULL_HANDLE, result);
```
EwanC, line 162: I can't see `UR_RESULT_ERROR_INVALID_NULL_HANDLE` in the YML errors for `UpdateKernelLaunchExp`, could you add it (wonder if this means we don't need the `UR_RESULT_ERROR_INVALID_COMMAND_BUFFER_COMMAND_HANDLE_EXP`  error code, but that doesn't need to stop this PR)

Bensuo: This is handled automatically by the validation layer with any handle parameter, and I believe the expected best practice is not to clutter the YAML or force runtime checks for this for that reason.

EwanC (approved):

Bensuo:
DPC++ PR: https://github.com/intel/llvm/pull/15522

scripts/core/exp-command-buffer.yml
```diff
@@ -84,6 +84,10 @@ etors:
             [uint32_t] Reference count of the command-buffer object.
             The reference count returned should be considered immediately stale.
             It is unsuitable for general use in applications. This feature is provided for identifying memory leaks.
+    - name: IS_UPDATABLE
+      desc: |
+            [$x_bool_t] Returns the value of $x_exp_command_buffer_desc_t::isUpdatable
+            that the command-buffer was created with.
```
EwanC, line 90: Preference for this query returning all of the `$x_exp_command_buffer_desc_t` struct rather than just the individual member, if possible, as it is a more extensible design if we want to query other parts of it in future. If it's a pain to reconstruct this in the adapters code though it's not a big deal though.

Bensuo: I've gone with the latter approach for now (returning false if ignored), but can easily change to storing the unsused properties if that's your preference.

EwanC (approved):

aarongreig:
looking at merging this soon if you can fix the conflicts @Bensuo

Bensuo:
> looking at merging this soon if you can fix the conflicts @Bensuo

Thanks! I've updated the branch and will keep an eye on the CI while it runs.

PR #2133, opened 2024-09-25
fix zeCommandListAppendWaitOnEvents failing in rare scenarios

pbalcer:
This will make it so that we don't pass any events to zeCommandListAppendWaitOnEvents in case of driver-in order lists, except for interop events created through a native handle.


pbalcer:
This also includes the `for` loop fix from #2130. This time I checked that it's working under a debugger and added a print for good measure :D

nrspruit:
preferred change:
https://github.com/oneapi-src/unified-runtime/pull/2134

PR #2134, opened 2024-09-25
[L0] Fix Multi Device Event handling and remove unhandled events from in order wait list

nrspruit:
- fix multi device event creation to be enqueued as a barrier waiting on the original event keeping the multi device event valid.
- Avoid wait on events on a multi device event given driver in order due to p2p errors for driver in order lists. Barrier from multi device event creation will guarantee in order execution.


raiyanla (approved):

pbalcer (approved):
I like this, makes sense!

PR #2137, opened 2024-09-26
Don't use empty match test files

RossBrunton:
This patch removes any empty ".match" files, and instructs ctest to
run the tests verbatim if there is no match file available.

As well as providing more useful output, this also handles the case
where the test startup and/or teardown causes the test runner to fail.
Previously, this would be ignored (as it produced no output).


pbalcer (approved):

martygrant (approved):

RossBrunton:
Merging without LLVM commit since this doesn't touch LLVM.

PR #2139, opened 2024-09-26
[L0] Pass and track event dependencies required before executing Memory Copy buffer inits

nrspruit:
- To ensure buffers are always updated with the correct buffer values, track the dependencies as a prerequisite before performing the memory copy.

### DIFF
--- a/source/adapters/level_zero/memory.cpp
+++ b/source/adapters/level_zero/memory.cpp
@@ -407,7 +407,8 @@ static ur_result_t enqueueMemImageCommandHelper(
 
     char *ZeHandleSrc = nullptr;
     UR_CALL(SrcMem->getZeHandle(ZeHandleSrc, ur_mem_handle_t_::read_only,
-                                Queue->Device));
+                                Queue->Device, EventWaitList,
+                                NumEventsInWaitList));
     ZE2UR_CALL(zeCommandListAppendImageCopyToMemory,
                (ZeCommandList, Dst, ur_cast<ze_image_handle_t>(ZeHandleSrc),
                 &ZeSrcRegion, ZeEvent, WaitList.Length, WaitList.ZeEventList));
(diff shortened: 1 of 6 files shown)


igchor (approved):
It would be good to have a UR-only test for this - also to avoid making the same mistake in L0 v2.

nrspruit:
> It would be good to have a UR-only test for this - also to avoid making the same mistake in L0 v2.

Agreed, I don't have a good test off the top of my head so I will have to think about that and I may have a test in a follow-up PR.

igchor:
It actually looks like we already have a simple test for this, we just haven't run it CI. This PR should enable it: https://github.com/oneapi-src/unified-runtime/pull/2140 Locally this tests fails randomly on upstream UR and on this PR I get a segfault.

We still might want to add some more complex tests in future.

nrspruit:
We want to enable https://github.com/oneapi-src/unified-runtime/pull/2140 to catch this type of issue going forward.

igchor (approved):

source/adapters/level_zero/memory.cpp
```diff
@@ -2151,6 +2175,33 @@ ur_result_t _ur_buffer::getZeHandle(char *&ZeHandle, access_mode_t AccessMode,
...
+          if (waitlist.ZeEventList == nullptr) {
+            waitlist.ZeEventList = new ze_event_handle_t[numWaitEvents];
+          }
+          waitlist.ZeEventList[EventListIndex] = phWaitEvents[i]->ZeEvent;
+          waitlist.Length++;
+          EventListIndex++;
+        }
+      }
+      if (waitlist.Length > 0) {
+        ZE2UR_CALL(zeCommandListAppendWaitOnEvents,
```
pbalcer, line 2200: We could use a test for this scenario, since I don't believe it's covered by anything existing.
I made a test that creates a native L0 event here: https://github.com/oneapi-src/unified-runtime/blob/main/test/adapters/level_zero/urEventCreateWithNativeHandle.cpp#L24. This could be generalized.

pbalcer (approved):

source/adapters/level_zero/memory.cpp
nrspruit: I will see about adding that in a follow-up PR.

nrspruit:
e2e "failures" seem unrelated, unexpected pass:
********************
Unexpectedly Passed Tests (2):
  SYCL :: DeviceCodeSplit/grf.cpp
  SYCL :: KernelAndProgram/target_register_alloc_mode.cpp

PR #2143, opened 2024-09-27
Return more accurate errors from CL's ProgramCreateWithIL

aarongreig:
clCreateProgramWithIL only returns CL_INVALID_VALUE in three specific cirumstances, based on this we can deduce a more informative error code when we get that return.

Fixes the urProgramCreateWithILTest.BuildInvalidProgram cts test.


aarongreig:
LLVM pr: https://github.com/intel/llvm/pull/15534

RossBrunton (approved):

PR #2146, opened 2024-09-27
Compute-benchmarks use already compiled UR

lslusarczyk:
Utilize support for using UR binary through find_package added in  https://github.com/intel/compute-benchmarks/pull/21

Before this change compute-benchmarks always builds UR from sources which is unnecessary since we already build UR as part of SYCL on UR CI.

Add installing adapters by "cmake --install" which is needed by new benchmarks code.

### DIFF
--- a/scripts/benchmarks/benches/base.py
+++ b/scripts/benchmarks/benches/base.py
@@ -15,12 +15,27 @@ import tarfile
 class Benchmark:
     def __init__(self, directory):
         self.directory = directory
-        self.adapter_path = os.path.join(options.ur_dir, 'build', 'lib', f"libur_adapter_{options.ur_adapter_name}.so")
+
+    @staticmethod
+    def get_adapter_full_path():
+        for libs_dir_name in ['lib', 'lib64']:
+            adapter_path = os.path.join(
+                options.ur_dir, libs_dir_name, f"libur_adapter_{options.ur_adapter_name}.so")
+            if os.path.isfile(adapter_path):
+                return adapter_path
+        assert False, \
+            f"could not find adapter file {adapter_path} (and in similar lib paths)"
 
     def run_bench(self, command, env_vars):
         env_vars_with_forced_adapter = env_vars.copy()
-        env_vars_with_forced_adapter.update({'UR_ADAPTERS_FORCE_LOAD': self.adapter_path})
-        return run(command=command, env_vars=env_vars_with_forced_adapter, add_sycl=True, cwd=options.benchmark_cwd).stdout.decode()
+        env_vars_with_forced_adapter.update(
+            {'UR_ADAPTERS_FORCE_LOAD': Benchmark.get_adapter_full_path()})
+        return run(
+            command=command,
+            env_vars=env_vars_with_forced_adapter,
+            add_sycl=True,
+            cwd=options.benchmark_cwd
+        ).stdout.decode()
 
     def create_data_path(self, name):
         data_path = os.path.join(self.directory, "data", name)
--- a/.github/workflows/benchmarks_compute.yml
+++ b/.github/workflows/benchmarks_compute.yml
@@ -152,20 +152,31 @@ jobs:
       run: cmake --build ${{github.workspace}}/sycl_build -j
 
     - name: Configure UR
-      working-directory: ${{github.workspace}}/ur-repo
       run: >
         cmake -DCMAKE_BUILD_TYPE=Release
-        -B${{github.workspace}}/ur-repo/build
+        -S${{github.workspace}}/ur-repo
+        -B${{github.workspace}}/ur_build
+        -DCMAKE_INSTALL_PREFIX=${{github.workspace}}/ur_install
         -DUR_BUILD_TESTS=OFF
         -DUR_BUILD_ADAPTER_L0=ON
         -DUR_BUILD_ADAPTER_L0_V2=ON
+        -DUMF_DISABLE_HWLOC=ON
 
     - name: Build UR
-      run: cmake --build ${{github.workspace}}/ur-repo/build -j $(nproc)
+      run: cmake --build ${{github.workspace}}/ur_build -j $(nproc)
+
+    - name: Install UR
+      run: cmake --install ${{github.workspace}}/ur_build
 
     - name: Run benchmarks
       id: benchmarks
-      run: numactl -N 0 ${{ github.workspace }}/ur-repo/scripts/benchmarks/main.py ~/bench_workdir ${{github.workspace}}/sycl_build ${{github.workspace}}/ur-repo ${{ matrix.adapter.str_name }} ${{ inputs.bench_script_params }}
+      run: >
+        numactl -N 0 ${{ github.workspace }}/ur-repo/scripts/benchmarks/main.py 
+        ~/bench_workdir 
+        ${{github.workspace}}/sycl_build
+        ${{github.workspace}}/ur_install
+        ${{ matrix.adapter.str_name }}
+        ${{ inputs.bench_script_params }}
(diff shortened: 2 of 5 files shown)


scripts/benchmarks/benches/base.py
```diff
@@ -16,7 +16,7 @@
 class Benchmark:
     def __init__(self, directory):
         self.directory = directory
-        self.adapter_path = os.path.join(options.ur_dir, 'build', 'lib', f"libur_adapter_{options.ur_adapter_name}.so")
+        self.adapter_path = os.path.join(options.ur_dir, 'lib64', f"libur_adapter_{options.ur_adapter_name}.so")
```
igchor, line 19: It can also be in 'lib' on some systems I think

source/adapters/cuda/CMakeLists.txt
```diff
@@ -100,7 +101,7 @@ if (UR_ENABLE_TRACING)
     XPTI_ENABLE_INSTRUMENTATION
     XPTI_STATIC_LIBRARY
     )
-  target_include_directories(${TARGET_NAME} PUBLIC
+  target_include_directories(${TARGET_NAME} PRIVATE
```
igchor, line 104: Could you move those unrelated changes to a separate commit?

test/adapters/level_zero/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ if(UR_BUILD_ADAPTER_L0)
         # Make L0 use CallMap from a seprate shared lib so that we can access the map
         # from the tests. This only seems to work on linux
         add_library(zeCallMap SHARED zeCallMap.cpp)
+        install_ur_library(zeCallMap)
```
igchor, line 48: why are we installing this?

source/adapters/cuda/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ add_ur_adapter(${TARGET_NAME}
     ${CMAKE_CURRENT_SOURCE_DIR}/../../ur/ur.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/../../ur/ur.hpp
 )
+install_ur_library(${TARGET_NAME})
```
igchor, line 48: Why did you add this to CUDA and HIP? This wasn't needed before, right?

scripts/benchmarks/benches/base.py
lslusarczyk: Applied, code made more robust wrt lib dir name.

source/adapters/cuda/CMakeLists.txt
lslusarczyk: These are related changes. Using compiled UR needs "install" working. And "install" fails unless this wrongly publicly included XPTI is fixed.

scripts/benchmarks/benches/base.py
```diff
@@ -16,12 +16,24 @@
 class Benchmark:
     def __init__(self, directory):
         self.directory = directory
-        self.adapter_path = os.path.join(options.ur_dir, 'build', 'lib', f"libur_adapter_{options.ur_adapter_name}.so")
+        for libs_dir_name in ['lib', 'lib64']:
```
pbalcer, line 19: I think the make install fix is fine, but, in this case, not doing the install makes this deterministic - the file will be located in the lib directory.
Why the change to do a `make install`?

pbalcer: can you make 'get lib path' into a generic function?

lslusarczyk: > I think the make install fix is fine, but, in this case, not doing the install makes this deterministic - the file will be located in the lib directory. Why the change to do a `make install`?

Not doing install will fail benchmark in an assert in the next line with message like _could not find adapter file /home/lslusarczyk/src2/ur_install/lib64/libur_adapter_level_zero.so (and in similar lib paths)_ which looks what I would like to se, that is sample full path which was searched and still being a short message

"make install" used as cleanest way of using binaries and includes of already built UR

> can you make 'get lib path' into a generic function?
yes, separate function added

source/adapters/cuda/CMakeLists.txt
lslusarczyk: No adapters were installed before. As I discussed on teams channel, they should be installed. 

Regardless of that reason, CUDA adapter is also need by my change because as I see the benchmark framework works (or at least pretends to work) on CUDA. See: https://github.com/oneapi-src/unified-runtime/blob/main/.github/workflows/benchmarks_compute.yml#L132

igchor: I see, that makes sense.

test/adapters/level_zero/CMakeLists.txt
igchor: We could think about some other alternative to creating zeCallMap library. Perhpas we could just define a public function in Level Zero adapter that would return the pointer to zeCallMap (only if UR_L0_CALL_COUNT_IN_TESTS  is defined)?

pbalcer: @igchor I agree with @lslusarczyk. This is a horrible hack that needs to removed. The L0 adapter library has a dependency on this test-only object (!!).

I really don't understand what we are trying to test. Piling more hacks on top of this doesn't seem like a good idea. We need to rethink this approach.

lslusarczyk: > The tests will compile fine but you will get failure on dlopen.

Indeed. I get error on dlopen.

> We need to rethink this approach.

It seems you've already designed a solution here: https://github.com/oneapi-src/unified-runtime/issues/1454

I think I should neither implement it as a part of this PR nor block this PR because we lack of clean solution. I'm reverting to my original install solution that installs zeCallMap library if level_zero_adapter was compiled with UR_L0_CALL_COUNT_IN_TESTS define. It is needed to be installed because otherwise installed libur_adapter_level_zero.so can not be loaded. I'll add a note to revert this when https://github.com/oneapi-src/unified-runtime/issues/1454 is done.

Good point is that hack is now more visible, "stings the eyes" more, so we will sooner fix it, I think.

pbalcer (commented):
overall lgtm, but it might be easier to merge if we were to split this into two PRs: one that fixes `make install` (and maybe adds a test?) and the second one that does the benchmarks change.

lslusarczyk:
I've compared level_zero_v2 benchmarks on main and on my PR.
- using main: https://github.com/oneapi-src/unified-runtime/actions/runs/11159687520/job/31018579072
- using this PR:  https://github.com/oneapi-src/unified-runtime/actions/runs/11157704223/job/31012487313
What passes and what fails inside looks similar.
Since now we don't recompile UR the benchmarking step is much faster, from _51min 14 sec_ to **49min 0 sec** !!! :1st_place_medal: :)

Results are also comparable. E.g. "api_overhead_benchmark_sycl/ur SubmitKernel out of order/in order.
- using main: 20.153 μs, 24.01 μs, 14.613 μs, 11.627 μs
- using this PR: 20.564 μs, 21.222 μs, 14.275 μs, 11.822  μs

So it seems I've not broken anything I think.

lslusarczyk:
> overall lgtm, but it might be easier to merge if we were to split this into two PRs: one that fixes `make install` (and maybe adds a test?) and the second one that does the benchmarks change.

separate PR: https://github.com/oneapi-src/unified-runtime/pull/2169
Please check if my way of testing _cmake install_ looks sufficient to you there.

pbalcer:
the cmake change got merged, please rebase.

lslusarczyk:
> the cmake change got merged, please rebase.

already rebased

Issue #1197, opened 2023-12-15
Improve compiler interface testing


aarongreig:
Our testing of the compilation related `Program*` APIs is fairly basic at the moment, there's a lot we could do with it to improve coverage and rigour. To that end we should track any ideas that come up for this kind of testing here.

This was inspired by #877 which fixes a bug in the ProgramLink tests by making them worse: it turns out we can't link together two binaries that come out of our device binary generation system, so pending a potentially quite tricky rework of that we've had to settle for just linking one program. Therefore:

- [ ] Implement a ProgramLink test which links together multiple programs

After a quick look at the ProgramBuild and ProgramCompile tests it also occurred to me that we should probably
- [ ] Verify program state in ProgramBuild and ProgramCompile tests with appropriate ProgramGetInfo and ProgramGetBuildInfo queries

When urProgramCreateWithBinary is modified by #2147 to allow multiple binaries and multiple devices we should test that properly:
- [ ] Write a multi-device ProgramCreateWithBinary test

PR #1195, opened 2023-12-15
Merge changes from multi device compile extension into core spec.

aarongreig:
We originally changed this interface a bit from PI: currently these entry points take a context which represents your list of devices. This turned out to cause issues for the level zero adapter so the extension was created to effectively revert back to the PI/CL style interface. This PR reverts the core interface back to PI style, everything takes an explicit device list instead of a context.

LLVM testing https://github.com/intel/llvm/pull/12536

PR #2147, opened 2024-09-28
Support UR program creation from multiple device binaries

againull:
To support multi-device AOT scenario in SYCL we need an ability to create UR program from multiple device binaries.

Changes in this PR:
* Modify the core function `urProgramCreateWithBinary` to support program creation from multiple device binaries.
* Add methods to `ur_program` to get/set per-device data like L0 module handle, build log etc. Otherwise any change in the structure of the class requires multiple changes in all UR functions which work with `ur_program`, in addition to this it allows to handle interop case (which implementation is an exception right now) in a single place. Also make State and some other info per-device because it is possible that UR program is associated with multiple devices and `urProgramBuildExp` is called multiple times for subsets, and we have to know the state per-device.

Corresponding intel/llvm PR with E2E tests: https://github.com/intel/llvm/pull/15546
Also outlined minimal changes just to uplift UR tag: https://github.com/intel/llvm/pull/15710


aarongreig:
There's an ongoing effort to remove this extension by making it the default behaviour for the affected entry points https://github.com/oneapi-src/unified-runtime/pull/1195 as it was a workaround for an issue discovered late in the last release cycle. Since it seems like there's a compelling use case I think this should be a change to `urProgramCreateWithBinary`, maybe with a device info flag for adapters report whether they support creating binaries for more than one device at a time

againull:
> There's an ongoing effort to remove this extension by making it the default behaviour for the affected entry points #1195 as it was a workaround for an issue discovered late in the last release cycle. Since it seems like there's a compelling use case I think this should be a change to `urProgramCreateWithBinary`, maybe with a device info flag for adapters report whether they support creating binaries for more than one device at a time

Thank you, I've changed the PR to change the core  `urProgramCreateWithBinary` instead.

source/adapters/cuda/program.cpp
```diff
@@ -483,12 +483,18 @@ UR_APIEXPORT ur_result_t UR_APICALL urProgramGetNativeHandle(
...
+    ur_context_handle_t hContext, uint32_t numDevices,
+    ur_device_handle_t *phDevices, size_t *pLengths, const uint8_t **ppBinaries,
+    const ur_program_properties_t *pProperties,
     ur_program_handle_t *phProgram) {
+  if (numDevices > 1)
+    return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
 
-  UR_CHECK_ERROR(
-      createProgram(hContext, hDevice, size, pBinary, pProperties, phProgram));
+  if (numDevices == 0)
```
aarongreig, line 493: we should add this to the yml so it gets generated and used by all adapters, I think it should be an `INVALID_SIZE` return:
```
diff --git a/scripts/core/program.yml b/scripts/core/program.yml
index 3fe2632c..2a0bc14a 100644
--- a/scripts/core/program.yml
+++ b/scripts/core/program.yml
@@ -158,6 +158,7 @@ returns:
         - "`NULL != pProperties && pProperties->count > 0 && NULL == pProperties->pMetadatas`"
         - "`NULL != pProperties && NULL != pProperties->pMetadatas && pProperties->count == 0`"
+        - "`numDevices == 0`"
     - $X_RESULT_ERROR_INVALID_NATIVE_BINARY:
         - "If any binary in `ppBinaries` isn't a valid binary for the corresponding device in `phDevices.`"
```

againull: Thanks, fixed.

kbenzie:
@againull looks like there are conformance tests which still need updated. Some conformance tests are only enabled when the `UR_DPCXX` CMake variable points at `clang++` from an intel/llvm build as they depend on generating program binaries.

againull:
> @againull looks like there are conformance tests which still need updated. Some conformance tests are only enabled when the `UR_DPCXX` CMake variable points at `clang++` from an intel/llvm build as they depend on generating program binaries.

Oh, I see, thanks, I will check and fix that.

againull:
> @againull looks like there are conformance tests which still need updated. Some conformance tests are only enabled when the `UR_DPCXX` CMake variable points at `clang++` from an intel/llvm build as they depend on generating program binaries.

Fixed remaining failures.

againull:
Sorry, I was formatting using [.clang-format in unified-runtime](https://github.com/oneapi-src/unified-runtime/blob/main/.clang-format) and latest clang-format, it turns out cppformat target needs to be used together with clang-format-15 to get the correct formatting, fixed that. 

Other than that only e2e tasks are failing because at least these changes  https://github.com/intel/llvm/pull/15637 are required in intel/llvm to use new function signature.

againull:
Can I ask UR team to take a look at this PR please. Most of the changes are related to the L0 adapter, so it would be nice if @pbalcer or @nrspruit  could provide a review.

aarongreig (approved):
Spec, loader, CTS and CL adapter changes LGTM. It would be nice to have a test that creates a program with multiple devices and binaries but that's going to require a bit of CTS work to accommodate so I've added it to #1197 for future work.

nrspruit:
So far, everything looks good to me, however, @againull can your create a devops branch to kick off testing with multi devices in intel/llvm? Since this is specifically for multi device, then running on PVC for the e2e tests would be ideal. By default I don't think the pre-commit runs on PVC or DG2 or anything other than GEN12 for the most part.

againull:
> So far, everything looks good to me, however, @againull can you create a devops branch to kick off testing with multi devices in intel/llvm? Since this is specifically for multi device, then running on PVC for the e2e tests would be ideal. By default I don't think the pre-commit runs on PVC or DG2 or anything other than GEN12 for the most part.

Ok, sure, I will invoke such testing.

pbalcer:
> By default I don't think the pre-commit runs on PVC or DG2 or anything other than GEN12 for the most part.

just fyi - we have a multi-device PVC system in UR CI. You just need to add tests, which this doesn't.

konradkusiak97 (approved):
cuda/hip ok.

againull:
> > By default I don't think the pre-commit runs on PVC or DG2 or anything other than GEN12 for the most part.
> 
> just fyi - we have a multi-device PVC system in UR CI. You just need to add tests, which this doesn't.

@nrspruit  @pbalcer  I have enabled existing "program" conformance test, as well as added new multi-device to test `urProgramCreateWithBinary`.

againull:
@oneapi-src/unified-runtime-level-zero-write @oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-command-buffer-write  Could you please take a look.

nrspruit (approved):
LGTM, thank you for adding the UR tests!

EwanC:
> @oneapi-src/unified-runtime-level-zero-write @oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-command-buffer-write Could you please take a look.

I can't see any files owned by @oneapi-src/unified-runtime-command-buffer-write in this PR diff, I assume the change has been refactored in some way to avoid touching these files since this reviewer group was requested.

PietroGhg (approved):
Native CPU lgtm, thank you

againull:
@oneapi-src/unified-runtime-maintain Could you please merge this PR.

kbenzie:
@oneapi-src/unified-runtime-command-buffer-write please review

PR #2154, opened 2024-10-01
[CUDA][HIP] Fix exceptions throwing from adapter

npmiller:
`UR_CHECK_ERROR` will throw an exception if an error is found, so it should only be used within a try/catch block, otherwise the exception may leak out of the adapter.

PR #2155, opened 2024-10-01
[CUDA][HIP] Fix exceptions throwing from adapter

npmiller:
`UR_CHECK_ERROR` will throw an exception if an error is found, so it should only be used within a try/catch block, otherwise the exception may leak out of the adapter.

Cherry-pick of: https://github.com/oneapi-src/unified-runtime/pull/2154


kbenzie (approved):

kbenzie:
I've cherry-picked this commit onto [v0.10.9rc](https://github.com/oneapi-src/unified-runtime/pull/2157), closing.

PR #2158, opened 2024-10-01
Split source checks into a seperate workflow

RossBrunton:
The "Build - Ubuntu" job has a stage that checks the source generated
from the spec matches the version stored in the repo. This is the
part of the job that takes the longest, and is ran multiple times.

Since it doesn't require a build, it has been brought out into its own
workflow which is only ran once. This should hopefully speed up builds.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -49,7 +49,7 @@ jobs:
     - name: Install apt packages
       run: |
         sudo apt-get update
-        sudo apt-get install -y doxygen ${{matrix.compiler.c}}
+        sudo apt-get install -y ${{matrix.compiler.c}}
 
     - name: Install libhwloc
       run: .github/scripts/install_hwloc.sh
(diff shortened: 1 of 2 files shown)


pbalcer:
Good idea. If possible, we should make this build passing a prerequisite for all the other CI jobs.

RossBrunton:
@pbalcer Ordinarily I'd agree, but the source check takes over five minutes, while the jobs it would be blocking can sometimes finish in under that time. So you'd be adding 5 minutes of latency to reduce congestion in the rare case where the generation doesn't match.

I can look into making it so that on failure it will cancel the entire run though. Or maybe split out clang-format/license checks into their own job.

pbalcer:
Makes sense. I suggest we cancel the other jobs if that's easy to do. If not, let's just go with your change as-is.

RossBrunton:
I had a quick look at cancelling the job, and it seems doable using the `gh` command line. But that requires fiddling around with permissions which I don't think is worth it.

kbenzie (approved):

PR #2163, opened 2024-10-02
Split some tests into their own workflows

RossBrunton:
This makes the list of jobs more organised.


kbenzie:
The e2e jobs have a dependency of the level zero and opencl jobs which have been moved out of the cmake.yml workflow.

RossBrunton:
Probably not worth doing.

PR #2164, opened 2024-10-02
Benchmark updates for faster run and more reliable results

mateuszpn:


scripts/benchmarks/benches/velocity.py
```diff
@@ -29,7 +28,11 @@ def download_deps(self):
 
     def setup(self):
         self.download_deps()
-
+        self.benchmark_bin = os.path.join(self.directory, self.bench_name, self.bin_name)
+        
+        if options.no_git:
```
pbalcer, line 33: I'm not the biggest fan of the way this option is implemented. Would it be possible to just do this inside of `git_clone` function?
That way, it will simply skip the fetch, and the cmake/make should not need to do anything.

mateuszpn: The option prevents from any directory removal, build etc. Can be removed in a version we consider stable, as cluster node is not destination environment for the script

scripts/benchmarks/benches/velocity.py
```diff
@@ -9,12 +9,11 @@
 from utils.utils import run, create_build_path
 from .options import options
 import os
-import re
 
 class VelocityBench:
     def __init__(self, directory):
         self.directory = directory
-        self.repo_path = git_clone(self.directory, "velocity-bench-repo", "https://github.com/oneapi-src/Velocity-Bench", "34ee4ebe18d91dfdd38b7d798fd986b41874fcbc")
+        self.repo_path = git_clone(self.directory, "velocity-bench-repo", "https://github.com/mateuszpn/Velocity-Bench", "66e8188")
```
pbalcer, line 16: your PR was merged:
https://github.com/oneapi-src/Velocity-Bench/commit/b22215c16f789100449c34bf4eaa3fb178983d69

Can you update?

Otherwise lgtm.

mateuszpn: Done

pbalcer: no? I still see your fork being used.

pbalcer: ok, I see the new push.

PR #2166, opened 2024-10-02
[L0 v2] make device allocation resident and support multi-device buffers

igchor:


test/adapters/level_zero/v2/memory_residency.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <map>
+#include <string>
+
+using urMemoryResidencyTest = uur::urMultiDeviceContextTestTemplate<1>;
+
+// Is this a Data Center GPU Max series (aka PVC)?
+// TODO: change to use
+// https://spec.oneapi.io/level-zero/latest/core/api.html#ze-device-ip-version-ext-t
+// when that is stable.
+bool isPVC(ur_device_handle_t hDevice) {
```
pbalcer, line 19: moves this to some commons.

source/adapters/level_zero/v2/memory.cpp
```diff
@@ -51,55 +51,81 @@ void *ur_host_mem_handle_t::getPtr(ur_device_handle_t hDevice) {
   return ptr;
 }
 
+ur_result_t ur_device_mem_handle_t::migrateBufferTo(ur_device_handle_t hDevice,
```
pbalcer, line 54: I don't think we should be doing the same thing as the current adapter. The memory copy should be added to the command list as a dependency for the related enqueue operation.

This whole synchronous migration when setting the parameter seems like a temporary hack that became permanent.

pbalcer (commented):
this stack of commits has become unreviewable.

source/adapters/level_zero/v2/memory.cpp
igchor: It might be the case that just leaving the memory on the device, without migrating between devices is a better option - we should run some benchmarks first. If we decide that we want to migrate the memory, then yes, we should make this asynchronous (and also modify getPtr() function which is used by enqueueKernel, etc.). For now, this function in only used for initialization for copying the data from host to the device - perhaps I should rename the function to avoid confusion?

pbalcer: Hm, ok, I see. No, the name is fine, I was just fixated on that scenario.

I think the best approach here would be to create the memory handle in a 'pending' state, to be populated with data asynchronously on first use. My only concern would be concurrency (i.e., what happens when two threads operate on such 'pending' operation at the same time. we'd have to come up with some way to synchronize...).
For now leave it as-is, maybe create a task so that we don't forget.

igchor: For initialization, we need to do it synchronously I think because the host pointer might not be valid after urMemBufferCreate. We could use some buffers in DRAM (as I did initially) but again, we would probably need to benchmark and see if it makes sense.

For any migration that happens after that, if we just pass all the required dependencies to enqueueMemcpy and either return a signal event or (for in-inorder) just execute that enqueueMemcpy on the commandList that's used for the next operation (kernelLaunch, etc.) we don't need any extra synchronization I think.

igchor: In general, I guess we could make the logic to decide whether to migrate or not a bit more intelligent that just yes or no 0 - perhaps we could keep some statistics to see how often the memory is accessed from which device and then migrate  (or not) based on that.

pbalcer: right, makes sense. let's just do the simple thing and do the copy synchronously. trying to do a copy in dram could make sense for small buffers, but will break down for large things.

The device to device migration path I think may be a bit trickier, because it may involve multiple command lists. We need to start with some tests I think.

igchor: Yes, we definitely need more tests - I will create a task for that.

test/adapters/level_zero/v2/memory_residency.cpp
igchor: Done.

igchor:
> this stack of commits has become unreviewable.

I'll rebase once #2016 is merged

igchor:
> > this stack of commits has become unreviewable.
> 
> I'll rebase once #2016 is merged

Done.

source/adapters/level_zero/v2/memory.cpp
```diff
@@ -51,55 +51,80 @@ void *ur_host_mem_handle_t::getPtr(ur_device_handle_t hDevice) {
...
-  // (depending on an env var setting).
-  //
-  // TODO: port this behavior or figure out if it makes sense to keep the memory
-  // in a host buffer (e.g. for smaller sizes).
+      deviceAllocations(hContext->getPlatform()->getNumDevices()),
+      activeAllocationDevice(nullptr) {
   if (hostPtr) {
-    buffer.assign(reinterpret_cast<char *>(hostPtr),
-                  reinterpret_cast<char *>(hostPtr) + size);
+    auto initialDevice = hContext->getDevices()[0];
```
pbalcer, line 87: doing the allocation to first device isn't ideal... Are we sure this hostPtr must not be valid for the lifetime of the buffer?
Anyway, let's keep it this way, but I think we might want to create a todo/issue to revisit this decision at some point.

igchor: Yes, I'm tracking this a part of one of the issues. For now, this behavior (allocating on the first device) is the same as in legacy behavior (without the migration).

source/common/umf_helpers.hpp
```diff
@@ -26,6 +26,8 @@
 #include <tuple>
 #include <utility>
 
+ur_result_t ze2urResult(uint32_t ZeResult);
```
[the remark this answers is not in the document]
igchor: Not needed, removed.

test/adapters/level_zero/v2/memory_residency.cpp
```diff
@@ -0,0 +1,64 @@
...
+        auto status = urUSMDeviceAlloc(
+            context, uur::DevicesEnvironment::instance->devices[0], nullptr,
+            nullptr, allocSize, &ptr);
+
+        ASSERT_SUCCESS(
+            urDeviceGetInfo(uur::DevicesEnvironment::instance->devices[0],
+                            UR_DEVICE_INFO_GLOBAL_MEM_FREE, sizeof(size_t),
+                            &globalMemFree, nullptr));
+
+        if (status == UR_RESULT_ERROR_OUT_OF_DEVICE_MEMORY) {
```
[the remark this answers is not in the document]
igchor: Done.

source/adapters/level_zero/v2/usm.cpp
```diff
@@ -66,6 +66,16 @@ descToDisjoinPoolMemType(const usm::pool_descriptor &desc) {
   }
 }
 
+static std::map<std::string, std::function<ur_result_t(int32_t)>> &
```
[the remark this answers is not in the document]
igchor: Okay, I defined native error handler as a regular function that needs to be implemented by each adapter. Now, the implementation is nop for each one expect for L0.

pbalcer (approved):

PR #2168, opened 2024-10-03
Improved adapter selection in CTS testing

RossBrunton:
The CTS tests now have an additional --backend option, and the --platform
option allows specifying platforms per-device. In addition, new targets
have been added that test the "loader"; that is, where the adapter isn't
forced.


RossBrunton:
Currently this doesn't do the loader testing in CI, I'm going to do that as a separate change.

test/conformance/adapter/adapter_adapter_native_cpu_loader.match
```diff
@@ -0,0 +1 @@
+adapter_adapter_native_cpu.match
```
aarongreig, line 1: are these somehow referencing the existing match file? did we already have something in the scripts that let us do that?

RossBrunton: This is a symlink. Github doesn't display them very well except in the sidebar.

test/conformance/source/environment.cpp
```diff
@@ -62,6 +97,10 @@ std::ostream &operator<<(std::ostream &out,
 uur::PlatformEnvironment::PlatformEnvironment(int argc, char **argv)
```
martygrant, line 97: This is a pretty meaty function doing a lot, maybe it could be split up to simplify?

test/conformance/CMakeLists.txt
```diff
@@ -66,29 +68,50 @@ function(add_conformance_test name)
...
+        endif()
+        if(UR_BUILD_ADAPTER_L0_V2)
+            add_test_adapter(${name} adapter_level_zero_v2_loader LEVEL_ZERO OFF)
+        endif()
+        if(UR_BUILD_ADAPTER_OPENCL OR UR_BUILD_ADAPTER_ALL)
+            add_test_adapter(${name} adapter_opencl_loader OPENCL OFF)
+        endif()
+        if(UR_BUILD_ADAPTER_NATIVE_CPU OR UR_BUILD_ADAPTER_ALL)
+            add_test_adapter(${name} adapter_native_cpu_loader NATIVE_CPU OFF)
+        endif()
```
martygrant, line 107: It might be good for this stuff to be very explicit but I wonder if there's value in refactoring this into a function (if possible, I'm not super experienced with CMake) as there's a lot of repetition?

.github/workflows/build-hw-reusable.yml
```diff
@@ -82,6 +82,7 @@ jobs:
         -DUR_ENABLE_TRACING=ON
         -DUR_DEVELOPER_MODE=ON
         -DUR_BUILD_TESTS=ON
+        -DUR_CONFORMANCE_TEST_LOADER=OFF
```
martygrant, line 85: `UR_CONFORMANCE_TEST_LOADER` should be added to the list of flags in `README.md`

test/conformance/source/environment.cpp
```diff
@@ -196,6 +287,26 @@ void uur::PlatformEnvironment::TearDown() {
 PlatformEnvironment::PlatformOptions
 PlatformEnvironment::parsePlatformOptions(int argc, char **argv) {
```
martygrant, line 288: This function isn't as big but has some dense sections that could maybe be split up too.

test/conformance/source/environment.cpp
```diff
@@ -205,21 +316,35 @@ PlatformEnvironment::parsePlatformOptions(int argc, char **argv) {
...
+    /* If a platform was not provided using the --platform/--backend command line options,
      * check if environment variable is set to use as a fallback. */
     if (options.platform_name.empty()) {
         auto env_platform = ur_getenv("UR_CTS_ADAPTER_PLATFORM");
         if (env_platform.has_value()) {
             options.platform_name = env_platform.value();
         }
     }
+    if (!options.platform_backend) {
+        auto env_backend = ur_getenv("UR_CTS_BACKEND");
```
martygrant, line 341: Is `UR_CTS_BACKEND` new? I don't see it listed in the docs, should be added to the list here https://oneapi-src.github.io/unified-runtime/core/INTRO.html#environment-variables

test/conformance/adapter/adapter_adapter_native_cpu_loader.match
aarongreig: aaah I see, I wonder if we could just stop doing the `EXTRA_NAME` thing here https://github.com/oneapi-src/unified-runtime/pull/2168/files#diff-1595fcdcdd0e6bf22312a2db5cbf98fb89265acbbc4bfa9887845dbe193a291fR22 I can't see any cts match files actually using it to have multiple different match files for a given adapter/test combo, and honestly I wouldn't want us to start doing that - the whole thing is fiddly enough as is

RossBrunton: EXTRA_NAME seems to be unused and so could probably be removed. But I think that should be a separate change.

aarongreig: isn't EXTRA_NAME the thing that's necessitating all these symlinks? what I'm getting at is it would be cleaner if we could avoid the clutter by changing how the targets are set up

test/conformance/source/environment.cpp
RossBrunton: UR_CTS_ADAPTER_PLATFORM is not on that list either, so I assume that environment variables that are only relevant to the CTS (and not broadly to end users) shouldn't be on that list.

test/conformance/adapter/adapter_adapter_native_cpu_loader.match
RossBrunton: EXTRA_NAME is unused, the symlinks are created because of the first parameter ("name"). I'm intentionally giving them a different match file because its possible that the loader could introduce (or fix) specific tests.

... Although, now I think about it, `{{opt}}` might be a better solution to that.

aarongreig: ah right, yeah I'd prefer leaving that for `{{opt}}` if it turns out to be necessary

test/conformance/source/environment.cpp
RossBrunton: I'm not sure. This lambda is only used in this one function, so IMO it doesn't really make sense to split it out into a free function.

martygrant: No problem, I just like to split functions up for readability once they get over a certain length

test/conformance/source/environment.cpp
martygrant: Might just have been missed? I'd like to think anything that's configurable should be documented and not just left in the code as it's easily forgotten

RossBrunton: Done, wrote a paragraph explaining it (hopefully with no typos).

test/conformance/source/environment.cpp
```diff
@@ -42,15 +67,6 @@ std::ostream &operator<<(std::ostream &out,
     return out;
 }
 
-std::ostream &operator<<(std::ostream &out, const ur_device_handle_t &device) {
```
RossBrunton, line 45: This has been removed since we are importing a few more headers which define this elsewhere.

scripts/core/INTRO.rst
```diff
@@ -404,6 +404,25 @@ Specific environment variables can be set to control the behavior of unified run
...
+.. envvar:: UR_CTS_ADAPTER_PLATFORM
+
+    A specifier list in the form of `[(backend):](platform name)[;[(backend)]:(platform name)]...`. For example:
+    `Fictional Corp(R);LEVEL_ZERO:Another Fictional Corp(R)`. If a backend specific specifier is present in the list and
+    the test is running for that backend, the device with the given name is chosen. Otherwise, it must match the name of
+    the specifier from the list with no backend.
+
+.. envvar:: UR_CTS_BACKEND
+
+    A backend to force the test to use. For example, `OPENCL`, `LEVEL_ZERO`, `HIP` and so on.
```
kbenzie, line 424: I feel like these names should be lower case.

scripts/core/INTRO.rst
```diff
@@ -404,6 +404,25 @@ Specific environment variables can be set to control the behavior of unified run
...
+
+The following environment variables are used by the CTS runner and can be used to specify the platform that the test
+framework should run on. This can be used during development and testing to run CTS tests in case multiple platforms
+are available. If both filters are specified, then they both must match a platform for it to be selected. If there are
+no valid platforms, then the tests will fail. Command line arguments take priority over these variables.
+
+.. envvar:: UR_CTS_ADAPTER_PLATFORM
+
+    A specifier list in the form of `[(backend):](platform name)[;[(backend)]:(platform name)]...`. For example:
+    `Fictional Corp(R);LEVEL_ZERO:Another Fictional Corp(R)`. If a backend specific specifier is present in the list and
```
kbenzie, line 418: This seems like a confusing schema to me. How come the backend is optional?

kbenzie (commented):
There's a conflict in the cmake.yml workflow that will need resolved.

scripts/core/INTRO.rst
RossBrunton: To preserve the original behaviour and provide a convenience for users who only want to run the tests on one device. I've expanded the documentation to hopefully explain things better.

scripts/core/INTRO.rst
RossBrunton: They're actually case insensitive. I've updated the docs to call that out and made the values here lower case.

scripts/core/INTRO.rst
```diff
@@ -404,6 +404,31 @@ Specific environment variables can be set to control the behavior of unified run
...
+no valid platforms, then the tests will fail. Command line arguments take priority over these variables.
+
+.. envvar:: UR_CTS_ADAPTER_PLATFORM
+
+    A specifier list in the form of `[(backend):](platform name)[;[(backend)]:(platform name)]...`. If a backend
+    specific specifier is present in the list and the test is running for that backend, the device with the given name
+    is chosen. Otherwise, it must match the name of the specifier from the list with no backend.
+
+    For example, if the test device has multiple platforms and you want to run tests on the "ABC Corp" backend when
+    testing OpenCL and "XYZ Org" when testing level zero, you'd use `OPENCL:ABC Corp;LEVEL_ZERO:XYZ Org`. This form is
```
kbenzie, line 422: Are the backend and platform matches case insensitive as well?

RossBrunton: Backends are not, but the platform must match exactly. Updated the docs.

kbenzie: I'd like to avoid ALL_CAPS in the backends please, or at least be consistent with `UR_CTS_BACKEND` and make it case insensitive.

RossBrunton: ... I completely misread that. Backends are indeed case insensitive.

kbenzie (approved):

PR #2169, opened 2024-10-03
Install adapters by cmake

lslusarczyk:
Add installing adapters. It is needed by https://github.com/oneapi-src/unified-runtime/pull/2146 and was already reviewed there and was just extracted to this separate PR (with some kind of test added).

### DIFF
--- a/.github/workflows/build-hw-reusable.yml
+++ b/.github/workflows/build-hw-reusable.yml
@@ -87,12 +87,17 @@ jobs:
         -DUR_STATIC_ADAPTER_${{matrix.adapter.name}}=${{matrix.adapter.static_adapter}}
         -DUR_DPCXX=${{github.workspace}}/dpcpp_compiler/bin/clang++
         -DUR_SYCL_LIBRARY_DIR=${{github.workspace}}/dpcpp_compiler/lib
+        -DCMAKE_INSTALL_PREFIX=${{github.workspace}}/install
         ${{ matrix.adapter.name == 'HIP' && '-DUR_CONFORMANCE_AMD_ARCH=gfx1030' || '' }}
         ${{ matrix.adapter.name == 'HIP' && '-DUR_HIP_PLATFORM=AMD' || '' }}
 
     - name: Build
       # This is so that device binaries can find the sycl runtime library
       run: cmake --build ${{github.workspace}}/build -j $(nproc)
+
+    - name: Install
+      # This is to check that install command does not fail
+      run: cmake --install ${{github.workspace}}/build
 
     - name: Test adapter specific
       working-directory: ${{github.workspace}}/build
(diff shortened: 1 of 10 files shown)


kbenzie:
With some additional changes in intel/llvm this could also resolve https://github.com/intel/llvm/issues/14924

.github/workflows/build-hw-reusable.yml
```diff
@@ -87,13 +87,18 @@ jobs:
...
         -DUR_SYCL_LIBRARY_DIR=${{github.workspace}}/dpcpp_compiler/lib
+        -DCMAKE_INSTALL_PREFIX=${{github.workspace}}/install
         ${{ matrix.adapter.name == 'HIP' && '-DUR_CONFORMANCE_AMD_ARCH=gfx1030' || '' }}
         ${{ matrix.adapter.name == 'HIP' && '-DUR_HIP_PLATFORM=AMD' || '' }}
 
     - name: Build
       # This is so that device binaries can find the sycl runtime library
       run: cmake --build ${{github.workspace}}/build -j $(nproc)
 
+    - name: Install
```
lukaszstolarczuk, line 98: looked into logs and it seems `gmock` and `gtest` is installed as well... I think it should rather not - if this is an easy fix you could add it here, if not probably just file an issue for that (the issue was here before your changes, of course)

lukaszstolarczuk (commented):
👍

.github/workflows/build-hw-reusable.yml
lslusarczyk: Good point. Fixed.

.github/workflows/build-hw-reusable.yml
```diff
@@ -87,13 +87,18 @@ jobs:
...
+      # This is to check that install command does not fail
+      run: cmake --install ${{github.workspace}}/build
+
```
pbalcer, line 101: Ideally we'd check output against a list of expected files.

lslusarczyk: I object. Too much logic to put in the test. Not worth doing it. Libs can be static or dynamic. Adapters may be built or not depending on cmake params. There are 100+ files installed, so check will have to selective and thus not complete.

If any client will depend on make install we can reconsider. Now the only client is benchmark framework.

pbalcer: ok, makes sense.

pbalcer:
If the changes are ready, please undraft them.

pbalcer (approved):

steffenlarsen (approved):
Looks reasonable! 🚀

PietroGhg (approved):
Native CPU lgtm, thank you

kbenzie (approved):

PR #2170, opened 2024-10-03
[CTS] extend tests for urEnqueueEventsWait

igchor:


test/conformance/enqueue/urEnqueueEventsWaitMultiDevice.cpp
```diff
@@ -0,0 +1,209 @@
...
+struct urEnqueueEventsWaitMultiDeviceTest : uur::urMultiQueueMultiDeviceTest {
+    void SetUp() override { SetUp(2); /* we need at least 2 devices */ }
+
+    void SetUp(size_t numDuplicateDevices) {
+        UUR_RETURN_ON_FATAL_FAILURE(uur::urMultiQueueMultiDeviceTest::SetUp(
+            uur::KernelsEnvironment::instance->devices, numDuplicateDevices));
+
+        ptrs.resize(devices.size());
+        for (size_t i = 0; i < devices.size(); i++) {
+            EXPECT_SUCCESS(urUSMSharedAlloc(context, devices[i], nullptr,
```
aarongreig, line 22: before we get to this point we should skip if shared USM isn't supported, like this: https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/usm/urUSMSharedAlloc.cpp#L22

igchor: Right, done.

pbalcer (approved):

PR #2174, opened 2024-10-04
[NFC][CUDA] Include STL headers in program.cpp

ProGTX:
This patch fixes the following build error on Windows:
```
source\adapters\cuda\program.cpp(73): error C2079: 'WorkGroupElements' uses undefined class 'std::array<uint32_t,3>'
```
Technically only `<array>` was sufficient for resolving it, but it's better to include everything that the file relies on.

### DIFF
--- a/source/adapters/cuda/program.cpp
+++ b/source/adapters/cuda/program.cpp
@@ -11,6 +11,17 @@
 #include "program.hpp"
 #include "ur_util.hpp"
 
+#include <algorithm>
+#include <array>
+#include <cstdint>
+#include <cstring>
+#include <memory>
+#include <new>
+#include <string>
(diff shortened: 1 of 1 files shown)


MartinWehking:
Could you please clarify which build error this fixes?

MartinWehking:
Does this also affect the HIP adapter?

kbenzie:
IS HIP adapter supported on Windows? I thought it wasn't.

MartinWehking:
> Is HIP adapter supported on Windows? I thought it wasn't.

Ah yes, that's true.

ProGTX:
I've updated the description with the error.

MartinWehking (approved):
LGTM

PR #2181, opened 2024-10-08
[CTS] Fix a platform selection bug for cts

omarahmed1111:
This should fix a bug where if we filtered with a platform and this platform is not available in the first adapter, it will fail with not finding the platform needed without searching other adapters first for this platform.


RossBrunton:
This is fixed as part of https://github.com/oneapi-src/unified-runtime/pull/2168

omarahmed1111:
> This is fixed as part of #2168

Ah okay, I didn't notice it. Thanks for mentioning that! Will close this PR then.

PR #2184, opened 2024-10-09
Remove old CMake settings - external adapter source dirs

martygrant:
I spotted these CMake settings which I believe were used when adapter sources lived in intel/llvm which is no longer the case so these should be able to be removed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,16 +68,6 @@ set(UR_CONFORMANCE_TARGET_TRIPLES "" CACHE STRING
     "List of sycl targets to build CTS device binaries for")
 set(UR_CONFORMANCE_AMD_ARCH "" CACHE STRING "AMD device target ID to build CTS binaries for")
 option(UR_CONFORMANCE_ENABLE_MATCH_FILES "Enable CTS match files" ON)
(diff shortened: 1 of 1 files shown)


pbalcer (approved):

RossBrunton:
Are these not used by `add_ur_adapter_subdirectory`?

martygrant:
@RossBrunton oh yes I didn't spot the usage there. I do think the original usage was for when the adapters didn't live in the UR repo, so while the options could be used they don't really have a purpose anymore. `add_ur_adapter_subdirectory` could probably be removed too, what do you think?

RossBrunton:
@martygrant 

I can see a use case being for when you want to add some "secret sauce" to an adapter but don't want to go through the trouble of git merging it into UR? Not sure how common people would want to do that though.

kbenzie:
These are used in [intel/llvm](https://github.com/intel/llvm/blob/sycl/sycl/cmake/modules/FetchUnifiedRuntime.cmake#L114-L115) - the variable names are dynamically generated so don't show up for `grep` results but these should not be removed.

martygrant:
@kbenzie oh okay thanks for that! I'll close this off.

PR #2186, opened 2024-10-09
[UR] Don't allow the loader to catch exceptions

hdelan:
Call std::abort if an exception is thrown to the loader. Since UR is a C interface, no exceptions should be thrown between different UR layers.

### DIFF
--- a/scripts/templates/libapi.cpp.mako
+++ b/scripts/templates/libapi.cpp.mako
@@ -65,7 +65,7 @@ try {
 
     return ${th.make_pfn_name(n, tags, obj)}( ${", ".join(th.make_param_lines(n, tags, obj, format=["name"]))} );
 %endif
-} catch(...) { return exceptionToResult(std::current_exception()); }
+} catch(...) { std::abort(); }
 %if 'condition' in obj:
 #endif // ${th.subt(n, tags, obj['condition'])}
 %endif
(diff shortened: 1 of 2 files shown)


pbalcer (commented):
Layers (validation, tracing, sanitizer etc) are compiled into the loader. So while it's true that the DDI interface between the adapters and the loader is C, there's potentially other code that can throw where must catch exceptions before returning.

Is there anything the existing implementation breaks?

igchor:
> Layers (validation, tracing, sanitizer etc) are compiled into the loader. So while it's true that the DDI interface between the adapters and the loader is C, there's potentially other code that can throw where must catch exceptions before returning.
> 
> Is there anything the existing implementation breaks?

Also, I believe here is already some code (at least in L0 adapter) that relies on loader catching and translating exceptions to errors. 

urGet*ProcAddrTable are indeed defined as 'extern "C"' but the actual function pointers in the did table point to regular C++ functions which should work fine.

kbenzie:
> Is there anything the existing implementation breaks?

We've more been discussion how to prevent adapters from letting exceptions escape as this has caused issues with the CUDA/HIP adapters letting exceptions escape a C interface, this was one idea how to do that.

hdelan:
Instead of changing the default behaviour as this PR does, I am going to add a UR layer that ensures that no exceptions are being passed from the UR adapter across the lib boundary. This layer will only be enabled when running CTS tests, and can be disabled for level zero if necessary.

PR #2193, opened 2024-10-11
Clarify spec around isNativeHandleOwned.

aarongreig:


scripts/core/PROG.rst
```diff
@@ -322,3 +322,39 @@ through ${x}_usm_desc_t structure. Allocations that specify different pool handl
...
+   Not all backends have a 1:1 equivalent for every ${X} handle type, as such
+   any ``CreateWithNativeHandle`` or ``GetNativeHandle`` entry point *may* fail
+   with the error code ${X}_RESULT_ERROR_UNSUPPORTED_FEATURE for a given
+   adapter.
+
+Native Handle Ownership
+-----------------------
+
+By default a ${X} object constructed from a native handle doesn't own the
+native handle, it is guaranteed not to modify the native handle's reference
```
pbalcer, line 346: I don't think we can guarantee that. The adapter may modify the reference count of the provided handles, but it will always leave the refcount of the native handle with the same value as it was when on create.

jchlanda (approved):
cuda/hip 👍

pbalcer (approved):

scripts/core/PROG.rst
aarongreig: We're effectively saying that UR abdicates any responsibility for native resources passed in while `isNativeHandleOwned == false`. Since it's on the user to ensure native handles are alive for the entire duration of the associated UR handle's lifetime, why would an adapter need to increment the reference count? The CL adapter does increment the reference count but that's a bug imo, one which should get fixed by https://github.com/oneapi-src/unified-runtime/pull/1176

pbalcer: I was just thinking that when you wrap a native handle with a UR handle, and that handle is then used with some API functions that happens to need to increment the refcount on, let's say a kernel (and it will drop that refcount later as part of the program), I don't think we should be creating some conditions here like `if (!nativeHandle) refcount++;` that diverge the execution flow for the different types of adapters. It just creates potential areas of issues for not much gain.

Especially since it shouldn't matter much, since the adapter, in a correct program, will leave the handle's internal refcount exactly the same. It may just increment/decrement it while the handle is being used.

aarongreig: I see what you mean, I've made the wording less restrictive

aarongreig:
ping @oneapi-src/unified-runtime-opencl-write

aarongreig:
I've decided this isn't quite NFC enough to dodge a tag bump https://github.com/intel/llvm/pull/16730

PR #2194, opened 2024-10-11
[NFC] Whitespace fixes in yml files

RossBrunton:
Trim trailing whitespace in the yml files. This also affected some
source files because our generation system assumes a line ending with
a space means to insert a line break on the word before (which I assume
is a bug).


kbenzie (approved):

isaacault (approved):

RossBrunton:
Merging without associated LLVM commit as this should be an NFC.

PR #2199, opened 2024-10-14
Rewrite cts_exe.py to handle match logic

RossBrunton:
cts_exe.py has been rewritten to include the logic for handling .match files. Specifically, it will divide the test space into two halves; tests that are expected to pass, and those that are expected to fail.

For the tests expected to pass, it will run them all in the same gtest invocation with the assumption that it succeeds. For the tests expected to fail, they will each be ran with individual gtest invocations. This allows them to freely segfault or abort without hurting other tests.

In this commit, the match files are (mostly) unchanged, and the passing and failing tests should be the same. The match file is treated as a list of failing tests with a few tokens that are replaced:
* `{{NONDETERMINISTIC}}` ignored, required for compatibility with the match checker.
* `{{OPT}}` this test may or may not fail. It's still ran seperately, but doesn't report an error on failure.
* `{{.*}}` replaced with `*`; converts "match" wildcard matches to "gtest" test name matches.
* `#` and empty lines are ignored and treated as a comment.
* `{{Segmentation` for compatibility, this will cause a failure in the "excepted success" execution to not count as an error. This matches the behaviour of the prior match test logic.

Some .match files have been fixed and empty ones have been removed.


EwanC (approved):
command-buffer match file changes look good

RossBrunton:
Added specific logic to handle `ctest_parser.py`, manually checked and it now reports the same results before and after this change.

kbenzie (approved):

RossBrunton:
Doesn't affect intel/llvm, merging without associated commit.

PR #2208, opened 2024-10-16
update to L0 loader version v1.18.3 and don't scan _deps with Trivy or CodeQL

pbalcer:


nrspruit (approved):

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,2 @@
+paths-ignore:
+ - '**/_deps/**'
```
RossBrunton, line 2: This file needs a newline at the end.

nrspruit:
I was addressing issues in the l0 spec itself so I did not want to update UR until those issues were resolved. I have "just" fixed those issues so this is now ok to move to 1.18.3, that was why I had left the previous PR at 1.17.x until I had resolved those, but this is "ok" now.

PR #2209, opened 2024-10-16
[NFC] Ignore "loader" tests in ctest_parser.py

RossBrunton:


kbenzie (approved):

RossBrunton:
Since the loader is off by default now, will close this.

PR #2210, opened 2024-10-16
[NFC] Make loader tests opt-in rather than opt-out

RossBrunton:


aarongreig:
were we running these in CI and if so do we need to update that so they keep running?

RossBrunton:
@aarongreig CI explicitly sets this variable to "OFF", so this shouldn't be a change in behaviour. Long term I want this to be tested in CI, but it is not currently.

aarongreig (approved):

RossBrunton:
Merging without LLVM change: This should have no effect on intel/llvm (and if it does, this should revert behaviour back to an earlier version).

PR #2213, opened 2024-10-16
Don't pass --backend for mock tests

RossBrunton:
--backend doesn't make sense for the mock "adapter", since it doesn't
advertise a consistent valid backend.


pbalcer (approved):

lukaszstolarczuk (commented):
LGTM

RossBrunton:
So apparently the mock adapter isn't actually supposed to be used in the conformance testing at all. I'll just make it not do that.

PR #2216, opened 2024-10-17
[UR][layer] Add exception sanitizer layer

hdelan:
Add basic exception sanitizer layer that will call std::abort if an exception is thrown. This layer can be used to make sure an adapter isn't throwing, which is prohibited for a C interface. The layer can be enabled by building with the option `-DUR_ENABLE_EXCEPTION_SANITIZER=ON`


source/loader/ur_lib.hpp
```diff
@@ -94,6 +97,9 @@ class __urdlllocal context_t : public AtomicSingleton<context_t> {
 #endif
 #if UR_ENABLE_SANITIZER
             ur_sanitizer_layer::context_t::getNames(),
+#endif
```
pbalcer, line 100: you also need to add the layer info to `const std::vector<LayerData> layers` above.

pbalcer (approved):

pbalcer (commented):
I suspect this was never tested :D So might be good to enable this in CI for CTS.

source/loader/ur_lib.hpp
hdelan: Thanks @pbalcer have done so

hdelan:
> I suspect this was never tested :D So might be good to enable this in CI for CTS.

Thanks have done so. I'm not sure what the failures are related to.

source/adapters/cuda/command_buffer.cpp
```diff
@@ -1361,49 +1361,55 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferUpdateKernelLaunchExp(
...
-  CUfunction CuFunc = KernelCommandHandle->Kernel->get();
-  auto Result = setKernelParams(
-      CommandBuffer->Context, CommandBuffer->Device,
-      KernelCommandHandle->WorkDim, KernelCommandHandle->GlobalWorkOffset,
-      KernelCommandHandle->GlobalWorkSize, LocalWorkSize,
-      KernelCommandHandle->Kernel, CuFunc, ThreadsPerBlock, BlocksPerGrid);
-  if (Result != UR_RESULT_SUCCESS) {
-    return Result;
-  }
+    UR_CHECK_ERROR(
```
EwanC, line 1367: Why was this case in particular caught by the sanitizer, but the cases like [here](https://github.com/oneapi-src/unified-runtime/blob/ebc7760029503b367ae80653913544c75ee1cc8a/source/adapters/cuda/command_buffer.cpp#L819) where we have
`UR_CHECK_ERROR(getNodesFromSyncPoints(` outside of a try/catch aren't? Not that this change is wrong.

hdelan: This test was failing so the `throw` was caught. If a test passes and doesn't throw then this layer won't know that it *might* have thrown. I'll see if I can find any more cases where `try...catch` isn't being used but the most reliable way we can find such cases is if they fail.

EwanC: The Codeplay SYCL-Graph team have an internal ticket to go through the CUDA/HIP adapters and fix all these cases, so don't feel obliged to do here. Do you have a link/reference to which test "this test" is?

hdelan: I've caught some more `UR_CHECK_ERROR`s outside of `try catch` and fixed them in the last commit.

hdelan: > The Codeplay SYCL-Graph team have an internal ticket to go through the CUDA/HIP adapters and fix all these cases, so don't feel obliged to do here. Do you have a link/reference to which test "this test" is? Then can try write more ones like that when we make these changes.

There isn't actually a specific test set for this layer yet. The layer will just check if the adapter has thrown, and if so call `std::abort`. Potentially a test that would highlight all times `UR_CHECK_ERROR` is called outside of a try.. catch block is just to redefine the macro as

```c++
#define UR_CHECK_ERROR(Res) \
   throw;
```

Then this would be caught by the exception sanitiser layer and std::abort called with an intelligible error message.

CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ option(UR_USE_TSAN "enable ThreadSanitizer" OFF)
 option(UR_ENABLE_TRACING "enable api tracing through xpti" OFF)
 option(UR_ENABLE_SANITIZER "enable device sanitizer" ON)
 option(UR_ENABLE_SYMBOLIZER "enable symoblizer for sanitizer" OFF)
+option(UR_ENABLE_EXCEPTION_SANITIZER "enable exception sanitizer" OFF)
```
kbenzie, line 46: Not sure we need a separate option, we could always enable this as part of `UR_DEVELOPER_MODE` since that's enabled in CI.

CMakeLists.txt
```diff
@@ -205,6 +206,18 @@ else()
...
+    if(APPLE)
+        message(WARNING "Exception sanitizer layer isn't supported on macOS")
+        set(UR_ENABLE_EXCEPTION_SANITIZER OFF)
+    elseif(WIN32)
+        message(WARNING "Exception sanitizer layer isn't supported on Windows")
+        set(UR_ENABLE_EXCEPTION_SANITIZER OFF)
+    else()
+        add_compile_definitions(UR_ENABLE_EXCEPTION_SANITIZER)
+    endif()
+endif()
```
kbenzie, line 219: Always enable building the layer on Linux when developer mode is enabled.

```suggestion
if(UR_DEVELOPER_MODE AND CMAKE_SYSTEM_NAME STREQUAL Linux)
    set(UR_ENABLE_EXCEPTION_SANITIZER ON)
    add_compile_definitions(UR_ENABLE_EXCEPTION_SANITIZER)
endif()
```

.github/workflows/build-hw-reusable.yml
```diff
@@ -81,6 +81,7 @@ jobs:
         -DCMAKE_BUILD_TYPE=${{matrix.build_type}}
         -DUR_ENABLE_TRACING=ON
         -DUR_DEVELOPER_MODE=ON
+        -DUR_ENABLE_EXCEPTION_SANITIZER=ON
```
kbenzie, line 84: Don't need to set this if its roled into `UR_DEVELOPER_MODE`.

```suggestion
```

kbenzie (commented):
So the missing part here is actually enabling the layer in the CTS environment setup, something along the lines of this:

```diff
diff --git a/test/conformance/source/environment.cpp b/test/conformance/source/environment.cpp
index 006ea09b..737d26d2 100644
--- a/test/conformance/source/environment.cpp
+++ b/test/conformance/source/environment.cpp
@@ -90,6 +90,14 @@ uur::PlatformEnvironment::PlatformEnvironment(int argc, char **argv)
             error = "Failed to enable validation layer";
             return;
         }
+#ifdef UR_ENABLE_EXCEPTION_SANITIZER
+        if (urLoaderConfigEnableLayer(config, "UR_LAYER_EXCEPTION_SANITIZER") !=
+            UR_RESULT_SUCCESS) {
+            urLoaderConfigRelease(config);
+            error = "Failed to enable exception sanitizer layer";
+            return;
+        }
+#endif
     } else {
         error = "Failed to create loader config handle";
         return;
```

CMakeLists.txt
hdelan: Thanks, have removed.

hdelan:
> ```diff
> +            UR_RESULT_SUCCESS) {
> ```

Interesting, how was the layer working [here](https://github.com/oneapi-src/unified-runtime/actions/runs/11443472281/job/31836216797#step:9:268) without this call?

hdelan:
> So the missing part here is actually enabling the layer in the CTS environment setup, something along the lines of this:
> 
> ```diff
> diff --git a/test/conformance/source/environment.cpp b/test/conformance/source/environment.cpp
> index 006ea09b..737d26d2 100644
> --- a/test/conformance/source/environment.cpp
> +++ b/test/conformance/source/environment.cpp
> @@ -90,6 +90,14 @@ uur::PlatformEnvironment::PlatformEnvironment(int argc, char **argv)
>              error = "Failed to enable validation layer";
>              return;
>          }
> +#ifdef UR_ENABLE_EXCEPTION_SANITIZER
> +        if (urLoaderConfigEnableLayer(config, "UR_LAYER_EXCEPTION_SANITIZER") !=
> +            UR_RESULT_SUCCESS) {
> +            urLoaderConfigRelease(config);
> +            error = "Failed to enable exception sanitizer layer";
> +            return;
> +        }
> +#endif
>      } else {
>          error = "Failed to create loader config handle";
>          return;
> ```

Interesting, how was the layer working [here](https://github.com/oneapi-src/unified-runtime/actions/runs/11443472281/job/31836216797#step:9:268) without this call?

pbalcer (approved):

scripts/templates/exceptionddi.cpp.mako
```diff
@@ -0,0 +1,130 @@
...
+        %endfor
+        return result;
+    }
+
+    %endfor
+    ${x}_result_t
+    context_t::init(ur_dditable_t *dditable,
+                    const std::set<std::string> &,
+                    codeloc_data) {
+        ${x}_result_t result = ${X}_RESULT_SUCCESS;
```
pbalcer, line 118: this needs:
```
    if (!enabledLayerNames.count(name)) {
        return result;
    }
```

scripts/templates/exceptionddi.cpp.mako
```diff
@@ -0,0 +1,130 @@
...
+        %endif
+
+        %endfor
+        return result;
+    }
+
+    %endfor
+    ${x}_result_t
+    context_t::init(ur_dditable_t *dditable,
+                    const std::set<std::string> &,
```
pbalcer, line 116:
```suggestion
                    const std::set<std::string> &enabledLayerNames,
```

pbalcer:
> Interesting, how was the layer working here without this call?

The init function of all layers gets called unconditionally, and the layer themselves need to check whether they are on the enabled list.
This is so that a layer can force-enable itself if it's needed as a dependency of something else (e.g., validation layers).

hdelan:
> > Interesting, how was the layer working here without this call?
> 
> The init function of all layers gets called unconditionally, and the layer themselves need to check whether they are on the enabled list. This is so that a layer can force-enable itself if it's needed as a dependency of something else (e.g., validation layers).

Aha thanks @pbalcer that explains it

hdelan:
Have updated. Should be good to go.

pbalcer (approved):

igchor:
I think we should actually have a separate option to enable/disable this layer. Right now, as I understand, there is no way to disable this layer for L0 and L0 v2 adapter (other than disabling developer mode). 

In L0 and L0 v2, we use exceptions to propagate errors, expecting that the loader will translate them into error codes, like here: https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/v2/kernel.cpp#L77 .There should be no abort for L0

kbenzie:
> In L0 and L0 v2, we use exceptions to propagate errors, expecting that the loader will translate them into error codes, like here: https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/v2/kernel.cpp#L77 .There should be no abort for L0

My personal view of this is it's a bad idea and adapters should never allow exceptions to escape the UR entry points because it is a C API. The reason this layer was created is due to exceptions crossing that ABI boundry causing crashes.

igchor:
> > In L0 and L0 v2, we use exceptions to propagate errors, expecting that the loader will translate them into error codes, like here: https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/v2/kernel.cpp#L77 .There should be no abort for L0
> 
> My personal view of this is it's a bad idea and adapters should never allow exceptions to escape the UR entry points because it is a C API. The reason this layer was created is due to exceptions crossing that ABI boundry causing crashes.

If so, I think we should then enforce each function in an adapter to be wrapped by try/catch (that is, move those auto-generated checks from loader: https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_libapi.cpp#L2422 to each adapter function). We could also consider making all functions (and ddi pointers) noexcept. This might allow compiler/coverity to complain if there is an uncaught exception.

kbenzie:
> If so, I think we should then enforce each function in an adapter to be wrapped by try/catch (that is, move those auto-generated checks from loader: https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_libapi.cpp#L2422 to each adapter function).

Since the adapters are not generated from the yaml source, that could get quite manual and annoying.

> We could also consider making all functions (and ddi pointers) noexcept. This might allow compiler/coverity to complain if there is an uncaught exception.

This is an interesting approach and should be much less of a burden to add to adapters. I doubt we could do that as part of the `ur_api.h` header though. If we can make compilers emit diagnostics, that would be ideal.

igchor:
> > If so, I think we should then enforce each function in an adapter to be wrapped by try/catch (that is, move those auto-generated checks from loader: https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_libapi.cpp#L2422 to each adapter function).
> 
> Since the adapters are not generated from the yaml source, that could get quite manual and annoying.

Well, we could also add this try/catch when creating a ddi tables in adapters (which are auto-generated). Something like this:

```
urGetEnqueueProcAddrTable(...) {
    pDdiTable->pfnEventsWait = [](auto... args) { 
      try {
          return urEnqueueEventsWait(args...);
      } catch(...) { return exceptionToResult(std::current_exception()); 
    };
  
    ...    
}
```

I belive the compiler will optimize it so there should be no additional overhead but we can verify that.

kbenzie:
For the static linking use case there don't be any DDI tables, this needs to happen in the adapters not in some wrapping layer.

igchor:
Okay, I see; then it's probably case by case. I will add the necessary try/catch blocks to the v2 adapter.

martygrant:
## Information

## PR Migration

---

martygrant:

---

Issue #2215, opened 2024-10-17
Improve CI testing for "mock" adapter


RossBrunton:
As seen in https://github.com/oneapi-src/unified-runtime/issues/2212 , a bug arose when trying to run the mock adapter. However, CI reports all green, especially with a commit that broke the mock adapter.

We should implement proper CI testing for the mock adapter.

PR #2218, opened 2024-10-17
[NFC] Don't run conformance tests on mock adapter

RossBrunton:
Closes: #2215

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,6 +225,14 @@ if(UR_USE_MSAN)
     add_sanitizer_flag(memory)
 endif()
 
+if(NOT (UR_BUILD_ADAPTER_CUDA OR UR_BUILD_ADAPTER_HIP
+    OR UR_BUILD_ADAPTER_L0 OR UR_BUILD_ADAPTER_OPENCL
+    OR UR_BUILD_ADAPTER_NATIVE_CPU OR UR_BUILD_ADAPTER_L0_V2
+    OR UR_BUILD_ADAPTER_ALL))
+    message(WARNING "No adapters have been enabled; conformance tests will not be ran")
+    message(STATUS "Consider setting UR_BUILD_ADAPTER_*")
+endif()
+
 # Check if clang-format (in correct version) is available for Cpp code formatting.
 if(UR_FORMAT_CPP_STYLE)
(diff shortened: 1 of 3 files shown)


.github/workflows/cmake.yml
```diff
@@ -127,7 +127,7 @@ jobs:
 
     - name: Test
       working-directory: ${{github.workspace}}/build
-      run: ctest -C ${{matrix.build_type}} --output-on-failure -L "umf|loader|validation|tracing|unit|urtrace"
+      run: ctest -C ${{matrix.build_type}} --output-on-failure
```
kbenzie, line 130: This doesn't seem like a change we should make in this PR.

martygrant (approved):

test/layers/sanitizer/CMakeLists.txt
```diff
@@ -34,4 +34,6 @@ function(add_sanitizer_test name)
     set_sanitizer_test_properties(${name})
 endfunction()
 
-add_sanitizer_test(asan asan.cpp)
+if(ANY_ADAPTER)
```
aarongreig, line 37: not too sure but since this is linking `${PROJECT_NAME}::mock` above I suspect it might be designed to work with the mock adapter, some of the layer and tracing tests are

.github/workflows/cmake.yml
RossBrunton: Shall I expand the scope of this MR (via adding it to the title/description) or update the testing in a separate MR?

test/layers/sanitizer/CMakeLists.txt
aarongreig: Just had a look and it seems like the test is designed to run with a normal adapter, so this change does make sense I think. Probably the mock inclusion was just copy/paste. I'm not seeing a segfault with any adapter including the mock adapter, and the test is pretty trivial so might be worth looking into that - could be something dodgy with your env

.github/workflows/cmake.yml
kbenzie: *If* we make this change, I think it should be a separate PR. This makes this not NFC, for testing coverage anyway.

test/layers/sanitizer/CMakeLists.txt
RossBrunton: Reverted it for now, we can look at it more if we decide to expand the testing later.

PR #2219, opened 2024-10-17
[CL] Support using old ICD loaders

kbenzie:
While we recommend using the latest OpenCL-ICD-Loader version this is not always possible, perhaps due to a system library being outdated with no ability to update due to lacking admin privileges. This can cause the OpenCL adapter library to silently fail to load or worse cause a crash.

This patch changes entry points adding in OpenCL 2.2 or later to be loaded from the OpenCL-ICD-Loader at runtime instead of during startup by the operating system. The entry point function pointers are stored in the OpenCL adapter's `ur_adapter_handle_t` and all previous uses are replaced with checked calls to these function pointers.

Previously `clGetExtensionFunctionAddressForPlatform` was being used, however this is not guaranteed to return valid function pointers for core OpenCL entry points even if they are present.

https://github.com/intel/llvm/pull/15741


source/adapters/opencl/adapter.cpp
```diff
@@ -8,16 +8,49 @@
...
+#ifdef _MSC_VER
+#include <Windows.h>
+#else
+#include <dlfcn.h>
+#endif
+
+ur_adapter_handle_t_::ur_adapter_handle_t_() {
+#ifdef _MSC_VER
+  // Loading OpenCL.dll increments the libraries internal reference count.
+  auto handle = LoadLibraryA("OpenCL.dll");
```
omarahmed1111, line 24: I think we could that this handle is not `NULL` and if so, we print an error that it failed to find the opencl module.

omarahmed1111: Is this guaranteed to bring the same `opencl.dll` that was used in the adapter initially. I am just concerned that this will load a system `opencl.dll` and the adapter is already using a local fetch of the opencl.dll that lives in `build/bin` folder and they would have a different version (not still sure if that would cause also a harm if that is the case)?

source/adapters/opencl/adapter.cpp
```diff
@@ -8,16 +8,49 @@
...
+  FUNC = reinterpret_cast<decltype(::FUNC) *>(GetProcAddress(handle, "FUNC"));
+#include "core_functions.def"
+#undef CL_CORE_FUNCTION
+
+  // So we can safely decrement it here wihtout actually unloading OpenCL.dll.
+  FreeLibrary(handle);
+#else
+  // Loading libOpenCL.so to get the library handle but don't dlclose it as
+  // this causes a segfault when attempting to call any OpenCL entry point.
+  auto handle = dlopen("libOpenCL.so", RTLD_LOCAL);
```
omarahmed1111, line 36: I have the same concern here that I have on `LoadLibraryA` on loading the same opencl.dll used.

source/adapters/opencl/context.cpp
```diff
@@ -169,6 +170,13 @@ UR_APIEXPORT ur_result_t UR_APICALL urContextCreateWithNativeHandle(
 UR_APIEXPORT ur_result_t UR_APICALL urContextSetExtendedDeleter(
     ur_context_handle_t hContext, ur_context_extended_deleter_t pfnDeleter,
     void *pUserData) {
+  if (!ur::cl::getAdapter()->clSetContextDestructorCallback) {
```
omarahmed1111, line 173: NIT: I feel `ur::cl::getAdapter()` could be saved in a variable but that is just styling thing so feel free to ignore.

source/adapters/opencl/adapter.cpp
kbenzie: > I think we could that this handle is not NULL and if so, we print an error that it failed to find the opencl module.

We've already loaded OpenCL.dll by this point, the adapter would fail to load if it were not present and this code would not be capable of being executed. We're just asking for the handle to the already loaded OpenCL.dll.

> Is this guaranteed to bring the same opencl.dll that was used in the adapter initially.

From the remarks section in [LoadLibraryA](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya):

> If the specified module is a DLL that is not already loaded for the calling process, the system calls the DLL's [DllMain](https://learn.microsoft.com/en-us/windows/desktop/Dlls/dllmain) function with the DLL_PROCESS_ATTACH value.

Thie implication being that if its already loaded it does not load the library a different library of the same name.

source/adapters/opencl/adapter.cpp
kbenzie: The way we link `libOpenCL.so` is already making use of the standard dynamic library loading used by `dlopen()`. We don't link to an absolute path but rather the `libOpenCL.so.1` name and that's used for the library lookup. For example, on my system it's found `/lib/x86_64-linux-gnu/libOpenCL.so.1`.

```
$ ldd ~build/install/lib/libur_adapter_opencl.so
...
        libOpenCL.so.1 => /lib/x86_64-linux-gnu/libOpenCL.so.1 (0x00007b389a819000)
...
```

But if I set `LD_LIBRARY_PATH` it will use that.

```
$ LD_LIBRARY_PATH=$Projects/KhronosGroup/OpenCL-ICD-Loader/build/bin \
    ldd ~build/install/lib/libur_adapter_opencl.so
...
        libOpenCL.so.1 => /home/benie/Projects/KhronosGroup/OpenCL-ICD-Loader/build/bin/libOpenCL.so.1 (0x0000771087e10000)
...
```

My point being, the look up for done by `ld` during initialization is exacrlt the same as the lookup done by `dlopen()`. Not only that but `dlopen()` won't load the library again if its already present.

But just to be sure, lets invoke @ldrumm to confirm or deny my understanding.

source/adapters/opencl/context.cpp
kbenzie: I will disagree with you on this point.

source/adapters/opencl/adapter.cpp
ldrumm: man dlopen:

>   If  the  same shared object is opened again with dlopen(), the same object handle is returned.

source/adapters/opencl/adapter.cpp
omarahmed1111: Cool, that make sense. Thanks for clarification!

source/adapters/opencl/adapter.cpp
omarahmed1111: make sense. Thanks for clarification!

omarahmed1111 (approved):

omarahmed1111 (approved):

PR #2220, opened 2024-10-17
No longer expect segfaults in match files

RossBrunton:
A number of match files check for `Segmentation Fault`, which happens
when the test runner crashes. This change removes that check and
replaces it with checks for the tests that actually cause the segfault.
As the test runner no longer segfaults, more tests are executed and so
more non-segfaulting fails needed to be added to the match files.

In addition, a number of match files had the check for `Segmentation
Fault` removed if they didn't actually segfault.

Note that this will have no effect for the `ctest_parser.py` script, as
it ignores the match file logic.

Note for anyone here early: This is based on my system, it's possible that some fails listed here are due to my setup, rather than actual level zero issues!


kbenzie (approved):

kbenzie (approved):

RossBrunton:
@oneapi-src/unified-runtime-level-zero-write and @oneapi-src/unified-runtime-native-cpu-write  , can I get a quick look at this? These aren't new failures, but failures that our way of running tests happened to mask.

test/conformance/memory/memory_adapter_level_zero.match
```diff
@@ -14,4 +13,22 @@ urMemBufferPartitionTest.InvalidValueBufferCreateInfoOutOfBounds/Intel_R__oneAPI
 {{OPT}}urMemImageCreateTestWithImageFormatParam.Success/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}___UR_IMAGE_CHANNEL_ORDER_RGBA__UR_IMAGE_CHANNEL_TYPE_UNSIGNED_INT16
 {{OPT}}urMemImageCreateTestWithImageFormatParam.Success/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}___UR_IMAGE_CHANNEL_ORDER_RGBA__UR_IMAGE_CHANNEL_TYPE_UNSIGNED_INT32
 {{OPT}}urMemImageCreateTestWithImageFormatParam.Success/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}___UR_IMAGE_CHANNEL_ORDER_RGBA__UR_IMAGE_CHANNEL_TYPE_HALF_FLOAT
-{{OPT}}urMemImageCreateTestWithImageFormatParam.Success/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}___UR_IMAGE_CHANNEL_ORDER_RGBA__UR_IMAGE_CHANNEL_TYPE_FLOAT
\ No newline at end of file
+{{OPT}}urMemImageCreateTestWithImageFormatParam.Success/Intel_R__oneAPI_Unified_Runtime_over_Level_Zero___{{.*}}___UR_IMAGE_CHANNEL_ORDER_RGBA__UR_IMAGE_CHANNEL_TYPE_FLOAT
+# These tests fail in the "Multi device testing" job, but pass in the hardware specific test
```
pbalcer, line 17: (not a comment to the PR)
@nrspruit I think this may be expected. Multi device testing is 2T PVC, which, afaik, doesn't support the image api, right?
But it shouldn't crash, so...

pbalcer (approved):

test/conformance/memory/memory_adapter_level_zero.match
RossBrunton: It's entirely possible that the crash is from another test that happens to run before the image tests.

pbalcer: > It's entirely possible that the crash is from another test that happens to run before the image tests.

In which case the test is missing a skip if the feature is unsupported.

nrspruit (approved):

PietroGhg (approved):

RossBrunton:
Merging without intel/llvm change, since this only affects UR CTS.

PR #2222, opened 2024-10-18
Enable -flto and -fsanitize=cfi in clang

RossBrunton:
Note that this flagged up a few issues, for which followup tickets
were created:
* https://github.com/oneapi-src/unified-runtime/issues/2323
* https://github.com/oneapi-src/unified-runtime/issues/2309
* https://github.com/oneapi-src/unified-runtime/issues/2324


RossBrunton:
@PatKamin is there any particular reason the fuzz testing explicitly sets "-g"? It seems to trigger a weird edge case when used in release mode when -lto is also enabled, which crashes the linker.

PatKamin:
> @PatKamin is there any particular reason the fuzz testing explicitly sets "-g"? It seems to trigger a weird edge case when used in release mode when -lto is also enabled, which crashes the linker.

I've added it after libFuzzer examples: https://llvm.org/docs/LibFuzzer.html#fuzzer-usage
I think that it is safe to remove it as mixing Release flags prepared by CMake with this debug flag might cause issues like the one you have.

test/fuzz/CMakeLists.txt
```diff
@@ -51,7 +51,9 @@ target_link_libraries(fuzztest-base
     ${PROJECT_NAME}::headers
     ${PROJECT_NAME}::common
     -fsanitize=fuzzer)
-target_compile_options(fuzztest-base PRIVATE -g -fsanitize=fuzzer)
+# When built with -g and -flto (which is required by some hardening flags), this causes a segfault in (upstream)
+# LLVM 14-15 while linking when CMAKE_BUILD_TYPE is Release
```
aarongreig, line 55: should we have an issue for this?

RossBrunton: I don't think so. It feels like an upstream compiler bug, and we don't need to build with `-g` here anyway.

isaacault: Maybe an upstream issue is appropriate then?

.github/workflows/build-fuzz-reusable.yml
```diff
@@ -58,6 +60,7 @@ jobs:
         -DUR_USE_ASAN=ON
         -DUR_USE_UBSAN=ON
         -DUR_BUILD_ADAPTER_L0=ON
+        -DCXX_HAS_CFI_SANITIZE=OFF
```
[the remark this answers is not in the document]
RossBrunton: This was only meant to be a temporary workaround for the CI failures, but I think it does make sense for this to be a cmake config option.

I've added `UR_USE_CFI`, which is default on.

EwanC (approved):
command-buffer match file changes LGTM, I've created an internal ticket for our team to look into the fails further.

RossBrunton:
@oneapi-src/unified-runtime-level-zero-v2-write This flags up an issue with the event pool tests, so I've disabled it for now if CFI is to be enabled.

Could you look and see if this is something you're okay with me doing?

aarongreig (approved):

martygrant (approved):

PR #2227, opened 2024-10-21
Handle INVALID_VALUE in OCL urPlatformGet

RossBrunton:
The OpenCL function clGetPlatformIDs may return CL_INVALID_VALUE, which
needs to be converted into appropriate UR returns as follows:

* If a non-null platforms list is provided, but the number of elements
  in it is 0, then return UR_RESULT_ERROR_INVALID_SIZE.
* If neither platform or sizes outputs are provided, then return
  UR_RESULT_ERROR_INVALID_VALUE.

This required a spec change.

This fixes a bug Intel are tracking internally as URT-831.

### DIFF
--- a/source/adapters/opencl/platform.cpp
+++ b/source/adapters/opencl/platform.cpp
@@ -96,6 +96,11 @@ urPlatformGet(ur_adapter_handle_t *, uint32_t, uint32_t NumEntries,
     }
   }
 
+  /* INVALID_VALUE is returned when the size is invalid, special case it here */
+  if (Result == CL_INVALID_VALUE && phPlatforms != nullptr && NumEntries == 0) {
+    return UR_RESULT_ERROR_INVALID_SIZE;
+  }
+
   return mapCLErrorToUR(Result);
 }
 
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -1054,6 +1054,8 @@ typedef enum ur_adapter_backend_t {
 ///         + `NULL == phAdapters`
 ///     - ::UR_RESULT_ERROR_INVALID_SIZE
 ///         + `NumEntries == 0 && phPlatforms != NULL`
+///     - ::UR_RESULT_ERROR_INVALID_VALUE
+///         + `pNumPlatforms == NULL && phPlatforms == NULL`
 UR_APIEXPORT ur_result_t UR_APICALL
 urPlatformGet(
     ur_adapter_handle_t *phAdapters,   ///< [in][range(0, NumAdapters)] array of adapters to query for platforms.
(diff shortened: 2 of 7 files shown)


source/adapters/opencl/platform.cpp
```diff
@@ -96,6 +96,18 @@ urPlatformGet(ur_adapter_handle_t *, uint32_t, uint32_t NumEntries,
     }
   }
 
+  if (Result == CL_INVALID_VALUE) {
+    // From OpenCL:
+    // if num_entries is equal to zero and platforms is not NULL or if both
+    // num_platforms and platforms are NULL.
+    if (NumEntries == 0 && phPlatforms != nullptr) {
```
AlexeySachkov, line 103: Do we really need to repeat this check if it is already happening within validation layer?

https://github.com/oneapi-src/unified-runtime/blob/c742ca49efb12380a35b8b0b467e6577ab8174ce/source/loader/layers/validation/ur_valddi.cpp#L234-L236

I'm asking mostly out of curiosity, because I don't know what is the error handling policy within UR

aarongreig: We don't enable the validation layer from the sycl RT, although I think that decision was made while we were trying to fix a performance regression that later turned out to be something totally unrelated. I don't like having this check twice, maybe we should look to see what the perf impact of enabling just basic parameter validation would be

kbenzie: The validation layer is opt-in and is disabled by default in the SYCL RT. If checks are required for correctness based on SYCL RT usage patterns then they need to exist in the adapter.

AlexeySachkov: > The validation layer is opt-in and is disabled by default in the SYCL RT. If checks are required for correctness based on SYCL RT usage patterns then they need to exist in the adapter.

https://github.com/intel/llvm/blob/sycl/sycl/source/detail/adapter.hpp#L103 - if you look at the code here, we do not have a check that `platformCount` returned to SYCL RT is 0. [`std::vector::data`](https://en.cppreference.com/w/cpp/container/vector/data) does not guarantee that `nullptr` would be returned for an empty vector, so theoretically we _can_ encounter a case that we check in the adapter. However, in my local testing I wasn't able to encounter it, because `vector::data` always returned `nullptr` (empty `vector` resized to `0` elements).

I think that we should definitely update the SYCL RT code here to avoid performing an incorrect call and therefore it should allow to simplify the code in the adapter.

RossBrunton: I've added that check now in the associated intel/llvm change.

AlexeySachkov (approved):

martygrant (approved):

PR #2229, opened 2024-10-22
[DeviceASAN] Fix incorrect symbolizer output for shared libraries

zhaomaosu:
intel/llvm part: https://github.com/intel/llvm/pull/15797


source/loader/layers/sanitizer/linux/symbolizer.cpp
```diff
@@ -27,6 +29,20 @@ llvm::symbolize::PrinterConfig GetPrinterConfig() {
     return Config;
 }
 
+uptr GetModuleBase(const char *ModuleName) {
+    ModuleInfo Data{ModuleName, 0};
+    dl_iterate_phdr(
+        [](struct dl_phdr_info *Info, size_t, void *Arg) {
```
pbalcer, line 35: It should be possible to just pass `ModuleName` into this lambda function instead of wrapping it into a struct.

source/loader/layers/sanitizer/linux/symbolizer.cpp
```diff
@@ -7,6 +7,8 @@
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  *
  */
+#include "common.hpp"
+#include "link.h"
```
yingcong-wu, line 11: Is `"link.h"` a system header?

source/loader/layers/sanitizer/linux/symbolizer.cpp
zhaomaosu: Thanks, done.

source/loader/layers/sanitizer/linux/symbolizer.cpp
zhaomaosu: updated

yingcong-wu (approved):

AllanZyne (approved):

PR #2230, opened 2024-10-22
don't call xpti if there are no subscribers

pbalcer:
This avoids the overhead of preparing data for xpti, and the cost of the xpti call itself, if nothing is subscribed to the ur.call xpti call stream.


pbalcer:
@oneapi-src/unified-runtime-maintain please review

aarongreig (approved):

PR #2232, opened 2024-10-23
[DeviceASAN][NFC] Code Restructure

AllanZyne:
LLVM: https://github.com/intel/llvm/pull/15843
code refactoring for implementing the msan layer


AllanZyne:
Hi @oneapi-src/unified-runtime-maintain, could you please prioritize the review of this PR?
This PR has no functional change, and is essential for our further development!
Thank you very much!

zhaomaosu (approved):
lgtm

AllanZyne:
I've made patch to fix this in LLVM part.

```
ninja: error: '/home/test-user/actions-runners/opencl-1/_work/unified-runtime/unified-runtime/ur-repo/source/loader/layers/sanitizer/asan_libdevice.hpp', needed by 'lib/libsycl-asan-amdgcn-amd-amdhsa.bc', missing and no known rule to make it
```

AllanZyne:
Hi @pbalcer, could you please review and merge this PR before other sanitizer related PRs?
I've synced with other members.
Thank you very much!

source/loader/layers/sanitizer/sanitizer_common/sanitizer_stacktrace.cpp
```diff
@@ -93,10 +92,9 @@ void StackTrace::print() const {
         BacktraceInfo BI = BacktraceSymbols[i];
 
         // Skip runtime modules
-        if (!getContext()->interceptor->getOptions().Debug &&
```
yingcong-wu, line 96: I think we should keep this, maybe by creating a base class of option that have the common options available for all sanitizers, and in this case, the "debug" switch. But that can be addressed later.

AllanZyne: I agree. Besides, I started to prefer to create a single option object for all sanitizers. 
We can discuss this later.

pbalcer:
> Hi @pbalcer, could you please review and merge this PR before other sanitizer related PRs? I've synced with other members. Thank you very much!

Will do, is the PR final? I see that you are still making changes.

AllanZyne:
> > Hi @pbalcer, could you please review and merge this PR before other sanitizer related PRs? I've synced with other members. Thank you very much!
> 
> Will do, is the PR final? I see that you are still making changes.

Yes, it's final. I just pushed a commit to address a comment by Yingcong.

pbalcer (approved):
lgtm, just a few nits.

Not insisting, but I think you should drop the prefixes on files if they are not necessary, so "common" instead of "sanitizer_common", "allocator.hpp" instead of "sanitizer_allocator.hpp" and so on.

pbalcer:
I'll merge this once you get an approval on your intel/llvm PR (you are missing llvm-reviewers-runtime, I'll approve on behalf of unified-runtime-reviewers).

AllanZyne:
> lgtm, just a few nits.
> 
> Not insisting, but I think you should drop the prefixes on files if they are not necessary, so "common" instead of "sanitizer_common", "allocator.hpp" instead of "sanitizer_allocator.hpp" and so on.

For file naming style, I completely use the same way in compiler_rt (https://github.com/intel/llvm/tree/sycl/compiler-rt/lib/sanitizer_common) for convenience.

pbalcer:
> > lgtm, just a few nits.
> > Not insisting, but I think you should drop the prefixes on files if they are not necessary, so "common" instead of "sanitizer_common", "allocator.hpp" instead of "sanitizer_allocator.hpp" and so on.
> 
> For file naming style, I completely use the same way in compiler_rt (https://github.com/intel/llvm/tree/sycl/compiler-rt/lib/sanitizer_common) for convenience.

Ok, makes sense, it's better to be consistent. Disregard that then.

source/loader/layers/sanitizer/asan/asan_interceptor.cpp
```diff
@@ -918,4 +920,24 @@ USMLaunchInfo::~USMLaunchInfo() {
     assert(Result == UR_RESULT_SUCCESS);
 }
 
+} // namespace asan
+
+using namespace asan;
+
+static AsanInterceptor *interceptor;
```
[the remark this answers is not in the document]
AllanZyne: > But can't we just leave the AsanInterceptor inside of the layer object?

Yes, I agree that using global variable is not a good idea. 
We are going to develop each sanitizer as independent as possible at the beginning, and when these features are stable, we will try to extract their common characteristics, and create some base classes then.
Now I want the outside world (layer context) don't know how each sanitizer is implemented, and just give it some simple interfaces.

PR #2177, opened 2024-10-07
Improvements to align CTS and Spec for Memory

martygrant:
- Add test for expecting UR_RESULT_ERROR_INVALID_NULL/HOST_POINTER for urMemImageCreate and urMemBufferCreate
- Add more checks for UR_RESULT_ERROR_INVALID_IMAGE_FORMAT_DESCRIPTOR for urMemImageCreate
- Added missing error condition to spec for urMemImageCreate - UR_STRUCTURE_TYPE_IMAGE_DESC != pImageDesc->stype
- Add tests for expecting UR_RESULT_ERROR_INVALID_NULL_HANDLE/POINTER for urMemBuffer/ImageCreateWithNativeHandle
- Remove skip for urMemImageCreateWithNativeHandle - buffer was invalid, creating it in the test fixture definition now
- Update image format used in urMemImageText fixture as it was invalid and would cause previously skipped tests to fail (they now pass)
- Fix DDI table entry for urMemImageCreateWithNativeHandle for OpenCL
- Add test for calling urMemBufferPartition with different ur_mem_flag_t flags
- Add missing UR_MEM_INFO_REFERENCE_COUNT to spec for urMemGetInfo. Added a test for this and checking ref count is updated in urMemRetain/Release

PR #2233, opened 2024-10-23
Improvements to align CTS and Spec for Memory

martygrant:
https://github.com/intel/llvm/pull/15619

- Add tests for UR_RESULT_ERROR_INVALID_NULL/HOST_POINTER for urMemImageCreate and urMemBufferCreate
- Add missing error condition to spec for urMemImageCreate for checking type of image description struct
- Add tests for UR_RESULT_ERROR_INVALID_NULL_HANDLE/POINTER for urMemBuffer/ImageCreateWithnativeHandle
- Remove skip for urMemImageCreateWithNativeHandle - buffer was invalid as it was never instantiated in fixture struct
- Update image format used in urMemImageTest fixture as it was invalid and likely to cause previously skipped test to fail (which now pass)
- Add missing DDI table entry for urMemImageCreateWithNativeHandle for OpenCL
- Add test for using different ur_mem_flag_t flags with urMemBufferPartition
- Add missing UR_MEM_INFO_REFERENCE_COUNT to spec for urMemGetInfo and added test for this in urMemRetain/Release
- Removed assert in L0 urMemGetInfo which would fail if mem type is not image or the query is not a context query
- Fixed potential bug with HIP urMemGetNativeHandle which would fail if the mem handle was an image because use of std::variant
-  Check if properties param is null in L0 urMemBufferCreateWithNativeHandle and set OwnNativeHandle to false by default


martygrant:
@RossBrunton @hdelan @PietroGhg apologies you all reviewed a previous PR (https://github.com/oneapi-src/unified-runtime/pull/2177) for this change but I rebased a number of times fighting with CI failures and made some changes so though it would be good to just have it reviewed afresh. The main notable change is in [the hip adapter](https://github.com/oneapi-src/unified-runtime/pull/2233/files#diff-cdda61bf8ead17f141255a7bceb4231d050653676e565fde5aa6449a30a428b2R321) so image buffer types can be handled in `urMemGetNativeHandle`, and not letting some tests pass when entry points are unsupported if they aren't optional.

PietroGhg (approved):

martygrant:
@oneapi-src/unified-runtime-cuda-write @oneapi-src/unified-runtime-hip-write @oneapi-src/unified-runtime-level-zero-write looking for a review for this PR please 🙏

kbenzie (approved):

konradkusiak97 (approved):
LGTM!

nrspruit (commented):
LGTM for L0

nrspruit (approved):

PR #2234, opened 2024-10-23
[L0] Phase 2 of Counter-Based Event Implementation

winstonzhang-intel:
LLVM/SYCL test: https://github.com/intel/llvm/pull/16095

### DIFF
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -803,7 +803,22 @@ urEventWait(uint32_t NumEvents, ///< [in] number of events in the event list
           } else {
             ZE2UR_CALL(zeHostSynchronize, (ZeEvent));
           }
-          Event->Completed = true;
+          if (Event->CounterBasedEventsEnabled &&
+              Event->CommandList.value()->second.ZeFence &&
+              Event->CommandList.value()->second.ZeFenceInUse) {
+            while (true) {
+              logger::debug("Event completed, checking fence status");
+              ze_result_t ZeResult =
+                  ZE_CALL_NOCHECK(zeFenceQueryStatus,
+                                  (Event->CommandList.value()->second.ZeFence));
+              if (ZeResult == ZE_RESULT_SUCCESS) {
+                Event->Completed = true;
+                break;
+              }
+            }
+          } else {
+            Event->Completed = true;
+          }
         }
       }
       if (auto Q = Event->UrQueue) {
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -1272,7 +1272,8 @@ ur_result_t EventCreate(ur_context_handle_t Context, ur_queue_handle_t Queue,
   }
 
   if (auto CachedEvent = Context->getEventFromContextCache(
-          HostVisible, ProfilingEnabled, Device, CounterBasedEventEnabled)) {
+          HostVisible, ProfilingEnabled, Device, CounterBasedEventEnabled,
+          UsingImmediateCommandlists)) {
     *RetEvent = CachedEvent;
     return UR_RESULT_SUCCESS;
   }
--- a/source/adapters/level_zero/queue.cpp
+++ b/source/adapters/level_zero/queue.cpp
@@ -1338,17 +1338,20 @@ ur_queue_handle_t_::executeCommandList(ur_command_list_ptr_t CommandList,
   }
 
   if (!UsingImmCmdLists) {
-    int numCompleted = 0;
-    for (auto &Event : CommandList->second.EventList) {
-      if (Event->ZeEvent) {
-        ze_result_t ZeResult =
-            ZE_CALL_NOCHECK(zeEventQueryStatus, (Event->ZeEvent));
-        if (ZeResult == ZE_RESULT_SUCCESS)
+    if (this->CounterBasedEventsEnabled) {
+      size_t numCompleted = 0;
+      for (auto &Event : CommandList->second.EventList) {
+        if (Event->ZeEvent) {
+          ze_result_t ZeResult =
+              ZE_CALL_NOCHECK(zeEventQueryStatus, (Event->ZeEvent));
+          if (ZeResult != ZE_RESULT_SUCCESS)
+            break;
           numCompleted++;
+        }
       }
+      if (numCompleted == CommandList->second.EventList.size())
+        return UR_RESULT_SUCCESS;
     }
-    if (numCompleted == CommandList->second.EventList.size())
-      return UR_RESULT_SUCCESS;
     // In this mode all inner-batch events have device visibility only,
     // and we want the last command in the batch to signal a host-visible
     // event that anybody waiting for any event in the batch will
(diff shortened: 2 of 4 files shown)


winstonzhang-intel:
Reuploaded. The failure in sycl/test-e2e was due to the following issue (https://github.com/intel/llvm/issues/15737) and not this patch.

nrspruit (approved):

pbalcer:
> Only tests that failed were in sycl/test-e2e/syclcompat and the failure was due to the following issue ([intel/llvm#15737](https://github.com/intel/llvm/issues/15737)) and not this patch.

That issue doesn't specify what was the root cause or the fix. It simply states that the issue was gone after updating PRs to tip of intel/llvm.

nrspruit:
> The CI on intel/llvm failed with syclcompat:
> 
> ```
> ********************
> Failed Tests (1):
>   SYCL :: syclcompat/util/util_shift_sub_group_right.cpp
> ```

The sycl compat tests are failing sporadically according to the compiler QA team. These are failing before this patch according to that team.

winstonzhang-intel:
Checking on my machine now. Will post log here later.

winstonzhang-intel:
Log of running util_shift_sub_group_right.cpp with llvm-lit 10 times posted below. 
[util_shift_sub_group_right.log](https://github.com/user-attachments/files/17546990/util_shift_sub_group_right.log)

winstonzhang-intel:
Additionally, all test in LLVM/sycl are passing now: https://github.com/intel/llvm/pull/15834

pbalcer:
I restarted the CI on the PR and I see a syclcompat failing again:
```
********************
Failed Tests (1):
  SYCL :: syclcompat/util/util_select_from_sub_group.cpp
```


> Log of running util_shift_sub_group_right.cpp with llvm-lit 10 times posted below. [util_shift_sub_group_right.log](https://github.com/user-attachments/files/17546990/util_shift_sub_group_right.log)

This is clearly a timing issue. When running the whole suite, the CI will, I think, spawn up to ncpu workers. That might affect the test.

callumfare:
@winstonzhang-intel @pbalcer This PR is labeled ready to merge, but given the last comment is that definitely the case? I want to check before selecting this PR to be merged next. It also looks like the linked intel/llvm PR has failed CI which should be resolved first.

pbalcer:
@callumfare apologies, you are right, I should have removed the tag. Winston is still investigating the issue, and we cannot merge this in its current state because it will introduce failures in sycl.

source/adapters/level_zero/queue.cpp
```diff
@@ -1338,6 +1338,17 @@ ur_queue_handle_t_::executeCommandList(ur_command_list_ptr_t CommandList,
   }
 
   if (!UsingImmCmdLists) {
+    int numCompleted = 0;
+    for (auto &Event : CommandList->second.EventList) {
+      if (Event->ZeEvent) {
+        ze_result_t ZeResult =
+            ZE_CALL_NOCHECK(zeEventQueryStatus, (Event->ZeEvent));
+        if (ZeResult == ZE_RESULT_SUCCESS)
```
nrspruit, line 1346: to speed up this check, you can reverse this logic and exit given one event not completed. That way the if check later is not needed.

MichalMrozek: querying status of events is expenisve operation, why is this needed?

MichalMrozek: why on execute we check for event status ? This doesn't seem to be related operations.

winstonzhang-intel: Going to add in the check for cb events and only query the status of counting events.

MichalMrozek:
I still don't understand why event state determines whether command list was executed or not.
This doesn't seem right, command list should known whether it was send for execution and this shouldn't be connected to event state at all.

nrspruit:
> I still don't understand why event state determines whether command list was executed or not. This doesn't seem right, command list should known whether it was send for execution and this shouldn't be connected to event state at all.

Winston and I did some more digging and we found the problem is that we are seeing the Event complete before the Fence for that command completed.

That should not be possible so we are checking if there is a state issue, but this could be pointing to a bug in the driver. It is very sporadic.

MichalMrozek:
events would be signaled earlier then fences, that would be expected
fence signaling is after event signaling

nrspruit:
> events would be signaled earlier then fences, that would be expected fence signaling is after event signaling

The event is signaling before the memory state was updated hence the data corruption being seen.

For a single command in a command list, the fence and event should signal around the same time.

The issue is that the event is signaling before the memory was actually updated.

This is where the event state is being updated too early and why we are checking the state is correct for the counting events.

nrspruit:
When the test fails is when the Fence for a single command in a single command list is unsignaled, but the event is signaled.

When this happens with counting events, the memory is not updated yet and we are seeing data corruption. 

This only occurs when the counting event signals before the fence.

MichalMrozek:
does the event have proper signaling flags to indicate host visiblity ?

nrspruit:
> does the event have proper signaling flags to indicate host visiblity ?

yes, the event being waited on is a host visible event. The state indicates it was unsignaled, then signaled after the execute completed with it signaled from the device, but the data is invalid at time of signaling. The issue occurs within 1 nanosecond of time before the memory gets updated. Small changes and delays in the code avoids/masks the issue.

This is our current focus since the memory is not properly updated in this scenario exposed by the sycl compat tests.

MichalMrozek:
but does it have signal scope set to host ?
on what HW is the problem ?
Have you collected AUB to see if the programming of event signal is correct?

nrspruit:
> but does it have signal scope set to host ? on what HW is the problem ? Have you collected AUB to see if the programming of event signal is correct?

Yes, the signal scope is set to host for these events.

Thus far this appears to only reproduce on Intel GEN12 Integrated GPUs.

Discrete gpus don't reproduce this issue thus far and on other Intel GEN12 machines it sometimes does not reproduce so this may be a KMD/UMD difference between systems.

We just found this as being the difference last night so the plan is to fully verify that this is the consistent issue and collect the AUB files.

If we can confirm this, then a bug will be filed internally.

winstonzhang-intel:
The latest patch provides a workaround for the issue of event completion without fence completion during urEventWait.

source/adapters/level_zero/event.cpp
```diff
@@ -803,7 +803,22 @@ urEventWait(uint32_t NumEvents, ///< [in] number of events in the event list
           } else {
             ZE2UR_CALL(zeHostSynchronize, (ZeEvent));
           }
-          Event->Completed = true;
+          if (Event->CounterBasedEventsEnabled &&
+              Event->CommandList.value()->second.ZeFence &&
+              Event->CommandList.value()->second.ZeFenceInUse) {
+            while (true) {
```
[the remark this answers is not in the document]
winstonzhang-intel: That makes sense, the issue is that there is the case where the fence is somehow signaled before the last event is signaled. Maybe we can check if it is the last event in the commandlist?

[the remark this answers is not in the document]
MichalMrozek: No , those WA's are not valid.
Either we fix the problem or don't enable the feature on affected GPUs.

martygrant:
## Information

## PR Migration

---

PR #2237, opened 2024-10-24
[DeviceAsan] Symbolizer may destruct before SanitizerInterceptor 

yingcong-wu:
We may try to use symbolizer in the destructor of SanitizerInterceptor, but since we are using a static Symbolizer, we cannot control it's destruction order and it can be destructed before SanitizerInterceptor layer. This patch tries to manually let the Symboilzer destruct at the very end of the exit process.


source/loader/layers/sanitizer/linux/symbolizer.cpp
```diff
@@ -12,9 +12,20 @@
...
 namespace ur_sanitizer_layer {
 
-llvm::symbolize::LLVMSymbolizer *GetSymbolizer() {
-    static llvm::symbolize::LLVMSymbolizer Symbolizer;
-    return &Symbolizer;
+llvm::symbolize::LLVMSymbolizer *GetSymbolizer(bool destruct = false) {
+    static auto *Instance = new llvm::symbolize::LLVMSymbolizer{};
+    if (destruct) {
+        Instance = nullptr;
+        delete Instance;
```
AllanZyne, line 19: swap line

yingcong-wu: nice catch.

AllanZyne (approved):

zhaomaosu (approved):
lgtm

source/loader/layers/sanitizer/linux/symbolizer.cpp
```diff
@@ -12,9 +12,21 @@
 
 namespace ur_sanitizer_layer {
 
+static llvm::symbolize::LLVMSymbolizer *Symbolizer = nullptr;
+
 llvm::symbolize::LLVMSymbolizer *GetSymbolizer() {
```
pbalcer, line 17: can't you simply hold this value inside of the layer's context? Those structures are carefully managed to avoid lifetime issues. I don't like having related globals if we can avoid them.

yingcong-wu: That would be another option. 

But I want to avoid having such value inside the layer's context because I think the layer does not have to know about this symbolizer and that makes them coupled. And this will only construct the symbolizer when needed.

Also, according to our teammates, we are having this "linux" folder because they may not be available for other platforms(although I don't think we will support other platforms in near future), and that is another thing. 

And having a symbolizer is not also true for such case because the symbolizer is not available by default. 

@zhaomaosu suggests that we can expose a weak function that destructs the symbolizer and calls it in the destructor of SanitizerInterceptor.

pbalcer: I know this makes the code less clean, but I think it's a compromise worth making. The constructor/destructor ordering in sycl rt is very very fragile, and 
I guarantee that relaying on existing mechanism (the layer contexts) will be less error prone long term.

pbalcer (approved):

source/loader/layers/sanitizer/linux/symbolizer.cpp
yingcong-wu: Updated already, please review, thanks.

yingcong-wu:
Hi @pbalcer , we have another internal discussion and we still think having a symbolizer within the context is strange and we don't want that. Instead, we will construct a symbolizer each time we try to use it. Symbolization happened in the error reporting stage and we don't think performance is a problem at that stage. 

By doing so we won't have the symbolizer in the context and won't have the destruction order problem too.

pbalcer:
> Hi @pbalcer , we have another internal discussion and we still think having a symbolizer within the context is strange and we don't want that. Instead, we will construct a symbolizer each time we try to use it. Symbolization happened in the error reporting stage and we don't think performance is a problem at that stage.
> 
> By doing so we won't have the symbolizer in the context and won't have the destruction order problem too.

Sure, makes sense. Another option could be using the `AtomicSingleton` type - https://github.com/oneapi-src/unified-runtime/blob/main/source/common/ur_util.hpp#L427, but that might be an overkill.

AllanZyne (approved):

zhaomaosu (approved):

PR #2238, opened 2024-10-24
Revert "update to L0 loader version v1.18.3 and don't scan _deps with Trivy or CodeQL"

pbalcer:
Reverts oneapi-src/unified-runtime#2208


pbalcer:
I don't think we need to wait for CI. @kbenzie ?

PR #2239, opened 2024-10-24
Add guards to InvalidNullPointerPlatform

RossBrunton:
This test calls `urPlatformGetNativeHandle`, which may report that the
feature is unsupported. Skip this test if the feature is unavailable.

### DIFF
--- a/test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
+++ b/test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
@@ -83,7 +83,8 @@ TEST_F(urPlatformCreateWithNativeHandleTest, SuccessWithUnOwnedNativeHandle) {
 TEST_F(urPlatformCreateWithNativeHandleTest, InvalidNullPointerPlatform) {
     for (auto platform : platforms) {
         ur_native_handle_t native_handle = 0;
-        ASSERT_SUCCESS(urPlatformGetNativeHandle(platform, &native_handle));
+        UUR_ASSERT_SUCCESS_OR_UNSUPPORTED(
+            urPlatformGetNativeHandle(platform, &native_handle));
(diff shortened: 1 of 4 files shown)


test/conformance/platform/urPlatformCreateWithNativeHandle.cpp
```diff
@@ -83,7 +83,10 @@ TEST_F(urPlatformCreateWithNativeHandleTest, SuccessWithUnOwnedNativeHandle) {
 TEST_F(urPlatformCreateWithNativeHandleTest, InvalidNullPointerPlatform) {
     for (auto platform : platforms) {
         ur_native_handle_t native_handle = 0;
-        ASSERT_SUCCESS(urPlatformGetNativeHandle(platform, &native_handle));
+        {
+            UUR_ASSERT_SUCCESS_OR_UNSUPPORTED(
+                urPlatformGetNativeHandle(platform, &native_handle));
+        }
```
kbenzie, line 89: What's the point of the additional scope?

RossBrunton: Not entirely sure, but the rest of the tests have it.

I think it's because the macro is defined like this:
```c++
#define UUR_ASSERT_SUCCESS_OR_UNSUPPORTED(ret)                                 \
    auto status = ret;                                                         \
    if (status == UR_RESULT_ERROR_UNSUPPORTED_FEATURE) {                       \
        GTEST_SKIP();                                                          \
    } else {                                                                   \
        ASSERT_EQ(status, UR_RESULT_SUCCESS);                                  \
    }
```

So I it's to prevent `status` from leaking out of the macro?

... Although that's probably only an issue if there are multiple `UUR_ASSERT_SUCCESS_OR_UNSUPPORTED` calls in the function. I'll just de-scope it.

aarongreig (approved):

RossBrunton:
Merging without llvm commit as this only affects testing.

PR #2242, opened 2024-10-25
[L0] Enable Sysman Thru Env by default and have zesInit be optional

nrspruit:
- Enable ZES_ENABLE_SYSMAN=1 by default unless
  UR_L0_ENABLE_SYSMAN_ENV_DEFAULT==0.
- Allows for Sysman support thru zeInit() without user input.
- Added UR_L0_ENABLE_ZESINIT_DEFAULT to allow for a customer to enable
  zesInit over sysman env init if the env allows it.
- NOTE: UR_L0_ENABLE_ZESINIT_DEFAULT=1 requires
  UR_L0_ENABLE_SYSMAN_ENV_DEFAULT==0 due to the env overwriting the
zesInit support.

### DIFF
--- a/source/adapters/level_zero/adapter.cpp
+++ b/source/adapters/level_zero/adapter.cpp
@@ -144,6 +144,43 @@ ur_result_t adapterStateInit() {
   return UR_RESULT_SUCCESS;
 }
 
+/*
+This constructor initializes the `ur_adapter_handle_t_` object and
+sets up the environment for Level Zero (L0) initialization.
+The behavior of the initialization process is influenced by two
+environment variables:
+`UR_L0_ENABLE_SYSMAN_ENV_DEFAULT` and `UR_L0_ENABLE_ZESINIT_DEFAULT`.
+
+| Environment Variable           | Value | Behavior                   |
+|--------------------------------|-------|----------------------------|
+| UR_L0_ENABLE_SYSMAN_ENV_DEFAULT| 1     | Enables the default SysMan |
+|                                |       | environment initialization |
+|                                |       | by setting                 |
+|                                |       | `ZES_ENABLE_SYSMAN` to "1".|
+|                                | 0     | Disables the default SysMan|
+|                                |       | environment initialization.|
+|                                | unset | Defaults to 1, enabling the|
+|                                |       | SysMan environment         |
+|                                |       | initialization.            |
+| UR_L0_ENABLE_ZESINIT_DEFAULT   | 1     | Enables the default SysMan |
+|                                |       | initialization by loading  |
+|                                |       | SysMan-related functions   |
+|                                |       | and calling `zesInit`.     |
+|                                | 0     | Disables the default SysMan|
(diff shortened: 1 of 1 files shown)


source/adapters/level_zero/adapter.cpp
```diff
@@ -223,15 +236,29 @@ ur_adapter_handle_t_::ur_adapter_handle_t_()
...
-            processHandle, "zesDriverGetDeviceByUuidExp");
-    GlobalAdapter->getSysManDriversFunctionPtr =
-        (zes_pfnDriverGet_t)ur_loader::LibLoader::getFunctionPtr(
-            processHandle, "zesDriverGet");
-    GlobalAdapter->sysManInitFunctionPtr =
-        (zes_pfnInit_t)ur_loader::LibLoader::getFunctionPtr(processHandle,
-                                                            "zesInit");
+
+    // Check if the user has enabled the default L0 SysMan initialization.
+    const int UrSysManInitEnabled = [] {
```
rwmcguir, line 241: Simple question, is there a way to reduce code duplication here? This is the same code block as above

rwmcguir: Oh, I see the subtle difference... Maybe if there was a better name here, to make it more obvious.. 
UrSysManApiInitEnable?

nrspruit: sure, I can change that.

nrspruit: Updated in latest to have a clear distinction between the variables.

aravindksg (approved):

source/adapters/level_zero/adapter.cpp
```diff
@@ -144,6 +144,43 @@ ur_result_t adapterStateInit() {
...
+|                                |       | initialization.            |
+| UR_L0_ENABLE_ZESINIT_DEFAULT   | 1     | Enables the default SysMan |
+|                                |       | initialization by loading  |
+|                                |       | SysMan-related functions   |
+|                                |       | and calling `zesInit`.     |
+|                                | 0     | Disables the default SysMan|
+|                                |       | initialization with zesInit|
+|                                | unset | Defaults to 0, disabling   |
+|                                |       | the SysMan initialization  |
+|                                |       | thru zesInit.              |
```
kbenzie, line 173: I feel like these should be described in the [env var docs](https://oneapi-src.github.io/unified-runtime/core/INTRO.html#environment-variables) but I don't see any of the other L0 adapter env vars their either.

I think the best place for these would be an L0 reference doc, similar to the [CUDA ref doc](https://oneapi-src.github.io/unified-runtime/core/CUDA.html) and [HIP ref doc](https://oneapi-src.github.io/unified-runtime/core/HIP.html) describing all the L0 adapter specific env vars. This could be done in a separate PR, though.

pbalcer (approved):
I think I'd preferred a single env variable with three states, but this is also fine.

PR #2243, opened 2024-10-25
[DeviceAsan] Enlarge nullpointer redzone to better detect problem

yingcong-wu:
Also, print out the triggering address in the error report.

### DIFF
--- a/source/loader/layers/sanitizer/asan_shadow.hpp
+++ b/source/loader/layers/sanitizer/asan_shadow.hpp
@@ -16,6 +16,8 @@
 #include "common.hpp"
 #include <unordered_set>
 
+#define NULLPTR_REDZONE_SIZE (4096)
+
 namespace ur_sanitizer_layer {
 
 struct ShadowMemory {
--- a/source/loader/layers/sanitizer/asan_shadow.hpp
+++ b/source/loader/layers/sanitizer/asan_shadow.hpp
@@ -16,8 +16,6 @@
 #include "common.hpp"
 #include <unordered_set>
 
-#define NULLPTR_REDZONE_SIZE (4096)
-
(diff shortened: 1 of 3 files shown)


source/loader/layers/sanitizer/asan_shadow.hpp
```diff
@@ -16,6 +16,8 @@
 #include "common.hpp"
 #include <unordered_set>
 
+#define NULLPTR_REDZONE_SIZE (4096)
```
AllanZyne, line 19: add a comment about why choose this size

AllanZyne: this macro needn't be in header file

yingcong-wu: I choose this size becuase it is a common page value, and I don't have any other good reasons for it.

AllanZyne: You can query the size of page by GetVirtualMemGranularity.

yingcong-wu: I don't think that is necessary. Even when we have the page size, how many pages should we use? I think some simple value is good enough for now.

zhaomaosu (approved):
lgtm

source/loader/layers/sanitizer/asan_shadow.hpp
AllanZyne: > I don't think that is necessary. Even when we have the page size, how many pages should we use?
I think it's reasonable to use 1 page, so that we don't waste the mapped physical memory.

Please use constexpr.

AllanZyne (approved):

pbalcer (approved):

Issue #2079, opened 2024-09-11
Type mismatch in p2p peer info enum


pbalcer:
The p2p extension defines two query info enums as uint32_t values:
```
    - name: UR_PEER_ACCESS_SUPPORTED
      desc: "[uint32_t] 1 if P2P access is supported otherwise P2P access is not supported."
    - name: UR_PEER_ATOMICS_SUPPORTED
      desc: "[uint32_t] 1 if atomic operations are supported over the P2P link, otherwise such operations are not supported."
```
However, the CUDA implementation uses ints:
```
  int value;
  CUdevice_P2PAttribute cuAttr;
  ...
    switch (propName) {
    case UR_EXP_PEER_INFO_UR_PEER_ACCESS_SUPPORTED: {
      cuAttr = CU_DEVICE_P2P_ATTRIBUTE_ACCESS_SUPPORTED;
      break;
    }
   ...

    UR_CHECK_ERROR(cuDeviceGetP2PAttribute(&value, cuAttr, commandDevice->get(),
                                           peerDevice->get()));

  return ReturnValue(value);
```

And the L0 implementation uses a bool:
```
  bool propertyValue = false;
  switch (propName) {
  case UR_EXP_PEER_INFO_UR_PEER_ACCESS_SUPPORTED: {
    ...
    ZE2UR_CALL(
        zeDeviceCanAccessPeer,
        (commandDevice->ZeDevice, peerDevice->ZeDevice, &p2pDeviceSupported));
    propertyValue = p2pAccessSupported && p2pDeviceSupported;
    break;


  return ReturnValue(propertyValue);
```

This causes the scripts to generate incorrect code, especially for print API:
`
---> urUsmP2PPeerAccessGetInfoExp(.commandDevice = 0x6885b80, .peerDevice = 0x6886180, .propName = UR_EXP_PEER_INFO_UR_PEER_ACCESS_SUPPORTED, .propSize = 1, .pPropValue = invalid size (is: 1, expected: >=4), .pPropSizeRet = nullptr) -> UR_RESULT_SUCCESS;
`
Notice the pPropValue.

JackAKirk:
See https://github.com/intel/llvm/pull/15847#issuecomment-2437443907
I'll try to fix the implementation.

PR #2246, opened 2024-10-25
[spec][usm-p2p] p2p info query return int instead of uint32_t, fix l0 impl

JackAKirk:
The level zero p2p impl currently returns boolean values instead of an integer. This PR fixes that to match the spec requirement.
I've also changed the spec requirement slightly so that all `ur_exp_peer_info_t` types are `int` instead of  `uint32_t`

The switch to `int` practically makes no real difference right now but I think is an improvement.
I think that I made a mistake choosing `uint32_t` originally since in theory the cuda/hip backends could make use of negative values in the future since they use `int` type; such that by restricting to an unsigned type we are technically restricting these backends.
Also by not specifying a 32 bit type we allow for any unexpected issues where a runtime platform could theoretically use e.g. 16 bit integers as a query return value.

Fixes https://github.com/oneapi-src/unified-runtime/issues/2079


JackAKirk:
Corresponding DPC++ change here: https://github.com/intel/llvm/pull/15873

pbalcer (approved):

scripts/core/exp-usm-p2p.yml
```diff
@@ -24,9 +24,9 @@ name: $x_exp_peer_info_t
 typed_etors: True
 etors:
     - name: UR_PEER_ACCESS_SUPPORTED
-      desc: "[uint32_t] 1 if P2P access is supported otherwise P2P access is not supported."
+      desc: "[int] 1 if P2P access is supported otherwise P2P access is not supported."
```
pbalcer, line 27: nit: maybe this should be int32_t to be more precise?

JackAKirk: > nit: maybe this should be int32_t to be more precise?

I purposely chose `int`, because I don't think there is any reason that non-32 bit integers can't be used (if the platform uses e.g. 16 bits for `int`). This matches what cuda and hip do, see e.g. https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__PEER__ACCESS.html#group__CUDA__PEER__ACCESS_1g4c55c60508f8eba4546b51f2ee545393

The runtime correspondingly retrieves value of size(int): see https://github.com/intel/llvm/pull/15873/files#diff-e5a5ec910bab410f31617da9dfb56ee9e435769e7d6a830e85094c07b80c5454R250

JackAKirk:
L0 Failures are completely unrelated and also occur e.g. here: https://github.com/oneapi-src/unified-runtime/actions/runs/11527149243/job/32093062503?pr=2242

scripts/core/exp-usm-p2p.yml
pbalcer: Makes sense. Thanks.

PR #2247, opened 2024-10-25
Improve cts_exe.py formatting

RossBrunton:
It now shows the error code and a sampler of the output on CI.


martygrant (approved):

RossBrunton:
Merging without LLVM change as this only affects UR testing.

PR #2128, opened 2024-09-25
[DeviceSanitizer] Only try to get backtrace symbols when needed

zhaomaosu:

PR #2251, opened 2024-10-28
[DeviceASAN] Move validation layer before sanitizer layer

zhaomaosu:
Avoid internal UR API calls from sanitizer layer being intercepted by validation layer for better performance.
For miniqmc, the runtime can be reduced by -21.46% with this change.

### DIFF
--- a/source/loader/ur_lib.hpp
+++ b/source/loader/ur_lib.hpp
@@ -72,14 +72,16 @@ class __urdlllocal context_t : public AtomicSingleton<context_t> {
 
     using LayerData = std::pair<proxy_layer_context_t *, void (*)()>;
     const std::vector<LayerData> layers = {
-        {ur_validation_layer::getContext(),
-         ur_validation_layer::context_t::forceDelete},
-    // Initialize tracing layer after sanitizer layer to make sure tracing
-    // layer will properly print all API calls.
+    // First initialize sanitizer layer for two reasons:
+    //   (1) make sure tracing layer will properly print all API calls.
+    //   (2) avoid internal UR API calls from sanitizer layer being intercepted
+    //       by validation layer for better performance.
 #if UR_ENABLE_SANITIZER
         {ur_sanitizer_layer::getContext(),
          ur_sanitizer_layer::context_t::forceDelete},
 #endif
+        {ur_validation_layer::getContext(),
+         ur_validation_layer::context_t::forceDelete},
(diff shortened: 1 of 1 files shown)


yingcong-wu:
Do we have some performance data here?

AllanZyne (approved):

zhaomaosu:
> Do we have some performance data here?

Updated the git messages

yingcong-wu:
> > Do we have some performance data here?
> 
> Updated the git messages

Great, a good improvement.

yingcong-wu (approved):

pbalcer (approved):

pbalcer (commented):
I'm not opposed, but I'm not sure I understand the concern about performance. The validation layer is disabled by default. Users can just chose to not enable it if they observe reduced performance.

zhaomaosu:
> I'm not opposed, but I'm not sure I understand the concern about performance. The validation layer is disabled by default. Users can just chose to not enable it if they observe reduced performance.

Eh....Actually, it's because intel openmp offload runtime will enable validation layer by default in their ur plugin. BTW, the overhead  comes mainly from 'backtrace_symbols'. Maybe we can also do same changes in validation layer as PR #2128

pbalcer:
> BTW, the overhead comes mainly from 'backtrace_symbols'. Maybe we can also do same changes in validation layer as PR #2128

At some point we should move the backtrace implementation into common and have only one implementation.

pbalcer:
> Eh....Actually, it's because intel openmp offload runtime will enable validation layer by default in their ur plugin

They should stop doing that. It hurts performance overall, not just when sanitizer is enabled. If they want to retain some lightweight validation, they should enable only parameter validation (which is reasonable to leave always enabled).

zhaomaosu:
> They should stop doing that. It hurts performance overall, not just when sanitizer is enabled. If they want to retain some lightweight validation, they should enable only parameter validation (which is reasonable to leave always enabled).

Thanks. I'll suggest them not to enable full validation layer by default.

zhaomaosu:
> This is regressing functionality (validation layer won't catch issues with API use in other layers) to workaround a performance problem caused by incorrect usage of the validation layer.
> 
> I'm currently leaning towards rejecting this patch, unless this cannot be solved properly in openmp.

Ok, I'm asking openmp team if they can make the changes. I'll convert this PR to draft firstly.

PR #2258, opened 2024-10-28
Use extension version of clGetKernelSubGroupInfo when necessary.

aarongreig:
We do support at least one device (fpga emulator) that reports a cl version of < 2.1 and supports subgroups, in that scenario we need query out and use the KHR version of the entry point.


aarongreig:
LLVM PR: https://github.com/intel/llvm/pull/15896

source/adapters/opencl/kernel.cpp
```diff
@@ -189,11 +190,45 @@ urKernelGetSubGroupInfo(ur_kernel_handle_t hKernel, ur_device_handle_t hDevice,
...
+        SubgroupExtSupported));
+    if (!SubgroupExtSupported) {
+      return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+    }
+    cl_platform_id Platform;
+    CL_RETURN_ON_FAILURE(clGetDeviceInfo(
+        cl_adapter::cast<cl_device_id>(hDevice), CL_DEVICE_PLATFORM,
+        sizeof(Platform), &Platform, nullptr));
+    GetKernelSubGroupInfo =
+        reinterpret_cast<ApiFuncT>(clGetExtensionFunctionAddressForPlatform(
```
kbenzie, line 218: We should use the extension function pointer cache for this.

aarongreig: Huh yeah forgot we had those, I cribbed from `ProgramCreateWithIL` when I wrote this since it's a similar cl version situation there, updated that as well

callumfare (approved):

martygrant (approved):

PR #2268, opened 2024-10-30
[HIP] Use copy direction workaround in urEnqueueUSMMemcpy2D up to HIP v6.1.0

GeorgeWeb:
The fix for the original HIP/ROCm issue reported in https://github.com/ROCm/clr/issues/40 has been fixed by commit https://github.com/ROCm/clr/commit/d3bfb55d7a934355257a72fab538a0a634b43cad which is included in HIP/ROCm releases starting from version 6.1.0 onwards. Hence, we add the maximum version of HIP up to which we will use our workaround.


hdelan (approved):

npmiller:
Test failure is unrelated, L0 failure when this only touches HIP, and that L0 failure also fails on other PRs, marking this ready to merge

Issue #1374, opened 2024-02-22
Tighten spec for extension queries


aarongreig:
`UR_DEVICE_INFO_EXTENSIONS` and `UR_PLATFORM_INFO_EXTENSIONS` both return strings denoting "extensions supported" by the underlying adapter. The current implementations can return some pretty weird stuff, like cuda reporting support for a couple of [OpenCL extensions](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/cuda/device.cpp#L611), and a PI extension the line below that. In fact, OpenCL extension strings are reported by all the adapters. I suspect this behaviour is relied on by some element of the sycl RT but it isn't a good mechanism for communicating what it's trying to communicate, and we should definitely aim to get rid of it.

Beyond that a conversation needs to be had about what these queries *are* for. We deliberately avoided calling the various experimental features extensions, although we are now seeing some of them appear in the output of these queries. If we want them to be a mechanism for reporting support for UR experimental features then the spec should reflect that. If we want them to return extensions supported by the underlying API (i.e. have the CL adapter continue reporting CL extension strings) then that should explicitly be in the spec as well. I question the usefulness of both of these. Unless there's a compelling use case for returning native extension strings (which you could still do if you really wanted, by getting the native handle out), or reporting experimental feature support via a string rather than a `ur_device_info_t` enum like we mostly [already do](https://oneapi-src.github.io/unified-runtime/core/api.html#ur-device-info-t), we should consider these queries vestigial and remove them.

JackAKirk:
I used the extension string for extension testing: https://github.com/oneapi-src/unified-runtime/pull/1369
Following command_buffer extension precedent.

aarongreig:
yeah that's partially what prompted this issue, I'm not going to argue against its use there since there's an established precedent

alycm:
Discussed 2024-02-28, removing OpenCL extension string here makes sense, but should check that this won't affect current SYCL-RT code.

PR #2269, opened 2024-10-30
Add device info query to report support for native asserts.

aarongreig:
This allows cuda and hip to stop reporting the opencl extension string, see issue #1374

### DIFF
--- a/source/adapters/cuda/device.cpp
+++ b/source/adapters/cuda/device.cpp
@@ -617,7 +617,6 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   case UR_DEVICE_INFO_EXTENSIONS: {
 
     std::string SupportedExtensions = "cl_khr_fp64 cl_khr_subgroups ";
-    SupportedExtensions += "cl_intel_devicelib_assert ";
     // Return supported for the UR command-buffer experimental feature
     SupportedExtensions += "ur_exp_command_buffer ";
     SupportedExtensions += "ur_exp_usm_p2p ";
@@ -1114,6 +1113,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   }
   case UR_DEVICE_INFO_LOW_POWER_EVENTS_EXP:
     return ReturnValue(false);
+  case UR_DEVICE_INFO_USE_NATIVE_ASSERT:
+    return ReturnValue(true);
+
   default:
     break;
   }
--- a/source/adapters/hip/device.cpp
+++ b/source/adapters/hip/device.cpp
@@ -558,12 +558,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     return ReturnValue("");
   }
   case UR_DEVICE_INFO_EXTENSIONS: {
-    // TODO: Remove comment when HIP support native asserts.
-    // DEVICELIB_ASSERT extension is set so fallback assert
-    // postprocessing is NOP. HIP 4.3 docs indicate support for
-    // native asserts are in progress
     std::string SupportedExtensions = "";
-    SupportedExtensions += "cl_intel_devicelib_assert ";
     SupportedExtensions += "ur_exp_usm_p2p ";
 
     int RuntimeVersion = 0;
@@ -1107,6 +1102,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   case UR_DEVICE_INFO_LOW_POWER_EVENTS_EXP: {
     return ReturnValue(false);
   }
+  case UR_DEVICE_INFO_USE_NATIVE_ASSERT:
+    return ReturnValue(true);
   default:
     break;
   }
--- a/source/adapters/native_cpu/device.cpp
+++ b/source/adapters/native_cpu/device.cpp
@@ -426,6 +426,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
 
   case UR_DEVICE_INFO_USM_POOL_SUPPORT:
     return ReturnValue(false);
+  case UR_DEVICE_INFO_USE_NATIVE_ASSERT:
+    return ReturnValue(false);
 
   case UR_DEVICE_INFO_LOW_POWER_EVENTS_EXP:
     return ReturnValue(false);
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -2194,6 +2194,9 @@ typedef enum ur_device_info_t {
   /// [::ur_bool_t] support the ::urProgramSetSpecializationConstants entry
   /// point
   UR_DEVICE_INFO_PROGRAM_SET_SPECIALIZATION_CONSTANTS = 121,
+  /// [::ur_bool_t] return true if the device has a native assert
+  /// implementation.
+  UR_DEVICE_INFO_USE_NATIVE_ASSERT = 122,
   /// [::ur_bool_t] Returns true if the device supports the use of
   /// command-buffers.
   UR_DEVICE_INFO_COMMAND_BUFFER_SUPPORT_EXP = 0x1000,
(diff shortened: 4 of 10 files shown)


aarongreig:
LLVM PR https://github.com/intel/llvm/pull/15929

source/adapters/cuda/device.cpp
```diff
@@ -1105,6 +1104,13 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
                              CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR) >= 9;
     return ReturnValue(static_cast<bool>(Value));
   }
+  case UR_DEVICE_INFO_USE_NATIVE_ASSERT:
+    // CUDA doesn't actually support asserts natively so this results in
```
JackAKirk, line 1108: What does native assert mean if cuda doesn't support it? cuda has an implementation for `std::assert` etc. It has a sycl test here https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/Assert/assert_in_kernels.hpp
Hip also supports it in theory, but has some issues on some old (but officially supported) devices https://github.com/ROCm/HIP/issues/3368

source/adapters/hip/device.cpp
```diff
@@ -933,6 +928,14 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
   }
   case UR_DEVICE_INFO_COMMAND_BUFFER_EVENT_SUPPORT_EXP:
     return ReturnValue(false);
+  case UR_DEVICE_INFO_USE_NATIVE_ASSERT:
+    // TODO: Remove comment when HIP supports native asserts.
```
JackAKirk, line 932: HIP supports native asserts of devices that are most relevant for the users of dpc++ AFAIK: the assert tests definitely pass on CDNA2 cards, and AMD devs implied they should be supported on all supported devices. There was a failure on a RDNA2 card but that has little relevance to dpc++ users.

JackAKirk (changes requested):

source/adapters/hip/device.cpp
aarongreig: they still have `UNSUPPORTED: hip` in the e2e test along with a tracker issue, I could change this to a note about it not necessarily being 100% supported and link the issue, or I'm happy to just delete the comment if you don't think it's useful to track that here

source/adapters/cuda/device.cpp
aarongreig: Nice, will delete. I was basing my assumptions on the discussion I found on https://github.com/intel/llvm/pull/4604 when I was trying to figure out why we were reporting support for that cl extension

source/adapters/hip/device.cpp
JackAKirk: > they still have `UNSUPPORTED: hip` in the e2e test along with a tracker issue, I could change this to a note about it not necessarily being 100% supported and link the issue, or I'm happy to just delete the comment if you don't think it's useful to track that here

Sure, whatever you think is best. The only issue with the current comment is that it is all out of date/ factually incorrect.

As far as hip tests marked unsupported in intel/llvm, for a lot of cases this is because the intel/llvm CI does not use an officially supported rocm card. This case is a bit different because the failure still does occur on some officially supported cards in addition to the unsupported ones but I think your wish to add a comment to state something along the lines of "this is mostly supported" is fine.

It is best to try not to take too much account of such failures, unfortunately we have to rely on other testing.

GeorgeWeb:
@aarongreig According to my latest testing CDNA2 AMD GPUs had reliable native asserts support with latest versions of ROCm. I am not sure how well things are with consumer grade but I suspect all supported cards should in theory have support these days.

Just a question/thought, not a requirement:
I am wondering whether we need an intel/llvm PR for this too to update the device config file here https://github.com/intel/llvm/blob/d13019638e4ed3078a52cf06a91080875d2e0d0e/llvm/include/llvm/SYCLLowerIR/DeviceConfigFile.td#L293 wrt native_assert aspect support for HIP. This Device Configuration File is used for SYCL/DPC++ `all_devices_have<sycl::aspect>` and `any_device_has<sycl::aspect>` compile-time aspect traits implementation (see https://intel.github.io/llvm/design/DeviceAspectTraitDesign.html). Maybe we can selectively add `AspectExt_oneapi_native_assert` to all of the `Officially supported:` HIP-AMD targets regardless of whether they are CDNA arches or whether we tested them all (because it's impossible for us to fully test that on all gpus so we ought to rely on ROCm actually supporting their supported hardware). Until now I've only added it to `gfx90a` because that's the only one I knew of passing the test reliably 🤷‍♂️.

source/adapters/hip/device.cpp
aarongreig: I decided to omit the comment, looking at https://github.com/intel/llvm/issues/7634 it does seem like the remaining fail there is a weird issue with how the hardware is set up, and based on Georgi's comment it sounds like confidence is pretty high that this should be upgraded to an unqualified "supported"

JackAKirk (approved):
LGTM for cuda/hip

aarongreig:
> @aarongreig According to my latest testing a CDNA2 AMD GPU had reliable native asserts support with latest versions of ROCm. However, I am not sure how well things are with consumer grade ones, but I suspect all supported cards should _in theory_ have support for the feature these days.
> 
> Just a question/thought for this ties to DPC++, but not a requirement: I am wondering whether we need an intel/llvm PR for this too to update the device config file here https://github.com/intel/llvm/blob/d13019638e4ed3078a52cf06a91080875d2e0d0e/llvm/include/llvm/SYCLLowerIR/DeviceConfigFile.td#L293 wrt native_assert aspect support for HIP. This Device Configuration File is used for SYCL/DPC++ `all_devices_have<sycl::aspect>` and `any_device_has<sycl::aspect>` compile-time aspect traits implementation (see https://intel.github.io/llvm/design/DeviceAspectTraitDesign.html). Maybe we can selectively add `AspectExt_oneapi_native_assert` to all of the `Officially supported:` HIP-AMD targets regardless of whether they are CDNA arches or whether we tested them all (because it's impossible for us to fully test that on all gpus so we ought to rely on ROCm actually supporting their supported hardware). Until now I've only added it to `gfx90a` because that's the only one I knew of passing the test reliably 🤷‍♂️.

That probably does need to be updated, although I don't really want to include that in https://github.com/intel/llvm/pull/15929 since it's supposed to just be updating how the internal reporting mechanism works

GeorgeWeb:
>
> That probably does need to be updated, although I don't really want to include that in [intel/llvm#15929](https://github.com/intel/llvm/pull/15929) since it's supposed to just be updating how the internal reporting mechanism works

Yeah, I agree. Guess the comment was more of a note to refer to in a future follow-up to that specific part of intel/llvm.

Regardless, all cuda/hip related changes look good! Cheers.

nrspruit (approved):
LGTM for L0

aarongreig:
ping @oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-opencl-write

kbenzie (approved):

source/adapters/native_cpu/device.cpp
```diff
@@ -426,6 +426,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
 
   case UR_DEVICE_INFO_USM_POOL_SUPPORT:
     return ReturnValue(false);
+  case UR_DEVICE_INFO_USE_NATIVE_ASSERT:
+    return ReturnValue(false);
```
coldav, line 430: Feels like we should support something like this? If there a specification for this?

aarongreig: the adapter never reported support for the cl_intel_devicelib_assert extension so this isn't a change in behaviour, you can see on the sycl side it's a straight change from checking for the extension string to checking for this query https://github.com/intel/llvm/pull/15929/files#diff-d0057c737493e4ed04a7761d08fbd154866eb67f5763bfbc032e5ff550214066L76

coldav (approved):

martygrant:
## Information

## PR Migration

---

PR #2274, opened 2024-11-01
[DeviceASAN] Simplify shadow mapping logic for GPU device

zhaomaosu:
intel/llvm part: https://github.com/intel/llvm/pull/15954


source/loader/layers/sanitizer/asan_shadow.cpp
```diff
@@ -256,12 +256,9 @@ ur_result_t ShadowMemoryGPU::ReleaseShadow(std::shared_ptr<AllocInfo> AI) {
...
 uptr ShadowMemoryPVC::MemToShadow(uptr Ptr) {
-    if (Ptr & 0xFF00000000000000ULL) { // Device USM
-        return ShadowBegin + 0x80000000000ULL +
-               ((Ptr & 0xFFFFFFFFFFFFULL) >> ASAN_SHADOW_SCALE);
-    } else { // Only consider 47bit VA
-        return ShadowBegin + ((Ptr & 0x7FFFFFFFFFFFULL) >> ASAN_SHADOW_SCALE);
-    }
+    return ShadowBegin + (((((Ptr & 0x8000'0000'0000'0000) >> 16) + Ptr) &
+                           0xf'ffff'ffff'ffff) >>
+                          ASAN_SHADOW_SCALE);
```
yingcong-wu, line 261: Could we add some comment here to explain the mapping logic?

zhaomaosu: Actually, we already have shadow mapping logic here https://github.com/zhaomaosu/unified-runtime/blob/simplify-shadow-mapping/source/loader/layers/sanitizer/asan_shadow.hpp#L99

yingcong-wu: okay

AllanZyne (approved):

AllanZyne:
Change the mapping logic of DG2 as well?

zhaomaosu:
> Let's change the mapping logic of DG2 as well?

Done

source/loader/layers/sanitizer/ur_sanddi.cpp
```diff
@@ -73,6 +73,12 @@ __urdlllocal ur_result_t UR_APICALL urAdapterGet(
         return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
     }
 
+    // Set env AllocateHostAllocationsInHeapExtendedHost to 0, then host and
+    // shared usm will have same allocation range on GPU device. This will
+    // simplify shadow memory mapping logic on GPU device.
+    setenv("NEOReadDebugKeys", "1", 1);
```
pbalcer, line 79: @nrspruit is it a good idea to set these env variables in the layer? If this is really required, wouldn't it be a better idea to make this into an L0 API?
Using a debug key for functionality doesn't seem like a good idea.

pbalcer: I don't feel comfortable merging this patch. We need confirmation from the NEO folks that this makes sense.
@MichalMrozek can you help?

MichalMrozek: No, this is not correct to set this.
Please do not merge it.
If you want to have specific release keys avaialable this needs discussion with Piotr Rozenfeld.

MichalMrozek: AllocateHostAllocationsInHeapExtendedHost is a debug key, debug keys cannot be used in production software. They are meant only for experiments. Driver team may remove or change this debug variable at any point of time without any notifications.

AllanZyne: Ok, I will send this request to driver team. Thanks.

zhaomaosu:
I'll try to find other optimization opportunity. I'm closing this PR.

PR #2278, opened 2024-11-01
[L0] Add Level Zero Reference Document

nrspruit:
- Add inital Level Zero Reference document with environment variable definitions.


igchor:
Shouldn't we also have default values listed here (at least for some variables)?

nrspruit:
> Shouldn't we also have default values listed here (at least for some variables)?

Some of them I had that set as 1 or unset, but I could add to some the defaults.

kbenzie (approved):

PR #2279, opened 2024-11-01
[L0] Update Level Zero Loader version to v1.18.5

nrspruit:


igchor (approved):

EwanC:
I tested this out locally, and still see the `SYCL Exception encountered: Native API failed. Native API returns: 2147483646 (UR_RESULT_ERROR_UNKNOWN)` error running sycl-ls

nrspruit:
> I tested this out locally, and still see the `SYCL Exception encountered: Native API failed. Native API returns: 2147483646 (UR_RESULT_ERROR_UNKNOWN)` error running sycl-ls

Let me check, I wonder if we are having a conflict since MCL symbols are now included.

nrspruit:
> > I tested this out locally, and still see the `SYCL Exception encountered: Native API failed. Native API returns: 2147483646 (UR_RESULT_ERROR_UNKNOWN)` error running sycl-ls
> 
> Let me check, I wonder if we are having a conflict since MCL symbols are now included.

We know this is working correctly with older drivers and libraries, so this must be specific to this configuration.

EwanC (approved):
Discovered that the error I'm seeing is to do with a DPC++ build issue I've documented in https://github.com/intel/llvm/issues/15986

PR #2285, opened 2024-11-05
[NFC] Clean up match files for conf. testing

RossBrunton:
Clean up the match files to hopefully make them more easily readable. As
we now use cts_exe.py, we don't need as many workarounds.

This changes the following:
* Removes the "{{NONDETERMINISTIC}}" tag, which isn't needed any more.
* Removes the notice about only being supported in cts_exe.py - all
  suites now require cts_exe.py.
* Device names have been simplified down to a single wildcard -
  should make it easier to read them and copy between files if needed.
* If all tests for a specific enum fail, it has been folded into a
  single wildcard.


martygrant:
Last week we discussed our original goal of removing use of the match files, we plan to have a meeting soon to put a plan together. I think in the meantime it might be a good idea to have some documentation on how they work (not sure if there is any) given there have been some recent changes and with more proposed in this PR

RossBrunton:
> Last week we discussed our original goal of removing use of the match files, we plan to have a meeting soon to put a plan together. I think in the meantime it might be a good idea to have some documentation on how they work (not sure if there is any) given there have been some recent changes and with more proposed in this PR

There's been a few updates to the documentation recently, is there anything you think is missing?

isaacault (approved):

EwanC (approved):

RossBrunton:
Doesn't affect LLVM, merging without associated commit.

PR #2287, opened 2024-11-06
[Command-Buffer] Fix update validation for L0 and OpenCL

fabiomestre:
Fixes bug where passing nullptr to hNewKernel results on UR_RESULT_ERROR_UNSUPPORTED_FEATURE being returned.

intel/llvm: https://github.com/intel/llvm/pull/15998/files


EwanC (commented):
Probably a bit late, but can we write a UR CTS test that would've caught this

fabiomestre:
> Probably a bit late, but can we write a UR CTS test that would've caught this

I have now added a test for it :+1: 

Has PVC been added to CI? If not, even this new test wouldn't have caught the bug unless it was run locally.

PR #2298, opened 2024-11-08
WIP: Cuda Fix for command-buffer local argument upate

EwanC:
After setting kernel arguments during update, we need to reset the amount of local memory used.

WIP as if we only set some local arguments in update, but not all, I'm not sure if this solution is correct. So am refining further.


source/adapters/cuda/kernel.hpp
```diff
@@ -61,10 +61,22 @@ struct ur_kernel_handle_t_ {
...
     using args_index_t = std::vector<void *>;
+    /// Storage shared by all args which is mem copied into when adding a new
+    /// argument.
     args_t Storage;
+    /// Aligned size of each parameter, including padding.
     args_size_t ParamSizes;
+    /// Byte offset into /p Storage allocation for each parameter.
     args_index_t Indices;
-    args_size_t OffsetPerIndex;
+    /// Aligned size in bytes for each local memory parameter after padding has
```
hdelan, line 71: Could you maybe add a brief comment explaining how local mem is instrumented for DPC++? ie there's a single local mem buffer, and we pass the offsets into this as kernel args.

test/conformance/exp_command_buffer/update/local_memory_update.cpp
```diff
@@ -8,15 +8,19 @@
...
         program_name = "saxpy_usm_local_mem";
         UUR_RETURN_ON_FATAL_FAILURE(
             urUpdatableCommandBufferExpExecutionTest::SetUp());
 
+        if (backend == UR_PLATFORM_BACKEND_LEVEL_ZERO) {
+            GTEST_SKIP()
+                << "Local memory argument update not supported on Level Zero.";
+        }
+
         // HIP has extra args for local memory so we define an offset for arg indices here for updating
```
aarongreig, line 24: apologies but as long as we're in here might as well fix the formatting
```suggestion
        // HIP has extra args for local memory so we define an offset for arg
        // indices here for updating
```

test/conformance/exp_command_buffer/update/local_memory_update.cpp
```diff
@@ -327,16 +544,345 @@ TEST_P(LocalMemoryUpdateTest, UpdateParametersAndLocalSize) {
...
+
+    // Verify that update occurred correctly
+    uint32_t *new_output = (uint32_t *)shared_ptrs[0];
+    uint32_t *new_X = (uint32_t *)shared_ptrs[3];
+    uint32_t *new_Y = (uint32_t *)shared_ptrs[4];
+    Validate(new_output, new_X, new_Y, new_A, global_size, new_local_size);
+}
+
+// Test updating A,X,Y parameters to new values and only one of the local memory
+// parameters, which is set to a new values. Then a separate update call for
```
aarongreig, line 703:
```suggestion
// parameters, which is set to a new value. Then a separate update call for
```

aarongreig (approved):
CTS LGTM

source/adapters/cuda/kernel.hpp
hdelan: Probably the main thing that I think isn't clear is that the CUDA/HIP `urKernelSetArgLocal` impls will end up passing offsets as kernel args to the device kernel. This means that kernels need to be written in such a way that the args contain offsets into the smem allocation, and that within device code, the user will combine these offsets with the extern ptr that represents the entire smem allocation. This is even true for the base offset into an smem allocation, which is always zero.

If local memory is to be used in the CUDA/HIP UR adapters, the compiler must generate a kernel like this that uses an offset (if `urKernelSetArgLocal` is to be used).

```c++
int main() {                                                                                                                       
  T *ptr;                                                                                                                          
  sycl::queue{}.submit([&](sycl::handler &cgh) {                                                                                   
    sycl::local_accessor<T> local_acc(1, cgh);                                                                                     
    cgh.single_task<foo>([=] { local_acc[0] += local_acc[0]; });                                                                   
  });                                                                                                                              
}    
```

```asm
//                                                                                                                                 
// Generated by LLVM NVPTX Back-End                                                                                                
//                                                                                                                                 
                                                                                                                                   
.version 8.5                                                                                                                       
.target sm_60                                                                                                                      
.address_size 64                                                                                                                   
                                                                                                                                   
        // .weak        _ZTS3foo                        // -- Begin function _ZTS3foo                                              
.extern .shared .align 4 .b8 _ZTS3foo_shared_mem[];                                                                                
                                        // @_ZTS3foo                                                                               
.weak .entry _ZTS3foo(                                                                                                             
        .param .u32 _ZTS3foo_param_0                                                                                               
)                                                                                                                                  
{                                                                                                                                  
        .reg .b32       %r<3>;                                                                                                     
        .reg .b64       %rd<4>;                                                                                                    
                                                                                                                                   
// %bb.0:                               // %entry                                                                                  
        ld.param.s32    %rd1, [_ZTS3foo_param_0];                                                                                  
        mov.u64         %rd2, _ZTS3foo_shared_mem;                                                                                 
        add.s64         %rd3, %rd2, %rd1;                                                                                          
        ld.shared.u32   %r1, [%rd3];                                                                                               
        shl.b32         %r2, %r1, 1;                                                                                               
        st.shared.u32   [%rd3], %r2;                                                                                               
        ret;                                                                                                                       
                                        // -- End function                                                                         
}

Comment (hdelan) [reply source/adapters/cuda/kernel.hpp:71]: I think in general we don't have good enough testing for this sort of behaviour in DPC++. From a quick grep I can't find any tests where you have multiple local accessors in the same command group that have different alignments.

File: source/adapters/cuda/kernel.hpp [hunk 4]
```diff
@@ -102,28 +115,83 @@ struct ur_kernel_handle_t_ {
...
+      std::tie(AlignedLocalSize, AlignedLocalOffset) =
+          calcAlignedLocalArgument(Index, Size);
+
+      // Store argument details
       addArg(Index, sizeof(size_t), (const void *)&(AlignedLocalOffset),
-             Size + (AlignedLocalOffset - LocalOffset));
+             AlignedLocalSize);
+
+      // For every existing local argument which follows at later argument
+      // indices, updated the offset and pointer into the kernel local memory.
```
Bensuo, line 170:
```suggestion
      // indices, update the offset and pointer into the kernel local memory.
```

source/adapters/hip/kernel.hpp
```diff
@@ -97,28 +110,83 @@ struct ur_kernel_handle_t_ {
...
+      size_t AlignedLocalSize, AlignedLocalOffset;
+      std::tie(AlignedLocalSize, AlignedLocalOffset) =
+          calcAlignedLocalArgument(Index, Size);
+
+      // Store argument details
+      addArg(Index, sizeof(size_t), (const void *)&(AlignedLocalOffset),
+             AlignedLocalSize);
+
+      // For every existing local argument which follows at later argument
+      // indices, updated the offset and pointer into the kernel local memory.
```
Bensuo, line 165:
```suggestion
      // indices, update the offset and pointer into the kernel local memory.
```

source/adapters/cuda/kernel.hpp
EwanC: Do you think you could write a snippet of a SYCL/OpenCL-C/SPIRV/PTX kernel that you think wouldn't work currently? I'm not understanding enough to come up with a concrete example, and would help specify what the restrictions are.

My current thinking on how to progress this is:
* add E2E DPC++ tests
  * multiple local accessors in the same command group that have different alignments.
  * Kernel written in OpenCL-C using multiple local arguments (core SYCL test without graphs and test with graphs)

* add UR CTS tests outwith command-buffers for local memory args
  * `urKernelSetArgLocal`  doesn't need to be called between multiple enqueues of a kernel.
  * Multiple `urKernelSetArgLocal`  args with different alignments
 

* Document in local mem arg impl and/or restrictions in CUDA/HIP reference doc

hdelan: CUDA kernel that won't work:

```c++

__global__ void kernel(int *a) {
  extern __shared__ int blah1[];
  a[0] = blah1[0];
}
```

As this produces the PTX:

```
//
// Generated by NVIDIA NVVM Compiler
//
// Compiler Build ID: CL-34841621
// Cuda compilation tools, release 12.6, V12.6.77
// Based on NVVM 7.0.1
//

.version 8.5
.target sm_52
.address_size 64

	// .globl	_Z6kernelPi
.extern .shared .align 16 .b8 blah1[];

.visible .entry _Z6kernelPi(
	.param .u64 _Z6kernelPi_param_0
)
{
	.reg .b32 	%r<2>;
	.reg .b64 	%rd<3>;


	ld.param.u64 	%rd1, [_Z6kernelPi_param_0];
	cvta.to.global.u64 	%rd2, %rd1;
	ld.shared.u32 	%r1, [blah1];
	st.global.u32 	[%rd2], %r1;
	ret;

}
```

This doesn't work because the UR adapters assume that every use of `urKernelSetArgLocal` corresponds to an `offset` param in the kernel. This kernel doesn't use an offset into the shared local memory since it is just using the unoffset base address of the local mem allocation.

There is no way to say "This kernel uses x bytes in dynamic local mem, but my kernel signature doesn't include an offset param with which I want to index into the local mem". By contrast this would be fine:

```c++
__global__ void kernel(int *a, unsigned offset) {
  extern __shared__ int
      blah1[]; // This should never be used unless offset with a kernel arg
  int *blah_ptr = &blah1[offset];
  a[0] = blah_ptr[0];
}
```

Once you have a param you can call `urKernelSetArgLocal` with the index of the offset param (in this case 1).

hdelan: openCL kernel that won't work (at least when run by the CUDA adapter):

```c++
__kernel void blah(
    __local float* local_mem)      // Output array in global memory
```

As the CUDA adapter assumes that the `Index` arg given to `urKernelSetArgLocal` will map to an `offset` param in the kernel signature. In this case the dynamic local mem doesn't have an `offset` param, but instead will have a ptr param which corresponds to the local ptr. Loading it at runtime with an offset param will fail.

hdelan: Also good find with the local arg alignment tests. Interesting that it's `UNSUPPORTED` for all targets. We should maybe pass this task to the OAC team?

EwanC: Ahh, I get you know - I read "instrumented" as "implemented" in your original comment.  Could you check if what i've put in `scripts/core/CUDA.rst` is the kinda thing you had in mind? 

> Also good find with the local arg alignment tests. Interesting that it's UNSUPPORTED for all targets. We should maybe pass this task to the OAC team?

Good segway to ping @oneapi-src/unified-runtime-cuda-write for a review on this PR

scripts/core/CUDA.rst
```diff
@@ -148,6 +148,36 @@ take the extra global offset argument. Use of the global offset is not
...
+----------------------
+
+In UR local memory is a region of memory shared by all the work-items in
+a work-group. A kernel function signature can include local memory address
+space pointer arguments, which are set by the user with
+``urKernelSetArgLocal`` with the number of bytes of local memory to allocate
+and make available from the pointer argument.
+
+The CUDA adapter implements local memory arguments to a kernel as a single
+``__shared__`` memory allocation, with each local address space pointer argument
```
hdelan, line 161: It's not actually a pointer arg, it ends up as a `u32 offset` arg, which is combined with the ptr within the kernel

hdelan: I mean you actually mention this elsewhere. So maybe this is OK

hdelan: Do you think we could add an overload for `urKernelSetArgLocalExt` that only takes a `Size`, not a `Size, Index`? Then the user could ask for dynamic shared mem without having to have extra offsets in the kernel args.

frasercrmck: I believe we've an internal ticket about changing this, and combining the ptr+offset on the host side? Is that right, @hdelan?

source/adapters/cuda/kernel.hpp
```diff
@@ -102,28 +115,83 @@ struct ur_kernel_handle_t_ {
...
+      const size_t NumArgs = Indices.size() - 1; // Accounts for implicit arg
+      for (auto SuccIndex = Index + 1; SuccIndex < NumArgs; SuccIndex++) {
+        const size_t OriginalLocalSize = OriginalLocalMemSize[SuccIndex];
+        if (OriginalLocalSize == 0) {
+          // Skip if successor argument isn't a local memory arg
+          continue;
+        }
+
+        // Recalculate alignment
+        size_t SuccAlignedLocalSize, SuccAlignedLocalOffset;
```
frasercrmck, line 181: I don't suppose `auto [SuccAlignedLocalSize, SuccAlignedLocalOffset] = calcAlignedLocalArgument(...)` is preferable? Matter of style I suppose.

EwanC: Yeah it definitely is, updated

scripts/core/CUDA.rst
EwanC: I've rephrased this description to say a `u32` type parameter early in the text.

> Do you think we could add an overload for urKernelSetArgLocalExt that only takes a Size, not a (Size, Index)? Then the user could ask for dynamic shared mem without having to have extra offsets in the kernel args.

I'd probably go for adding an `urKernelSetExecInfo` property for this because at that point, we're not talking about kernel arguments. That might be a viable alternative usage, but I'm not sure how much value it really buys. Defer to @oneapi-src/unified-runtime-cuda-write on if you want us to create an issue somewhere on this topic.

hdelan: Sounds good to me. Nice one

frasercrmck (approved):

PR #2299, opened 2024-11-08
[CUDA][Bindless] Fix memory leak in interop mapping

cppchedy:
* added a map to ur_device_handle_t_.
* Capture the leaking Cumipmappedarray in urBindlessImagesMapExternalArrayExp into map.
* Update urBindlessImagesImageFreeExp to check if the Cuarray is derived, then destroy the corresponding Cumipmappedarray.

### DIFF
--- a/source/adapters/cuda/image.cpp
+++ b/source/adapters/cuda/image.cpp
@@ -430,6 +430,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageFreeExp(
   ScopedContext Active(hDevice);
   try {
     UR_CHECK_ERROR(cuArrayDestroy((CUarray)hImageMem));
+    if (auto it = hDevice->ChildCuarrayFromMipmapMap.find((CUarray)hImageMem);
+        it != hDevice->ChildCuarrayFromMipmapMap.end()) {
+      UR_CHECK_ERROR(cuMipmappedArrayDestroy((CUmipmappedArray)it->second));
+      hDevice->ChildCuarrayFromMipmapMap.erase(it);
+    }
   } catch (ur_result_t Err) {
     return Err;
   } catch (...) {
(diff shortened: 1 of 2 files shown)


przemektmalon (approved):

DBDuncan (approved):

ProGTX:
Friendly ping @oneapi-src/unified-runtime-cuda-write

source/adapters/cuda/image.cpp
```diff
@@ -430,6 +430,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageFreeExp(
   ScopedContext Active(hDevice);
   try {
     UR_CHECK_ERROR(cuArrayDestroy((CUarray)hImageMem));
+    if (auto it = hDevice->ChildCuarrayFromMipmapMap.find((CUarray)hImageMem);
```
jchlanda, line 433: Not a huge fan of c-style casts and in other parts of the codebase we try to actively avoid them, but seems like in this file it's a done thing, so am OK with it.

source/adapters/cuda/image.cpp
```diff
@@ -430,6 +430,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageFreeExp(
...
+        it != hDevice->ChildCuarrayFromMipmapMap.end()) {
```
jchlanda, line 434: This seems to be a bug. You are ignoring the result of a check for `ChildCuarrayFromMipmapMap.end()`, as it is placed inside of the conditional.

source/adapters/cuda/image.cpp
```diff
@@ -430,6 +430,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urBindlessImagesImageFreeExp(
...
+      UR_CHECK_ERROR(cuMipmappedArrayDestroy((CUmipmappedArray)it->second));
```
jchlanda, line 435: So `it` here can be `ChildCuarrayFromMipmapMap.end()`.

jchlanda (commented):
Sean is on holidays till the end of the year, so I took over. Thank you.

jchlanda (approved):

source/adapters/cuda/image.cpp
ProGTX: I agree, we'll need to get rid of the C-style casts separately.

PR #2301, opened 2024-11-08
[CMake]: Use check_linker_flag for /DEPENDENTLOADFLAG

Maetveis:
PR #2100 changed the WIN32 check to check from `if(WIN32)` to checking if link.exe is used via `CMAKE_CXX_COMPILER_LINKER_ID`. This has two problems:
- CMAKE_CXX_COMPILER_LINKER_ID is only supported starting with CMake 3.29, but UR still claims to CMake versions from 3.20 (`cmake_minimum_required` is called with this version). This results in the flag being silently dropped in earlier versions of CMake.
- There are other linkers that also support this flag for example LLD.

Using check_linker_flag resolves these issues without hard-coding a list of known linkers.

### DIFF
--- a/source/adapters/level_zero/CMakeLists.txt
+++ b/source/adapters/level_zero/CMakeLists.txt
@@ -3,6 +3,8 @@
 # See LICENSE.TXT
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+include(CheckLinkerFlag)
+
 if(UR_BUILD_ADAPTER_L0)
     set(ADAPTER_LIB_TYPE SHARED)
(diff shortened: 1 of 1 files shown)


Maetveis:
ping @kbenzie

kbenzie:
> ping @kbenzie

FYI I'm not a code owner for this change.

Maetveis:
@oneapi-src/unified-runtime-level-zero-write

> > ping @kbenzie
> 
> FYI I'm not a code owner for this change.

NP I thought I'd add you as the author of #2100.

@oneapi-src/unified-runtime-level-zero-write

PR #2303, opened 2024-11-08
[L0] Implement Support for zeInitDrivers

nrspruit:
- As of v1.10 of the L0 spec, zeInit and zeDriverGet is the old init pathway and the desired init api is zeInitDrivers. This new api allows for multi heterogenous drivers to coexist in a single L0 Process.

### DIFF
--- a/cmake/FetchLevelZero.cmake
+++ b/cmake/FetchLevelZero.cmake
@@ -40,7 +40,7 @@ if (NOT DEFINED LEVEL_ZERO_LIBRARY OR NOT DEFINED LEVEL_ZERO_INCLUDE_DIR)
         set(UR_LEVEL_ZERO_LOADER_REPO "https://github.com/oneapi-src/level-zero.git")
     endif()
     if (UR_LEVEL_ZERO_LOADER_TAG STREQUAL "")
-        set(UR_LEVEL_ZERO_LOADER_TAG v1.18.5)
+        set(UR_LEVEL_ZERO_LOADER_TAG v1.19.2)
     endif()
 
     # Disable due to a bug https://github.com/oneapi-src/level-zero/issues/104
--- a/.github/workflows/build-fuzz-reusable.yml
+++ b/.github/workflows/build-fuzz-reusable.yml
@@ -41,7 +41,7 @@ jobs:
 
     - name: Build level zero with gcc
       run: |
-        git clone -b v1.17.6 --depth=1 https://github.com/oneapi-src/level-zero.git ${{github.workspace}}/level-zero
+        git clone -b v1.18.5 --depth=1 https://github.com/oneapi-src/level-zero.git ${{github.workspace}}/level-zero
         cd ${{github.workspace}}/level-zero
         cmake -B build -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++
         cmake --build build -j $(nproc)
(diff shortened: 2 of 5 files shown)


nrspruit:
Found a race condition here: https://github.com/oneapi-src/level-zero/pull/232

Once the loader update is made and released, I will rebase this change.

nrspruit:
CI issues appears to be unreleated.

nrspruit:
@oneapi-src/unified-runtime-maintain please review.

cmake/FetchLevelZero.cmake
```diff
@@ -40,7 +40,7 @@ if (NOT DEFINED LEVEL_ZERO_LIBRARY OR NOT DEFINED LEVEL_ZERO_INCLUDE_DIR)
         set(UR_LEVEL_ZERO_LOADER_REPO "https://github.com/oneapi-src/level-zero.git")
     endif()
     if (UR_LEVEL_ZERO_LOADER_TAG STREQUAL "")
-        set(UR_LEVEL_ZERO_LOADER_TAG v1.18.5)
+        set(UR_LEVEL_ZERO_LOADER_TAG ea296eb72e760b9c04b93e008063ce95cdfe6488)
```
kbenzie, line 43: How come this is a commit hash but in the fuzz workflow the v0.18.5 tag is used?

I'm Wondering if the discrepancy may cause issues?

kbenzie (approved):
One query, otherwise LGTM.

cmake/FetchLevelZero.cmake
nrspruit: The fuzz workflow just needed at least v1.18 headers to compile and test, but the version of the loader needed in the Fetch requires a newer version that has several fixes for backwards compatibility.

winstonzhang-intel (approved):

nrspruit:
> @nrspruit This is marked ready to merge but two L0 jobs have failed (not counting the E2E tests) - are these failures expected?

@callumfare the errors are unrelated and fail sporadically in multiple PRs. 

I can rebase and repush the changes to see if I can clear it out. I will remove the read to merge until that is complete.

nrspruit:
@winstonzhang-intel , I found a couple of things that I needed to change, can you re-review?

nrspruit:
@callumfare , I will need @pbalcer's help to determine why only the static lib no platform test is hanging when it passes everywhere else.

nrspruit:
> @callumfare , I will need @pbalcer's help to determine why only the static lib no platform test is hanging when it passes everywhere else.

Based off an offline discussion, because the loader test error for loader_platform_test_no_platforms cannot be reproduced elsewhere we are disabling it until the issues with ctest can be worked out.

nrspruit:
Adding back ready to merge label. The intel/llvm ci is passing and the related L0 test CI are passing. The failures are unrelated, the CI is failing to download one of the packages. I am awaiting the CI to finally finish before I restart it, but the failures are unrelated.

nrspruit:
There was a race condition in the code, I have re-enabled the test that was failing on windows to see if that change fixes the issue. If so, then I will add back in the ready_to_merge, awaiting update of the CI results to be sure.

nrspruit:
> There was a race condition in the code, I have re-enabled the test that was failing on windows to see if that change fixes the issue. If so, then I will add back in the ready_to_merge, awaiting update of the CI results to be sure.

the fix to the code did not address the race condition with the loader no platforms test, I disabled the test again. Once the CI finishes I will add the ready_to_merge label again since this has been reviewed again offline.

nrspruit:
Based off the intel/llvm CI result here: https://github.com/intel/llvm/pull/16034 and the updated results in the CI, this is ready to merge again.

The failures in the e2e CI in UR are unrelated and have been seen before.

PR #2312, opened 2024-11-11
[UR] Add support for bidirectional USM prefetching

ianayl:
This PR implements the previously unimplemented `ur_usm_migration_flags_t` enum, and adds the logic for prefetching for from the device to host for each implemented adapter.

Note that level zero currently does not support this behavior: it remains unimplemented for now until level zero implements a way to prefetch from device to host.

Corresponding intel/llvm PR for testing: https://github.com/intel/llvm/pull/16047


source/adapters/level_zero/command_buffer.cpp
```diff
@@ -1300,7 +1300,12 @@ ur_result_t urCommandBufferAppendUSMPrefetchExp(
   std::ignore = EventWaitList;
   std::ignore = Event;
   std::ignore = Command;
-  std::ignore = Flags;
+
+  if (Flags == UR_USM_MIGRATION_FLAG_DEVICE_TO_HOST) {
+    logger::warning("USM migration from device to host is not currently "
+                    "supported by level zero.");
+    return UR_RESULT_SUCCESS;
+  }
```
EwanC, line 1308: This early return is a problem for out-of-order command-lists in that it doesn't respect the command dependencies or signal dependent commands when finished.

In the `else` branch below think we just want to opt-out of the actual `zeCommandListAppendMemoryPrefetch` but keep the `zeCommandListAppendWaitOnEvents` and `zeCommandListAppendSignalEvent` commnads.

ianayl: Hey thanks for reviewing! Admittedly I'm not the most familiar with how things work in L0; would I also need to do this for [source/adapters/level_zero/memory.cpp:1243](https://github.com/oneapi-src/unified-runtime/blob/3aa5249e071f23c261a8589eef6ea22ece024546/source/adapters/level_zero/memory.cpp#L1243)?

jchlanda (approved):
CUDA/HIP 👍

source/adapters/level_zero/command_buffer.cpp
EwanC: Yes I think that the same issue with an early return, the `OutEvent` output parameter won't be set and it's not respecting `EventWaitList` dependencies

test/conformance/enqueue/enqueue_adapter_cuda.match
```diff
@@ -1,10 +1,13 @@
...
 urEnqueueUSMAdviseTest.MultipleParamsSuccess/*
-urEnqueueUSMPrefetchWithParamTest.Success/*__UR_USM_MIGRATION_FLAG_DEFAULT
-urEnqueueUSMPrefetchWithParamTest.CheckWaitEvent/*__UR_USM_MIGRATION_FLAG_DEFAULT
+urEnqueueUSMPrefetchWithParamTest.Success/*__UR_USM_MIGRATION_FLAG_HOST_TO_DEVICE
+urEnqueueUSMPrefetchWithParamTest.Success/*__UR_USM_MIGRATION_FLAG_DEVICE_TO_HOST
+urEnqueueUSMPrefetchWithParamTest.CheckWaitEvent/*__UR_USM_MIGRATION_FLAG_HOST_TO_DEVICE
+urEnqueueUSMPrefetchWithParamTest.CheckWaitEvent/*__UR_USM_MIGRATION_FLAG_DEVICE_TO_HOST
 urEnqueueTimestampRecordingExpTest.Success/*
-urEnqueueTimestampRecordingExpTest.SuccessBlocking/*
+urEnqueueTimestampRecordingExpTest.SuccessBlocking/*
```
kbenzie, line 13: We've recently merged #2479 which removes uses of match files from the CTS, the new mechanism for tracking know failures is described [here](https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/README.md#known-failures).

kbenzie (commented):
Looks like there's conflicts in generated files, please pull in the latest changes from the main branch and during conflit resolution run the generate target to resolve those conflicts.

test/conformance/exp_command_buffer/commands.cpp
```diff
@@ -124,8 +124,13 @@ TEST_P(urCommandBufferCommandsTest, urCommandBufferAppendMemBufferFillExp) {
 
 TEST_P(urCommandBufferCommandsTest, urCommandBufferAppendUSMPrefetchExp) {
   ASSERT_SUCCESS(urCommandBufferAppendUSMPrefetchExp(
-      cmd_buf_handle, device_ptrs[0], allocation_size, 0, 0, nullptr, 0,
-      nullptr, nullptr, nullptr, nullptr));
+      cmd_buf_handle, device_ptrs[0], allocation_size,
+      UR_USM_MIGRATION_FLAG_HOST_TO_DEVICE, 0, nullptr, 0, nullptr, nullptr,
+      nullptr, nullptr));
+  ASSERT_SUCCESS(urCommandBufferAppendUSMPrefetchExp(
```
reble, line 130: My understanding is that changes include calling prefetch here twice for host and device in both directions. Is that consistent across all tests?

ianayl: As far as I'm aware I'm checking both directions for only tests checking functionality; e.g. I am not checking both directions if the test is checking how the function handles invalid input. Let me know if you'd like the tests changed to always use both directions regardless.

martygrant:

---

PR #2317, opened 2024-11-12
[CI] Move coverity job to public GHA runner

lukaszstolarczuk:
For this change to happen it was required to re-write the workflow:
- install dependencies for adapters and UMF,
- download coverity tool from the project's page,
- push tarball to Coverity's scan webpage via curl.

FYI, Coverity's been disabled on main for some time, since no proper self-hosted runner was enabled.


.github/workflows/coverity.yml
```diff
@@ -1,63 +1,81 @@
...
 
+      - name: Download Coverity
+        run: |
+          wget -O coverity_tool.tgz -nv https://scan.coverity.com/download/linux64 \
+            --post-data "token=${{ secrets.COVERITY_SCAN_TOKEN }}&project=oneapi-src%2Funified-runtime"
+
+      - name: Extract Coverity
+        run: tar xzf coverity_tool.tgz
+
+      # TODO: enable HIP adapter as well (requires proper package(s) installation)
```
lukaszstolarczuk, line 44: all packages I tried to install for HIP was too big (storage on public runners should be around ~14GB) - any hints what I could install here to enable HIP adapter's build...?

lukaszstolarczuk: @oneapi-src/unified-runtime-hip-write, any hint what minimal packages I could install that fit within 14GBs and enables UR build...?

PR #2320, opened 2024-11-13
[NFC] Make macro in conf. test no longer leak

RossBrunton:
`UUR_ASSERT_SUCCESS_OR_UNSUPPORTED` previously leaked a `status`
variable to the calling scope. This change makes it no longer do that.

### DIFF
--- a/test/conformance/testing/include/uur/fixtures.h
+++ b/test/conformance/testing/include/uur/fixtures.h
@@ -23,13 +23,15 @@
     (void)0
 
 #define UUR_ASSERT_SUCCESS_OR_UNSUPPORTED(ret)                                 \
-    auto status = ret;                                                         \
-    if (status == UR_RESULT_ERROR_UNSUPPORTED_FEATURE ||                       \
-        status == UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION) {                   \
-        GTEST_SKIP();                                                          \
-    } else {                                                                   \
-        ASSERT_EQ(status, UR_RESULT_SUCCESS);                                  \
-    }
+    do {                                                                       \
+        auto status = ret;                                                     \
+        if (status == UR_RESULT_ERROR_UNSUPPORTED_FEATURE ||                   \
+            status == UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION) {               \
+            GTEST_SKIP();                                                      \
+        } else {                                                               \
+            ASSERT_EQ(status, UR_RESULT_SUCCESS);                              \
+        }                                                                      \
+    } while (0)
 
 namespace uur {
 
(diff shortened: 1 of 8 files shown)


test/conformance/testing/include/uur/fixtures.h
```diff
@@ -22,14 +22,14 @@
...
-#define UUR_ASSERT_SUCCESS_OR_UNSUPPORTED(ret)                                 \
+#define UUR_ASSERT_SUCCESS_OR_UNSUPPORTED(ret) do {                            \
     auto status = ret;                                                         \
     if (status == UR_RESULT_ERROR_UNSUPPORTED_FEATURE ||                       \
         status == UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION) {                   \
         GTEST_SKIP();                                                          \
     } else {                                                                   \
         ASSERT_EQ(status, UR_RESULT_SUCCESS);                                  \
-    }
+    } } while(0)                                                               \
```
aarongreig, line 32: I know it's semantically the same but I'd prefer just a scope like https://github.com/oneapi-src/unified-runtime/pull/2193/files#diff-3c41a24f2744a3a369eebf302dd6359004331bb34e5828ea2033c4aeafc5ca3aR26

test/conformance/testing/include/uur/fixtures.h
```diff
@@ -22,14 +22,14 @@
     }                                                                          \
     (void)0
 
-#define UUR_ASSERT_SUCCESS_OR_UNSUPPORTED(ret)                                 \
+#define UUR_ASSERT_SUCCESS_OR_UNSUPPORTED(ret) do {                            \
```
martygrant, line 25: I guess because `status` wasn't enclosed in any scope it was leaking and causing the errors elsewhere? Could the macro just be enclosed in `{ }` instead of a loop maybe?

test/conformance/testing/include/uur/fixtures.h
RossBrunton: do .. while(0) is traditionally used to ensure semicolons are handled correctly. For example:

```c++
if (something)
  UUR_ASSERT_SUCCESS_OR_UNSUPPORTED(ret);
else
  something_else();
```

https://stackoverflow.com/questions/154136/why-use-apparently-meaningless-do-while-and-if-else-statements-in-macros

aarongreig: ah I see, it's ugly but if it works it works

aarongreig (approved):

test/conformance/testing/include/uur/fixtures.h
martygrant: Yep thanks for the link!

martygrant (approved):

RossBrunton:
Doesn't affect LLVM, merging without LLVM change.

PR #2326, opened 2024-11-14
Index into property array in urDevicePartition for OpenCL

martygrant:
In `urDevicePartition` in the OpenCL adapter we loop through the array of properties but were not actually indexing into the array, only using the first element of the array.

Also added a new CTS test to verify this behaviour.

https://github.com/intel/llvm/pull/16083


callumfare (approved):

alycm:
Is it possible to have some kind of CTS or regression test for this?

kbenzie (approved):

martygrant:
@alycm added a CTS test for this

PR #2327, opened 2024-11-14
[CommandBuffer] Make double finalizing fail

hdelan:
A command buffer should fail if finalize is called multiple times on the same command buffer. This matches openCL behaviour.

DPC++ PR: https://github.com/intel/llvm/pull/16097


source/adapters/cuda/command_buffer.cpp
```diff
@@ -401,6 +401,8 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
 
 UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
+  UR_ASSERT(!hCommandBuffer->IsFinalized,
+            UR_RESULT_ERROR_INVALID_COMMAND_BUFFER_EXP);
```
EwanC, line 405: Slight preference for `UR_RESULT_ERROR_INVALID_OPERATION` to be the error code to match this OpenCL error code from [clFinalizeCommandBufferKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clFinalizeCommandBufferKHR)

> [CL_INVALID_OPERATION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_OPERATION) if command_buffer is not in the [Recording](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#recording) state.

We should also specify that this error will be returned here https://github.com/oneapi-src/unified-runtime/blob/main/scripts/core/exp-command-buffer.yml#L331

EwanC: Nice one, expect you'll need to run `ninja generate` now you've touch yml

source/adapters/cuda/command_buffer.cpp
```diff
@@ -401,6 +401,7 @@ urCommandBufferReleaseExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
 
 UR_APIEXPORT ur_result_t UR_APICALL
 urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
+  UR_ASSERT(!hCommandBuffer->IsFinalized, UR_RESULT_ERROR_INVALID_OPERATION);
```
EwanC, line 404: We might be able to check if the `CudaGraphExec` member is not `nullptr`, rather than needing to add a new member variable for `IsFinalized`. i.e `UR_ASSERT(hCommandBuffer->CudaGraphExec == nullptr, UR_RESULT_ERROR_INVALID_OPERATION);`

source/adapters/cuda/command_buffer.cpp
hdelan: Good catch. Have updated autogenerated stuff.

source/adapters/cuda/command_buffer.cpp
hdelan: Aha sorry I thought it was alive before this point. Yes that sounds good to me.

test/conformance/exp_command_buffer/commands.cpp
```diff
@@ -204,3 +204,14 @@ TEST_P(urCommandBufferAppendKernelLaunchExpTest, Basic) {
         ASSERT_EQ(result, ptrZ[i]);
     }
 }
+
+TEST_P(urCommandBufferAppendKernelLaunchExpTest, FinalizeTwice) {
```
EwanC, line 208: For native CPU the kernel tests don't work, so think this will need to be added to the match file next to the test above https://github.com/oneapi-src/unified-runtime/blob/4fb1a847b39e2cd2c5efac952e72ec3cf9dd8ddd/test/conformance/exp_command_buffer/exp_command_buffer_adapter_native_cpu.match#L7

(Bonus points if you want to look into why that's happening, but it's not a change for this PR)

aarongreig (approved):

test/conformance/exp_command_buffer/commands.cpp
hdelan: Good shout. Have updated match file

EwanC (approved):

EwanC:
Can you whack a link to the DPC++ PR for this in the description, ta

callumfare:
@hdelan This PR is next in the queue to be merged, could you resolve the conflict?

EwanC:
> @hdelan This PR is next in the queue to be merged, could you resolve the conflict?

Hugh is on holiday until Monday, so might be better to jump to another ready-to-merge PR until then

hdelan:
Have rebased

martygrant:
@hdelan there's a failing command buffer test, just checking this is expected before this can be merged?

hdelan:
The error given is:

```
Error: ZERO>[ERROR]: urCommandBufferCreateExp function not implemented!
```

Which I reckon is due to old/incompatible drivers on the runner. I'll kick it off again and see if it passes.

PR #1600, opened 2024-05-14
[L0] Refactoring of boolean event parameters

winstonzhang-intel:
CI in LLVM/SYCL: https://github.com/intel/llvm/pull/14536

PR #2334, opened 2024-11-16
[L0] Interrupt-based event implementation

winstonzhang-intel:
To expose this functionality in UR, we want two ways of enabling low-power events:

1. Queue-wide enabling so all events created on the queue are low-powered.
2. As a property passed to urEnqueueEventsWaitWithBarrier making the resulting event a low-power event. This will require the existing interface to be extended with properties, potentially through a new experimental function.

SYCL/LLVM: https://github.com/intel/llvm/pull/16252

### DIFF
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -478,7 +478,8 @@ static const uint32_t MaxNumEventsPerPool = [] {
 ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
     ze_event_pool_handle_t &Pool, size_t &Index, bool HostVisible,
     bool ProfilingEnabled, ur_device_handle_t Device,
-    bool CounterBasedEventEnabled, bool UsingImmCmdList) {
+    bool CounterBasedEventEnabled, bool UsingImmCmdList,
+    bool InterruptBasedEventEnabled) {
   // Lock while updating event pool machinery.
   std::scoped_lock<ur_mutex> Lock(ZeEventPoolCacheMutex);
 
@@ -536,6 +537,14 @@ ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
       logger::debug("ze_event_pool_desc_t counter based flags set to: {}",
                     counterBasedExt.flags);
       ZeEventPoolDesc.pNext = &counterBasedExt;
+    }
+    if (InterruptBasedEventEnabled) {
+      ze_intel_event_sync_mode_exp_desc_t eventSyncMode = {
+          ZE_INTEL_STRUCTURE_TYPE_EVENT_SYNC_MODE_EXP_DESC, nullptr, 0};
+      eventSyncMode.syncModeFlags =
+          ZE_INTEL_EVENT_SYNC_MODE_EXP_FLAG_LOW_POWER_WAIT |
+          ZE_INTEL_EVENT_SYNC_MODE_EXP_FLAG_SIGNAL_INTERRUPT;
+      ZeEventPoolDesc.pNext = &eventSyncMode;
     }
 
     std::vector<ze_device_handle_t> ZeDevices;
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -574,8 +574,9 @@ ur_event_handle_t ur_context_handle_t_::getEventFromContextCache(
     bool HostVisible, bool WithProfiling, ur_device_handle_t Device,
     bool CounterBasedEventEnabled, bool InterruptBasedEventEnabled) {
   std::scoped_lock<ur_mutex> Lock(EventCacheMutex);
-  auto Cache = getEventCache(HostVisible, WithProfiling, Device,
-                             CounterBasedEventEnabled);
+  auto Cache =
+      getEventCache(HostVisible, WithProfiling, Device,
+                    CounterBasedEventEnabled, InterruptBasedEventEnabled);
   if (Cache->empty()) {
     logger::info("Cache empty (Host Visible: {}, Profiling: {}, Counter: {}, "
                  "Device: {})",
@@ -611,9 +612,9 @@ void ur_context_handle_t_::addEventToContextCache(ur_event_handle_t Event) {
     Device = Event->UrQueue->Device;
   }
 
-  auto Cache =
-      getEventCache(Event->isHostVisible(), Event->isProfilingEnabled(), Device,
-                    Event->CounterBasedEventsEnabled);
+  auto Cache = getEventCache(
+      Event->isHostVisible(), Event->isProfilingEnabled(), Device,
+      Event->CounterBasedEventsEnabled, Event->InterruptBasedEventsEnabled);
   logger::info("Inserting {} event (Host Visible: {}, Profiling: {}, Counter: "
                "{}, Device: {}) into cache {}",
                Event, Event->HostVisibleEvent, Event->isProfilingEnabled(),
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -579,27 +579,24 @@ ur_event_handle_t ur_context_handle_t_::getEventFromContextCache(
                     CounterBasedEventEnabled, InterruptBasedEventEnabled);
   if (Cache->empty()) {
     logger::info("Cache empty (Host Visible: {}, Profiling: {}, Counter: {}, "
-                 "Device: {})",
-                 HostVisible, WithProfiling, CounterBasedEventEnabled, Device);
+                 "Interrupt: {}, Device: {})",
+                 HostVisible, WithProfiling, CounterBasedEventEnabled,
+                 InterruptBasedEventEnabled, Device);
     return nullptr;
   }
 
   auto It = Cache->begin();
   ur_event_handle_t Event = *It;
-  if (Event->CounterBasedEventsEnabled != CounterBasedEventEnabled) {
-    return nullptr;
-  }
-  if (Event->InterruptBasedEventsEnabled != InterruptBasedEventEnabled) {
-    return nullptr;
-  }
+
   Cache->erase(It);
   // We have to reset event before using it.
   Event->reset();
 
   logger::info("Using {} event (Host Visible: {}, Profiling: {}, Counter: {}, "
-               "Device: {}) from cache {}",
+               "Interrupt: {}, Device: {}) from cache {}",
                Event, Event->HostVisibleEvent, Event->isProfilingEnabled(),
-               Event->CounterBasedEventsEnabled, Device, Cache);
+               Event->CounterBasedEventsEnabled,
+               Event->InterruptBasedEventsEnabled, Cache);
 
   return Event;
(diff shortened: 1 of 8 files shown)


source/adapters/level_zero/device.cpp
```diff
@@ -485,6 +485,8 @@ ur_result_t urDeviceGetInfo(
   case UR_DEVICE_INFO_BUILT_IN_KERNELS:
     // TODO: To find out correct value
     return ReturnValue("");
+  case UR_DEVICE_INFO_LOW_POWER_EVENTS_EXP:
+    return ReturnValue(UR_DEVICE_INFO_LOW_POWER_EVENTS_EXP);
```
EwanC, line 489: the return type for `UR_DEVICE_INFO_LOW_POWER_EVENTS_EXP` is a `ur_bool_t`, but it looks like we're just returning the enum back when being queried.

winstonzhang-intel: when this flag is set, we return true now.

source/adapters/level_zero/event.cpp
```diff
@@ -434,8 +456,22 @@ ur_result_t urEnqueueEventsWaitWithBarrierExt(
         *OutEvent ///< [in,out][optional] return an event object that identifies
                   ///< this particular command instance.
 ) {
-  return ur::level_zero::urEnqueueEventsWaitWithBarrier(
-      Queue, NumEventsInWaitList, EventWaitList, OutEvent);
+  bool InterruptBased =
+      EnqueueExtProp &&
+      (EnqueueExtProp->flags & UR_EXP_ENQUEUE_EXT_FLAG_LOW_POWER_EVENTS);
+  if (InterruptBased) {
```
pbalcer, line 462: I'm not sure I understand the intention here or why we need to cast the function.

source/adapters/level_zero/event.cpp
```diff
@@ -434,8 +456,22 @@ ur_result_t urEnqueueEventsWaitWithBarrierExt(
...
+    return static_cast<ur_result_t (*)(
+        ur_queue_handle_t, uint32_t, const ur_event_handle_t *,
+        ur_event_handle_t *, bool)>(EnqueueEventsWaitWithBarrier)(
+        Queue, NumEventsInWaitList, EventWaitList, OutEvent, true);
+  } else {
+    return static_cast<ur_result_t (*)(
+        ur_queue_handle_t, uint32_t, const ur_event_handle_t *,
+        ur_event_handle_t *, bool)>(EnqueueEventsWaitWithBarrier)(
+        Queue, NumEventsInWaitList, EventWaitList, OutEvent,
+        Queue ? Queue->InterruptBasedEventsEnabled : false);
```
pbalcer, line 473: `Queue ? Queue->InterruptBasedEventsEnabled : false`
I don't believe Queue can be null here. So this should be simply `Queue->InterruptBasedEventsEnabled`.

source/adapters/level_zero/context.cpp
```diff
@@ -574,6 +583,9 @@ ur_event_handle_t ur_context_handle_t_::getEventFromContextCache(
   if (Event->CounterBasedEventsEnabled != CounterBasedEventEnabled) {
     return nullptr;
   }
+  if (Event->InterruptBasedEventsEnabled != InterruptBasedEventEnabled) {
```
pbalcer, line 586: This will not work unless the interrupt based events are somehow first on the Cache container. The same is true of the counter-based events. There needs to be separate pool for interrupt-driven events (or maybe they shouldn't be cached at all?).

source/adapters/level_zero/context.hpp
```diff
@@ -225,17 +245,29 @@ struct ur_context_handle_t_ : _ur_object {
...
+
+    HostVisibleInterruptBasedRegularCacheType,
+    HostInvisibleInterruptBasedRegularCacheType,
+    HostVisibleInterruptBasedImmediateCacheType,
+    HostInvisibleInterruptBasedImmediateCacheType,
+
+    HostVisibleInterruptAndCounterBasedRegularCacheType,
+    HostInvisibleInterruptAndCounterBasedRegularCacheType,
+    HostVisibleInterruptAndCounterBasedImmediateCacheType,
+    HostInvisibleInterruptAndCounterBasedImmediateCacheType
```
pbalcer, line 258: This is exactly why I've asked you to do the refactor. This will grow exponentially with new event parameters.

pbalcer (approved):

source/adapters/level_zero/event.cpp
nrspruit: I agree, this is not needed, please remove and simplify.

source/adapters/level_zero/context.hpp
nrspruit: I agree, @winstonzhang-intel has https://github.com/oneapi-src/unified-runtime/pull/1600 for the refactor that has been moved back due to other issues, but I have asked him to finish that this week with your recent changes to the cache.

I think the refactor though should be in that refactor PR after this change to avoid over complicating this feature review.

source/adapters/level_zero/queue.cpp
```diff
@@ -1192,10 +1192,20 @@ ur_queue_handle_t_::ur_queue_handle_t_(
     }
     return std::atoi(UrRet) != 0;
   }();
+  static const bool useInterruptBasedEvents = [] {
+    const char *UrRet = std::getenv("UR_L0_USE_INTERRUPT_BASED_EVENTS");
```
nrspruit, line 1196: I don't think the env is necessary, this event type creation is based on input flags from the user. They are not automatically created and should only event be created when the user requested it.

source/adapters/level_zero/queue.cpp
```diff
@@ -1192,10 +1192,20 @@ ur_queue_handle_t_::ur_queue_handle_t_(
...
+      return true;
+    }
+    return std::atoi(UrRet) != 0;
+  }();
   this->CounterBasedEventsEnabled =
       UsingImmCmdLists && isInOrderQueue() && Device->useDriverInOrderLists() &&
       useDriverCounterBasedEvents &&
       Device->Platform->ZeDriverEventPoolCountingEventsExtensionFound;
+  this->InterruptBasedEventsEnabled = useInterruptBasedEvents &&
+                                      isLowPowerEvents() && isInOrderQueue() &&
```
nrspruit, line 1207: This should be simplified to:
isLowPowerEvents() && isInOrderQueue() &&  Device->useDriverInOrderLists();

The user directly requests for this feature so we don't need another variable to ignore the user.

source/adapters/level_zero/event.cpp
winstonzhang-intel: -We make sure that EnqueueExtProp is not a nullptr then check the flag to see if we want to use LOW_POWER.
-EnqueueEventsWaitWithBarrier(ur_queue_handle_t_*&, uint32_t&, ur_event_handle_t_* const*&, ur_event_handle_t_**&, bool is overloaded therefore we needed to specify.

source/adapters/level_zero/context.cpp
winstonzhang-intel: good catch, added to getEventCache to address interrupt case.

source/adapters/level_zero/context.hpp
winstonzhang-intel: Since there will be some refactoring done with this PR, I will work on the refactor patch after this is in. https://github.com/oneapi-src/unified-runtime/pull/1600

source/adapters/level_zero/queue.cpp
winstonzhang-intel: Gotcha, removed.

EwanC (approved):
command_buffer.cpp LGTM

source/adapters/level_zero/context.cpp
nrspruit: So, just to confirm, now there is a separate cache for both counting events and interrupt based events correct?

If the cache retrieved must contain only the types of events that can be used ie counter based events etc, then the if checks are not needed. ie a counter based event cache would retrieve counter based events, interrupt based event cache will have interrupt based events.

source/adapters/level_zero/event.cpp
nrspruit: I think this would be cleaner if you move the implementation of urEnqueueEventsWaitWithBarrier() to urEnqueueEventsWaitWithBarrierExt() and have urEnqueueEventsWaitWithBarrier call urEnqueueEventsWaitWithBarrierExt with the check for the queue property for interrupt based.

source/adapters/level_zero/context.cpp
winstonzhang-intel: > So, just to confirm, now there is a separate cache for both counting events and interrupt based events correct?
> 
> If the cache retrieved must contain only the types of events that can be used ie counter based events etc, then the if checks are not needed. ie a counter based event cache would retrieve counter based events, interrupt based event cache will have interrupt based events.

So after the rebase, getEventCache now returns the cached event for the given profiling mode and provided scope. I simply added another case of interrupt based. As for the eventPoolCache, I've already added another pool for interrupt based. 
Like you said, we need to make sure counter based event cache retrieves counter based events and likewise for interrupt based event. That's why this check is in place (Event->InterruptBasedEventsEnabled != InterruptBasedEventEnabled).

winstonzhang-intel: Alternatively, we can always remove the caching of interrupt based events. If this is what we want, I can simply remove this portion of the patch.

source/adapters/level_zero/event.cpp
winstonzhang-intel: Good idea, let me add that in. Thanks!

nrspruit (approved):
LGTM, thank you for the updates.

nrspruit (approved):
Thanks for the syntax fixes.

nrspruit (commented):
CI is failing reporting no devices sporadically. It appears unrelated.

source/adapters/level_zero/context.cpp
```diff
@@ -537,6 +538,14 @@ ur_result_t ur_context_handle_t_::getFreeSlotInExistingOrNewPool(
                     counterBasedExt.flags);
       ZeEventPoolDesc.pNext = &counterBasedExt;
     }
+    if (InterruptBasedEventEnabled) {
+      ze_intel_event_sync_mode_exp_desc_t eventSyncMode = {
+          ZE_INTEL_STRUCTURE_TYPE_EVENT_SYNC_MODE_EXP_DESC, nullptr, 0};
+      eventSyncMode.syncModeFlags =
+          ZE_INTEL_EVENT_SYNC_MODE_EXP_FLAG_LOW_POWER_WAIT |
+          ZE_INTEL_EVENT_SYNC_MODE_EXP_FLAG_SIGNAL_INTERRUPT;
+      ZeEventPoolDesc.pNext = &eventSyncMode;
```
al42and, line 547: Hi! This stores a pointer to a temporary in `ZeEventPoolDesc.pNext`, and also potentially (with `CounterBasedEventEnabled == true`) overwrites the previously stored value.

PR #2335, opened 2024-11-18
logging full path of loaded adapter

lslusarczyk:


source/common/linux/ur_lib_loader.cpp
```diff
@@ -50,7 +54,14 @@ LibLoader::loadAdapterLibrary(const char *name) {
         logger::info("failed to load adapter '{}' with error: {}", name,
                      err ? err : "unknown error");
     } else {
-        logger::info("loaded adapter 0x{} ({})", handle, name);
+#if defined(ADD_FULL_PATH_LOG)
```
pbalcer, line 57: might be easier to just do `realpath` on `name`.

pbalcer: good point, realpath would just resolve the relative path of name. nvm.

lslusarczyk:
PR on LLVM pass with this change. Failure in CI here is not related to my change but it is Windows compilation broken in general. Please merge.

pbalcer (approved):

PR #2336, opened 2024-11-18
[CUDA][HIP][CommandBuffer] Add try catches to anything that can throw

hdelan:
`UR_CHECK_ERROR` must be called within a try catch block, as it can throw.

For functions that don't return a `ur_result_handle_t`, and which might throw, make sure these funcs are only called within try catches in the caller funcs, and that they are not exposed to the user.

DPC++ PR: https://github.com/intel/llvm/pull/16105

### DIFF
--- a/source/adapters/cuda/command_buffer.cpp
+++ b/source/adapters/cuda/command_buffer.cpp
@@ -21,7 +21,7 @@
 
 namespace {
 ur_result_t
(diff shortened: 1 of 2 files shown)


hdelan:
Thanks @martygrant have rebased.

martygrant:
This will get pulled in as part of https://github.com/intel/llvm/pull/15965

PR #2352, opened 2024-11-19
Fix a number of issues from the latest coverity scan.

aarongreig:


kbenzie (approved):

PietroGhg (approved):
Native CPU lgtm, thank you

source/adapters/hip/common.cpp
```diff
@@ -164,13 +164,19 @@ void detail::ur::assertion(bool Condition, const char *pMessage) {
...
 // Global variables for UR_RESULT_ADAPTER_SPECIFIC_ERROR
 thread_local ur_result_t ErrorMessageCode = UR_RESULT_SUCCESS;
-thread_local char ErrorMessage[MaxMessageSize];
+thread_local char ErrorMessage[MaxMessageSize]{};
 
 // Utility function for setting a message and warning
 [[maybe_unused]] void setErrorMessage(const char *pMessage,
                                       ur_result_t ErrorCode) {
   assert(strlen(pMessage) < MaxMessageSize);
+  // Copy MaxMessageSize - 1 bytes to ensure the resultant string is always null
```
frasercrmck, line 173: Maybe "copy *at most* MaxMessageSize - 1" bytes?

frasercrmck: At least, that's how I infer `strncpy_s` to work. I see that `strncpy` seems to copy all MaxMessageSize - 1 characters after all.

frasercrmck (commented):
CUDA/HIP LGTM other than a nit about that comment

aarongreig:
LLVM PR https://github.com/intel/llvm/pull/16158

aarongreig:
ping @oneapi-src/unified-runtime-level-zero-write

source/adapters/level_zero/common.cpp
```diff
@@ -332,15 +332,21 @@ template <> zes_structure_type_t getZesStructureType<zes_mem_properties_t>() {
 
 // Global variables for ZER_EXT_RESULT_ADAPTER_SPECIFIC_ERROR
 thread_local ur_result_t ErrorMessageCode = UR_RESULT_SUCCESS;
-thread_local char ErrorMessage[MaxMessageSize];
+thread_local char ErrorMessage[MaxMessageSize]{};
 thread_local int32_t ErrorAdapterNativeCode;
 
 // Utility function for setting a message and warning
 [[maybe_unused]] void setErrorMessage(const char *pMessage,
```
pbalcer, line 339: this same function is repeated multiple times. can we move this to common?

source/adapters/level_zero/common.cpp
```diff
@@ -332,15 +332,21 @@ template <> zes_structure_type_t getZesStructureType<zes_mem_properties_t>() {
...
                                       ur_result_t ErrorCode,
                                       int32_t AdapterErrorCode) {
-  assert(strlen(pMessage) <= MaxMessageSize);
-  strcpy(ErrorMessage, pMessage);
+  assert(strlen(pMessage) < MaxMessageSize);
+  // Copy at most MaxMessageSize - 1 bytes to ensure the resultant string is
+  // always null terminated.
+#if defined(_WIN32)
+  strncpy_s(ErrorMessage, MaxMessageSize - 1, pMessage, strlen(pMessage));
```
pbalcer, line 346: tbh I'd just ignore coverity complaining about not using safe windows string functions, and use `strncpy` on all platforms. The `strlen(pMessage)` does not make this any safer.

aarongreig: I copied this from other places with strcpy ([here](https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_print.cpp#L33) and [here](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/mock/ur_mock.cpp#L38)), I remember there being issues on some windows configs with warnings being generated if you don't use strncpy_s. I didn't actually check if that was still a concern, happy to delete if it isn't

kbenzie: > tbh I'd just ignore coverity complaining about not using safe windows string functions, and use `strncpy` on all platforms.

I'm happy with this approach.

pbalcer: > I copied this from other places with strcpy ([here](https://github.com/oneapi-src/unified-runtime/blob/main/source/loader/ur_print.cpp#L33) and [here](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/mock/ur_mock.cpp#L38)), I remember there being issues on some windows configs with warnings being generated if you don't use strncpy_s. I didn't actually check if that was still a concern, happy to delete if it isn't

Well, the theory is that if you know the length of the string (which we don't here), you can supply it to `strncpy_s` and it can do bounds checking without having to iterate over the cstring. This protects against malformed cstrings that are not 0-terminated.

aarongreig: deleted the strncpy_s path

nrspruit (approved):

omarahmed1111 (approved):

PR #2353, opened 2024-11-19
[L0] Enable zesInit by default given newer L0 IP version

nrspruit:
- If the user has not overwritten the sysman init type, then the adapter will check the device ip version to determine if zesInit should be used and the env disabled.


winstonzhang-intel (approved):

nrspruit:
I added ready to merge since the CI failures are unrelated. I will rekick the CI once it allows it....

martygrant:
intel/llvm PR for UR tag update - https://github.com/oneapi-src/unified-runtime/pull/2353

PR #2354, opened 2024-11-19
[L0 v2] implement native handle API for event

igchor:


test/adapters/level_zero/CMakeLists.txt
```diff
@@ -3,89 +3,108 @@
 # See LICENSE.TXT
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-if(UR_BUILD_ADAPTER_L0)
+if(NOT WIN32 AND NOT UR_STATIC_ADAPTER_L0)
```
pbalcer, line 6: this is terrible, and we should remove the test that requires this.

igchor: I know... Right now we have 2 tests that use this I belive. I was thinking that perhaps until we have a better way we could just move the checks outside of the binary to some script - that is, just print some kind of markers in the test binary and redirect it along with ZE traces to a file and then do matching in python. Anyway, that's something for a separate PR.

source/adapters/level_zero/v2/event.cpp
```diff
@@ -183,6 +237,10 @@ ur_result_t urEventGetInfo(ur_event_handle_t hEvent, ur_event_info_t propName,
 
   switch (propName) {
   case UR_EVENT_INFO_COMMAND_EXECUTION_STATUS: {
+    // We don't have NOT_SUBMITTED status in UR
+    if (!hEvent->getQueue())
```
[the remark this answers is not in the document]
igchor: Right, there's no reason not to support this. Done. I also added a test for this.

source/adapters/level_zero/v2/event.cpp
```diff
@@ -300,4 +361,32 @@ ur_result_t urEventGetProfilingInfo(
...
+  if (!hNativeEvent) {
+    assert((hContext->nativeEventsPool.getFlags() & v2::EVENT_FLAGS_COUNTER) ==
+           0);
+
+    *phEvent = hContext->nativeEventsPool.allocate();
+    ZE2UR_CALL(zeEventHostSignal, ((*phEvent)->getZeEvent()));
+  } else {
+    *phEvent = new ur_native_event_t(hNativeEvent, hContext, pProperties);
+  }
+  return UR_RESULT_SUCCESS;
```
pbalcer, line 387: `(*phEvent)->retain();`

igchor: That should not be needed. I think the problem was with Mutex not being unlocked before destroying the event. Should be fixed now.

pbalcer: I verified, it works.

test/adapters/level_zero/CMakeLists.txt
igchor: Done: https://github.com/oneapi-src/unified-runtime/pull/2431

pbalcer (approved):

PR #2356, opened 2024-11-20
[HIP] Disable SYCL images by default

npmiller:
Align the HIP adapter with the CUDA adapter and report SYCL images disabled by default, they're implemented at about the same level as the ones in CUDA, so there's no reason to treat them differently on HIP.

Also any future work to improve the HIP image support should focus on bindless images rather than this SYCL image support, so it makes sense to consider this essentially "deprecated".

### DIFF
--- a/source/adapters/hip/device.cpp
+++ b/source/adapters/hip/device.cpp
@@ -12,6 +12,7 @@
 #include "adapter.hpp"
 #include "context.hpp"
 #include "event.hpp"
+#include "logger/ur_logger.hpp"
 
 #include <sstream>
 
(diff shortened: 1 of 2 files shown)


source/adapters/hip/device.cpp
```diff
@@ -223,7 +224,19 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
     return ReturnValue(uint64_t{MaxAlloc});
   }
   case UR_DEVICE_INFO_IMAGE_SUPPORTED: {
-    return ReturnValue(ur_bool_t{true});
+    bool Enabled = false;
+
+    if (std::getenv("UR_HIP_ENABLE_IMAGE_SUPPORT") != nullptr) {
```
frasercrmck, line 229: I note that CUDA also has `SYCL_PI_CUDA_ENABLE_IMAGE_SUPPORT`. Do we need to mirror that?

npmiller: `SYCL_PI_CUDA_ENABLE_IMAGE_SUPPORT` is a leftover from the PI interface that pre-dates UR, it's in the CUDA adapter for legacy reasons, since we're introducing the variable for HIP I think it makes sense to just have the UR variant.

frasercrmck: Fair enough, thanks

frasercrmck (approved):

aarongreig (approved):

npmiller:
intel/llvm change + testing: https://github.com/intel/llvm/pull/16147

npmiller:
E2E failures also happen on other PRs and don't happen on the intel/llvm PR testing so this should be ready although it does need the matching intel/llvm change

PR #2357, opened 2024-11-20
[HIP] Fix error code for unsupported program info

npmiller:
Unsupported program info should return unsupported enumeration instead of unsupported feature.

This aligns with the CUDA adapter and makes the CTS understand that these properties aren't failing but just unsupported.

### DIFF
--- a/source/adapters/hip/program.cpp
+++ b/source/adapters/hip/program.cpp
@@ -408,7 +408,10 @@ urProgramGetInfo(ur_program_handle_t hProgram, ur_program_info_t propName,
   case UR_PROGRAM_INFO_BINARIES:
     return ReturnValue(&hProgram->Binary, 1);
   case UR_PROGRAM_INFO_KERNEL_NAMES:
-    return getKernelNames(hProgram);
+    /* TODO: Add implementation for getKernelNames */
+    UR_ASSERT(getKernelNames(hProgram), UR_RESULT_ERROR_UNSUPPORTED_FEATURE);
+    return UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION;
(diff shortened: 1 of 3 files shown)


source/adapters/hip/program.cpp
```diff
@@ -408,7 +408,10 @@ urProgramGetInfo(ur_program_handle_t hProgram, ur_program_info_t propName,
   case UR_PROGRAM_INFO_BINARIES:
     return ReturnValue(&hProgram->Binary, 1);
   case UR_PROGRAM_INFO_KERNEL_NAMES:
-    return getKernelNames(hProgram);
+    /* TODO: Add implementation for getKernelNames */
+    UR_ASSERT(getKernelNames(hProgram), UR_RESULT_ERROR_UNSUPPORTED_FEATURE);
```
frasercrmck, line 412: I'm aware this was copied from CUDA but I don't think I like using `UR_ASSERT` for checking error codes. This is essentially saying that if `getKernelNames` returns `UR_SUCCESS`, it should in fact return `UR_RESULT_ERROR_UNSUPPORTED_FEATURE`? That's confusing, especially as `getKernelNames` unconditionally returns `UR_RESULT_ERROR_UNSUPPORTED_FEATURE` anyway.

kbenzie (approved):

source/adapters/hip/program.cpp
npmiller: Agreed, that code is a bit silly, just updated it for CUDA/HIP to remove the extra function and asserts.

npmiller:
L0 failures look unrelated, this patches doesn't touch the L0 adapter at all

PR #2360, opened 2024-11-20
[Cuda] Implement device info extensions for HW_THREADS_PER_EU and EU_SIMD_WIDTH on Cuda

GeorgeWeb:
Implements the `UR_DEVICE_INFO_GPU_HW_THREADS_PER_EU` and `UR_DEVICE_INFO_GPU_EU_SIMD_WIDTH` device info property queries on Cuda, which were originally supported only on LevelZero.

https://github.com/intel/llvm/pull/16148


jinz2014:
I have a question. What is EU in the following diagram ?   

image

GeorgeWeb:
> I have a question. What is EU in the following diagram ?
> 
I don't think there's a good 1:1 mapping here. This PR was just an experiment that we deemed is not going to work because the concepts don't match at all. Here I interpreted EU as SM for the sake of getting some values but it's all conceptually incorrect so we'll do things through separate queries. 
Apologies @jinz2014 if that's caused any confusion to you, I was meant to close this PR already but have forgotten.

Issue #2323, opened 2024-11-13
`-flto -fsanitize=cfi` causes linker errors on the CI fuzz testing job


RossBrunton:
When CFI sanitization is enabled as part of the `build-fuzz-reusable` job, the linker fails when trying to find symbols:

```
[...]

PR #2366, opened 2024-11-21
Disable CFI checks when ASAN is specified

RossBrunton:
This causes linker errors, and presumably isn't meaningful anyway.

Closes: #2323


cmake/helpers.cmake
```diff
@@ -63,6 +63,12 @@ if(CMAKE_SYSTEM_NAME STREQUAL Linux)
     check_cxx_compiler_flag("-fstack-clash-protection" CXX_HAS_FSTACK_CLASH_PROTECTION)
 endif()
 
+if (UR_USE_CFI AND UR_USE_ASAN)
+    message(WARNING "Both UR_USE_CFI and UR_USE_ASAN are ON."
```
aarongreig, line 67: cmake doesn't insert a space or line break between the strings
```suggestion
    message(WARNING "Both UR_USE_CFI and UR_USE_ASAN are ON. "
```

martygrant (approved):

PR #2372, opened 2024-11-22
[Benchmarks] support CPU counter measurements

igchor:
for api_overhead_benchmark_ur

### DIFF
--- a/scripts/benchmarks/benches/compute.py
+++ b/scripts/benchmarks/benches/compute.py
@@ -263,6 +263,7 @@ class StreamMemory(ComputeBenchmark):
             f"--memoryPlacement={self.placement}",
             "--useEvents=0",
             "--contents=Zeros",
+            "--multiplier=1",
         ]
 
 class VectorSum(ComputeBenchmark):
--- a/scripts/benchmarks/benches/compute.py
+++ b/scripts/benchmarks/benches/compute.py
@@ -19,7 +19,7 @@ class ComputeBench(Suite):
         if options.sycl is None:
             return
 
-        repo_path = git_clone(self.directory, "compute-benchmarks-repo", "https://github.com/intel/compute-benchmarks.git", "c80ddec9f0b4905bcbeb0f264f710093dc70340d")
+        repo_path = git_clone(self.directory, "compute-benchmarks-repo", "https://github.com/intel/compute-benchmarks.git", "df38bc342641d7e83fbb4fe764a23d21d734e07b")
         build_path = create_build_path(self.directory, 'compute-benchmarks-build')
 
         configure_command = [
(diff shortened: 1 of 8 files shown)


scripts/benchmarks/benches/compute.py
```diff
@@ -71,17 +71,20 @@ def benchmarks(self) -> list[Benchmark]:
...
+                SubmitKernelURWithTime(self, 1),
+                SubmitKernelURWithCPUCounter(self, 0),
+                SubmitKernelURWithCPUCounter(self, 1),
             ]
 
         return benches
 
 class ComputeBenchmark(Benchmark):
-    def __init__(self, bench, name, test):
+    def __init__(self, bench, name, test, unit="time [us]"):
```
pbalcer, line 83: this looks very odd, please do an "enum" ([example](https://github.com/oneapi-src/unified-runtime/blob/main/scripts/benchmarks/benches/options.py#L4))

scripts/benchmarks/benches/compute.py
```diff
@@ -175,7 +210,7 @@ def __init__(self, bench, ioq, isCopyOnly, source, destination, size):
         self.source = source
         self.destination = destination
         self.size = size
-        super().__init__(bench, "api_overhead_benchmark_sycl", "ExecImmediateCopyQueue")
+        super().__init__(bench, "api_overhead_benchmark_sycl", "ExecImmediateCopyQueue", "time [us]")
```
pbalcer, line 213: You've set a default unit, do you need to set these?

scripts/benchmarks/benches/compute.py
```diff
@@ -148,14 +162,35 @@ def bin_args(self) -> list[str]:
             "--KernelExecTime=1"
         ]
 
-class SubmitKernelUR(ComputeBenchmark):
+class SubmitKernelURWithTime(ComputeBenchmark):
```
pbalcer, line 165: I don't think you need to duplicate these. Simply return two values from the run() with the extra label. In fact, I think all you have to do is remove the if at line 117 and everything will just work without duplicated benchmark classes.

scripts/benchmarks/benches/compute.py
igchor: Not longer needed.

scripts/benchmarks/benches/compute.py
igchor: I also needed to remove unit() function and set it on the result directly. Done.

scripts/benchmarks/benches/compute.py
```diff
@@ -77,6 +77,13 @@ def benchmarks(self) -> list[Benchmark]:
 
         return benches
 
+def parse_unit_type(compute_unit):
+    if "[count]" in compute_unit:
+        return "CPU count"
```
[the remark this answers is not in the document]
igchor: Sure, done.

pbalcer (approved):

PR #2384, opened 2024-11-26
Failing e2e tests no longer kill the whole build

RossBrunton:


.github/workflows/e2e_core.yml
```diff
@@ -190,6 +190,7 @@ jobs:
 
     - name: Run e2e tests
       id: tests
+      continue-on-error: true
```
kbenzie, line 193: This has been tried before, it doesn't do what we want here.

.github/workflows/e2e_core.yml
```diff
@@ -190,6 +190,7 @@ jobs:
...
       run: ninja -C build-e2e check-sycl-e2e
```
kbenzie, line 194: This is what we need.
```suggestion
      run: ninja -C build-e2e check-sycl-e2e || true
```

.github/workflows/e2e_core.yml
kbenzie: i.e. this will still show the jobs as red.

pbalcer (approved):

kbenzie (approved):

PR #2389, opened 2024-11-27
[L0] Set exec info for all L0 kernels in UR kernel

againull:
Currently at urKernelSetExecInfo function a flag is set only for a single L0 kernel handle associated with one of the devices. In multi-device case there might be multiple handles and we have to set the flag for all of them.

Currently this issue causes SYCL "free function kernel" tests to fail in multi-device env because indirect access flag is not set for some of the kernel handles.

intel/llvm PR: https://github.com/intel/llvm/pull/16241

### DIFF
--- a/source/adapters/level_zero/kernel.cpp
+++ b/source/adapters/level_zero/kernel.cpp
@@ -947,35 +947,36 @@ ur_result_t urKernelSetExecInfo(
   std::ignore = PropSize;
   std::ignore = Properties;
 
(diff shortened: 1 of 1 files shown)


againull:
@oneapi-src/unified-runtime-maintain could you please merge this PR.

kbenzie:
I pushed a clang-format fix to get the CI passing.

nrspruit (approved):

PR #2394, opened 2024-11-28
[DevASAN] Do allocation with USM pool to reduce memory overhead

zhaomaosu:
Release mapped physical memory according to its dependency may cause
some problems. So, we decide to use USM pool to do allocation to
reduce memory overhead.

### DIFF
--- a/source/loader/layers/sanitizer/asan/asan_interceptor.cpp
+++ b/source/loader/layers/sanitizer/asan/asan_interceptor.cpp
@@ -36,7 +36,8 @@ AsanInterceptor::~AsanInterceptor() {
     // We must release these objects before releasing adapters, since
     // they may use the adapter in their destructor
     for (const auto &[_, DeviceInfo] : m_DeviceMap) {
-        DeviceInfo->Shadow->Destory();
+        [[maybe_unused]] auto URes = DeviceInfo->Shadow->Destory();
+        assert(URes == UR_RESULT_SUCCESS);
     }
 
     m_Quarantine = nullptr;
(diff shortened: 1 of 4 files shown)


aarongreig:
It might be worth considering that this will raise the bar to entry for supporting dev asan on different platforms in the future, the native_cpu and opencl adapters don't currently have usm pool support and I don't know of any plans to implement it in the near future.

If that is a concern I suggest a fallback path if pooling isn't supported, if it isn't feel free to ignore.

zhaomaosu:
> It might be worth considering that this will raise the bar to entry for supporting dev asan on different platforms in the future, the native_cpu and opencl adapters don't currently have usm pool support and I don't know of any plans to implement it in the near future.
> 
> If that is a concern I suggest a fallback path if pooling isn't supported, if it isn't feel free to ignore.

Thanks for your suggestion. But original goal of this PR is to reduce memory overhead on L0 GPU device, opencl adapter doesn't support usm pool is ok for us now.

AllanZyne (approved):

AllanZyne:
Hi, I have a patch on syclos which needs this PR to merge, so I update my UR tag to this one.
Feel free to link them together.

LLVM: https://github.com/intel/llvm/pull/16279

AllanZyne:
Please update to 3ad235dee9588152a58980bb73c8827a3d03a9e8.

zhaomaosu:
Hi @oneapi-src/unified-runtime-maintain, could you please help review this PR? We need this PR in v0.11.x, thanks.

kbenzie (approved):

Issue #2393, opened 2024-11-27
Level Zero adapter crashes on old L0 loader versions


sarnex:
In intel/llvm, we are seeing that any program that loads the L0 adapter causes a program crash. Specifically, we see it in our [nightly](https://github.com/intel/llvm/actions/runs/12043066294/job/33607577044) testing on OCL CPU (not sure why, maybe we don't install a newer L0 loader in this case)

The crash is [here](https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/adapter.cpp#L389), in the access of the `ZeInitDriversResult` field on `GlobalAdapter`.

The problem is the L0 loader version is `1.18.5` in the crashing case, which is less than `1.19.2`,  so we set `useInitDrivers` to `false` and this never put a real value in the `std::optional` `ZeInitDriversResult` field.

The crash is that we just blindly try to access the contained value of `ZeInitDriversResult` without checking if it has a contained value.

sarnex:
FYI @oneapi-src/unified-runtime-level-zero-write, hopefully we can get a quick fix since this is causing every SYCL nightly to fail.

PR #2396, opened 2024-11-28
[L0] Add nullopt check before init results access

kswiecicki:
ZeInitResult and ZeInitDriversResult could potentially be uninitialized, causing undefined behavior when accessed.

Fixes: https://github.com/oneapi-src/unified-runtime/issues/2393.

### DIFF
--- a/source/adapters/level_zero/adapter.cpp
+++ b/source/adapters/level_zero/adapter.cpp
@@ -256,6 +256,9 @@ Behavior Summary:
 */
 ur_adapter_handle_t_::ur_adapter_handle_t_()
     : logger(logger::get_logger("level_zero")) {
+  ZeInitDriversResult = ZE_RESULT_ERROR_UNINITIALIZED;
+  ZeInitResult = ZE_RESULT_ERROR_UNINITIALIZED;
+  ZesResult = ZE_RESULT_ERROR_UNINITIALIZED;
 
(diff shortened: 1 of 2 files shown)


source/adapters/level_zero/adapter.cpp
```diff
@@ -385,8 +385,10 @@ ur_adapter_handle_t_::ur_adapter_handle_t_()
         }
       }
 
-      if (*GlobalAdapter->ZeInitResult == ZE_RESULT_SUCCESS ||
-          *GlobalAdapter->ZeInitDriversResult == ZE_RESULT_SUCCESS) {
+      if ((GlobalAdapter->ZeInitResult != std::nullopt &&
```
pbalcer, line 388: Originally we used `GlobalAdapter->ZeResult` being empty to check whether initialization already happened. These other variables don't seem to be serving the same function. My suggestion would to not use std::optional for them, and set them initially to `ZE_RESULT_ERROR_UNINITIALIZED` or `ZE_RESULT_ERROR_UNKNOWN` or something.

pbalcer: or, now that I thought about for a second, we can just use bools for these. I don't think we care about specific values apart from "SUCCESS".

pbalcer (approved):

PR #2397, opened 2024-11-28
[NFC] Add a 2 minute per-test timeout in e2e tests

RossBrunton:
Some tests hang, this ensures that the e2e run ends in a timely amount
of time.


RossBrunton:
Since the CUDA test is now disabled, closing this unless L0/OpenCL turn out to also be super slow.

PR #2404, opened 2024-12-02
Add UR_ADAPTER_INFO_VERSION query

kbenzie:
In order to support in-code management of known failures we need a query to differentiate between the two versions of the Level Zero adapter. This patch adds the `UR_ADAPTER_INFO_VERSION` query which returns a `uint32_t` which is set to return `1` for all adapters except the Level Zero adapter when being compiled in V2 mode.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -967,6 +967,9 @@ typedef enum ur_adapter_info_t {
                                          ///< The reference count returned should be considered immediately stale.
                                          ///< It is unsuitable for general use in applications. This feature is
                                          ///< provided for identifying memory leaks.
+    UR_ADAPTER_INFO_VERSION = 2,         ///< [uint32_t] Specifies the adapter version, initial value of 1 and
+                                         ///< incremented unpon major changes, e.g. when multiple versions of an
+                                         ///< adapter may exist in parallel.
     /// @cond
     UR_ADAPTER_INFO_FORCE_UINT32 = 0x7fffffff
     /// @endcond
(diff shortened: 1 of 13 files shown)


pbalcer (approved):

martygrant (approved):

include/ur_api.h
```diff
@@ -963,7 +963,8 @@ urAdapterGetLastError(
 typedef enum ur_adapter_info_t {
     UR_ADAPTER_INFO_BACKEND = 0,         ///< [::ur_adapter_backend_t] Identifies the native backend supported by
                                          ///< the adapter.
-    UR_ADAPTER_INFO_REFERENCE_COUNT = 1, ///< [uint32_t] Reference count of the adapter.
+    UR_ADAPTER_INFO_VERSION = 1,         ///< [uint32_t] Specifies the adapter version, defaults to 1.
```
aarongreig, line 966: maybe this should come after REFERENCE_COUNT in the yml to preserve the old enum value

frasercrmck (approved):

include/ur_api.h
RossBrunton, line 966: What does "defaults to 1" actually mean here? It seems all adapters set it explicitly.

PietroGhg (approved):

include/ur_api.h
RossBrunton: "Default" implies that it is set automatically if you don't set it, which is not the case here.

Perhaps "Starts at 1 and should be incremented on every major adapter release" would be clearer?

kbenzie: Updated.

PR #2410, opened 2024-12-02
[L0] Fix cached and evicted timestamp recordings

steffenlarsen:
This commit fixes two issues with the level zero implementation of timestamp recording events:
* Events allocated for timestamp recordings may have been previously used, which may lead the implementation to think that the recordings of the old timestamp are right. The implementation will now reset the value of it.
* To avoid cases where timestamp recordings could conflict in the recordings buffer, unfinished recordings of dead events are now moved to another map, to be evicted fully on queue synchronization or death.

### DIFF
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -574,8 +574,7 @@ ur_result_t urEventGetProfilingInfo(
 
       // End time needs to be adjusted for resolution and valid bits.
       uint64_t ContextEndTime =
-          (EndTimeRecording.RecordEventEndTimestamp & TimestampMaxValue) *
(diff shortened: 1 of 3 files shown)


nrspruit (approved):

nrspruit:
Hello @steffenlarsen , can you post an intel/llvm draft PR pointing to your UR change? Otherwise it looks good to me.

steffenlarsen:
> Hello @steffenlarsen , can you post an intel/llvm draft PR pointing to your UR change? Otherwise it looks good to me.

Oh, sorry! Evening brain let it slip my mind. https://github.com/intel/llvm/pull/16233

PR #2414, opened 2024-12-04
[DeviceSanitizer] Workaround for CFI errors

yingcong-wu:
`fun:*__aligned_buffer*` is for a common `make_shared` false positive.

`src:*llvm/include/llvm/Support/raw_ostream.h` is for sanitizer symbolizer, we think it is because we don't meet the requirement of enabling https://clang.llvm.org/docs/ControlFlowIntegrity.html#non-virtual-member-function-call-checking , and we will disable it for good.


AllanZyne (approved):

zhaomaosu (approved):
lgtm

PR #2415, opened 2024-12-04
[DeviceASAN] Fix ASAN with kernel assert

AllanZyne:
LLVM: https://github.com/intel/llvm/pull/16256

Even when the kernel is not instrumented, we still need to execute the same routine just like instrumented kernel.
That's because some of arguments have been intercepted by the sanitizer layer and are no longer compatible with the native level zero's value. For example, membuffer.


zhaomaosu (approved):
lgtm

AllanZyne:
Hi @oneapi-src/unified-runtime-maintain, please review and add this to 2025.1.
Thanks!

source/loader/layers/sanitizer/asan/asan_ddi.cpp
```diff
@@ -1443,8 +1422,8 @@ __urdlllocal ur_result_t UR_APICALL urKernelSetArgValue(
     std::shared_ptr<KernelInfo> KernelInfo;
     if (argSize == sizeof(ur_mem_handle_t) &&
         (MemBuffer = getAsanInterceptor()->getMemBuffer(
-             *ur_cast<const ur_mem_handle_t *>(pArgValue))) &&
-        (KernelInfo = getAsanInterceptor()->getKernelInfo(hKernel))) {
+             *ur_cast<const ur_mem_handle_t *>(pArgValue)))) {
+        auto KernelInfo = getAsanInterceptor()->getKernelInfo(hKernel);
```
pbalcer, line 1426: this is aliasing an already existing variable that appears to be otherwise unused:
`std::shared_ptr<KernelInfo> KernelInfo;` << I think you should be able to just remove this.

source/loader/layers/sanitizer/asan/asan_ddi.cpp
```diff
@@ -1474,8 +1453,8 @@ __urdlllocal ur_result_t UR_APICALL urKernelSetArgMemObj(
 
     std::shared_ptr<MemBuffer> MemBuffer;
     std::shared_ptr<KernelInfo> KernelInfo;
-    if ((MemBuffer = getAsanInterceptor()->getMemBuffer(hArgValue)) &&
-        (KernelInfo = getAsanInterceptor()->getKernelInfo(hKernel))) {
+    if ((MemBuffer = getAsanInterceptor()->getMemBuffer(hArgValue))) {
+        auto KernelInfo = getAsanInterceptor()->getKernelInfo(hKernel);
```
pbalcer, line 1457: same.

source/loader/layers/sanitizer/asan/asan_ddi.cpp
```diff
@@ -1542,8 +1522,8 @@ __urdlllocal ur_result_t UR_APICALL urKernelSetArgPointer(
         pArgValue);
 
     std::shared_ptr<KernelInfo> KI;
-    if (getAsanInterceptor()->getOptions().DetectKernelArguments &&
-        (KI = getAsanInterceptor()->getKernelInfo(hKernel))) {
+    if (getAsanInterceptor()->getOptions().DetectKernelArguments) {
+        auto KI = getAsanInterceptor()->getKernelInfo(hKernel);
```
pbalcer, line 1526: same issue.

pbalcer (commented):
just a few nits, but I do have a general comment - this looks like a pretty big change in logic, with the not instrumented kernels now executing more code that was previously short-circuited with `if (not instrumented) return KernelLaunch(...)`. Which might be fine, but there's no explanation in the commit description.

Do you have SYCL tests that exercise this path with non-instrumented kernels?

AllanZyne:
> just a few nits, but I do have a general comment - this looks like a pretty big change in logic, with the not instrumented kernels now executing more code that was previously short-circuited with `if (not instrumented) return KernelLaunch(...)`. Which might be fine, but there's no explanation in the commit description.

Ok, I will add explanation.

> Do you have SYCL tests that exercise this path with non-instrumented kernels?

Thanks for reminder, I enhanced the kernel-filter test.

AllanZyne:
Hi @pbalcer, the commit description is updated.

pbalcer (approved):

PR #2416, opened 2024-12-04
[L0] Disabling Driver In Order Lists by default

nrspruit:
- Due to L0 Driver issues related to performance using driver in order lists, the feature is being disabled for the time being.
- Once issues with the L0 Drivers implementing this feature are resolved, then this feature will be re-enabled.

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -598,16 +598,15 @@ ur_result_t createMainCommandList(ur_context_handle_t Context,
  */
 bool canBeInOrder(ur_context_handle_t Context,
                   const ur_exp_command_buffer_desc_t *CommandBufferDesc) {
-  std::ignore = Context;
-  std::ignore = CommandBufferDesc;
+  const char *UrRet = std::getenv("UR_L0_USE_DRIVER_INORDER_LISTS");
   // In-order command-lists are not available in old driver version.
-  // bool CompatibleDriver =
-  // Context->getPlatform()->isDriverVersionNewerOrSimilar(
-  //     1, 3, L0_DRIVER_INORDER_MIN_VERSION);
-  // return CompatibleDriver
-  //            ? (CommandBufferDesc ? CommandBufferDesc->isInOrder : false)
-  //            : false;
-  return false;
+  bool DriverInOrderRequested = UrRet ? std::atoi(UrRet) != 0 : false;
+  bool CompatibleDriver = Context->getPlatform()->isDriverVersionNewerOrSimilar(
+      1, 3, L0_DRIVER_INORDER_MIN_VERSION);
+  bool CanUseDriverInOrderLists = CompatibleDriver && DriverInOrderRequested;
+  return CanUseDriverInOrderLists
+             ? (CommandBufferDesc ? CommandBufferDesc->isInOrder : false)
+             : false;
 }
 
 /**
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -598,12 +598,16 @@ ur_result_t createMainCommandList(ur_context_handle_t Context,
  */
 bool canBeInOrder(ur_context_handle_t Context,
                   const ur_exp_command_buffer_desc_t *CommandBufferDesc) {
+  std::ignore = Context;
+  std::ignore = CommandBufferDesc;
   // In-order command-lists are not available in old driver version.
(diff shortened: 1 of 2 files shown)


pbalcer (approved):

EwanC:
In command-buffer code we use `ZE_COMMAND_LIST_FLAG_IN_ORDER` command-lists without a precondition on this `useDriverInOrderLists()` check (https://github.com/oneapi-src/unified-runtime/blob/3fdf7e3998b029fbf6cdb584e552a71a678f3146/source/adapters/level_zero/command_buffer.cpp#L575). Is that something that will suffer from the problem that's motivating this change?

Bensuo:
> In command-buffer code we use `ZE_COMMAND_LIST_FLAG_IN_ORDER` command-lists without a precondition on this `useDriverInOrderLists()` check (it's based on a flag passed from SYCL-RT)
> 
> https://github.com/oneapi-src/unified-runtime/blob/3fdf7e3998b029fbf6cdb584e552a71a678f3146/source/adapters/level_zero/command_buffer.cpp#L575
> Is that something that will suffer from the problem that's motivating this change?

We do the same check [here](https://github.com/oneapi-src/unified-runtime/blob/3fdf7e3998b029fbf6cdb584e552a71a678f3146/source/adapters/level_zero/command_buffer.cpp#L599) that's removed in this PR to confirm whether we can use inorder command lists or not, so I would guess that should also be changed.

nrspruit:
> In command-buffer code we use `ZE_COMMAND_LIST_FLAG_IN_ORDER` command-lists without a precondition on this `useDriverInOrderLists()` check (it's based on a flag passed from SYCL-RT)
> 
> https://github.com/oneapi-src/unified-runtime/blob/3fdf7e3998b029fbf6cdb584e552a71a678f3146/source/adapters/level_zero/command_buffer.cpp#L575
> 
> . Is that something that will suffer from the problem that's motivating this change?

If this is based on a flag passed from SYCL RT, then it should not be changed. This is about changing the default behavior to use standard lists and not in order by default.

nrspruit:
Ah, I missed the "canBeInOrder" let me change that too.

nrspruit:
@EwanC , please check the command buffer changes, now both are disabled for now.

nrspruit:
> @EwanC , please check the command buffer changes, now both are disabled for now.

had a problem in my patch, please re-check, unfortunately this will disable driver in order in command buffer for now until the driver issues can be resolved.

nrspruit:
@EwanC / @Bensuo please re-review so we can get this change in as soon as possible to create the release branch.

This updated PR will disable in both common code and command buffers.

Bensuo:
> @EwanC / @Bensuo please re-review so we can get this change in as soon as possible to create the release branch.
> 
> This updated PR will disable in both common code and command buffers.



> @EwanC / @Bensuo please re-review so we can get this change in as soon as possible to create the release branch.
> 
> This updated PR will disable in both common code and command buffers.

I'm not sure offhand why we don't have this already, but could you add the check for `UR_L0_USE_DRIVER_INORDER_LISTS` to `canBeInOrder()` so that we can still enable them if needed? Since that's still in the general implementation I think that should be fine for now.

nrspruit:
> > @EwanC / @Bensuo please re-review so we can get this change in as soon as possible to create the release branch.
> > This updated PR will disable in both common code and command buffers.
> 
> I'm not sure offhand why we don't have this already, but could you add the check for `UR_L0_USE_DRIVER_INORDER_LISTS` to `canBeInOrder()` so that we can still enable them if needed? Since that's still in the general implementation I think that should be fine for now.

Ok, let me add that in such that in both cases it is "possible" to use the feature.

Bensuo (approved):
LGTM, thanks for making those changes!

nrspruit:
> > > @EwanC / @Bensuo please re-review so we can get this change in as soon as possible to create the release branch.
> > > This updated PR will disable in both common code and command buffers.
> > 
> > 
> > I'm not sure offhand why we don't have this already, but could you add the check for `UR_L0_USE_DRIVER_INORDER_LISTS` to `canBeInOrder()` so that we can still enable them if needed? Since that's still in the general implementation I think that should be fine for now.
> 
> Ok, let me add that in such that in both cases it is "possible" to use the feature.

@EwanC , I have updated the check in command buffer such that it also checks the env for driver in order. This will enable both paths to work now.

nrspruit:
This change is ready to merge, but we are fighting CI issues with a very long delay in execution.

EwanC:
Sorry @nrspruit I logged off for the day after leaving that comment so didn't see your replies, but Ben has reviewed this on behalf of the command-buffer group and happy with the final patch.

PR #2425, opened 2024-12-05
[CUDA] Fix potential issue with command buffer fills on CUDA

Bensuo:
- Fix a potential issue where decomposed fill nodes for large patterns would overwrite external event dependencies provided by the user when stored in a command handle, this could lead to an error when events were updatd.
- Also store decomposed nodes in fill command handles for future use when updating.
- Add missing event_sync tests with and without event update for large pattern fills (> 4 bytes)

### DIFF
--- a/source/adapters/cuda/command_buffer.hpp
+++ b/source/adapters/cuda/command_buffer.hpp
@@ -172,12 +172,19 @@ struct usm_memcpy_command_handle : ur_exp_command_buffer_command_handle_t_ {
 struct usm_fill_command_handle : ur_exp_command_buffer_command_handle_t_ {
   usm_fill_command_handle(ur_exp_command_buffer_handle_t CommandBuffer,
                           CUgraphNode Node, CUgraphNode SignalNode,
-                          const std::vector<CUgraphNode> &WaitNodes)
+                          const std::vector<CUgraphNode> &WaitNodes,
+                          const std::vector<CUgraphNode> &DecomposedNodes = {})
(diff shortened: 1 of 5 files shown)


source/adapters/cuda/command_buffer.hpp
```diff
@@ -172,12 +172,18 @@ struct usm_memcpy_command_handle : ur_exp_command_buffer_command_handle_t_ {
...
-                                                WaitNodes) {}
+                                                WaitNodes),
+        DecomposedNodes(std::move(DecomposedNodes)) {}
   CommandType getCommandType() const noexcept override {
     return CommandType::USMFill;
   }
+
+  // If this fill command was decomposed into multiple nodes, this vector
+  // contains all of those nodes in the order they were added to the graph.
+  std::vector<CUgraphNode> DecomposedNodes;
```
EwanC, line 186: Do we need to store this, do we ever reference this member?

Bensuo: Yes I figured since they are created as a separate vector now and we will need them in future to update a decomposed fill command that it made sense to just store them. Since the CUDA adapter already has some future proofing for updating different types of commands it seemed easy to just store them here for now.

I've updated the comments for these members to reflect this.

Bensuo:
DPC++ PR: https://github.com/intel/llvm/pull/16305

martygrant:
@Bensuo got 2 failing jobs, might need an update to the match files possibly?

martygrant:
@Bensuo no problem, cheers!

martygrant:
Pulled in with https://github.com/intel/llvm/pull/16410

PR #2426, opened 2024-12-05
[UMF] Bump version and update DisjointPool params handling

lukaszstolarczuk:
using UMF with PR https://github.com/oneapi-src/unified-memory-framework/pull/969

### DIFF
--- a/source/common/umf_pools/disjoint_pool_config_parser.cpp
+++ b/source/common/umf_pools/disjoint_pool_config_parser.cpp
@@ -1,5 +1,7 @@
 //===--- disjoint_pool_config_parser.cpp -configuration for USM memory pool-==//
 //
+// Copyright (C) 2023-2024 Intel Corporation
+//
 // Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
--- a/source/common/umf_pools/disjoint_pool_config_parser.cpp
+++ b/source/common/umf_pools/disjoint_pool_config_parser.cpp
@@ -9,11 +9,23 @@
 //===----------------------------------------------------------------------===//
 
 #include "disjoint_pool_config_parser.hpp"
+#include "logger/ur_logger.hpp"
 
+#include <cassert>
 #include <iomanip>
 #include <iostream>
 #include <limits>
 #include <string>
+
+// Don't expect any failures in setting params; the only 2 reasons for failure
+// are not enough memory for allocating 'name' or a NULL params handle.
+#define UMF_CALL(RET)                                                          \
+    if (RET != UMF_RESULT_SUCCESS) {                                           \
+        if (RET == UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY) {                      \
+            throw std::runtime_error("DisjointPool name allocation failed");   \
+        }                                                                      \
+        throw std::runtime_error("DisjointPool setting params failed");        \
+    }
 
 namespace usm {
 constexpr auto operator""_B(unsigned long long x) -> size_t { return x; }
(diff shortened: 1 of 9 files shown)


source/common/umf_pools/disjoint_pool_config_parser.cpp
```diff
@@ -25,45 +34,81 @@ constexpr auto operator""_GB(unsigned long long x) -> size_t {
...
+        CheckConfigRet(ret);
+
+        Config = params;
     }
 
-    Configs[DisjointPoolMemType::Host].Name = "Host";
-    Configs[DisjointPoolMemType::Device].Name = "Device";
-    Configs[DisjointPoolMemType::Shared].Name = "Shared";
-    Configs[DisjointPoolMemType::SharedReadOnly].Name = "SharedReadOnly";
+    ret = umfDisjointPoolParamsSetName(Configs[DisjointPoolMemType::Host], "Host");
```
igchor, line 59: nit: you can just do `CheckConfigRet(umfDisjointPoolParamsSetName(Configs[DisjointPoolMemType::Host], "Host"))`.

Also, I would rename CheckConfigRet to `UMF_CALL` or something like that so it's shorter and more consistent with UR_CALL/ZE_CALL

source/common/umf_pools/disjoint_pool_config_parser.cpp
```diff
@@ -118,34 +163,46 @@ DisjointPoolAllConfigs parseDisjointPoolConfig(const std::string &config,
...
         }
 
-        bool More = ParamParser(Params, AllConfigs.Configs[LM].MaxPoolableSize,
-                                ParamWasSet);
+        bool More = ParamParser(Params, TmpValue, ParamWasSet);
         if (ParamWasSet && memType == DisjointPoolMemType::All) {
             for (auto &Config : AllConfigs.Configs) {
-                Config.MaxPoolableSize = AllConfigs.Configs[LM].MaxPoolableSize;
+                ret = umfDisjointPoolParamsSetMaxPoolableSize(Config, TmpValue);
+                CheckConfigRet(ret);
```
igchor, line 177: This return will only exit from this lambda, not from the entire parseDisjointPoolConfig function.

source/common/umf_pools/disjoint_pool_config_parser.cpp
```diff
@@ -222,8 +279,8 @@ DisjointPoolAllConfigs parseDisjointPoolConfig(const std::string &config,
         umfDisjointPoolSharedLimitsDestroy);
 
     for (auto &Config : AllConfigs.Configs) {
-        Config.SharedLimits = AllConfigs.limits.get();
-        Config.PoolTrace = trace;
+        umfDisjointPoolParamsSetSharedLimits(Config, AllConfigs.limits.get());
+        umfDisjointPoolParamsSetTrace(Config, trace);
```
igchor, line 283: no return check?

source/common/umf_pools/disjoint_pool_config_parser.cpp
```diff
@@ -1,18 +1,27 @@
...
 
 #include <iomanip>
 #include <iostream>
 #include <limits>
 #include <string>
 
+#define CheckConfigRet(umf_ret)                  \
+    if (umf_ret != UMF_RESULT_SUCCESS) {                                    \
+        logger::error("DisjointPool params failed");                        \
+        return;                                                             \
```
igchor, line 22: Should we just ignore the error? If any of the setParams function fail and we will just exit the function we might use incomplete config later on.

For calls in DisjointPoolAllConfigs::DisjointPoolAllConfigs I think we should have an assert and for calls in parseDisjointPoolConfig if there is any error we should catch it, log it, and probably revert all configs to the default settings.

source/common/umf_pools/disjoint_pool_config_parser.cpp
lukaszstolarczuk: Done

source/common/umf_pools/disjoint_pool_config_parser.cpp
lukaszstolarczuk: Done

source/common/umf_pools/disjoint_pool_config_parser.cpp
lukaszstolarczuk: Done

source/common/umf_pools/disjoint_pool_config_parser.cpp
lukaszstolarczuk: Done

source/common/umf_pools/disjoint_pool_config_parser.cpp
```diff
@@ -234,39 +305,39 @@ DisjointPoolAllConfigs parseDisjointPoolConfig(const std::string &config,
...
-        << std::setw(15) << "Capacity" << std::setw(12)
-        << AllConfigs.Configs[DisjointPoolMemType::Host].Capacity
-        << std::setw(12)
-        << AllConfigs.Configs[DisjointPoolMemType::Device].Capacity
-        << std::setw(12)
-        << AllConfigs.Configs[DisjointPoolMemType::Shared].Capacity
-        << std::setw(12)
-        << AllConfigs.Configs[DisjointPoolMemType::SharedReadOnly].Capacity
-        << std::endl;
+    // TODO: fixme, accessing config values directly is no longer allowed - API's changed
```
lukaszstolarczuk, line 308: I tried to do this nicely, but I'm not sure if I can... @igchor tried to make the pre-generated content in ctor, but it would have to be updated somehow in the parsing function... If this is not crucial I could update that in a sep. PR...?

lukaszstolarczuk:
replaced with https://github.com/oneapi-src/unified-runtime/pull/2436

PR #2429, opened 2024-12-05
[L0] Fixed getImmCmdList returning cmdlist with wrong properties

winstonzhang-intel:
When using counterbased events, sometimes when retrieving immediate commandlist, we would return a commandlist that is does not have the ZE_COMMAND_QUEUE_FLAG_IN_ORDER flag set which causes an invalid argument when trying to use that commandlist later. 
SYCL/LLVM: https://github.com/intel/llvm/pull/16291


nrspruit (approved):
LGTM, thanks for the patch!

igchor (approved):

PR #2430, opened 2024-12-06
[UR] Fix correct usage of In Order sync list given counting events

nrspruit:
- Fixes a race condition when the wait events are counting events when executing the context's synchronous immediate command list.
- If a counting event is used in a non immediate command list, then an error will be returned.

### DIFF
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -319,9 +319,16 @@ ur_result_t ur_context_handle_t_::initialize() {
 
   ZeCommandQueueDesc.index = 0;
   ZeCommandQueueDesc.mode = ZE_COMMAND_QUEUE_MODE_SYNCHRONOUS;
+  if (Device->useDriverInOrderLists() &&
+      Device->useDriverCounterBasedEvents()) {
+    logger::debug(
+        "L0 Synchronous Immediate Command List needed with In Order property.");
+    ZeCommandQueueDesc.flags |= ZE_COMMAND_LIST_FLAG_IN_ORDER;
+  }
   ZE2UR_CALL(
       zeCommandListCreateImmediate,
       (ZeContext, Device->ZeDevice, &ZeCommandQueueDesc, &ZeCommandListInit));
+
   return UR_RESULT_SUCCESS;
 }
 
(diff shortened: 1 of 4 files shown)


winstonzhang-intel (approved):

source/adapters/level_zero/memory.cpp
```diff
@@ -2181,9 +2187,14 @@ ur_result_t _ur_buffer::getZeHandle(char *&ZeHandle, access_mode_t AccessMode,
         }
       }
       if (waitlist.Length > 0) {
-        ZE2UR_CALL(zeCommandListAppendWaitOnEvents,
-                   (UrContext->ZeCommandListInit, waitlist.Length,
-                    waitlist.ZeEventList));
+        ze_command_list_handle_t ZeCommandListForInit =
+            UrContext->ZeCommandListInit;
+        if (InOrderListRequired) {
+          ZeCommandListForInit = UrContext->ZeCommandListInitInOrder;
```
pbalcer, line 2193: can we just create the correct type of cmdlist in the first place in `UrContext->ZeCommandListInit`? Will there ever be a situation where two different queues in a single context require different types of events (thus command lists)?

nrspruit: the user can request an out of order queue so the envs would not be taken so we would have an out of order queue, then an in order immediate command list being used for init.

There is no "correct" in this case. They would be different in this case anyways.

If we want the init list to always be determined by the defaults of the two envs, then  we can change this such that is the envs are set, then the init command list is also in order.

I was worried about the impact of performance for these calls, but if that is not a concern, then this can be made that way.

nrspruit: @pbalcer , please re-review, I decided to just simplify to the single list when both envs are set. that would then be a catch all. Less configurable to swap inbetween, but this is cleaner.

winstonzhang-intel (approved):

source/adapters/level_zero/memory.cpp
pbalcer: lgtm, thanks for making the changes!

martygrant:
Created intel/llvm PR for this https://github.com/intel/llvm/pull/16327

PR #2431, opened 2024-12-06
[Tests] replace zeCallsMap with zelTracer

igchor:


igchor:
For some reason multi_device_event_cache_test.cpp segfaults when I removed linking with zeCallsMap... Need to investigate

igchor:
@nrspruit could you please take a look? I had to modify the multi_device_event_cache_tests.cpp before I was getting a Segfault. I think that what the test was trying to do (use event created on one context associated with device1, on a queue associated with a different device is not allowed) but maybe I'm missing something.

Also, the GivenMultiDeviceWithQueuePerDeviceThenMultiDeviceEventIsCreated test is failing on my machine (event without the changes on this PR) - the number of zeCommandListAppendWaitOnEvents is now 2 instead of the expected 3. Is this correct?

nrspruit (approved):
Wonderful! Glad you are using the tracing layer for this, looks good to me!

nrspruit:
> zeCommandListAppendWaitOnEvents
the 2 vs 3 is because of no in order lists. there was a multi device workaround for driver in order lists before, which is now not applying that additional wait.

test/adapters/level_zero/multi_device_event_cache_tests.cpp
```diff
@@ -40,12 +54,12 @@ TEST_F(urMultiQueueMultiDeviceEventCacheTest,
...
     uur::raii::Context context1 = nullptr;
-    ASSERT_SUCCESS(
-        urContextCreate(1, &sub_devices[0], nullptr, context1.ptr()));
+    ASSERT_SUCCESS(urContextCreate(sub_devices.size(), &sub_devices[0], nullptr,
+                                   context1.ptr()));
     ASSERT_NE(nullptr, context1);
     uur::raii::Context context2 = nullptr;
-    ASSERT_SUCCESS(
-        urContextCreate(1, &sub_devices[1], nullptr, context2.ptr()));
+    ASSERT_SUCCESS(urContextCreate(sub_devices.size(), &sub_devices[0], nullptr,
```
nrspruit, line 61: Just noticed this, So, you meant to change this to not use the second sub device?

nrspruit: line 67 would then be creating the queue on context 2 which is on sub device 0, with the queue on sub device 1

igchor: We're still creating the second queue on the second context, and the second sub_device:
```
ASSERT_SUCCESS(urQueueCreate(context2, sub_devices[1], 0, &queue2));
```
The difference is just that both contexts now contain all subdevices. Alternative would be to just use a single context but I don't know if that matches the use-case this test was supposed to validate.

nrspruit: ahh, I understand, I missed the count, then it looks good to me.

pbalcer (approved):

PR #2441, opened 2024-12-10
[L0] Remove zero size check from getImageRegionHelper

wenju-he:
A SYCL bindless image test, that has ext_oneapi_copy call with copy region width = 0, fails in UR level-zero adapter due to the check.

On the other hand, the test can run successfully on cuda adapter. In order to make test behavior consistent across different adapters, this PR leaves the check to the level-zero backend and the backend is supposed to return success for this case.

### DIFF
--- a/source/adapters/level_zero/image.cpp
+++ b/source/adapters/level_zero/image.cpp
@@ -578,8 +578,6 @@ ur_result_t getImageRegionHelper(ze_image_desc_t ZeImageDesc,
                 (ZeImageDesc.type == ZE_IMAGE_TYPE_3D),
             UR_RESULT_ERROR_INVALID_VALUE);
 
-  UR_ASSERT(Region->width && Region->height && Region->depth,
-            UR_RESULT_ERROR_INVALID_VALUE);
   UR_ASSERT(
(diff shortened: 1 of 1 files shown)


przemektmalon:
Hi @wenju-he, apologies for the late response, and thanks for bringing this up.

Do you know which test fails due to the copy region width being `0`?

I'm thinking that catching of this type of incorrect input should happen within the SYCL RT, and we should throw an exception there, before we get to the UR adapter code.

wenju-he:
Thanks @przemektmalon for the review.

It is a test migrated from https://github.com/NVIDIA/cuda-samples/tree/master/Samples/0_Introduction/simpleTexture to sycl bindless image by SYCLomatic. The SYCL bindless image test has run fail on intel L0 backend but can run pass on cuda backend.
The fail expose two issues:
1. SYCLomatic should not generate a copy of width 0. This is already fixed in SYCLomatic.
2. L0 backend shouldn't fail when width is 0, in order to align with cuda backend behavior that it is ok to have a width of 0.

The purpose of this PR is to align with cuda behavior.
I agree with you that we can also reject the case of width 0 in sycl runtime.

PR #2452, opened 2024-12-12
[L0] Fixed event leak when outevent is given and is completed

winstonzhang-intel:
During urEnqueueEventsWait, when an OutEvent is given we do not properly clean up after the event is completed. This patch addresses that. 

llvm/sycl: https://github.com/intel/llvm/pull/16339


source/adapters/level_zero/event.cpp
```diff
@@ -145,6 +145,10 @@ ur_result_t urEnqueueEventsWait(
     std::unique_lock<ur_shared_mutex> Lock(Queue->Mutex);
     resetCommandLists(Queue);
   }
+  if (OutEvent && (*OutEvent)->Completed) {
```
nrspruit, line 148: Can you add a comment about why this is needed and why this does not get cleaned up later?

nrspruit:
looks like this causes a segfault in the multi device tests:

    Start 29: test-adapter-level_zero_ze_calls
2/4 Test #29: test-adapter-level_zero_ze_calls ......***Exception: SegFault  0.23 sec
ZE_LOADER_DEBUG_TRACE:zeInitDrivers called first, but not supported by driver, returning uninitialized.
Selected platform: [LEVEL_ZERO] Intel(R) oneAPI Unified Runtime over Level-Zero
[==========] Running 48 tests from 1 test suite.

is it possible the out event is being freed twice?

nrspruit (approved):
thanks, LGTM

PR #2462, opened 2024-12-13
Improve specification of command-buffer update errors

EwanC:
The `UR_DEVICE_INFO_COMMAND_BUFFER_UPDATE_SUPPORT_EXP` query no longer exists, so error defined for trying to make an updatable command-buffer when a device doesn't support it needs reworded to say if no capabilities are supported.

Additionally, for each individual update capability, specify that an error is thrown if update is attempted using that characteristic.

UR CTS added to verify the above clarifications.

DPC++ PR https://github.com/intel/llvm/pull/16383

### DIFF
--- a/scripts/core/exp-command-buffer.yml
+++ b/scripts/core/exp-command-buffer.yml
@@ -1206,6 +1206,7 @@ returns:
     - $X_RESULT_ERROR_UNSUPPORTED_FEATURE:
         - "If $X_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_KERNEL_ARGUMENTS is not supported by the device, but any of `pUpdateKernelLaunch->numNewMemObjArgs`, `pUpdateKernelLaunch->numNewPointerArgs`, or `pUpdateKernelLaunch->numNewValueArgs` are not zero."
         - "If $X_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_LOCAL_WORK_SIZE is not supported by the device but `pUpdateKernelLaunch->pNewLocalWorkSize` is not nullptr."
+        - "If $X_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_LOCAL_WORK_SIZE is not supported by the device but `pUpdateKernelLaunch->pNewLocalWorkSize` is nullptr and `pUpdateKernelLaunch->pNewGlobalWorkSize` is not nullptr."
         - "If $X_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_GLOBAL_WORK_SIZE is not supported by the device but `pUpdateKernelLaunch->pNewGlobalWorkSize` is not nullptr"
         - "If $X_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_GLOBAL_WORK_OFFSET is not supported by the device but `pUpdateKernelLaunch->pNewGlobalWorkOffset` is not nullptr."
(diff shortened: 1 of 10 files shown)


kbenzie (approved):

scripts/core/exp-command-buffer.yml
```diff
@@ -1203,11 +1204,15 @@ params:
       desc: "[in] Struct defining how the kernel command is to be updated."
 returns:
     - $X_RESULT_ERROR_UNSUPPORTED_FEATURE:
-        - "If update functionality is not supported by the device."
+        - "If $X_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_KERNEL_ARGUMENTS is not supported by the device, but any of `pUpdateKernelLaunch->numNewMemObjArgs`, `pUpdateKernelLaunch->numNewPointerArgs`, or `pUpdateKernelLaunch->numNewValueArgs` are not zero."
+        - "If $X_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_LOCAL_WORK_SIZE is not supported by the device but `pUpdateKernelLaunch->pNewLocalWorkSize` is not nullptr."
```
Bensuo, line 1208: In the update descriptor wording we have the following:

>  If `pNewGlobalWorkSize` is set and `pNewLocalWorkSize` is nullptr,
        then the runtime implementation will choose the local work size.

If the device doesn't support `UR_DEVICE_COMMAND_BUFFER_UPDATE_CAPABILITY_FLAG_LOCAL_WORK_SIZE` are we just assuming that the implementation won't generate a new local work size here (which may be different from the previous value) or that it would do that but throw an error because updating the local size is not possible?

If the former we should probably change that quoted wording to reflect that it is not guaranteed to generate a new local size if not supported, or document the latter error here.

test/conformance/exp_command_buffer/update/invalid_update.cpp
```diff
@@ -224,3 +224,254 @@ TEST_P(InvalidUpdateTest, InvalidDimensions) {
         UR_RESULT_ERROR_INVALID_VALUE,
         urCommandBufferUpdateKernelLaunchExp(command_handle, &update_desc));
 }
+
+// Tests that an error is thrown when trying to update a kernel capability
+// that isn't supported.
+struct urInvalidUpdateCommandBufferExpExecutionTest
```
Bensuo, line 230: Nit but is there a reason for using the `ur` prefix here for the fixture name? This differs from all our other tests so would prefer consistency if possible.

scripts/core/exp-command-buffer.yml
EwanC: Good shout, I went with an error because the old local size might not be compatable with the new global size if we can't update it under the hood.

source/adapters/opencl/command_buffer.cpp
EwanC, line 1: Putting back into draft because I think we need more granular checking in this adapter for the unsupported cases in update()

source/adapters/opencl/command_buffer.cpp
```diff
@@ -540,18 +540,72 @@ void updateKernelArgs(std::vector<cl_mutable_dispatch_arg_khr> &CLArgs,
...
 
+ur_result_t validateCommandDesc(
+    ur_exp_command_buffer_command_handle_t Command,
+    const ur_exp_command_buffer_update_kernel_launch_desc_t *UpdateDesc) {
+  // Kernel handle updates are not yet supported.
+  if (UpdateDesc->hNewKernel && UpdateDesc->hNewKernel != Command->Kernel) {
+    return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+  }
+
+  // Error if work-dim has change but a new global size/offset hasn't been set
```
Bensuo, line 551:
```suggestion
  // Error if work-dim has changed but a new global size/offset hasn't been set
```

Bensuo (approved):

PR #2474, opened 2024-12-17
[L0] Fix use after free with Module build strings

nrspruit:
- Address Sanitizer HASAN fails with temporary strings even after last use after free patch. This change resolves all use after free of these strings satifsying the address sanitzer checks with HASAN.


nrspruit:
The CI in the intel/llvm patch is working, the multi device CI is hung until the systems are rebooted.

PR #2476, opened 2024-12-18
[L0] Fix Event Memory Leak due to no destroy on delete

nrspruit:
- Given Internal Event that is not cached we must destroy the ze event
  at handle destroy. This can only occur if the associated queue does
not have discard events enabled.
- during context release and when event caching is disabled, the destroy
  sets the event handle to nullptr to avoid double cleanup.

### DIFF
--- a/source/adapters/level_zero/context.cpp
+++ b/source/adapters/level_zero/context.cpp
@@ -422,6 +422,7 @@ ur_result_t ur_context_handle_t_::finalize() {
     for (auto &EventCache : EventCaches) {
       for (auto &Event : EventCache) {
         auto ZeResult = ZE_CALL_NOCHECK(zeEventDestroy, (Event->ZeEvent));
+        Event->ZeEvent = nullptr;
         // Gracefully handle the case that L0 was already unloaded.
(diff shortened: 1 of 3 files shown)


pbalcer (approved):

nrspruit:
I need to check that this change does not interfere with the "discard events" feature. This definitely resolves a memory leak issue with internal events, but I need to ensure I did not miss something with the discard feature that does not allow this cleanup. I will have this in draft until that is confirmed.

winstonzhang-intel (approved):

nrspruit:
> I need to check that this change does not interfere with the "discard events" feature. This definitely resolves a memory leak issue with internal events, but I need to ensure I did not miss something with the discard feature that does not allow this cleanup. I will have this in draft until that is confirmed.

Fixed the issue in my latest. The additional Destroy can only occur if the associated queue is NOT using Discard Events. If it is using Discard Events, then the zeEvent is not meant to be destroyed, but instead it is reused in new UR Events.

The latest accounts for this such that the known event handle leak is still fixed, but does not break the discard events use case.

nrspruit:
Added ready to merge since the CI is passing related to discard events now after my additional check.

PR #2480, opened 2024-12-18
Use UMF CUDA memory provider in UR

ldorau:
llvm: https://github.com/intel/llvm/pull/16761

### DIFF
--- a/source/adapters/cuda/context.hpp
+++ b/source/adapters/cuda/context.hpp
@@ -19,6 +19,7 @@
 
 #include "common.hpp"
 #include "device.hpp"
+#include "umf_helpers.hpp"
 
 #include <umf/memory_pool.h>
 
@@ -74,6 +75,31 @@ typedef void (*ur_context_extended_deleter_t)(void *user_data);
 ///  if necessary.
 ///
 ///
+
+static ur_result_t
+CreateHostMemoryProvider(ur_device_handle_t_ *DeviceHandle,
+                         umf_memory_provider_handle_t *MemoryProviderHost) {
+  umf_cuda_memory_provider_params_handle_t CUMemoryProviderParams = nullptr;
+
+  *MemoryProviderHost = nullptr;
+  CUcontext context = DeviceHandle->getNativeContext();
+
+  umf_result_t UmfResult =
+      umfCUDAMemoryProviderParamsCreate(&CUMemoryProviderParams);
+  UMF_RETURN_UR_ERROR(UmfResult);
+
+  umf::cuda_params_unique_handle_t CUMemoryProviderParamsUnique(
+      CUMemoryProviderParams, umfCUDAMemoryProviderParamsDestroy);
+
+  // create UMF CUDA memory provider for the host memory (UMF_MEMORY_TYPE_HOST)
+  UmfResult = umf::createMemoryProvider(
+      CUMemoryProviderParamsUnique.get(), 0 /* cuDevice */, context,
+      UMF_MEMORY_TYPE_HOST, MemoryProviderHost);
+  UMF_RETURN_UR_ERROR(UmfResult);
+
+  return UR_RESULT_SUCCESS;
+}
+
 struct ur_context_handle_t_ {
 
   struct deleter_data {
@@ -86,14 +112,25 @@ struct ur_context_handle_t_ {
   std::vector<ur_device_handle_t> Devices;
   std::atomic_uint32_t RefCount;
 
+  // UMF CUDA memory provider for the host memory (UMF_MEMORY_TYPE_HOST)
+  umf_memory_provider_handle_t MemoryProviderHost = nullptr;
+
   ur_context_handle_t_(const ur_device_handle_t *Devs, uint32_t NumDevices)
       : Devices{Devs, Devs + NumDevices}, RefCount{1} {
     for (auto &Dev : Devices) {
       urDeviceRetain(Dev);
     }
+
+    // Create UMF CUDA memory provider for the host memory
+    // (UMF_MEMORY_TYPE_HOST) from any device (Devices[0] is used here, because
+    // it is guaranteed to exist).
+    UR_CHECK_ERROR(CreateHostMemoryProvider(Devices[0], &MemoryProviderHost));
   };
 
   ~ur_context_handle_t_() {
+    if (MemoryProviderHost) {
+      umfMemoryProviderDestroy(MemoryProviderHost);
+    }
     for (auto &Dev : Devices) {
       urDeviceRelease(Dev);
     }
--- a/source/adapters/cuda/device.hpp
+++ b/source/adapters/cuda/device.hpp
@@ -11,6 +11,8 @@
 
 #include <ur/ur.hpp>
 
+#include <umf/memory_provider.h>
+
 #include "common.hpp"
 
 struct ur_device_handle_t_ {
@@ -79,9 +81,20 @@ public:
     // CUDA doesn't really have this concept, and could allow almost 100% of
     // global memory in one allocation, but is dependent on device usage.
     UR_CHECK_ERROR(cuDeviceTotalMem(&MaxAllocSize, cuDevice));
+
+    MemoryProviderDevice = nullptr;
+    MemoryProviderShared = nullptr;
   }
 
-  ~ur_device_handle_t_() { cuDevicePrimaryCtxRelease(CuDevice); }
+  ~ur_device_handle_t_() {
+    if (MemoryProviderDevice) {
+      umfMemoryProviderDestroy(MemoryProviderDevice);
+    }
+    if (MemoryProviderShared) {
+      umfMemoryProviderDestroy(MemoryProviderShared);
+    }
+    cuDevicePrimaryCtxRelease(CuDevice);
+  }
 
   native_type get() const noexcept { return CuDevice; };
 
@@ -117,6 +130,12 @@ public:
 
   // bookkeeping for mipmappedArray leaks in Mapping external Memory
   std::map<CUarray, CUmipmappedArray> ChildCuarrayFromMipmapMap;
+
+  // UMF CUDA memory provider for the device memory (UMF_MEMORY_TYPE_DEVICE)
+  umf_memory_provider_handle_t MemoryProviderDevice;
+
+  // UMF CUDA memory provider for the shared memory (UMF_MEMORY_TYPE_SHARED)
+  umf_memory_provider_handle_t MemoryProviderShared;
 };
 
 int getAttribute(ur_device_handle_t Device, CUdevice_attribute Attribute);
--- a/source/adapters/cuda/memory.cpp
+++ b/source/adapters/cuda/memory.cpp
@@ -14,6 +14,7 @@
 #include "context.hpp"
 #include "enqueue.hpp"
 #include "memory.hpp"
+#include "umf_helpers.hpp"
 
 /// Creates a UR Memory object using a CUDA memory allocation.
 /// Can trigger a manual copy depending on the mode.
@@ -49,7 +50,8 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemBufferCreate(
           cuMemHostRegister(HostPtr, size, CU_MEMHOSTREGISTER_DEVICEMAP));
       AllocMode = BufferMem::AllocMode::UseHostPtr;
     } else if (flags & UR_MEM_FLAG_ALLOC_HOST_POINTER) {
-      UR_CHECK_ERROR(cuMemAllocHost(&HostPtr, size));
+      UMF_CHECK_ERROR(umfMemoryProviderAlloc(hContext->MemoryProviderHost, size,
+                                             0, &HostPtr));
       AllocMode = BufferMem::AllocMode::AllocHostPtr;
     } else if (flags & UR_MEM_FLAG_ALLOC_COPY_HOST_POINTER) {
       AllocMode = BufferMem::AllocMode::CopyIn;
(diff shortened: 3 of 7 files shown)


source/adapters/cuda/context.hpp
```diff
@@ -85,12 +86,47 @@ struct ur_context_handle_t_ {
...
+        nullptr;
+    umf_result_t umf_result =
+        umfCUDAMemoryProviderParamsCreate(&cu_memory_provider_params);
+    if (umf_result != UMF_RESULT_SUCCESS) {
+      Result = umf::umf2urResult(umf_result);
+      throw Result;
+    }
+
+    umf_result =
+        umf::createUmfCUDAprovider(cu_memory_provider_params, device0, context0,
```
bratpiorka, line 114: Please pass 0 as a device. In UMF we should check if mem type is HOST and ignore device in this case

source/adapters/cuda/memory.cpp
```diff
@@ -49,7 +50,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urMemBufferCreate(
           cuMemHostRegister(HostPtr, size, CU_MEMHOSTREGISTER_DEVICEMAP));
       AllocMode = BufferMem::AllocMode::UseHostPtr;
     } else if (flags & UR_MEM_FLAG_ALLOC_HOST_POINTER) {
-      UR_CHECK_ERROR(cuMemAllocHost(&HostPtr, size));
+      // UR_CHECK_ERROR(cuMemAllocHost(&HostPtr, size));
+      umf_memory_provider_handle_t umfCUDAprovider =
```
bratpiorka, line 54: name it just "provider" or even call umfMemoryProviderAlloc(hContext->memoryProviderHost, size, 0, &HostPtr);

source/adapters/cuda/device.hpp
```diff
@@ -23,6 +25,7 @@ struct ur_device_handle_t_ {
   std::atomic_uint32_t RefCount;
   ur_platform_handle_t Platform;
   uint32_t DeviceIndex;
+  umf_memory_provider_handle_t umfCUDAprovider[UMF_MEMORY_TYPE_SHARED];
```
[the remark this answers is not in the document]
ldorau: There is no such code now. You have reviewed the previous revision.

source/adapters/cuda/device.hpp
```diff
@@ -115,6 +118,16 @@ struct ur_device_handle_t_ {
 
   uint32_t getNumComputeUnits() const noexcept { return NumComputeUnits; };
 
+  void setUmfCUDAprovider(umf_usm_memory_type_t memType,
+                          umf_memory_provider_handle_t _umfCUDAprovider) {
```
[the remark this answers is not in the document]
ldorau: There is no such code now. You have reviewed the previous revision.

source/adapters/cuda/memory.cpp
ldorau: Done

ldorau:
llvm: https://github.com/intel/llvm/pull/16761

source/adapters/cuda/usm.cpp
```diff
@@ -102,26 +102,47 @@ urUSMSharedAlloc(ur_context_handle_t hContext, ur_device_handle_t hDevice,
...
-    if (IsManaged || Type == CU_MEMORYTYPE_DEVICE) {
-      // Memory allocated with cuMemAlloc and cuMemAllocManaged must be freed
-      // with cuMemFree
-      UR_CHECK_ERROR(cuMemFree((CUdeviceptr)Pointer));
+
+    std::vector<ur_device_handle_t> Devices = hContext->getDevices();
+    unsigned int NumDevices = Devices.size();
+    UR_ASSERT(DeviceOrdinal < NumDevices, UR_RESULT_ERROR_INVALID_DEVICE);
+
+    ur_device_handle_t Device = Devices[DeviceOrdinal];
```
npmiller, line 126: I don't think the CUDA DeviceOrdinal works here because the UR context might not contain all the devices available, using the DeviceOrdinal with the platform might be better, I think the order in the platform is correct but I'm not 100% sure.

source/adapters/cuda/usm.cpp
```diff
@@ -102,26 +102,47 @@ urUSMSharedAlloc(ur_context_handle_t hContext, ur_device_handle_t hDevice,
...
+      memoryProvider = Device->memoryProviderShared;
+    } else if (Type == CU_MEMORYTYPE_DEVICE) {
+      // Memory allocated with cuMemAlloc must be freed
+      // cuMemFree((CUdeviceptr)Pointer)
+      memoryProvider = Device->memoryProviderDevice;
     } else {
-      // Memory allocated with cuMemAllocHost must be freed with cuMemFreeHost
-      UR_CHECK_ERROR(cuMemFreeHost(Pointer));
+      // Memory allocated with cuMemAllocHost must be freed with
+      // cuMemFreeHost(Pointer)
```
npmiller, line 139: Should we update and/or remove these comments here since this code isn't directly using the CUDA APIs anymore?

source/common/umf_helpers.hpp
```diff
@@ -279,6 +280,39 @@ inline ur_result_t umf2urResult(umf_result_t umfResult) {
   };
 }
 
+inline umf_result_t createUmfCUDAprovider(
```
frasercrmck, line 283: having `provider` be lower-case stands out a bit. I realise having `CUDAProvider` is also a little awkward.

frasercrmck (commented):
In general a lot of the variables introduced by this patch don't align with the naming conventions found in the rest of the surrounding code: `MemoryProvider`, `CUMemoryProviderParams`, etc.

source/adapters/cuda/usm.cpp
ldorau: I will remove them.

ldorau:
> In general a lot of the variables introduced by this patch don't align with the naming conventions found in the rest of the surrounding code: `MemoryProvider`, `CUMemoryProviderParams`, etc.

There no variables like `*MemoryProviderParams*`.
There are only: `memoryProvider`, `memoryProviderHost`, `memoryProviderDevice` and `memoryProviderShared`.
I can replace them, but please propose an alternative names for them.

frasercrmck:
> > In general a lot of the variables introduced by this patch don't align with the naming conventions found in the rest of the surrounding code: `MemoryProvider`, `CUMemoryProviderParams`, etc.
> 
> There no variables like `*MemoryProviderParams*`. There are only: `memoryProvider`, `memoryProviderHost`, `memoryProviderDevice` and `memoryProviderShared`. I can replace them, but please propose an alternative names for them.

Apologies for the short-hand. I meant that those were the preferred names: `memoryProvider` -> `MemoryProvider`, `cu_memory_provider_params` -> `CUMemoryProviderParams`, etc.

I don't personally have a preference of naming style, and don't know if the CUDA adapter as a whole has a preferred style, but I tend just to prefer fitting in with the surrounding code in the absence of any rules.

ldorau:
> > > In general a lot of the variables introduced by this patch don't align with the naming conventions found in the rest of the surrounding code: `MemoryProvider`, `CUMemoryProviderParams`, etc.
> > 
> > 
> > There no variables like `*MemoryProviderParams*`. There are only: `memoryProvider`, `memoryProviderHost`, `memoryProviderDevice` and `memoryProviderShared`. I can replace them, but please propose an alternative names for them.
> 
> Apologies for the short-hand. I meant that those were the preferred names: `memoryProvider` -> `MemoryProvider`, `cu_memory_provider_params` -> `CUMemoryProviderParams`, etc.
> 
> I don't personally have a preference of naming style, and don't know if the CUDA adapter as a whole has a preferred style, but I tend just to prefer fitting in with the surrounding code in the absence of any rules.

OK.
And can `umf_result_t umf_result` stay as is, or should I replace it with `umf_result_t UmfResult` ?

source/adapters/cuda/platform.cpp
```diff
@@ -115,6 +116,51 @@ urPlatformGet(ur_adapter_handle_t *, uint32_t, uint32_t NumEntries,
...
+            }
+
+            umfCUDAMemoryProviderParamsDestroy(cu_memory_provider_params);
+          } catch (ur_result_t Err) {
+            Result = Err;
+            throw Err;
+          } catch (...) {
+            Result = UR_RESULT_ERROR_OUT_OF_RESOURCES;
+            throw;
+          }
```
omarahmed1111, line 163: I think this could also be wrapped in its own function which will make the `urPlatformGet` code more concise.

source/common/umf_helpers.hpp
```diff
@@ -279,6 +280,39 @@ inline ur_result_t umf2urResult(umf_result_t umfResult) {
...
+    umf_cuda_memory_provider_params_handle_t cu_memory_provider_params,
+    int cuDevice, void *cuContext, umf_usm_memory_type_t memType,
+    umf_memory_provider_handle_t *provider) {
+  umf_result_t umf_result = umfCUDAMemoryProviderParamsSetContext(
+      cu_memory_provider_params, cuContext);
+  if (umf_result != UMF_RESULT_SUCCESS) {
+    return umf_result;
+  }
```
omarahmed1111, line 291: [NIT] I think we could wrap all if condition of this kind in a new macro that works like `UR_CHECK_RESULT`, or we could change this function to return `ur_result_t` and wrap this in `UR_CHECK_RESULT(umf::umf2urResult(umf_result))` so that makes the code more concise. I don't have strong opinions which one of this 2 approaches is better. but I think this function is used in UR entry points anyway, so the second approach might make its usage easier like instead of `UR_CHECK_ERROR(umf::umf2urResult(createUmfCUDAprovider(...)))`, it becomes `UR_CHECK_ERROR(createUmfCUDAprovider(...))`

source/adapters/cuda/platform.cpp
omarahmed1111: We could also make this inside the current `try/catch` block in `urPlatformGet` without the need to make a new one.

omarahmed1111 (commented):
Just left some stylish fixes, but functionality wise, it looks good.

source/adapters/cuda/usm.cpp
ldorau: This is the `ur_result_t USMFreeImpl(ur_context_handle_t hContext, void *Pointer)` function. We have access to `ur_context_handle_t hContext` only here, but AFAIK `ur_context_handle_t` does not have access to the platform, or am I missing something? Is there a way to get the platform from `ur_context_handle_t`?

npmiller: I don't think you can get the platform from the context, but I think you can get it from the device, if that works you could get the platform from any device in the context and then use that to find the correct device.

ldorau: @npmiller Thanks!

Done.

source/adapters/cuda/usm.cpp
ldorau: Done.

source/common/umf_helpers.hpp
ldorau: Done.

source/adapters/cuda/context.hpp
```diff
@@ -86,11 +87,48 @@ struct ur_context_handle_t_ {
...
+      }
+
+      umfCUDAMemoryProviderParamsDestroy(cu_memory_provider_params);
+    } catch (ur_result_t Err) {
+      Result = Err;
+      throw Err;
+    } catch (...) {
+      Result = UR_RESULT_ERROR_OUT_OF_RESOURCES;
+      throw;
+    }
```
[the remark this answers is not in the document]
ldorau: Done.

source/adapters/cuda/context.hpp
```diff
@@ -86,11 +87,48 @@ struct ur_context_handle_t_ {
...
+  // UMF CUDA memory provider for the host memory (UMF_MEMORY_TYPE_HOST)
+  umf_memory_provider_handle_t memoryProviderHost = nullptr;
+
   ur_context_handle_t_(const ur_device_handle_t *Devs, uint32_t NumDevices)
       : Devices{Devs, Devs + NumDevices}, RefCount{1} {
     for (auto &Dev : Devices) {
       urDeviceRetain(Dev);
     }
+
+    ur_result_t Result = UR_RESULT_SUCCESS;
```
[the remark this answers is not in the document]
ldorau: Done.

source/common/umf_helpers.hpp
frasercrmck: Thanks!

source/adapters/cuda/platform.cpp
ldorau: Done.

source/common/umf_helpers.hpp
ldorau: Done.

frasercrmck:
> OK. And can `umf_result_t umf_result` stay as is, or should I replace it with `umf_result_t UmfResult` ?

As best as I can see, by far the most common style in the CUDA adapter is e.g., `ur_result_t Result = ...` so I'd lean towards `UmfResult` as my preference. Perhaps the CUDA/HIP adapters are the odd ones out in terms of naming conventions.

ldorau:
https://github.com/intel/llvm/pull/16761 is updated

npmiller (approved):
LGTM

ldorau:
> `UmfResult`

ldorau:
> > OK. And can `umf_result_t umf_result` stay as is, or should I replace it with `umf_result_t UmfResult` ?
> 
> As best as I can see, by far the most common style in the CUDA adapter is e.g., `ur_result_t Result = ...` so I'd lean towards `UmfResult` as my preference. Perhaps the CUDA/HIP adapters are the odd ones out in terms of naming conventions.

@frasercrmck Replaced `umf_result` with `UmfResult`.

ldorau:
@frasercrmck @omarahmed1111 re-review please

frasercrmck (approved):

frasercrmck (approved):

ldorau:
@omarahmed1111 @frasercrmck @npmiller Thank you very much! :-)

bratpiorka (approved):

ldrumm:
Are we considering that UMF isn't capable of expressing allocation flags like `CU_MEM_ALLOC_WRITE_COMBINED`, `CL_MEM_ALLOC_WRITE_COMBINED_INTEL`, `hipHostMallocNonCoherent`, or similar flags - and therefore that this patch neuters our ability to control how the adapter configures the properties of the allocation?

Pools might be nice, but I think it's better to be in direct control of the allocation at this level.

Indeed I was about to submit a patch to take advantage of such interfaces, and got a merge conflict that led me to this patch. 

The draft is [here](https://github.com/ldrumm/unified-runtime/tree/luke/ext_usm_properties) for reference

To be clear I'm not against using UMF pools, but I think UMF itself needs to be mature enough top express what the drivers are already exposing.

PR #2481, opened 2024-12-18
Fix potential deadlock in the WaitEvent path of CmdBuffers

fabiomestre:
Add barriers to the SignalCommandList that guarantee that resetting the WaitEvent is done at the right time.

This fixes a potential race condition where, if the SignalCommandList executes before the ComputeCommandList, the WaitEvent could be reset before the ComputeCommandList can wait on it and, consequently, create a deadlock.

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -1683,10 +1683,15 @@ ur_result_t enqueueWaitEventPath(ur_exp_command_buffer_handle_t CommandBuffer,
         (ZeCopyCommandQueue, 1, &CommandBuffer->ZeCopyCommandList, nullptr));
   }
 
+  ZE2UR_CALL(zeCommandListAppendBarrier,
+             (SignalCommandList->first, nullptr, 1,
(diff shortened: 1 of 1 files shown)


EwanC (commented):
Looks like the `urCommandBufferFillCommandsTest.Buffer/`  CTS test is timing out in CI in a few different L0 conditions.

fabiomestre:
> Looks like the `urCommandBufferFillCommandsTest.Buffer/` CTS test is timing out in CI in a few different L0 conditions.

I guess that moving the resets to the ComputeCommandList doesn't work when using the CopyCommandList because it might reset the events before the CopyCommandList starts. For that to work we would have to add extra synchronization between the CopyCommandList and the ComputeCommandList. 

I'm trying a different at the moment with extra barriers in the SignalCommandList.

EwanC (approved):
Makes sense to me

I think https://github.com/intel/llvm/blob/sycl/sycl/doc/design/images/L0_UR_command-buffer-v5.jpg could be updated in the DPC++ PR but there's probably a whole bunch of places that doc section isn't quite right, and not sure it's updating giving we're trying to depreciate this path, so leave it up to you whether to do that or not.

PR #2496, opened 2024-12-20
[UR][Bindless] Initial implementation of bindless images for HIP

GeorgeWeb:
This PR implements the entry points in the HIP adapter for the Bindless Images extension.

Some features, while implemented, only partially pass the End-to-end tests for them in DPC++ (the PR that enables the support and testing in DPC++: https://github.com/intel/llvm/pull/16439). This is because the HIP runtime doesn't support, for example, some parameter combinations with some image memory types. The not-fully supported features are marked with `TODO`s in the device info queries for further work.

Signed-off-by: Georgi Mirazchiyski <georgi.mirazchiyski@codeplay.com>

Co-authored-by: Peter Žužek <peter@codeplay.com>

### DIFF
--- a/source/adapters/hip/device.cpp
+++ b/source/adapters/hip/device.cpp
@@ -14,6 +14,7 @@
 #include "event.hpp"
 #include "logger/ur_logger.hpp"
 
+#include <hip/hip_runtime.h>
 #include <sstream>
 
 int getAttribute(ur_device_handle_t Device, hipDeviceAttribute_t Attribute) {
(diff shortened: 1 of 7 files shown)


GeorgeWeb:
@oneapi-src/unified-runtime-hip-write Heyo, that's a whole lot of changes and additions, so I am just pinging as a reminder in case it was lost in messages over the Winter holidays. There is no rush in reviewing immediately, I just wanted to remind about it. Thank you!

source/adapters/hip/device.cpp
```diff
@@ -785,6 +786,145 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
...
+    // TODO: DPC++ doesn't implement the required builtins for SYCL.
+    return ReturnValue(ur_bool_t{false});
+  }
+  case UR_DEVICE_INFO_CUBEMAP_SEAMLESS_FILTERING_SUPPORT_EXP: {
+    // HIP supports cubemap seamless filtering.
+    // TODO: DPC++ doesn't implement the required builtins for SYCL.
+    return ReturnValue(ur_bool_t{false});
+  }
+  case UR_DEVICE_INFO_BINDLESS_SAMPLED_IMAGE_FETCH_1D_USM_EXP: {
+    // HIP does support fetching 1D USM sampled image data.
```
frasercrmck, line 882: This may be a personal preference but I prefer `supports` (as in the comment(s) above) to `does support` (below). I initially parsed it as `does not support`.

source/adapters/hip/device.cpp
```diff
@@ -785,6 +786,145 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
...
+    // TODO: DPC++ doesn't implement the required builtins for SYCL.
+    return ReturnValue(ur_bool_t{false});
+  }
+  case UR_DEVICE_INFO_BINDLESS_SAMPLED_IMAGE_FETCH_1D_EXP: {
+    // HIP does not support fetching 1D non-USM sampled image data.
+    // TODO: DPC++ doesn't implement the required builtins for SYCL.
```
frasercrmck, line 888: Is there a reason to have a TODO here if HIP doesn't support it anyway?

source/adapters/hip/image.cpp
```diff
@@ -8,178 +8,1335 @@
...
+  switch (order) {
+  case ur_image_channel_order_t::UR_IMAGE_CHANNEL_ORDER_A:
+  case ur_image_channel_order_t::UR_IMAGE_CHANNEL_ORDER_R:
+    *NumChannels = 1;
+    return UR_RESULT_SUCCESS;
+  case ur_image_channel_order_t::UR_IMAGE_CHANNEL_ORDER_RG:
+  case ur_image_channel_order_t::UR_IMAGE_CHANNEL_ORDER_RA:
+    *NumChannels = 2;
+    return UR_RESULT_SUCCESS;
+  case ur_image_channel_order_t::UR_IMAGE_CHANNEL_ORDER_RGB:
```
frasercrmck, line 35: Is there a reason that `RGB` is kept separate from the other unsupported formats?

source/adapters/hip/image.cpp
```diff
@@ -8,178 +8,1335 @@
...
+hipToUrImageChannelFormat(hipArray_Format hip_format,
+                          ur_image_channel_type_t *return_image_channel_type) {
+
+  switch (hip_format) {
+#define HIP_TO_UR_IMAGE_CHANNEL_TYPE(FROM, TO)                                 \
+  case FROM: {                                                                 \
+    *return_image_channel_type = TO;                                           \
+    return UR_RESULT_SUCCESS;                                                  \
+  }
+    HIP_TO_UR_IMAGE_CHANNEL_TYPE(HIP_AD_FORMAT_UNSIGNED_INT8,
```
frasercrmck, line 131: I note that this could also map to `UR_IMAGE_CHANNEL_TYPE_UNORM_INT8` (going by `urToHipImageChannelFormat` above). Is it a problem that we can't round-trip these? Does the UR spec expect a unique mapping? Should we document this somewhere?

source/adapters/hip/image.cpp
```diff
@@ -8,178 +8,1335 @@
...
+  }
+  return UR_RESULT_SUCCESS;
+}
+
+ur_result_t
+hipToUrImageChannelFormat(hipArray_Format hip_format,
+                          ur_image_channel_type_t *return_image_channel_type) {
+
+  switch (hip_format) {
+#define HIP_TO_UR_IMAGE_CHANNEL_TYPE(FROM, TO)                                 \
```
frasercrmck, line 126: Might wanna `undef` this later.

source/adapters/hip/image.cpp
```diff
@@ -8,178 +8,1335 @@
...
+}
+
+ur_result_t urTextureCreate(ur_sampler_handle_t hSampler,
+                            const ur_image_desc_t *pImageDesc,
+                            const HIP_RESOURCE_DESC &ResourceDesc,
+                            const unsigned int normalized_dtype_flag,
+                            ur_exp_image_native_handle_t *phRetImage) {
+
+  try {
+    /// pi_sampler_properties
```
frasercrmck, line 159: This comment looks out of date? I wouldn't expect a `pi` here.

source/adapters/hip/image.cpp
```diff
@@ -8,178 +8,1335 @@
...
+                      memType == hipMemoryTypeArray,
+                  UR_RESULT_ERROR_INVALID_VALUE);
+
+        size_t CopyExtentBytes = PixelSizeBytes * pCopyRegion->copyExtent.width;
+        const char *SrcWithOffset = static_cast<const char *>(pSrc) +
+                                    (pCopyRegion->srcOffset.x * PixelSizeBytes);
+
+        if (memType == hipMemoryTypeArray) {
+          // HIP doesn not provide async copies between host and image arrays
+          // memory in versions early than 6.2.
```
frasercrmck, line 628:
```suggestion
          // memory in versions earlier than 6.2.
```

source/adapters/hip/image.cpp
```diff
@@ -8,178 +8,1335 @@
...
+                  UR_RESULT_ERROR_INVALID_VALUE);
+
+        size_t CopyExtentBytes = PixelSizeBytes * pCopyRegion->copyExtent.width;
+        void *DstWithOffset =
+            static_cast<void *>(static_cast<char *>(pDst) +
+                                (PixelSizeBytes * pCopyRegion->dstOffset.x));
+
+        if (memType == hipMemoryTypeArray) {
+          // HIP doesn not provide async copies between image arrays and host
+          // memory in versions early than 6.2.
```
frasercrmck, line 736:
```suggestion
          // memory in versions earlier than 6.2.
```

source/adapters/hip/image.cpp
```diff
@@ -8,178 +8,1335 @@
...
+        }
+      } else if (pSrcImageDesc->type == UR_MEM_TYPE_IMAGE2D) {
+        hip_Memcpy2D cpy_desc = {};
+        cpy_desc.srcXInBytes = pCopyRegion->srcOffset.x * PixelSizeBytes;
+        cpy_desc.srcY = pCopyRegion->srcOffset.y;
+        cpy_desc.dstXInBytes = pCopyRegion->dstOffset.x * PixelSizeBytes;
+        cpy_desc.dstY = pCopyRegion->dstOffset.y;
+        cpy_desc.dstMemoryType = hipMemoryTypeHost;
+        cpy_desc.dstHost = pDst;
+        if (pSrcImageDesc->rowPitch == 0) {
```
frasercrmck, line 767: Is it okay that `cpy_desc.srcPitch` is uninitialized in this path? I might have missed similar instances of this question before/after.

source/adapters/hip/image.cpp
```diff
@@ -8,178 +8,1335 @@
...
+        extMemDesc.type = hipExternalMemoryHandleTypeOpaqueFd;
+      } else if (BaseDesc->stype == UR_STRUCTURE_TYPE_EXP_WIN32_HANDLE) {
+        auto Win32Handle = static_cast<const ur_exp_win32_handle_t *>(pNext);
+
+        switch (memHandleType) {
+        case UR_EXP_EXTERNAL_MEM_TYPE_WIN32_NT:
+          extMemDesc.type = hipExternalMemoryHandleTypeOpaqueWin32;
+          break;
+        case UR_EXP_EXTERNAL_MEM_TYPE_WIN32_NT_DX12_RESOURCE:
+          // Memory descriptor flag values such as hipExternalMemoryDedicatedare
```
frasercrmck, line 1098:
```suggestion
          // Memory descriptor flag values such as hipExternalMemoryDedicated are
```

source/adapters/hip/image.cpp
```diff
@@ -8,178 +8,1335 @@
...
+      } else if (BaseDesc->stype == UR_STRUCTURE_TYPE_EXP_WIN32_HANDLE) {
+        auto Win32Handle = static_cast<const ur_exp_win32_handle_t *>(pNext);
+        switch (semHandleType) {
+        case UR_EXP_EXTERNAL_SEMAPHORE_TYPE_WIN32_NT:
+          extSemDesc.type = hipExternalSemaphoreHandleTypeOpaqueWin32;
+          break;
+        case UR_EXP_EXTERNAL_SEMAPHORE_TYPE_WIN32_NT_DX12_FENCE:
+          extSemDesc.type = hipExternalSemaphoreHandleTypeD3D12Fence;
+          break;
+        case UR_EXP_EXTERNAL_SEMAPHORE_TYPE_OPAQUE_FD:
```
frasercrmck, line 1232: I wonder if this case falling through to `default` would be better served with a `[[fallthrough]]` annotation. Or left out? There's at least one more case of this further up if so.

frasercrmck (commented):
I can't say I was able to give the most thorough review with respect to specific HIP APIs or all of the ins and outs of the image properties and logic concerning rows and pitches and so on.

source/adapters/hip/device.hpp
```diff
@@ -88,6 +95,13 @@ struct ur_device_handle_t_ {
   int getConcurrentManagedAccess() const noexcept {
     return ConcurrentManagedAccess;
   };
+
+  bool supportsHardwareImages() const noexcept {
+    return HardwareImageSupport ? true : false;
```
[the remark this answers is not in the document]
GeorgeWeb: Yeah, there's another one stored as integer, which isn't an issue but since it is going to be used as boolen in any situation I can imagine, it may make more sense to just store as boolean. I'll do that. Thanks.

source/adapters/hip/device.hpp
```diff
@@ -88,6 +95,13 @@ struct ur_device_handle_t_ {
...
+  }
+
+  // Used for bookkeeping for mipmapped array leaks in mapping external memory.
+  std::map<hipArray_t, hipMipmappedArray_t> ChildHipArrayFromMipmapMap;
```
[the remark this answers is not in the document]
GeorgeWeb: That's a fair question. I don't know the answer to this, it is a solution I was asked to adapt from the Cuda backend implementation that fixes a race/leak. I will double check and come back to it. Hence, I didn't change the data structure. I will double check with the team to see if the order is important and if not I'll switch to `std::unordered_map`.

source/adapters/hip/sampler.hpp
```diff
@@ -30,4 +41,30 @@ struct ur_sampler_handle_t_ {
...
+
+  ur_sampler_filter_mode_t getFilterMode() const noexcept {
+    return static_cast<ur_sampler_filter_mode_t>((Props >> 1) & 0b1);
+  }
+
+  ur_sampler_addressing_mode_t getAddressingMode() const noexcept {
+    return static_cast<ur_sampler_addressing_mode_t>((Props >> 2) & 0b111);
+  }
+
+  ur_sampler_addressing_mode_t getAddressingModeDim(size_t i) const noexcept {
```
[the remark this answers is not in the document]
GeorgeWeb: Good eye! Sure.

source/adapters/hip/image.cpp
GeorgeWeb: So I think here we want it (`srcPitch`) zero-initialised. I thought it would be init to that, originally, from the looking at the `hip_Memcpy2D cpy_desc = {};` construction (another one taken from the Cuda adapter counterpart). That was on the idea that, assuming that's a POD type, it's all zero-init. However, if my understanding is wrong it may be better to do it explictly.

DBDuncan: In cuda `CU_MEMORYTYPE_ARRAY`, which is equivalent to `hipMemoryTypeArray`, arrays are opaque memory and do not have (at least exposed) notion of pitch.

So it is not set as it is not expected for the pitch field to be used at all.

GeorgeWeb: Yeah that's correct. However, I think Fraser's concern was a little more general here - that we might end up with a garbage value if left unitialised (which isn't good practice) - ideally we want it zero-init in such cases. Luckily, since C++11 with the brace-init syntax via the uniform init-brackets `{}` that are used for `cpy_desc` here (considering `hip_Memcpy2D` is a POD or trivially default-constructible), the members that are not explicitly initialised should get zero-init (or `nullptr` for pointers).

source/adapters/hip/device.hpp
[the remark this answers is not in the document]
GeorgeWeb: Oh, I had another look and this can be entirely omitted for now. It's not possible to use now because due to a mipmap creation from array API in the hip runtime for `urBindlessImagesMapExternalArrayExp`. This change is entirely unneeded until that's implemented. Actually there wasn't even going to be anything in the map ever in the current state of this implementation. Cheers for bringing this to my attention!

source/adapters/hip/image.cpp
GeorgeWeb: Yeah the whole line here is irrelevant and out-of-date.

source/adapters/hip/image.cpp
GeorgeWeb: RGB is a 3-channel format which is not supported by either Cuda or HIP. I think currently only Level Zero supports this. Hence, it's handled as a separate case here.

source/adapters/hip/image.cpp
GeorgeWeb: Yup.

GeorgeWeb:
> I can't say I was able to give the most thorough review with respect to specific HIP APIs or all of the ins and outs of the image properties and logic concerning rows and pitches and so on.

@frasercrmck Firstly, apologies for the delay in addressing the feedback. Now this feedback was actually quite thorough and I really appreciate it, so thank you for this!

frasercrmck (approved):

DBDuncan (approved):

PR #2501, opened 2024-12-20
[L0] Fixes potential overwrite in ZeEventPoolDesc.pNext

winstonzhang-intel:
When both counterbased and interruptbased events are enabled there is a ZeEventPoolDesc.pNext will get overwritten in
getFreeSlotInExistingOrNewPool. This fixes that.


nrspruit (approved):

martygrant:
Pulled into intel/llvm with https://github.com/intel/llvm/pull/16512

PR #2502, opened 2024-12-23
CUTensorMap is only in CUDA v12

ldrumm:
The documentation at https://docs.nvidia.com/cuda/cutensor/latest/ say it's supported in 11.0, 11.8, and 12.x

Our local testing against 11.7 fails because none of the required types or functions are declared in the header.

It's safer to remain buildable with older versions but with this feature disabled than requiring a version bump that isn't otherwise warranted.

### DIFF
--- a/source/adapters/cuda/tensor_map.cpp
+++ b/source/adapters/cuda/tensor_map.cpp
@@ -13,6 +13,24 @@
 
 #include "context.hpp"
 
+#if CUDA_VERSION < 12000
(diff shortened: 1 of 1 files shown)


omarahmed1111 (approved):

ldrumm:
Note that this code path is not yet used by dpc++ since the related extension is [still under discussion](https://github.com/intel/llvm/pull/16247) - so I don't think we need an intel/llvm PR to go with it

martygrant:
Pulled into intel/llvm with https://github.com/intel/llvm/pull/16512

PR #2504, opened 2024-12-23
[L0] Remove handle Translation in MCL if loader extension used

fabiomestre:
- Fixes the translation of handles which is not needed when the loader extension of MCL is used. This also fixes the problem during multi driver scenarios.

intel/llvm PR: https://github.com/intel/llvm/pull/16459

### DIFF
--- a/source/adapters/level_zero/platform.hpp
+++ b/source/adapters/level_zero/platform.hpp
@@ -96,6 +96,12 @@ struct ur_platform_handle_t_ : public _ur_platform {
   // associated with particular Level Zero driver, store this extension here.
   struct ZeMutableCmdListExtension {
     bool Supported = false;
+    // If LoaderExtension is true, the L0 loader is aware of the MCL extension.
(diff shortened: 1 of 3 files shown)


source/adapters/level_zero/platform.hpp
```diff
@@ -96,6 +96,7 @@ struct ur_platform_handle_t_ : public _ur_platform {
   // associated with particular Level Zero driver, store this extension here.
   struct ZeMutableCmdListExtension {
     bool Supported = false;
+    bool LoaderExtension = false;
```
Bensuo, line 99: Could we add a brief comment here to indicate what this value means? It's explained in a few places in the code where this value is used, but not generally.

Bensuo (approved):
LGTM just one minor comment.

fabiomestre:
@oneapi-src/unified-runtime-level-zero-write could someone review this PR?

nrspruit (approved):

PR #2505, opened 2024-12-24
Yc test/1223 context qanrantine

yingcong-wu:


AllanZyne:
Please sync these fix to msan as well.

yingcong-wu:
> Please sync these fix to msan as well.

Sync some fixes to msan.

source/loader/layers/sanitizer/asan/asan_interceptor.cpp
```diff
@@ -862,10 +884,15 @@ AsanInterceptor::findAllocInfoByAddress(uptr Address) {
...
-           "Wrong AllocInfo for the address");
+    // Make sure we got the right AllocInfo, or any other follow-up ops are based on wrong info.
+    bool IsTheRightAllocInfo =
+        Address >= It->second->AllocBegin &&
+        Address < It->second->AllocBegin + It->second->AllocSize;
+    if (!IsTheRightAllocInfo) {
+        getContext()->logger.error(
+            "Failed to find the right AllocInfo for address: {}", Address);
+        die("Address Sanitizer Fatal Error");
+    }
```
AllanZyne, line 895: is it possible that `It == m_AllocationMap.end()`?

source/loader/layers/sanitizer/asan/asan_interceptor.cpp
```diff
@@ -887,6 +914,28 @@ bool ProgramInfo::isKernelInstrumented(ur_kernel_handle_t Kernel) const {
...
+ur_result_t
+AsanInterceptor::releaseAllocationNoCheck(ur_context_handle_t Context,
+                                          std::shared_ptr<AllocInfo> AI,
+                                          bool IsFromQuarantined) {
+    if (IsFromQuarantined) {
+        getContext()->logger.info("Quarantine Free: {}",
+                                  (void *)AI->AllocBegin);
+    } else {
+        getContext()->logger.info("Free: {}", (void *)AI->AllocBegin);
+    }
```
AllanZyne, line 929: I feel not comfortable that "IsFromQuarantined" is just for different logger, it'd be better to move logger before calling "releaseAllocationNoCheck".

AllanZyne (approved):
LGTM

source/loader/layers/sanitizer/asan/asan_interceptor.cpp
yingcong-wu: That is a good question. I will rework the logic here.

yingcong-wu: After thinking, I think that case is not possible here. I have added comments about this.

source/loader/layers/sanitizer/asan/asan_interceptor.cpp
yingcong-wu: That is just a helper function to reduce duplicated codes. If we split that logic to caller, then we would have to repeat the same code in different places, which is quite again the purpose of the function.

AllanZyne: I mean you can just move "getContext()->logger" before calling this function.

yingcong-wu: > I mean you can just move "getContext()->logger" before calling this function.

Yes, that is what I mean. If we move that logging logic to caller to this help function, then we would have duplicated codes.

AllanZyne: Got it.

PR #2506, opened 2024-12-24
[DeviceASAN] Fix urKernelCreateWithNativeHandle segfault

AllanZyne:
LLVM: https://github.com/intel/llvm/pull/16473

Since we can't get nullptr when calling `getKernelInfo`, I changed the return type of this function to reference.

To make it support "urKernelCreateXXX" APIs automatically, I changed `getKernelInfo` to `getOrCreateKernelInfo`, so that we needn't intercept "urKernelCreateXXX" anymore.

Sync those changes to msan.

### DIFF
--- a/source/loader/layers/sanitizer/asan/asan_interceptor.cpp
+++ b/source/loader/layers/sanitizer/asan/asan_interceptor.cpp
@@ -648,8 +648,8 @@ KernelInfo &AsanInterceptor::getOrCreateKernelInfo(ur_kernel_handle_t Kernel) {
     }
 
     // Create new KernelInfo
-    auto hProgram = GetProgram(Kernel);
-    auto PI = getAsanInterceptor()->getProgramInfo(hProgram);
+    auto Program = GetProgram(Kernel);
+    auto PI = getProgramInfo(Program);
     bool IsInstrumented = PI->isKernelInstrumented(Kernel);
 
     std::scoped_lock<ur_shared_mutex> Guard(m_KernelMapMutex);
--- a/source/loader/layers/sanitizer/asan/asan_interceptor.cpp
+++ b/source/loader/layers/sanitizer/asan/asan_interceptor.cpp
@@ -639,22 +639,26 @@ ur_result_t AsanInterceptor::eraseProgram(ur_program_handle_t Program) {
     return UR_RESULT_SUCCESS;
 }
 
-ur_result_t AsanInterceptor::insertKernel(ur_kernel_handle_t Kernel) {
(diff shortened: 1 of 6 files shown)


source/loader/layers/sanitizer/asan/asan_interceptor.cpp
```diff
@@ -639,22 +639,26 @@ ur_result_t AsanInterceptor::eraseProgram(ur_program_handle_t Program) {
...
+    // Create new KernelInfo
+    auto Program = GetProgram(Kernel);
+    auto PI = getProgramInfo(Program);
     bool IsInstrumented = PI->isKernelInstrumented(Kernel);
 
+    std::scoped_lock<ur_shared_mutex> Guard(m_KernelMapMutex);
     m_KernelMap.emplace(Kernel,
-                        std::make_shared<KernelInfo>(Kernel, IsInstrumented));
-    return UR_RESULT_SUCCESS;
+                        std::make_unique<KernelInfo>(Kernel, IsInstrumented));
```
yingcong-wu, line 657: Why bother using a `unique_ptr` if there is no ownership change? We can just use raw pointer here.

AllanZyne: We need to manually delete them if it's raw pointer. `unique_ptr` is more convenient here.

yingcong-wu: Then we can just use the object created by the map. It should be deleted when being erased from then map.

AllanZyne:
@oneapi-src/unified-runtime-maintain, please add to 2025.1

source/loader/layers/sanitizer/asan/asan_interceptor.cpp
AllanZyne: > Then we can just use the object created by the map. It should be deleted when being erased from then map.

I've tried this but encountered a difficult-to-solve compilation error.
Using `unqiue_ptr` is a simple solution here.

yingcong-wu: Then I am fine with this as a temporary solution. We plan to refactor related code in near future anyway. We can discuss for any better solution in that time.

yingcong-wu:
Hi @zhaomaosu , please help review this PR.

yingcong-wu (approved):

zhaomaosu (approved):
lgtm

kbenzie (approved):

kbenzie:
Changes pulled in by https://github.com/intel/llvm/pull/16478.

PR #2512, opened 2025-01-02
[NativeCPU] Handle local args.

hvdijk:
Depending on the number of available threads, NativeCPU goes through different code paths for launching kernels. Some of these were missing the call to `kernel.handleLocalArgs`, resulting in local pointers being left as `nullptr`.

https://github.com/intel/llvm/pull/16500

### DIFF
--- a/source/adapters/native_cpu/kernel.hpp
+++ b/source/adapters/native_cpu/kernel.hpp
@@ -142,7 +142,9 @@ struct ur_kernel_handle_t_ : RefCounted {
     _localMemPoolSize = reqSize;
   }
 
-  // To be called before executing a work group
+  bool hasLocalArgs() const { return !_localArgInfo.empty(); }
+
+  // To be called before executing a work group if local args are present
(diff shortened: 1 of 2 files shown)


coldav (approved):

source/adapters/native_cpu/kernel.hpp
```diff
@@ -142,7 +142,9 @@ struct ur_kernel_handle_t_ : RefCounted {
     _localMemPoolSize = reqSize;
   }
 
-  // To be called before executing a work group
+  bool hasLocalArgs() { return !_localArgInfo.empty(); }
```
uwedolinsky, line 145: Could this method be `const` ?

hvdijk:
> I think #2367 is a fuller version of this.

Agreed, if we can get that, we should. As discussed, since that will take a while to get reviewed, still going ahead with this as a short term solution.

source/adapters/native_cpu/kernel.hpp
hvdijk: Done.

PR #2516, opened 2025-01-02
[L0 v2][CTS] Fix problems reported by SYCL e2e tests

igchor:


source/adapters/level_zero/v2/common.hpp
```diff
@@ -65,7 +81,9 @@ struct ze_handle_wrapper {
     }
 
     if (ownZeHandle) {
-      auto zeResult = ZE_CALL_NOCHECK(destroy, (handle));
+      auto zeResult =
+          ZeCall().doCall(destroy(handle), zeDestroyFunctionName<ZeHandleT>(),
```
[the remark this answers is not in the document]
igchor: Rebased.

igchor:
Hm, looks like change in [[CTS][L0 v2] adjust urEnqueueMemBufferMap](https://github.com/oneapi-src/unified-runtime/pull/2516/commits/d8b2a29cdab2be16a18519f3282a33e83951dc8c) should is specific to L0 adapter only (for other adapters SYCL doesn't expect this behavior).

I think that for now, I will make the test L0-specific but the proper solution should be to fix the SYCL runtime to not rely on this behavior (it;s not in the spec).

pbalcer:
> Hm, looks like change in [[CTS][L0 v2] adjust urEnqueueMemBufferMap](https://github.com/oneapi-src/unified-runtime/pull/2516/commits/d8b2a29cdab2be16a18519f3282a33e83951dc8c) is specific to L0 adapter only (for other adapters SYCL doesn't expect this behavior).
> 
> I think that for now, I will make the test L0-specific but the proper solution should be to fix the SYCL runtime to not rely on this behavior (it;s not in the spec).

Which SYCL test is failing? We'd have to fix the legacy adapter. And I imagine it was changed to address some issue (or something someone thought was an issue).

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
```diff
@@ -872,17 +880,20 @@ ur_result_t ur_queue_immediate_in_order_t::enqueueUSMMemcpy2D(
...
-  std::ignore = dstPitch;
-  std::ignore = pSrc;
-  std::ignore = srcPitch;
-  std::ignore = width;
-  std::ignore = height;
-  std::ignore = numEventsInWaitList;
-  std::ignore = phEventWaitList;
-  std::ignore = phEvent;
-  return UR_RESULT_ERROR_UNSUPPORTED_FEATURE;
+  TRACK_SCOPE_LATENCY("ur_queue_immediate_in_order_t::enqueueUSMMemcpy2D");
```
pbalcer, line 883: can you also implement `enqueueUSMFill2D` and add a test? I checked and there isn't one in SYCL.

igchor: I actually don't think enqueueUSMFill2D is implemented by any adapter (and it's not used by SYCL). There are a few functions that are not supported even by the legacy L0 adapter and for now I didn;t want to include them in v2.

pbalcer (approved):

pbalcer:
> > > Hm, looks like change in [[CTS][L0 v2] adjust urEnqueueMemBufferMap](https://github.com/oneapi-src/unified-runtime/pull/2516/commits/d8b2a29cdab2be16a18519f3282a33e83951dc8c) is specific to L0 adapter only (for other adapters SYCL doesn't expect this behavior).
> > > I think that for now, I will make the test L0-specific but the proper solution should be to fix the SYCL runtime to not rely on this behavior (it;s not in the spec).
> > 
> > 
> > Which SYCL test is failing? We'd have to fix the legacy adapter. And I imagine it was changed to address some issue (or something someone thought was an issue).
> 
> This one: https://github.com/intel/llvm/blob/sycl/sycl/test-e2e/Regression/host_unified_memory.cpp I haven;t looked into the implementation too much but it seems like SYCL might handle this by it's own for other backends.
> 
> The legacy adapter works fine (it passes the UR test that I added and the SYCL test) but this is because it implements this logic that's not in the spec.

Ok, but please create a ticket so we don't forget.

Issue #2290, opened 2024-11-06
Restructure GetInfo CTS tests to use separate tests instead of a switch


martygrant:
As per https://github.com/oneapi-src/unified-runtime/pull/2272#discussion_r1830918465 most of the GetInfo CTS tests use a switch statement to test all of the possible info type enumerations (e.g. https://github.com/oneapi-src/unified-runtime/blob/main/test/conformance/context/urContextGetInfo.cpp#L43). It may be better that these are modified to be separate and individual tests.

- [x] adapter
- [x] context - https://github.com/oneapi-src/unified-runtime/pull/2518
- [x] device - https://github.com/intel/llvm/pull/16746
- [x] event - https://github.com/oneapi-src/unified-runtime/pull/2488 + https://github.com/oneapi-src/unified-runtime/pull/2626
- [x] kernel - https://github.com/oneapi-src/unified-runtime/pull/2444
- [x] memory - https://github.com/oneapi-src/unified-runtime/pull/2521
- [x] memory(image) - https://github.com/oneapi-src/unified-runtime/pull/2619
- [x] platform - https://github.com/oneapi-src/unified-runtime/pull/2523
- [x] program - https://github.com/intel/llvm/pull/17094
- [x] queue - https://github.com/oneapi-src/unified-runtime/pull/2282
- [x] sampler - https://github.com/oneapi-src/unified-runtime/pull/2536
- [x] usm - https://github.com/oneapi-src/unified-runtime/pull/2538
- [x] virtual memory - https://github.com/oneapi-src/unified-runtime/pull/2272 + https://github.com/oneapi-src/unified-runtime/pull/2541

Additional: 
- [x] https://github.com/oneapi-src/unified-runtime/pull/2565

PR #2518, opened 2025-01-03
Move urContextGetInfo success test from a switch to individual tests

martygrant:
For https://github.com/oneapi-src/unified-runtime/issues/2290

### DIFF
--- a/test/conformance/context/urContextGetInfo.cpp
+++ b/test/conformance/context/urContextGetInfo.cpp
@@ -12,7 +12,6 @@ TEST_P(urContextGetInfoTest, SuccessNumDevices) {
     size_t size = 0;
 
     ASSERT_SUCCESS(urContextGetInfo(context, info_type, 0, nullptr, &size));
-    ASSERT_NE(size, 0);
     ASSERT_EQ(size, sizeof(uint32_t));
 
     uint32_t nDevices = 0;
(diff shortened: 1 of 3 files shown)


test/conformance/context/urContextGetInfo.cpp
```diff
@@ -4,116 +4,147 @@
...
+}
+
+TEST_P(urContextGetInfoTest, SuccessDevices) {
+    ur_context_info_t info_type = UR_CONTEXT_INFO_DEVICES;
+    size_t size = 0;
+
+    ASSERT_SUCCESS(urContextGetInfo(context, info_type, 0, nullptr, &size));
+    ASSERT_NE(size, 0);
+
+    if (size > 0) {
```
aarongreig, line 32: I think we can just unconditionally do the part in here and get rid of the `else`, we've already checked that `size != 0` above

test/conformance/context/urContextGetInfo.cpp
```diff
@@ -4,116 +4,147 @@
...
 }
 
-using urContextGetInfoTest = uur::urContextTest;
-UUR_INSTANTIATE_DEVICE_TEST_SUITE_P(urContextGetInfoTest);
+TEST_P(urContextGetInfoTest, SuccessUSMMemCpy2DSupport) {
+    ur_context_info_t info_type = UR_CONTEXT_INFO_USM_MEMCPY2D_SUPPORT;
+    size_t size = 0;
+
+    ASSERT_SUCCESS(urContextGetInfo(context, info_type, 0, nullptr, &size));
+    ASSERT_NE(size, 0);
```
aarongreig, line 57: for these ones I think we can drop this check, if size is zero it'll get caught by the check below and we'll still get the same information (i.e. it'll print the actual vs. expected values and that will tell us that `size` is 0)

martygrant: yeah good shout, definitely redundant having this extra check. I done the same in Event and Queue so I've removed them from there too

test/conformance/context/urContextGetInfo.cpp
```diff
@@ -4,116 +4,136 @@
...
-                 uur::deviceTestWithParamPrinter<ur_context_info_t>);
-
-TEST_P(urContextGetInfoTestWithInfoParam, Success) {
-    ur_context_info_t info = getParam();
-    size_t info_size = 0;
-    ASSERT_SUCCESS_OR_OPTIONAL_QUERY(
-        urContextGetInfo(context, info, 0, nullptr, &info_size), info);
-    ASSERT_NE(info_size, 0);
+    ASSERT_SUCCESS(urContextGetInfo(context, info_type, 0, nullptr, &size));
+    ASSERT_NE(size, 0);
```
aarongreig, line 15: missed one

aarongreig (approved):

PR #2522, opened 2025-01-06
Update urProgramGetFunctionPointer test

RossBrunton:
The spec says that implementations may return
`UR_RESULT_ERROR_FUNCTION_ADDRESS_NOT_AVAILABLE` for this function, but
that wasn't handled by this test.

### DIFF
--- a/test/conformance/program/urProgramGetFunctionPointer.cpp
+++ b/test/conformance/program/urProgramGetFunctionPointer.cpp
@@ -21,10 +21,13 @@ struct urProgramGetFunctionPointerTest : uur::urProgramTest {
 UUR_INSTANTIATE_DEVICE_TEST_SUITE_P(urProgramGetFunctionPointerTest);
 
 TEST_P(urProgramGetFunctionPointerTest, Success) {
-    UUR_KNOWN_FAILURE_ON(uur::OpenCL{"Intel(R) UHD Graphics 770"});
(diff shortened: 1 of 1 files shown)


RossBrunton:
> This is a valid return but I'm not sure it's necessarily what we're expecting under the circumstances set up by the test for all platforms, I'm a bit worried a genuine error condition could go unnoticed down the line with this change. What platform were you seeing the error on?

I'm seeing it on my local system, which identifies as:
```
[opencl:gpu][opencl:0] Intel(R) OpenCL Graphics, Intel(R) UHD Graphics 770 OpenCL 3.0 NEO  [24.39.31294.12]
```

My reading of the spec is that implementations are allowed to return this value if they can't get the address of the function for some internal technical reason. Which basically means that, from the point of view of UR, implementations may (at their discretion) return `UR_RESULT_ERROR_FUNCTION_ADDRESS_NOT_AVAILABLE` provided the function is present.

aarongreig:
Ah, looks like this is one of the CL extensions with no documentation anywhere. Just to satisfy my curiosity, could you try creating a kernel for the function before trying to retrieve the function pointer? I'm wondering if maybe we aren't using the entry point right. Otherwise I guess we can just accept that return as tantamount to "not supported"

aarongreig (approved):

PR #2524, opened 2025-01-06
[L0]: Fix Out Event in Enqueue Wait Events to prevent reuse

nrspruit:
- Fix Out Event created in Enqueue Wait Events to ensure the event is
  not reused.
- Removed completion logic that was allowing caching of the event.
- event_handle_t zeEventDestroy handles the memory leak issue that this
  branch was attempting to fix.

### DIFF
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -145,10 +145,6 @@ ur_result_t urEnqueueEventsWait(
     std::unique_lock<ur_shared_mutex> Lock(Queue->Mutex);
     resetCommandLists(Queue);
   }
-  if (OutEvent && (*OutEvent)->Completed) {
-    UR_CALL(CleanupCompletedEvent((*OutEvent), false, false));
-    UR_CALL(urEventReleaseInternal((*OutEvent)));
-  }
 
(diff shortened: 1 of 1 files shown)


nrspruit:
I just realized a previous patch of mine will fix the memory leak that the patch by Winston had originally fixed, I am going to try that first and may simplify this code.

source/adapters/level_zero/event.cpp
```diff
@@ -1137,7 +1140,8 @@ ur_result_t urEventReleaseInternal(ur_event_handle_t Event) {
   if (DisableEventsCaching || !Event->OwnNativeHandle) {
     delete Event;
   } else {
-    Event->Context->addEventToContextCache(Event);
```
igchor, line 1140: Aren't we leaking the Event now, since it's neither `delete`ed nor added to the cache?

nrspruit: I simplified this. I had actually fixed the original memory leak that Winston was addressing here:

https://github.com/oneapi-src/unified-runtime/pull/2476

Please re-review, the patch now is a revert and minor update of the patch from Winston.

igchor (approved):

PR #2527, opened 2025-01-07
Wrap urEventSetCallback when ran through loader

RossBrunton:
The events returned by the loader do not match the events returned by
openCL (or other backends), which causes issues when adding a callback
handler. This adds an intermediate wrapper to replace the event with the
loader event.

### DIFF
--- a/source/loader/ur_ldrddi.cpp
+++ b/source/loader/ur_ldrddi.cpp
@@ -13,6 +13,7 @@
 #include "ur_loader.hpp"
 
 namespace ur_loader {
+
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Intercept function for urAdapterGet
 __urdlllocal ur_result_t UR_APICALL urAdapterGet(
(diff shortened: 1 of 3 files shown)


source/loader/ur_ldrddi.cpp
```diff
@@ -4423,6 +4440,13 @@ __urdlllocal ur_result_t UR_APICALL urEventSetCallback(
 
     [[maybe_unused]] auto context = getContext();
 
+    // Replace the callback with a wrapper function that gives the callback the loader event rather than a
+    // backend-specific event
+    auto wrapper_data =
+        new event_callback_wrapper_data_t{pfnNotify, hEvent, pUserData};
```
RossBrunton, line 4446: Yes, I know. Manual memory management.

But I'm not sure if there's a better C++-y way of doing this, does anyone have any ideas?

RossBrunton: I don't think so. It needs to survive a roundtrip to and from a bare pointer to be used in the API.

pbalcer: yeah, i don't think we can rely on raii here, unless you'd like to manually play around with constructing/destroying unique pointers, but I don't see much point.

pbalcer (approved):

PR #2531, opened 2025-01-08
Fix calls to ze destructors in level-zero adapter v2

Xewar313:
Currently, when using level zero v2 adapter, all destructions reported with use of UR_L0_DEBUG flag are labeled as "destroy(handle)". It provides no information on which objects are destroyed and which level zero functions are called. It also breaks leak detection using UR_L0_LEAKS_DEBUG flag - it does not report any objects freed this way.

This is a bit crude workaround, however I wasn't able to come up with any cleaner solution.

### DIFF
--- a/source/adapters/level_zero/v2/common.hpp
+++ b/source/adapters/level_zero/v2/common.hpp
@@ -15,12 +15,21 @@
 
 #include "../common.hpp"
 #include "logger/ur_logger.hpp"
+namespace {
+  const char* desturctorNames[] = {
+  "zeKernelDestroy",
+  "zeEventDestroy",
+  "zeEventPoolDestroy",
+  "zeContextDestroy",
+  "zeCommandListDestroy"
+  };
+}
 
 namespace v2 {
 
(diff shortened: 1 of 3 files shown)


source/adapters/level_zero/v2/common.hpp
```diff
@@ -90,19 +96,19 @@ struct ze_handle_wrapper {
 };
 
 using ze_kernel_handle_t =
-    ze_handle_wrapper<::ze_kernel_handle_t, zeKernelDestroy>;
+    ze_handle_wrapper<::ze_kernel_handle_t, zeKernelDestroy, 0>;
```
pbalcer, line 99: yeah, no, this is too awkward to use ;) I'd rather we create some sort of macro (or some clever lambda as parametr, dunno) that wraps all the destroy functions and calls ZE_CALL_NOCHECK inside of the wrapper.

igchor (approved):

pbalcer (approved):

PR #2532, opened 2025-01-08
Add basic command buffer support to level zero adapter v2

Xewar313:
This pull request implements basic calls to command buffer in level zero v2 adapter. These calls are required by sycl graph functionality implemented inside llvm, such as record and replay.


source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,309 @@
...
+// given Context and Device.
+void checkImmediateAppendSupport(ur_context_handle_t Context,
+                                 ur_device_handle_t Device) {
+  // TODO The L0 driver is not reporting this extension yet. Once it does,
+  // switch to using the variable zeDriverImmediateCommandListAppendFound.
+
+  // Minimum version that supports zeCommandListImmediateAppendCommandListsExp.
+  constexpr uint32_t MinDriverVersion = 30898;
+  bool DriverSupportsImmediateAppend =
+      Context->getPlatform()->isDriverVersionNewerOrSimilar(1, 3,
```
EwanC, line 28: This is out of sync from the changes in https://github.com/oneapi-src/unified-runtime/pull/2498 could we share this `checkImmediateAppendSupport` helper between v1 and v2 adapters to avoid duplicating code and it becoming inconsistent.

source/adapters/level_zero/v2/command_buffer.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include <unordered_set>
+
+#include "common.hpp"
+
+#include "context.hpp"
+#include "kernel.hpp"
+#include "queue_api.hpp"
+
+struct command_buffer_profiling_t {
+  ur_exp_command_buffer_sync_point_t NumEvents;
```
igchor, line 25: nit: we are naming variables/params using with all lower-case in v2

source/adapters/level_zero/v2/command_buffer.hpp
```diff
@@ -0,0 +1,73 @@
...
+                  uint32_t numWaitEvents);
+  // Releases the resources associated with the command-buffer before the
+  // command-buffer object is destroyed.
+  void cleanupCommandBufferResources();
+
+  // UR context associated with this command-buffer
+  ur_context_handle_t Context;
+  // Device associated with this command buffer
+  ur_device_handle_t Device;
+  ze_command_list_handle_t ZeCommandList;
```
igchor, line 49: please use v2::raii::ze_command_list_handle_t from v2/common.hpp

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+/**
+ * Creates a L0 command list
+ * @param[in] Context The Context associated with the command-list
+ * @param[in] Device  The Device associated with the command-list
+ * @param[in] IsUpdatable Whether the command-list should be mutable.
+ * @param[out] CommandList The L0 command-list created by this function.
+ * @return UR_RESULT_SUCCESS or an error code on failure
+ */
+ur_result_t createMainCommandList(ur_context_handle_t Context,
```
igchor, line 118: Would it make sense to cache command lists like we do for non-command-buffer path?

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,309 @@
...
+  checkImmediateAppendSupport(Context, Device);
+
+  if (IsUpdatable) {
+    UR_ASSERT(Context->getPlatform()->ZeMutableCmdListExt.Supported,
+              UR_RESULT_ERROR_UNSUPPORTED_FEATURE);
+  }
+  
+  ze_command_list_handle_t ZeCommandList = nullptr;
+  UR_CALL(createMainCommandList(Context, Device, IsUpdatable, ZeCommandList));
+  try {
```
igchor, line 160: please move the try/catch block to cover the entire function and add try/catch to other functions as well, i.e. `urCommandBufferCreateExp(...) try { ... } catch (...) {  return exceptionToResult(std::current_exception()); }` In v2, we have helpers functions that might throw so it's best to wrap every function with try/catch.

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,309 @@
...
+    ur_context_handle_t Context, ur_device_handle_t Device,
+      ze_command_list_handle_t CommandList,
+      const ur_exp_command_buffer_desc_t *Desc)
+    : Context(Context), Device(Device), ZeCommandList(CommandList),
+      IsUpdatable(Desc ? Desc->isUpdatable : false) {
+  UR_CALL_THROWS(ur::level_zero::urContextRetain(Context));
+  UR_CALL_THROWS(ur::level_zero::urDeviceRetain(Device));
+}
+
+void ur_exp_command_buffer_handle_t_::cleanupCommandBufferResources() {
```
igchor, line 90: I don't think `urContextRelease/urDeviceRelease` can actually fail.

I would suggest just removing this entire function and moving the logic to the destructor.

source/adapters/level_zero/v2/command_buffer.hpp
```diff
@@ -0,0 +1,73 @@
...
+  // This list is needed to release all kernels retained by the
+  // command_buffer.
+  std::vector<ur_kernel_handle_t> KernelsList;
+};
+
+struct ur_exp_command_buffer_command_handle_t_ : public _ur_object {
+  ur_exp_command_buffer_command_handle_t_(ur_exp_command_buffer_handle_t,
+                                          uint64_t);
+
+  virtual ~ur_exp_command_buffer_command_handle_t_();
```
igchor, line 67: Doesn;t need to be virtual I think

source/adapters/level_zero/v2/command_buffer.cpp
EwanC: `UR_CALL_THROWS` can throw an exception, which is bad practice to do in C++ destructors. So if we want to do this move I think it can be a straight copy/paste.

source/adapters/level_zero/v2/command_buffer.cpp
EwanC: I don't know what the non-command-buffer path does, but for reference creating a pool of command-lists to use has been an idea we've had for the v1 adapter but never got around to. [TODO comment that has since been removed](https://github.com/intel/llvm/pull/9992/commits/c385466a68e5192ad38d5bfd4a4c2a6dd36ac4c7#r1247439851)

source/adapters/level_zero/v2/command_buffer.hpp
Xewar313: In that case, I believe that command_list_cache should also be changed, right? (Its fields are named starting with upper-case)

source/adapters/level_zero/v2/command_buffer.cpp
Xewar313: What about urKernelRelease? I also need to release it when releasing buffer, so I would also need to know it can fail, and I am not sure if it is a case

source/adapters/level_zero/v2/queue_api.hpp
```diff
@@ -14,6 +14,16 @@
 
 #include <ur_api.h>
 
+#include "../common.hpp"
```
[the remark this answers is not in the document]
Xewar313: I have a small problem with removing all includes. As far as I understand, the queue_api.hpp.mako is being generated based on some file containing declarations (ur_api.hpp?), but all of these declarations only use structures that were defined inside UR (all starting with ur_*). However, I had to add function that uses ze_command_list_handle_t, because there is no respective class with ur_ prefix. And the problem is, that the ze_command_list_handle_t must be included, so I have to at least include "common.hpp" or "../common.hpp" - without that this simply won't work. Other option would be creating something like ur_command_list_handle_t, but I believe that should be a separate PR, because of scope of the change.

source/adapters/level_zero/v2/command_buffer.cpp
EwanC: That makes sense, if v2 unconditionally uses immediate command-lists then we don't need to be so conservative. The `isPVC` check was also there, because that's the platform where we've verified this path for v1, but again for v2 purposes I don't think we need to be as conservative.

source/adapters/level_zero/v2/command_buffer.cpp
igchor: In v2, we have https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/v2/command_list_cache.hpp which could be used here.

source/adapters/level_zero/v2/command_buffer.hpp
igchor: Yes, good point, we never got to fixing it

source/adapters/level_zero/v2/command_buffer.cpp
igchor: Well, technically, urKernelRelease can fail if zeKernelDestroy fails but if that happens and we throw from this function we can have a memory leak (if there are more unreleased kernels in the vector).

Also, is it really that useful to return an error from urCommandBufferReleaseExp? 

Perhaps we could just log all the failures from `urKernelRelease`, `urContextRelease`, etc and always return UR_RESULT_SUCCESS in `urCommandBufferReleaseExp` This would ensure that all resources are (attempted to be) freed, even if there is some error during release of one of them.

source/adapters/level_zero/v2/queue_api.hpp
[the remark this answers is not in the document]
igchor: If you need to use `ze_command_list_handle_t` then it should be enough to just include `ze_api.h`, just add it (and the enqueueCommandBuffer declaration to the https://github.com/oneapi-src/unified-runtime/blob/main/scripts/templates/queue_api.hpp.mako and call `make generate`

scripts/templates/queue_api.cpp.mako
```diff
@@ -19,6 +19,8 @@ from templates import helper as th
  *
  */
 
+// This file was generated basing on scripts/templates/queue_api.cpp.mako
```
pbalcer, line 22: nit:
```
Do not edit. This file is auto generated from a template:
scripts/templates/queue_api.cpp.mako
```

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,290 @@
...
+                  "the current driver does not support the "
+                  "zeCommandListImmediateAppendCommandListsExp entrypoint.");
+    std::abort();
+  }
+}
+
+} // namespace
+
+std::pair<ze_event_handle_t *, uint32_t>
+ur_exp_command_buffer_handle_t_::getWaitListView(
```
pbalcer, line 35: This is identical to the implementation in queue. Please create simple WaitListView abstraction usable in both.

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,290 @@
...
+ * @param[in] isUpdatable Whether the command-list should be mutable.
+ * @param[out] commandList The L0 command-list created by this function.
+ * @return UR_RESULT_SUCCESS or an error code on failure
+ */
+ur_result_t createMainCommandList(ur_context_handle_t context,
+                                  ur_device_handle_t device, bool isUpdatable,
+                                  ze_command_list_handle_t &commandList) {
+
+  using queue_group_type = ur_device_handle_t_::queue_group_info_t::type;
+  // that should be call to queue getZeOrdinal,
```
pbalcer, line 85: This, together with the fact that we have no way to allocate events from the correct queue, makes me think we either need to defer creation of these objects until the first enqueue of the command buffer or `urCommandBufferCreateExp` should take a queue.
@EwanC thoughts?

I'm also thinking whether it wouldn't make sense just to make the CommandBuffer allocate a whole pool of events from the context. That way, when we need an event, we don't have to acquire context locks.

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,290 @@
...
+ur_result_t
+urCommandBufferCreateExp(ur_context_handle_t context, ur_device_handle_t device,
+                         const ur_exp_command_buffer_desc_t *commandBufferDesc,
+                         ur_exp_command_buffer_handle_t *commandBuffer) {
+  try {
+    bool isUpdatable = commandBufferDesc && commandBufferDesc->isUpdatable;
+    checkImmediateAppendSupport(context);
+
+    if (isUpdatable) {
+      UR_ASSERT(context->getPlatform()->ZeMutableCmdListExt.Supported,
```
pbalcer, line 116: Please don't use these "asserts". Do a normal if and return `UR_RESULT_ERROR_UNSUPPORTED_FEATURE`.
I just dislike overloading the term "assert" to mean fail with error.

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Creates a L0 command list
+ * @param[in] context The Context associated with the command-list
+ * @param[in] device  The Device associated with the command-list
+ * @param[in] isUpdatable Whether the command-list should be mutable.
+ * @param[out] commandList The L0 command-list created by this function.
+ * @return UR_RESULT_SUCCESS or an error code on failure
+ */
+ur_result_t createMainCommandList(ur_context_handle_t context,
```
pbalcer, line 80: this isn't used anywhere.

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,290 @@
...
+  std::ignore = numSyncPointsInWaitList;
+  std::ignore = syncPointWaitList;
+  std::ignore = retSyncPoint;
+
+  // TODO
+  std::ignore = numKernelAlternatives;
+  std::ignore = kernelAlternatives;
+  std::ignore = command;
+  try {
+    UR_ASSERT(hKernel, UR_RESULT_ERROR_INVALID_NULL_HANDLE);
```
pbalcer, line 204: This is near identical to the queue enqueue implementation. I imagine a lot of other similar enqueue methods will be likewise similar.

We need to come up with an abstraction that lets us share this implementation between queue and command buffers.

While reviewing this, I had a thought that a command buffer is a subset of the queue functionality with some extra bits. This sounds like this should be solvable with composition:
```
class cmds { // enqueuable? command_list?
  WaitViewList waitlist;
  CmdList cmdlist;
  event_pool events;
  enqueue_kernel enqueue_kernel(...);
  ...
}

class queue {
  cmds cmd;
}

class command_buffer {
  cmds cmd;
}
```
@igchor thoughts?

source/adapters/level_zero/v2/command_buffer.hpp
```diff
@@ -0,0 +1,70 @@
...
+      v2::raii::command_list_unique_handle &&commandList,
+      const ur_exp_command_buffer_desc_t *desc);
+  ~ur_exp_command_buffer_handle_t_();
+  ur_event_handle_t getSignalEvent(ur_event_handle_t *hUserEvent,
+                                   ur_command_t commandType);
+
+  std::pair<ze_event_handle_t *, uint32_t>
+  getWaitListView(const ur_event_handle_t *phWaitEvents,
+                  uint32_t numWaitEvents);
+
```
pbalcer, line 41: private:

source/adapters/level_zero/v2/command_buffer.hpp
```diff
@@ -0,0 +1,70 @@
...
+  // command_buffer.
+  std::vector<ur_kernel_handle_t> kernelsList;
+};
+
+struct ur_exp_command_buffer_command_handle_t_ : public _ur_object {
+  ur_exp_command_buffer_command_handle_t_(ur_exp_command_buffer_handle_t,
+                                          uint64_t);
+
+  ~ur_exp_command_buffer_command_handle_t_();
+
```
pbalcer, line 65: private:

source/adapters/level_zero/v2/command_buffer.cpp
EwanC: > This, together with the fact that we have no way to allocate events from the correct queue, makes me think we either need to defer creation of these objects until the first enqueue of the command buffer or urCommandBufferCreateExp should take a queue.
@EwanC thoughts?

I'm not that keen on `urCommandBufferCreateExp` taking a queue. 1) it doesn't match the SYCL API, where we don't have a queue object when creating the UR command-buffer. 2) It is the opposite direction to where the  OpenCL WG is going with command-buffers in https://github.com/KhronosGroup/OpenCL-Docs/pull/1292 to separate the queues used on command-buffer creation and enqueue.

Are there only 2 types of queue ordinal relevant here, compute and copy? If so I would suggest creating command-lists for both. This PR doesn't have the v1 functionality of splitting commands from the UR command-buffers into compute and copy command-lists. But we saw this having good perf benefits on V1, so would be imagine creating a copy engine command-list is something we'll need to do at some point anyway.

EwanC: cool, letting the UMD handle this consideration definitely sounds like it will simplify the adapter code

source/adapters/level_zero/v2/command_buffer.cpp
igchor: I agree, also, we can simplify the enqueue operations by combining getSignalEvent and getWaitListView functions. I would move this functionality to the ur_command_list_handler_t (we might want to rename it) and implement it like this:

```
struct ur_command_list_handler_t {
  ur_command_list_handler_t(ur_context_handle_t hContext,
                            ur_device_handle_t hDevice,
                            const ur_queue_properties_t *pProps);

  ur_command_list_handler_t(ze_command_list_handle_t hZeCommandList,
                            bool ownZeHandle);

  std::tuple<ze_event_handle_t, uint32_t, ze_event_handle_t *>
  getSignalEventAndWaitList(ur_event_handle_t *hUserEvent,
                            ur_command_t commandType,
                            const ur_event_handle_t *phWaitEvents,
                            uint32_t numWaitEvents);

  raii::command_list_unique_handle commandList;
  std::vector<ze_event_handle_t> waitList;
};
```

then, in enqueue* we can do:

```
auto [signalEvent, numWaitEvents, waitEvent] = getSignalEventAndWaitList(...);
...
ZE2UR_CALL(zeCommandListAppenSomething(..., signalEvent, numWaitEvents, waitEvent));
```

igchor: Also, right now, getSIgnalEvent returns ur_event_handle_t (not ze_event_handle_t) because in enqueueTimestampRecordingExp, we need to call timestamp-related functions on it. But actually, we already have a pointer to the ur_event - it's the same one that we pass as a first argument to the getSignalEvent.

It's better to make getSignalEventAndWaitList return ze_event_handle_t so that we can avoid checking for nullptr.

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,290 @@
...
+        context, device, std::move(zeCommandList), commandBufferDesc);
+
+  } catch (const std::bad_alloc &) {
+    return exceptionToResult(std::current_exception());
+  }
+  return UR_RESULT_SUCCESS;
+}
+ur_result_t
+urCommandBufferRetainExp(ur_exp_command_buffer_handle_t hCommandBuffer) {
+  try {
```
igchor, line 137: please move the try before the first `{` so it's like this:

```
ur_result_t
urCommandBufferRetainExp(ur_exp_command_buffer_handle_t hCommandBuffer) try {
...
} catch(..._ {
...
}
```

This way there's one less indentation level.

source/loader/layers/sanitizer/asan/asan_report.hpp
```diff
@@ -2,9 +2,9 @@
  *
  * Copyright (C) 2024 Intel Corporation
  *
- * Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
- * See LICENSE.TXT
- * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ * Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM
+ * Exceptions. See LICENSE.TXT SPDX-License-Identifier: Apache-2.0 WITH
```
pbalcer, line 6: invalid rebase?

source/adapters/level_zero/v2/command_list_manager.cpp
```diff
@@ -0,0 +1,146 @@
...
+  auto zeSignalEvent = signalEvent ? signalEvent->getZeEvent() : nullptr;
+
+  ZE2UR_CALL(zeCommandListImmediateAppendCommandListsExp,
+             (zeCommandList.get(), numCommandLists, phCommandLists,
+              zeSignalEvent, numWaitEvents, pWaitEvents));
+
+  return UR_RESULT_SUCCESS;
+}
+
+ur_result_t ur_command_list_manager::closeCommandList() {
```
pbalcer, line 135: this should be inside of command buffer. Immediate command lists never need to be closed.

source/adapters/level_zero/v2/command_list_manager.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  std::scoped_lock<ur_shared_mutex> Lock(this->Mutex);
+  auto signalEvent = getSignalEvent(phEvent, callerCommand);
+
+  auto [pWaitEvents, numWaitEvents] =
+      getWaitListView(phEventWaitList, numEventsInWaitList);
+
+  auto zeSignalEvent = signalEvent ? signalEvent->getZeEvent() : nullptr;
+
+  ZE2UR_CALL(zeCommandListImmediateAppendCommandListsExp,
```
pbalcer, line 128: this is command buffer specific.

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,200 @@
...
+                  "the current driver does not support the "
+                  "zeCommandListImmediateAppendCommandListsExp entrypoint.");
+    std::abort();
+  }
+}
+
+} // namespace
+
+std::pair<ze_event_handle_t *, uint32_t>
+ur_exp_command_buffer_handle_t_::getWaitListView(
```
pbalcer, line 35: this can be removed, right?

source/loader/layers/sanitizer/asan/asan_report.hpp
Xewar313: This is strange one - for some reason code formatting formats this header, but it shouldn't be formatted

source/adapters/level_zero/v2/command_list_manager.cpp
Xewar313: Actually it is queue specific (since we need to enqueue command buffer, by appending command list to execution list), however since it is modifying the command list, I feel like it makes more sense to add it here, instead of queue implementation. What do you think?

pbalcer: Ah, right. I'd move it to queue then. The "command list manager" is not meant to be a clean command list abstraction, but merely a place where we store and retrieve command list and associated state. Otherwise we risk pushing everything into this class.

pbalcer (commented):
directionally lgtm, but we can't do this refactoring piecemeal with cmdlist-related state duplicated between queue class and the new manager. That might lead to inconsistency issues/duplicating event pools etc.

source/adapters/level_zero/v2/command_buffer.cpp
Xewar313: I changed getSignalEvent to use ze_event_handle_t instead of ur_event_handle_t, but I don't think that merging getSignalEvent with getWaitListView makes sense - due to some C++ weirdness, referencing variables bound using [] inside lambda is impossible. So following code is not compiling:
```
auto [signalEvent, numWaitEvents, waitEvent] = getSignalEventAndWaitList(...);
  hMem->unmapHostPtr(pMappedPtr, [&](void *src, void *dst, size_t size) {
    ZE2UR_CALL_THROWS(zeCommandListAppendMemoryCopy,
                      (commandListManager.getZeCommandList(), dst, src, size,
                       nullptr, numWaitEvents, waitEvent));
    memoryMigrated = true;
  });
}
```
This pattern is used across whole queue implementation, so I would have to call `std::get` on tuple inside function, what would be even less readable.

igchor: Right, I forgot about this. In that case, instead of the tuple we can just use a custom structure:

```
struct {
    ze_event_handle_t signalEvent;
    size_t numWaitEvents;
    ze_event_handle_t *waitEvents;
};
```

and then, the code just becomes:

```
auto e = getSignalEventAndWaitList(...);
  hMem->unmapHostPtr(pMappedPtr, [&](void *src, void *dst, size_t size) {
    ZE2UR_CALL_THROWS(zeCommandListAppendMemoryCopy,
                      (commandListManager.getZeCommandList(), dst, src, size,
                       nullptr, e.numWaitEvents, e.waitEvent));
    memoryMigrated = true;
  });
}
```

Xewar313: Another problem that I have noticed with that solution is inside enqueueTimestampRecordingExp - We call there only getWaitListView without getSignalEvent, because ze_event_handle_t is obtained from ur_event_handle_t method. So we will get signal event without needing it there (so the most optimal solution would be again splitting this merged method into two)

pbalcer: Let's skip this change for now. Returning 24 byte-sized structure might prevent register use.

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,185 @@
...
+      device->QueueGroup[queue_group_type::Compute].ZeOrdinal;
+  v2::raii::command_list_unique_handle zeCommandList =
+      context->commandListCache.getRegularCommandList(device->ZeDevice, true,
+                                                      queueGroupOrdinal, true);
+
+  *commandBuffer = new ur_exp_command_buffer_handle_t_(
+      context, device, std::move(zeCommandList), commandBufferDesc);
+  return UR_RESULT_SUCCESS;
+
+} catch (const std::bad_alloc &) {
```
igchor, line 75: I would change this to catch `...` - the cache can throw other things as well

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,185 @@
...
+  return UR_RESULT_SUCCESS;
+} catch (...) {
+  return exceptionToResult(std::current_exception());
+}
+
+ur_result_t
+urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t hCommandBuffer) try {
+  UR_ASSERT(hCommandBuffer, UR_RESULT_ERROR_INVALID_NULL_POINTER);
+  UR_ASSERT(!hCommandBuffer->isFinalized, UR_RESULT_ERROR_INVALID_OPERATION);
+  hCommandBuffer->closeCommandList();
```
igchor, line 102: UR_CALL?

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,185 @@
...
+  delete hCommandBuffer;
+  return UR_RESULT_SUCCESS;
+} catch (...) {
+  return exceptionToResult(std::current_exception());
+}
+
+ur_result_t
+urCommandBufferFinalizeExp(ur_exp_command_buffer_handle_t hCommandBuffer) try {
+  UR_ASSERT(hCommandBuffer, UR_RESULT_ERROR_INVALID_NULL_POINTER);
+  UR_ASSERT(!hCommandBuffer->isFinalized, UR_RESULT_ERROR_INVALID_OPERATION);
```
igchor, line 101: this is not thread safe, I think you should move this (and setting isFinalized to closeCommandList).

source/adapters/level_zero/v2/command_list_manager.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include "kernel.hpp"
+
+ur_command_list_manager::ur_command_list_manager(
+    ur_context_handle_t context, ur_device_handle_t device,
+    v2::raii::command_list_unique_handle &&commandList, v2::event_flags_t flags,
+    ur_queue_handle_t_ *queue)
+    : context(context), device(device),
+      eventPool(context->eventPoolCache.borrow(device->Id.value(), flags)),
+      zeCommandList(
+          std::forward<v2::raii::command_list_unique_handle>(commandList)),
```
igchor, line 24: std::move()

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
```diff
@@ -88,29 +63,35 @@ ur_queue_immediate_in_order_t::ur_queue_immediate_in_order_t(
...
-              ownZeQueue) {}
-
-ur_event_handle_t
+      commandListManager(
+          hContext, hDevice,
+          raii::command_list_unique_handle(
+              reinterpret_cast<ze_command_list_handle_t>(hNativeHandle),
+              [ownZeQueue](ze_command_list_handle_t hZeCommandList) {
+                if (ownZeQueue) {
+                  zeCommandListDestroy(hZeCommandList);
```
igchor, line 86: ZE_CALL_NOCHECK, this is needed to have proper logs

source/adapters/level_zero/v2/command_buffer.cpp
igchor: @Xewar313 Yeah, I don't think your current implementation is correct. You should still call getSignalEvent() as you need to somehow allocate the event (the passed handle is out handle only, it doesn't point to a valid event).

So, with `getSignalEventAndWaitList` you would do:
```
auto e = getSignalEventAndWaitList(phEvent, ...);

(*phEvent)->recordStartTimestamp();
assert(phEvent->getZeHandle() == e.signalEvent);
...
```

igchor: @pbalcer I don't think there should be any difference with current usage where we call those function one after another and only then use the resulting values.

If that's really something that you'd like to optimize we can just move the implementation to the header and make it inline.

igchor: @Xewar313 you can leave the implementation as is for now (just fix enqueueTimestampRecordingExp) and we can implement getSignalEventAndWaitList in another PR.

pbalcer (commented):
lgtm, just few minor comments

source/adapters/level_zero/v2/command_buffer.cpp
```diff
@@ -0,0 +1,183 @@
...
+  std::ignore = command;
+  UR_CALL(commandBuffer->commandListManager.appendKernelLaunch(
+      hKernel, workDim, pGlobalWorkOffset, pGlobalWorkSize, pLocalWorkSize, 0,
+      nullptr, nullptr));
+  return UR_RESULT_SUCCESS;
+} catch (...) {
+  return exceptionToResult(std::current_exception());
+}
+
+ur_result_t urCommandBufferEnqueueExp(
```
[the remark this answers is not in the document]
EwanC: that's a good spot, I've created https://github.com/oneapi-src/unified-runtime/issues/2600 to sort this out separately

PR #2533, opened 2025-01-08
Make profiling info optional and update tests

RossBrunton:
This patch turns all of the values returned by urEventGetProfilingInfo
to be optional and updates adapters to handle this by returning the
appropriate enum when it is not supported.

The tests have also been updated, to ensure that returning a counter of
"0" or values equal to the previous profiling event is no longer
considered a failure.


martygrant (approved):

scripts/core/event.yml
```diff
@@ -117,15 +117,15 @@ name: $x_profiling_info_t
 typed_etors: True
 etors:
     - name: COMMAND_QUEUED
-      desc: "[uint64_t] A 64-bit value of current device counter in nanoseconds when the event is enqueued"
+      desc: "[uint64_t][optional-query] A 64-bit value of current device counter in nanoseconds when the event is enqueued"
     - name: COMMAND_SUBMIT
-      desc: "[uint64_t] A 64-bit value of current device counter in nanoseconds when the event is submitted"
+      desc: "[uint64_t][optional-query] A 64-bit value of current device counter in nanoseconds when the event is submitted"
     - name: COMMAND_START
```
aarongreig, line 123: going off the match files it looks like START/END could be considered non-optional

RossBrunton: Perhaps, but I think it's simpler to have everything be optional rather than have to retrofit it into a future driver that doesn't have profiling info.

aarongreig: Alright I think I'm ok with this. Initially I was thinking it might be worth enforcing a distinction between supporting the entry point but not supporting all its enums (by having some portion < 100% of enums be optional), and just not supporting the entry point at all (by the usual mechanism of having a device info query that you check before calling it). Looking at how sycl actually uses this though, it kind of seems like the distinction would go unnoticed. I also had a look at the vulkan spec to see if there was evidence for the "it's unlikely a future adapter would dip below the current baseline of support" argument, but no it looks like there's a good chance implementing any of these is going to at least be difficult if not impossible on vulkan.

aarongreig (approved):

RossBrunton:
@oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-hip-write @oneapi-src/unified-runtime-cuda-write @oneapi-src/unified-runtime-level-zero-write

Can I get this looked at?

frasercrmck (approved):
CUDA/HIP LGTM

RossBrunton:
@oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-level-zero-write

Can I get this looked at?

hvdijk (approved):

pbalcer (approved):

PR #2537, opened 2025-01-09
[v0.11.x] Bump UMF to v0.10.1 patch release

lukaszstolarczuk:
It contains 2 significant fixes:
- Setting symbols' versioning, which will help preventing breaking changes
- Removing incorrect assert in utils_align_ptr_up_size_down

At best, this would go to v0.11.x, but I'm not sure of the current state of releasing this branch...?

### DIFF
--- a/source/common/CMakeLists.txt
+++ b/source/common/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2022-2023 Intel Corporation
+# Copyright (C) 2022-2025 Intel Corporation
 # Part of the Unified-Runtime Project, under the Apache License v2.0 with LLVM Exceptions.
 # See LICENSE.TXT
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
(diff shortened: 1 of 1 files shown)


kbenzie:
> undrafted, ready for review.
> 
> I tried to verify this PR on llvm repo, but they are already moved on to using UR from main... so my PR is doomed to fail...? Any hints what would be the best approach at the moment...? ping @oneapi-src/unified-runtime-maintain

There's a few things to consider here:

1. We need to have these changes reflected in UR main and therefore intel/llvm as well before we can merged to v0.11.x release branch. Would this involve divergent UMF branches?
2. There needs to be a Jira (likely high exposure) associated with any bug fixes before they can be included in the internal release.
3. This will trigger a number of cherry-picks to be included in the internal release.

lukaszstolarczuk:
Thanks @kbenzie

I've discussed this PR with the team and I believe it's not crucial to have this one in v0.11.x

PR #2539, opened 2025-01-09
Added `DEVICE_INFO_PROGRAM_SET_SPECIALIZATION_CONSTANTS`

RossBrunton:
THis is similar to the existing
`DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS` device info, only
it applies to programs rather than kernels. All of the adapters should
be updated to report it correctly, and tests have been added.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -2198,6 +2198,9 @@ typedef enum ur_device_info_t {
   /// to the `USMPool` entry points and usage of the `pool` parameter of the
   /// USM alloc entry points.
   UR_DEVICE_INFO_USM_POOL_SUPPORT = 119,
+  /// [::ur_bool_t] support the ::urProgramSetSpecializationConstants entry
+  /// point
+  UR_DEVICE_INFO_PROGRAM_SET_SPECIALIZATION_CONSTANTS = 120,
   /// [::ur_bool_t] Returns true if the device supports the use of
   /// command-buffers.
   UR_DEVICE_INFO_COMMAND_BUFFER_SUPPORT_EXP = 0x1000,
@@ -5739,6 +5742,10 @@ typedef struct ur_specialization_constant_info_t {
 /// @brief Set an array of specialization constants on a Program.
 ///
 /// @details
+///     - This entry point is optional, the application should query for support
+///       with device query
+///       ::UR_DEVICE_INFO_PROGRAM_SET_SPECIALIZATION_CONSTANTS passed to
+///       ::urDeviceGetInfo.
 ///     - The application may call this function from simultaneous threads for
 ///       the same device.
 ///     - The implementation of this function should be thread-safe.
@@ -5758,6 +5765,9 @@ typedef struct ur_specialization_constant_info_t {
 ///         + `NULL == pSpecConstants`
 ///     - ::UR_RESULT_ERROR_INVALID_SIZE
 ///         + `count == 0`
+///     - ::UR_RESULT_ERROR_UNSUPPORTED_FEATURE
+///         + If ::UR_DEVICE_INFO_PROGRAM_SET_SPECIALIZATION_CONSTANTS query is
+///         false
 ///     - ::UR_RESULT_ERROR_INVALID_VALUE
 ///         + A pSpecConstant entry contains a size that does not match that of
 ///         the specialization constant in the module.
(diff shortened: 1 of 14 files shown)


frasercrmck (approved):
CUDA/HIP LGTM

include/ur_api.h
```diff
@@ -1632,28 +1632,30 @@ typedef enum ur_device_info_t {
...
+    UR_DEVICE_INFO_PROGRAM_SET_SPECIALIZATION_CONSTANTS = 107,       ///< [::ur_bool_t] support the ::urProgramSetSpecializationConstants entry
+                                                                     ///< point
+    UR_DEVICE_INFO_MEMORY_BUS_WIDTH = 108,                           ///< [uint32_t][optional-query] return the width in bits of the memory bus
                                                                      ///< interface of the device.
-    UR_DEVICE_INFO_MAX_WORK_GROUPS_3D = 108,                         ///< [size_t[3]] return max 3D work groups
-    UR_DEVICE_INFO_ASYNC_BARRIER = 109,                              ///< [::ur_bool_t] return true if Async Barrier is supported
-    UR_DEVICE_INFO_MEM_CHANNEL_SUPPORT = 110,                        ///< [::ur_bool_t] return true if specifying memory channels is supported
-    UR_DEVICE_INFO_HOST_PIPE_READ_WRITE_SUPPORTED = 111,             ///< [::ur_bool_t] Return true if the device supports enqueueing commands
+    UR_DEVICE_INFO_MAX_WORK_GROUPS_3D = 109,                         ///< [size_t[3]] return max 3D work groups
```
aarongreig, line 1639: I'd rather keep these the same if we can, think this can be avoided by appending the yml definition to the end of the list

RossBrunton: Done.

test/conformance/device/urDeviceGetInfo.cpp
```diff
@@ -109,6 +109,7 @@ static std::unordered_map<ur_device_info_t, size_t> device_info_size_map = {
     {UR_DEVICE_INFO_BFLOAT16, sizeof(ur_bool_t)},
     {UR_DEVICE_INFO_MAX_COMPUTE_QUEUE_INDICES, sizeof(uint32_t)},
     {UR_DEVICE_INFO_KERNEL_SET_SPECIALIZATION_CONSTANTS, sizeof(ur_bool_t)},
+    {UR_DEVICE_INFO_PROGRAM_SET_SPECIALIZATION_CONSTANTS, sizeof(ur_bool_t)},
```
aarongreig, line 112: needs adding to `::testing::Values` below as well (we do have a refactor coming down the pipe that eliminates the two list system)

aarongreig (approved):

RossBrunton:
@oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-level-zero-write Can you quickly look at this? I've just set it so that Native CPU and Level Zero don't advertise support for specialization constants.

coldav:
> @oneapi-src/unified-runtime-native-cpu-write @oneapi-src/unified-runtime-level-zero-write Can you quickly look at this? I've just set it so that Native CPU and Level Zero don't advertise support for specialization constants.

It might makes sense to get @PietroGhg to review (not sure if he is still in these groups or not) - we will want specialized constants for native cpu though I think.

kbenzie:
> we will want specialized constants for native cpu though I think.

It doesn't support them today so it should say so to the user.

RossBrunton:
> > > we will want specialized constants for native cpu though I think.
> > 
> > 
> > It doesn't support them today so it should say so to the user.
> 
> I thought we did, just that we fail tests.

Both `urKernelSetSpecializationConstants` and `urProgramSetSpecializationConstants` return `UR_RESULT_ERROR_UNSUPPORTED_FEATURE` in native CPU.

RossBrunton:
@oneapi-src/unified-runtime-level-zero-write I want to get this merged, can you look at it?

@coldav @PietroGhg It should be as simple as implementing it and returning `true` for this device info. Once that is done, it should "just work" and (properly written) user code will switch over to using spec constants. If you're okay with it, can I get an approval?

PietroGhg:
> > > > we will want specialized constants for native cpu though I think.
> > > 
> > > 
> > > It doesn't support them today so it should say so to the user.
> > 
> > 
> > I thought we did, just that we fail tests.
> 
> Both `urKernelSetSpecializationConstants` and `urProgramSetSpecializationConstants` return `UR_RESULT_ERROR_UNSUPPORTED_FEATURE` in native CPU.

For Native CPU we are going through the "emulated" path for Spec Constants, which means (I may be wrong, it's been a while since I looked into it) that the values for them are set as kernel arguments by the SYCL runtime, so I think that the values for `urKernelSetSpecializationConstants` and `urProgramSetSpecializationConstants` won't affect Native CPU, so this change looks fine to me

PietroGhg (approved):

pbalcer (approved):

PR #2543, opened 2025-01-10
Fix Unsupported feature returns

isaacault:
* urKernelCreateWithNativeHandle handle param is [nocheck]. Remove invalid handle as reason to return unsupported feature.
* Mark urKernelGetNativeHandle handle param as [nocheck]. Remove invalid handle as reason to return unsupported feature.


isaacault:
`UNSUPPORTED_FEATURE` is a valid return for when the adapter does not have a notion of "native handles". As for the lack of support for `UR_ERROR_INVALID_KERNEL`, this can be addressed at a later date and is logged by https://github.com/oneapi-src/unified-runtime/issues/2555.

PR #2552, opened 2025-01-13
[L0 v2] introduce raii wrapper for UR handles

igchor:
Some entities (e.g. devices) do not need to be retained as they are owned by the platform. For such cases, only validate RefCount instead of acutally increasing/decreasing it.

It doesn't seem to be necessary to retain context whenever it's used to construct queue, etc. However, the spec is not very clear on this. I'm wondering if we should we update the spec or add retains on context?

Also, there are a few instances where I think calling retain() might be justified (we are not calling it right now) - e.g. when passing ur_mem_handle_t to urKernelSetArgMemObj / urKernelSetArgSampler.

### DIFF
--- a/source/adapters/level_zero/v2/common.hpp
+++ b/source/adapters/level_zero/v2/common.hpp
@@ -14,6 +14,7 @@
 #include <ze_api.h>
 
 #include "../common.hpp"
+#include "../ur_interface_loader.hpp"
 #include "logger/ur_logger.hpp"
 namespace {
 #define DECLARE_DESTROY_FUNCTION(name)                                         \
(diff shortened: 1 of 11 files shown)


source/adapters/level_zero/v2/event.hpp
```diff
@@ -98,13 +98,14 @@ struct ur_event_handle_t_ : _ur_object {
   uint64_t getEventEndTimestamp();
 
 protected:
+  // Do not use ref couting on context to avoid circular dependency.
```
pbalcer, line 101: do we need something like a weak pointer? You added quite a few comments like that. It would be better to enforce this using a type.

pbalcer (commented):
> It doesn't seem to be necessary to retain context whenever it's used to construct queue, etc. However, the spec is not very clear on this. I'm wondering if we should we update the spec or add retains on context?

hm, I think we should tie lifetime of context to queues/command buffers/programs/pools and all the other entities created from a context. But not kernels or individual enqueues. These retain their parent objects (like e.g., a queue), so the lifetime relationship between, e.g., an enqueue and context is transitive. So it should be safe.

Generally, my opinion is that it should be safe to release the context while there are some pending operations.

source/adapters/level_zero/v2/event.hpp
igchor: I guess I could just do an alias for this, and add some explanation in the comment above that alias definition.

pbalcer: My suggestion would be to create a weak<> type that has deleted copy constructors/operators and can only be created from a ref-counted ptr (using something like .downgrade() or whatever).

igchor:
> > It doesn't seem to be necessary to retain context whenever it's used to construct queue, etc. However, the spec is not very clear on this. I'm wondering if we should we update the spec or add retains on context?
> 
> hm, I think we should tie lifetime of context to queues/command buffers/programs/pools and all the other entities created from a context. But not kernels or individual enqueues. These retain their parent objects (like e.g., a queue), so the lifetime relationship between, e.g., an enqueue and context is transitive. So it should be safe.
> 
> Generally, my opinion is that it should be safe to release the context while there are some pending operations.

Yeah, that makes sense, I think at least CUDA and HIP are doing something similar. I will add this and perhaps some tests for that.

source/adapters/level_zero/v2/event.hpp
igchor: Yeah, but in some situations there might be no rc<> instance. For example, when you create an eventPool in the context constructor, you need to pass ur_context_handle_t (this) to the eventPool. If it would only accept weak<> and if weak<> can only be created from rc<>, you would need a temporary rc<>. This is a bit verbose but on the other hand, you can use rc_val_only (to do ref count validation) so this could provider us with some more checks.

See the latest commit for how it would look like. Also, I guess if we decide to do that, then probably all handles in v2 should be replaced with rc<> rc_val_only<> or weak<>.

igchor: Having weak<> could also allow us to check the pointers for NULL (only allow rc<> and weak<> to be != NULL). If we ever need the value to be null (2 places right now), we can use optional (see latest commit).

source/adapters/level_zero/v2/common.hpp
```diff
@@ -104,5 +104,114 @@ using ze_context_handle_t =
...
+  };
+
+// This version of ref_counted calls retain/release functions.
+template <typename URHandle>
+using rc = ref_counted<URHandle, ref_counted_traits<URHandle>::retain,
+                       ref_counted_traits<URHandle>::release>;
+
+// This version of ref_counted validates that the ref count is not zero on every
+// release and retain in debug mode, and does nothing in the release mode.
+// Used for types that should always be alibe during the adapter lifetime (e.g.
```
pbalcer, line 195:
```suggestion
// Used for types that should always be alive during the adapter lifetime (e.g.
```

source/adapters/level_zero/v2/common.hpp
```diff
@@ -104,5 +104,114 @@ using ze_context_handle_t =
...
+// devices).
+template <typename URHandle>
+using rc_val_only =
```
pbalcer, line 198: hmm, I don't like the name :D

`durable`, `NonNull`, `Live`, hm, dunno. Basically what we want to say is that this object's lifetime is guaranteed to exceed that of the containing object.

source/adapters/level_zero/v2/event.hpp
pbalcer: I think I'd prefer to have a weak be nullable, like normal std::weak. So that when the original rc pointer goes out of scope, we can detect that.

> For example, when you create an eventPool in the context constructor, you need to pass ur_context_handle_t (this) to the eventPool.

That's a scenario for the `rc_val_only` type, because context must outlive a queue/command buffer associated with the eventPool. I think weak is for when we create objects that might possibly outlive their "parent" objects.

igchor: > That's a scenario for the rc_val_only type, because context must outlive a queue/command buffer associated with the eventPool. I think weak is for when we create objects that might possibly outlive their "parent" objects.

Yeah, but the problem is that if you use `rc_val_only` then it will always fail during `release` because of how urContextRelease (and other functions are implemented) - we first decrement the RefCount (to 0) and only then call a destructor. We could implement a real shared_ptr like in c++ (where the control block is in separate allocation) so we could change the order of those ops but I don't know if it's worth it :d

source/adapters/level_zero/v2/common.hpp
igchor: Yeah, me too. `non_null`/`not_null` might good, or maybe `checked`?

source/adapters/level_zero/v2/event.hpp
igchor: Hm, but on the other hand, probably every use case of rc_val_only will be like this so we could just check if the ref count is >= 0

source/adapters/level_zero/v2/common.hpp
pbalcer: i like `checked`.

source/adapters/level_zero/v2/event.hpp
igchor: Tbh, I think I'll remove `rc_val_only`, as it's only applicable to device and platforms and there's a small chance we will have lifetime issues with those. We could just add some checks to the device/platform classes if we really want to.

martygrant:
## Information

## PR Migration

---

martygrant:

---

PR #2557, opened 2025-01-14
Add size validation for USM device/shared allocs

RossBrunton:
The spec states that an allocation size higher than
`DEVICE_INFO_MAX_MEM_ALLOC_SIZE` should report an error. This has been
added to the cuda, hip and native cpu backends now.

The tests have also been updated, which sadly introduces a new
regression for level zero.


kbenzie:
I don't think we can reliably have this check due to limitations in how the different adapters report max alloc size.

RossBrunton:
> I don't think we can reliably have this check due to limitations in how the different adapters report max alloc size.

What do you mean? The spec says that `MAX_MEM_ALLOC_SIZE` is the largest USM allocation that can be made.

frasercrmck (approved):
CUDA/HIP LGTM

PR #2559, opened 2025-01-14
[CUDA][HIP] Fix kernel arguments being overwritten when added out of order

fabiomestre:
In the Cuda and Hip adapter, when kernel arguments are added out of order (e.g. argument at index 1 is added before argument at index 0), the existing arguments are currently being overwritten. This happens because some of the argument sizes might not be known when adding them out of order and the code relies on those sizes to choose where to store the argument.

This PR avoids this issue by storing the arguments in the same order that they are added and accessing them using pointer offsets.

intel/llvm: https://github.com/intel/llvm/pull/16733

### DIFF
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -492,7 +492,7 @@ enqueueKernelLaunch(ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel,
       UR_CHECK_ERROR(RetImplEvent->start());
     }
 
-    auto &ArgIndices = hKernel->getArgIndices();
+    auto &ArgIndices = hKernel->getArgPointers();
     UR_CHECK_ERROR(cuLaunchKernel(
         CuFunc, BlocksPerGrid[0], BlocksPerGrid[1], BlocksPerGrid[2],
         ThreadsPerBlock[0], ThreadsPerBlock[1], ThreadsPerBlock[2], LocalSize,
@@ -680,7 +680,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunchCustomExp(
       UR_CHECK_ERROR(RetImplEvent->start());
     }
 
-    auto &ArgIndices = hKernel->getArgIndices();
+    auto &ArgPointers = hKernel->getArgPointers();
 
     CUlaunchConfig launch_config;
     launch_config.gridDimX = BlocksPerGrid[0];
@@ -696,7 +696,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunchCustomExp(
     launch_config.numAttrs = launch_attribute.size();
 
     UR_CHECK_ERROR(cuLaunchKernelEx(&launch_config, CuFunc,
-                                    const_cast<void **>(ArgIndices.data()),
+                                    const_cast<void **>(ArgPointers.data()),
                                     nullptr));
 
     if (phEvent) {
--- a/source/adapters/cuda/enqueue.cpp
+++ b/source/adapters/cuda/enqueue.cpp
@@ -492,11 +492,11 @@ enqueueKernelLaunch(ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel,
       UR_CHECK_ERROR(RetImplEvent->start());
     }
 
-    auto &ArgIndices = hKernel->getArgPointers();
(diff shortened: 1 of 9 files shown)


EwanC (commented):
Could we add CTS tests for this? Probably in `test/conformance/exp_command_buffer/update/local_memory_update.cpp` & `test/conformance/kernel/urKernelSetArgLocal.cpp` 

Might make it easier if we want to merge this fix earlier than the https://github.com/intel/llvm/pull/16573 feature

source/adapters/cuda/kernel.hpp
```diff
@@ -109,14 +112,21 @@ struct ur_kernel_handle_t_ {
...
-      // Update the stored value for the argument
-      std::memcpy(&Storage[InsertPos], Arg, Size);
-      Indices[Index] = &Storage[InsertPos];
-      AlignedLocalMemSize[Index] = LocalSize;
+
+      // Copy new argument to storage if it hasn't been added before.
+      if (ParamSizes[Index] == 0) {
+        ParamSizes[Index] = Size;
+        std::memcpy(&Storage[InsertPos], Arg, Size);
+        Indices[Index] = &Storage[InsertPos];
```
frasercrmck, line 120: This isn't an issue introduced by this PR, but the description of `Indices` as `Byte offset into /p Storage allocation for each parameter.` is a little confusing, isn't it? An index, to me, implies a relative value from the beginning of the storage, not an absolute address. Do you think we should do something about this? I note this PR is embedding the use of `Indices` more deeply into the code so it might be best to fix now.

source/adapters/hip/kernel.hpp
```diff
@@ -151,20 +161,11 @@ struct ur_kernel_handle_t_ {
...
-          calcAlignedLocalArgument(Index, Size);
-
-      // Store argument details
-      addArg(Index, sizeof(size_t), (const void *)&(AlignedLocalOffset),
-             AlignedLocalSize);
-
-      // For every existing local argument which follows at later argument
-      // indices, update the offset and pointer into the kernel local memory.
-      // Required as padding will need to be recalculated.
+    // Iterate over all existing local argument which follows StartIndex
```
frasercrmck, line 164:
```suggestion
    // Iterate over all existing local arguments which follow StartIndex
```

```suggestion
    // Iterate over each existing local argument which follows StartIndex
```

either works

source/adapters/cuda/kernel.hpp
fabiomestre: Yes, I don't mind changing the variable name and related getters. Do you have a preference for the name? Perhaps `ParamPointers` or `ParamOffsets` `?

test/conformance/exp_command_buffer/update/local_memory_update.cpp
```diff
@@ -1105,3 +1105,223 @@ TEST_P(LocalMemoryMultiUpdateTest, UpdateWithoutBlocking) {
...
+    }
+
+    void TearDown() override {
+        if (command_handle) {
+            EXPECT_SUCCESS(urCommandBufferReleaseCommandExp(command_handle));
+        }
+
+        UUR_RETURN_ON_FATAL_FAILURE(
+            LocalMemoryUpdateTestBaseOutOfOrder::TearDown());
+    }
```
EwanC, line 1220: Nothing to do right now, but heads up this will conflict with https://github.com/oneapi-src/unified-runtime/pull/2578

EwanC (approved):

source/adapters/cuda/kernel.hpp
fabiomestre: The existing getter was called `getArgIndices()` so I decided to rename the variable to `ArgPointers`.  Hopefully that name makes more sense.

frasercrmck: Sounds good to me, thank you

aarongreig (approved):

source/adapters/cuda/enqueue.cpp
```diff
@@ -492,7 +492,7 @@ enqueueKernelLaunch(ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel,
       UR_CHECK_ERROR(RetImplEvent->start());
     }
 
-    auto &ArgIndices = hKernel->getArgIndices();
+    auto &ArgIndices = hKernel->getArgPointers();
```
frasercrmck, line 495: `ArgPointers`?

frasercrmck (approved):

PR #2560, opened 2025-01-14
Graph API benchmarks added

mateuszpn:


scripts/benchmarks/benches/compute.py
```diff
@@ -355,3 +355,30 @@ def bin_args(self) -> list[str]:
...
+
+    def name(self):
+        return f"graph_api_benchmark_sycl SubmitExecute"
+
+    def bin_args(self) -> list[str]:
+        return [
+            "--iterations=1000",
+            "--numKernels=100",
+        ]   
+
```
pbalcer, line 384: missing newline

scripts/benchmarks/benches/compute.py
```diff
@@ -355,3 +355,30 @@ def bin_args(self) -> list[str]:
...
+class GraphApiSinKernelSYCL(ComputeBenchmark):
+    def __init__(self, bench):
+        super().__init__(bench, "graph_api_benchmark_sycl", "SinKernel")
+
+    def name(self):
+        return f"graph_api_benchmark_sycl SinKernel"
+
+    def bin_args(self) -> list[str]:
+        return [
+            "--iterations=1000",
```
pbalcer, line 368: is that enough iterations for the benchmark to be stable?

scripts/benchmarks/benches/compute.py
```diff
@@ -355,3 +355,30 @@ def bin_args(self) -> list[str]:
...
+            "--numKernels=100",
```
pbalcer, line 369: should we add more scenarios with different number of kernels? e.g., with 1 kernel, to see the cost of the whole machinery end-to-end.

scripts/benchmarks/benches/compute.py
```diff
@@ -74,6 +74,26 @@ def benchmarks(self) -> list[Benchmark]:
             MemcpyExecute(self, 10, 16, 1024, 10000, 0, 1, 1),
             MemcpyExecute(self, 4096, 1, 1024, 10, 0, 1, 0),
             MemcpyExecute(self, 4096, 4, 1024, 10, 0, 1, 0),
+            GraphApiSinKernelSYCL(self, 0, 1),
```
pbalcer, line 77: hm, this might be too much :D

scripts/benchmarks/benches/compute.py
```diff
@@ -355,3 +375,38 @@ def bin_args(self) -> list[str]:
...
         ]
+
+class GraphApiSinKernelSYCL(ComputeBenchmark):
+    def __init__(self, bench, graphs, numKernels):
+        self.graphs = graphs
+        self.numKernels = numKernels
+        super().__init__(bench, "graph_api_benchmark_sycl", "SinKernel")
+
+    def name(self):
+        return f"graph_api_benchmark_sycl SinKernel"
```
pbalcer, line 386: names need to reflect arguments, otherwise the benchmarks won't be unique in output.

pbalcer (commented):
The graph benchmarks didn't run because compute-benchmarks is too old. You need to update the commit being used (line 25).

pbalcer:
This failed with:
```
RequestError [HttpError]: Validation Failed: {"resource":"IssueComment","code":"unprocessable","field":"data","message":"Body is too long (maximum is 65536 characters)"}
```

If this keeps happening after you've reduced the number of scenarios, I suggest we temporarily remove output (just comment it out) from the markdown. I plan on eventually creating an HTML file per PR and then link to it in the markdown, and that will give us the ability to have longer content with all the details.

pbalcer:
All graph-related benchmarks failed with:
```
Abort was called at 80 line in file:
./shared/source/command_stream/linear_stream.h
```
and then I think the gpu crashed:
```
terminate called after throwing an instance of 'sycl::_V1::exception'
  what():  No device of requested type available.
```

Have you seen that before?

PR #2561, opened 2025-01-14
[L0][CMDBUF] Optimize fence/event waits during update

Bensuo:
- Track whether a command buffer was recently synchronized so we can skip unnecessary fence/event waits. Previous consecutive calls to update a command buffer would result in calling `zeFenceHostSynchronize` for each call unnecessarily.

### DIFF
--- a/source/adapters/level_zero/command_buffer.hpp
+++ b/source/adapters/level_zero/command_buffer.hpp
@@ -141,6 +141,10 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
   // This list is needed to release all kernels retained by the
   // command_buffer.
   std::vector<ur_kernel_handle_t> KernelsList;
+  // Track whether synchronization is required when updating the command buffer
+  // Set this value to true when a command buffer is enqueued, and false after
+  // any fence or event synchronization to avoid repeated calls to synchronize.
+  bool NeedsUpdateSynchronization = false;
 };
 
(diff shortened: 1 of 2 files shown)


Bensuo:
DPC++ PR: https://github.com/intel/llvm/pull/16629

source/adapters/level_zero/command_buffer.hpp
```diff
@@ -141,6 +141,10 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
   // This list is needed to release all kernels retained by the
   // command_buffer.
   std::vector<ur_kernel_handle_t> KernelsList;
+  // Track whether synchronization is required when updating the command buffer
+  // Set this value to true when a command buffer is enqueued, and false after
+  // any fence or event synchronization to avoid repeated calls to synchronize.
+  bool NeedSynchronization = false;
```
EwanC, line 147: Slightly more verbose (and a nitpick) but slight preference for "update" in the variable name, something like `NeedsUpdateSynchronization` . There's a whole bunch of possible sync that could be going on, so think being explicit about which specific sync I think makes code a bit more readable.

EwanC (approved):

source/adapters/level_zero/command_buffer.hpp
Bensuo: Good call, done.

PR #2563, opened 2025-01-15
Add umf benchmarks: preloaded umfProxy

EuphoricThinking:
Extend UMF benchmarks with a base class for preloaded benchmarks and umfProxy benchmark.

### DIFF
--- a/scripts/benchmarks/benches/umf.py
+++ b/scripts/benchmarks/benches/umf.py
@@ -37,6 +37,7 @@ class UMFSuite(Suite):
         
         benches = [
             GBench(self),
+            GBenchUmfProxy(self),
         ]
 
         return benches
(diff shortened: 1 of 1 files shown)


scripts/benchmarks/benches/umf.py
```diff
@@ -159,3 +160,57 @@ def parse_output(self, output):
...
+                raise ValueError(f"Error parsing output: {e}")
+
+        return results
+    
+
+class GBenchUmfProxy(GBenchPreloaded):
+    def __init__(self, bench):
+        super().__init__(bench)
+
+        self.lib_to_be_replaced = "glibc"
```
pbalcer, line 211: looks like these should be constructor arguments.

scripts/benchmarks/benches/umf.py
```diff
@@ -159,3 +160,57 @@ def parse_output(self, output):
...
+class GBenchPreloaded(GBench):
+    def __init__(self, bench):
+        super().__init__(bench)
+
+        self.lib_to_be_replaced = None
+        self.replacing_lib = None 
+
+    def bin_args(self):
+        full_args = super().bin_args()
+        full_args.append("--benchmark_filter=glibc")
```
pbalcer, line 174:
```suggestion
        full_args.append(f"--benchmark_filter={self.lib_to_be_replaced}")
```
?

PR #2564, opened 2025-01-15
Update spec to make kernel validation optional

RossBrunton:
Several adapters don't support validating kernel signatures when
enqueued. To handle this, we now allow urEnqueueKernelLaunch to return
`SUCCESS` even when parameters are invalid.

### DIFF
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -7429,6 +7429,11 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventSetCallback(
 ///////////////////////////////////////////////////////////////////////////////
 /// @brief Enqueue a command to execute a kernel
 ///
+/// @details
+///     - Adapters may perform validation on the number of arguments set to the
+///       kernel, but are not required to do so and may return
+///       `::UR_RESULT_SUCCESS` even for invalid invocations.
+///
 /// @remarks
 ///   _Analogues_
 ///     - **clEnqueueNDRangeKernel**
@@ -7456,8 +7461,9 @@ UR_APIEXPORT ur_result_t UR_APICALL urEventSetCallback(
 ///     - ::UR_RESULT_ERROR_INVALID_WORK_DIMENSION
 ///     - ::UR_RESULT_ERROR_INVALID_WORK_GROUP_SIZE
 ///     - ::UR_RESULT_ERROR_INVALID_VALUE
-///     - ::UR_RESULT_ERROR_INVALID_KERNEL_ARGS - "The kernel argument values
-///     have not been specified."
+///     - ::UR_RESULT_ERROR_INVALID_KERNEL_ARGS
+///         + The kernel argument values have not been specified and the adapter
+///         is able to detect this.
 ///     - ::UR_RESULT_ERROR_OUT_OF_HOST_MEMORY
 ///     - ::UR_RESULT_ERROR_OUT_OF_RESOURCES
 UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
--- a/scripts/core/enqueue.yml
+++ b/scripts/core/enqueue.yml
@@ -16,6 +16,9 @@ type: function
 desc: "Enqueue a command to execute a kernel"
 class: $xEnqueue
 name: KernelLaunch
+details:
+  - "Adapters may perform validation on the number of arguments set to the kernel, but are not required to do so and may
+    return `$X_RESULT_SUCCESS` even for invalid invocations."
 ordinal: "0"
 analogue:
     - "**clEnqueueNDRangeKernel**"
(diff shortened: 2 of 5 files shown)


source/adapters/cuda/enqueue.cpp
```diff
@@ -465,10 +465,18 @@ enqueueKernelLaunch(ur_queue_handle_t hQueue, ur_kernel_handle_t hKernel,
...
-        CuFunc, BlocksPerGrid[0], BlocksPerGrid[1], BlocksPerGrid[2],
-        ThreadsPerBlock[0], ThreadsPerBlock[1], ThreadsPerBlock[2], LocalSize,
-        CuStream, const_cast<void **>(ArgPointers.data()), nullptr));
+    try {
+      UR_CHECK_ERROR(cuLaunchKernel(
+          CuFunc, BlocksPerGrid[0], BlocksPerGrid[1], BlocksPerGrid[2],
+          ThreadsPerBlock[0], ThreadsPerBlock[1], ThreadsPerBlock[2], LocalSize,
+          CuStream, const_cast<void **>(ArgPointers.data()), nullptr));
+    } catch (ur_result_t Err) {
+      // Cuda returns UR_RESULT_ERROR_INVALID_VALUE if the args are incorrect
```
frasercrmck, line 474: This may be true, but it's not *only* if the args are incorrect. I suspect `cuLaunchKernel` can return `CUDA_ERROR_INVALID_VALUE` -> `UR_RESULT_ERROR_INVALID_VALUE` for many reasons (the docs don't say, obviously). Would that be a problem, if we return `UR_RESULT_ERROR_INVALID_KERNEL_ARGS` mistakenly in some cases?

frasercrmck: Also just as a matter of style, should we say that "Cuda" returns a UR error code? Could that mislead people? Its' really `UR_CHECK_ERROR` that's translating `CUDA_ERROR_INVALID_VALUE` to `UR_RESULT_ERROR_INVALID_VALUE`

RossBrunton: I'm not sure there's a better option; `INVALID_VALUE` isn't a correct return for the test (since all the arguments to `enqueueKernelLaunch` are valid - It's just the arguments passed to the kernel itself that are wrong).

Should I leave it like it is in this MR (and issues can be patched when they are found), or revert this change and mark the test as XFAIL until a better solution is found?

frasercrmck: I've searched the internet a bit for `cuLaunchKernel` and `INVALID_VALUE` and all results I've seen are because people are compiling for a different architecture than the one they're running on ([1]) or they're launching with incorrect threads/dimensions ([2], [3]) or even CUDA context mismanagement ([4]). I've not seen anything saying that this error is returning when arguments are invalid.

There's also [5] which states:

```
The error CUDA_ERROR_INVALID_VALUE will be returned if kernel parameters are specified with both kernelParams and extra (i.e. both kernelParams and extra are non-NULL).
```

Of course, some of the situations I've encountered above may well never occur in the corresponding UR usage, because they'd be caught by other UR APIs.

But still my gut is telling me we're going to report false positives, which is less useful than just accepting the CUDA API's weaknesses. If we can't be sure, I'd lean towards keeping is a known test failure.

I'd take @npmiller's thoughts on this.

[[1]](https://forums.developer.nvidia.com/t/cudaapierror-1-call-to-culaunchkernel-results-in-cuda-error-invalid-value-in-python/44767/2), [[2]](https://forums.developer.nvidia.com/t/cudaapierror-1-call-to-culaunchkernel-results-in-cuda-error-invalid-value-with-numba/200699/2), [[3]](https://github.com/NVIDIA/NeMo/issues/4882#issuecomment-1238628439), [[4]](https://github.com/iree-org/iree/issues/9515), [[5]](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXEC.html#group__CUDA__EXEC_1gb8f3dc3031b40da29d5f9a7139e52e15)

npmiller: Oh that's a tough one, in practice I suspect that error code mostly comes from [1], but we can, and likely already are, ensuring the sizes are correct in the adapter before calling `cuLaunchKernel`, and it still goes through `UR_CHECK_ERROR` so the actual error code information still goes in the logger no matter what code we actually return.

That being said I think I also lean towards just keeping the error code as-is and leaving the test XFAIL, ultimately there are still some cases where this error code might get returned that aren't about the kernel arguments and that would be very confusing. Since this is optional now I think it makes more sense to just not support it for the CUDA adapter as we can't support it properly.

RossBrunton: I'll make it XFAIL for the moment, and it can be fixed properly in a separate task.

PR #2566, opened 2025-01-15
ShadowMemory should increment/decrement device handle

RossBrunton:
ShadowMemory holds a reference to the appropriate device, but did not
retain/release it properly, resulting in use-after-frees in the asan
testing.


martygrant (approved):

RossBrunton:
Merging without associated LLVM commit. In theory this could break LLVM, but in practice I think that's unlikely.

PR #2571, opened 2025-01-16
Enable the registry of level_zero_v2

omarahmed1111:
This PR enable us to register level_zero_v2 in the list of UR adapters, so it could be used with dpc++.

### DIFF
--- a/source/loader/ur_adapter_registry.hpp
+++ b/source/loader/ur_adapter_registry.hpp
@@ -121,8 +121,9 @@ private:
   // to load the adapter.
   std::vector<std::vector<fs::path>> adaptersLoadPaths;
 
-  static constexpr std::array<const char *, 5> knownAdapterNames{
+  static constexpr std::array<const char *, 6> knownAdapterNames{
       MAKE_LIBRARY_NAME("ur_adapter_level_zero", "0"),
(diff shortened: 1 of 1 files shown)


pbalcer:
The issue with this is that SYCL will now see every L0 device twice, so applications would have to be able to select one or the other.
I'm not opposed, since this is the simplest approach to exposing the v2 adapter, but I think we should gate this behind an env variable, e.g.:
`UR_ADAPTER_LEVEL_ZERO_V2=1`. If set, only v2 adapter is loaded, else we only load current adapter.

@igchor ping

pbalcer (approved):
lgtm, although long-term I think we should think how we can achieve the same thing using `ONEAPI_DEVICE_SELECTOR`.

kbenzie:
> lgtm, although long-term I think we should think how we can achieve the same thing using `ONEAPI_DEVICE_SELECTOR`.

The implementation in UR is not currently functional, unfortunately.

igchor (approved):

kbenzie:
I've taken the ready to merge tag off this, add it back again when https://github.com/intel/llvm/pull/16656 is approved.

pbalcer (approved):

Issue #2544, opened 2025-01-10
SYCL_PI_LEVEL_ZERO_USM_ALLOCATOR: presence of device in env change behavior of shared


Alexandr-Konovalov:
I'm trying to use USM pooling and the results are not seems match [the documentation](https://intel.github.io/llvm/EnvironmentVariables.html). I see that presence of device in `SYCL_PI_LEVEL_ZERO_USM_ALLOCATOR` changes pooling behavior for shared memory. I.e., with explicit device:4M (that is a default), shared memory pooling is on, while without it, there is no more pooling.

This is reproduced under Linux for release compiler 2025.0.0.20241008 as well as for manual sycl build that uses UR [dbd168cb](https://github.com/oneapi-src/unified-runtime/commit/dbd168cbed2d2590b47904728cd5762f1c2f4c6b).

```
$ cat umf.cpp
#include <sycl/sycl.hpp>

int main() {
    sycl::queue queue;
    char *buf = sycl::aligned_alloc_shared<char>(64, 8, queue);
    for (auto iteration = 0u; iteration < 4; iteration++) {
        char *buf1 = sycl::aligned_alloc_shared<char>(64, 8, queue);
        printf("%ld\n", buf - buf1);
        buf = buf1;
    }
}

$ icpx -fsycl umf.cpp
$ SYCL_PI_LEVEL_ZERO_USM_ALLOCATOR="1;device:4M;shared:64K,4,64K" SYCL_PI_LEVEL_ZERO_USM_ALLOCATOR_TRACE=3 ./a.out
USM Pool Settings (Built-in or Adjusted by Environment Variable)
      Parameter        Host      Device   Shared RW   Shared RO
    SlabMinSize       65536       65536       65536     2097152
MaxPoolableSize     2097152     4194304       65536     4194304
       Capacity           4           4           4           4
    MaxPoolSize18446744073709551615
  EnableBuffers           1

Allocated        8 Shared bytes aligned at 64 from Provider ->0x7f749b6d0000
Allocated        8 Shared bytes aligned at 64 from Pool ->0x7f749b6d0200
-512
Allocated        8 Shared bytes aligned at 64 from Pool ->0x7f749b6d0400
-512
Allocated        8 Shared bytes aligned at 64 from Pool ->0x7f749b6d0600
-512
Allocated        8 Shared bytes aligned at 64 from Pool ->0x7f749b6d0800
-512
Shared memory statistics
[...]

PR #2574, opened 2025-01-16
fix parseDisjointPoolConfig and add tests

bratpiorka:
fix parseDisjointPoolConfig() (which parses the SYCL_PI_LEVEL_ZERO_USM_ALLOCATOR environment variable) + add basic tests

this is a fix for #2544 

intel/llvm: https://github.com/intel/llvm/pull/16708

### DIFF
--- a/source/common/umf_pools/disjoint_pool_config_parser.cpp
+++ b/source/common/umf_pools/disjoint_pool_config_parser.cpp
@@ -174,47 +174,36 @@ DisjointPoolAllConfigs parseDisjointPoolConfig(const std::string &config,
     MemParser(Params, M);
   };
 
-  size_t MaxSize = (std::numeric_limits<size_t>::max)();
-
   // Update pool settings if specified in environment.
+  size_t MaxSize = (std::numeric_limits<size_t>::max)();
   size_t EnableBuffers = 1;
-  if (config != "") {
-    std::string Params = config;
-    size_t Pos = Params.find(';');
-    if (Pos != std::string::npos) {
-      if (Pos > 0) {
-        GetValue(Params, Pos, EnableBuffers);
+
+  bool EnableBuffersSet = false;
+  bool MaxSizeSet = false;
+  size_t Start = 0;
+  size_t End = config.find(';');
+  while (true) {
+    std::string Param = config.substr(Start, End - Start);
+    if (!EnableBuffersSet && (Param == "" || isdigit(Param[0]))) {
+      if (Param != "") {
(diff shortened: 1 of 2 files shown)


source/common/umf_pools/disjoint_pool_config_parser.cpp
```diff
@@ -177,46 +177,31 @@ DisjointPoolAllConfigs parseDisjointPoolConfig(const std::string &config,
...
-                } while (true);
-            } else {
-                // set MaxPoolSize for all configs
-                GetValue(Params, Params.size(), MaxSize);
-            }
+    bool EnableBuffersSet = false;
+    bool MaxSizeSet = false;
+    size_t Start = 0;
+    size_t End = config.find(';');
+    while (true) {
```
igchor, line 187: We actually already have a generic parser for env variables here: https://github.com/oneapi-src/unified-runtime/blob/main/source/common/ur_util.hpp#L212C33-L212C46

Can we use it? We could replace the entire logic above as well. Probably we would need to make a version of getenv_to_vec that supports ';' or just accepts strings instead of env var names and we would do the splitting here.

bratpiorka: I thought about it, but in SYCL_PI_LEVEL_ZERO_USM_ALLOCATOR we got a mix of named and unnamed params.
So neither getenv_to_map nor getenv_to_vec will work without changes. It would be better to change the definition of env var instead, but this PR is just a simple fix.

igchor: The only unnamed param is the first option (EnableBuffers) - we should be able to apply getenv_to_map on the rest of the string, right?

igchor: I know that currently getenv_to_map  accepts a name of the env var, but it should be fairly simple to refactor out a function that accepts a string instead.

bratpiorka: we have two unnamed params - there is also MaxPoolSize

igchor: ah right, but it would still be fairly easy to extract the string to pass to getenv_to_map (or a string-accepting) version - just find a ';' followed by a letter (if any).

pbalcer (approved):

kbenzie:
Pulled in by https://github.com/intel/llvm/pull/16791 instead.

PR #2583, opened 2025-01-20
 Fix Coverity COPY_INSTEAD_OF_MOVE defects.

isaacault:


pbalcer (approved):
I'm fine with the change, but this shouldn't really make a difference so ignoring the coverity issue for these types of issues, in non-perf critical places like this, in the future might be a better option.

PR #2584, opened 2025-01-20
Fix Coverity INVALIDATE_ITERATOR defects.

isaacault:


source/common/ur_singleton.hpp
```diff
@@ -76,21 +76,25 @@ template <typename singleton_tn, typename key_tn> class singleton_factory_t {
 
   void retain(key_tn key) {
     std::lock_guard<std::mutex> lk(mut);
-    auto iter = map.find(getKey(key));
-    assert(iter != map.end());
-    iter->second.ref_count++;
+    if (auto iter = map.find(getKey(key)); iter != map.end()) {
+      iter->second.ref_count++;
+    } else {
+      __builtin_unreachable();
```
pbalcer, line 82: there's a function:
https://github.com/oneapi-src/unified-runtime/blob/8b7a99578966eb691a961d9620ea38d235196b2f/source/common/ur_util.hpp#L339

RossBrunton, line 82: I'm not sure removing the assert (what this is functionally equivalent to) is the correct resolution here - it's possible for an error in UR code to retain an error value, and I think flagging that issue in a debug build is useful. I don't think this also resolves the Coverity error either.

If we do want to improve the error checking and have it work in release mode, I think calling `abort()` with an error message is the best move.

isaacault: Done. Error message sent to the logger.

isaacault: No longer using unreachable in favour of `abort()` - see below.

RossBrunton:
This file will be removed by https://github.com/oneapi-src/unified-runtime/pull/2622

martygrant:

---

PR #2585, opened 2025-01-20
[DeivceASAN] Make ShadowMemory one instance per type

yingcong-wu:
Previously we had one `ShadowMemory` for every `DeviceInfo`. That is causing problem because it's release process is called too late. And only the first `ShadowMemory` would do the actual shadow memory bookkeeping.

This is a quick fix but also is heading in the direction we plan, which is one `ShadowMemory` for each device type. This relies on the fact that for the virtual memory with L0, it can access the address from different device and within different context with newer driver. And as for opencl CPU device, since it is host memory, it does not have to worry about this issue.

We will purpose PR later to do a better refactor, but for now we are trying to fix the issue with minimum changes to avoid any possible regression.

Intel/llvm PR: https://github.com/intel/llvm/pull/16687

### DIFF
--- a/source/loader/layers/sanitizer/asan/asan_interceptor.cpp
+++ b/source/loader/layers/sanitizer/asan/asan_interceptor.cpp
@@ -310,21 +310,19 @@ ur_result_t AsanInterceptor::postLaunchKernel(ur_kernel_handle_t Kernel,
 std::shared_ptr<ShadowMemory>
 AsanInterceptor::getOrCreateShadowMemory(ur_device_handle_t Device,
                                          DeviceType Type) {
+  std::scoped_lock<ur_shared_mutex> Guard(m_ShadowMapMutex);
   if (m_ShadowMap.find(Type) == m_ShadowMap.end()) {
-    std::scoped_lock<ur_shared_mutex> Guard(m_ShadowMapMutex);
-    if (m_ShadowMap.find(Type) == m_ShadowMap.end()) {
-      ur_context_handle_t InternalContext;
-      auto Res = getContext()->urDdiTable.Context.pfnCreate(1, &Device, nullptr,
-                                                            &InternalContext);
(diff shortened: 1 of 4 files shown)


yingcong-wu:
Hi @unified-runtime-maintain, this patch is targeting v0.11, please help tag this.

source/loader/layers/sanitizer/asan/asan_interceptor.cpp
```diff
@@ -36,8 +36,7 @@ AsanInterceptor::~AsanInterceptor() {
   // We must release these objects before releasing adapters, since
   // they may use the adapter in their destructor
   for (const auto &[_, DeviceInfo] : m_DeviceMap) {
-    [[maybe_unused]] auto URes = DeviceInfo->Shadow->Destory();
-    assert(URes == UR_RESULT_SUCCESS);
+    DeviceInfo->Shadow = nullptr;
```
AllanZyne, line 39: clear the "m_DeviceMap" directly?

yingcong-wu: Yes, that should also work.

wenju-he:
> called to late.

typo: too late

yingcong-wu:
> > called to late.
> 
> typo: too late

Updated, thanks.

yingcong-wu:
Thank you @pbalcer , we will hold this PR for now to discuss an issue during our meeting tomorrow.

zhaomaosu (approved):
lgtm

AllanZyne (approved):

yingcong-wu:
Some CI tasks fail at apt install stage.

pbalcer:
> Some CI tasks fail at apt install stage. Suspecting infrastructure problem.

Yeah, looks like GitHub infra issue.

PR #2595, opened 2025-01-21
Reduce CI matrix in preparation for repo move

kbenzie:
* Remove build only Linux and Windows jobs
* Split out SYCL E2E workflow
* Split out fuzz testing workflow
* Only do a single UR CTS run per adapter config
* Rename `cmake.yml` to `unified-runtime.yml`
* Convert `source-checks.yml` into a reusable workflow & trigger from `unified-runtinme.yml`


.github/workflows/build-hw-reusable.yml
```diff
@@ -40,19 +40,21 @@ env:
...
+            name: "${{inputs.adapter_name}}",
+            other_name: "${{inputs.other_adapter_name}}",
+            platform: "${{inputs.platform}}",
+            static_Loader: "${{inputs.static_loader}}",
+            static_adapter: "${{inputs.static_loader}}"
+          }
+        ]
+        build_type: [Release]
+        compiler: [{c: gcc, cxx: g++}]
         # TODO: The latest L0 loader segfaults when built with clang.
```
lukaszstolarczuk, line 58: the exclusions below with clang should be removed now

.github/workflows/unified-runtime.yml
```diff
@@ -0,0 +1,119 @@
...
+
+on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
```
lukaszstolarczuk, line 11: not sure why `pull-requests: write` was here, but most likely should be removed

.github/workflows/build-hw-reusable.yml
lukaszstolarczuk: hmm maybe even all exclusions should be removed? something here is causing L0 static job failing, I think

kbenzie: I've removed those, we'll see how that affects the L0 static job.

.github/workflows/unified-runtime.yml
kbenzie: Removed.

lukaszstolarczuk (approved):
LGTM

martygrant (approved):

PR #2597, opened 2025-01-22
Improvements to align CTS and Spec for Device

martygrant:
- Rework urDeviceGetInfoTest to move all enums to their own tests instead of a switch case - https://github.com/oneapi-src/unified-runtime/issues/2290
- Remove the UR/OpenCL Device enum map function - no point maintaining both a switch case and a separate mapping function
- Update some spec wording for consistency
- Add missing Device info enums to OpenCL adapater
- Add new urDevicePartition test for checking UR_DEVICE_INFO_PARENT_DEVICE
- Move UUR_RETURN_ON_FATAL_FAILURE and UUR_ASSERT_SUCCESS_OR_UNSUPPORTED to join similar macros in checks.h
- Update spec to say info queries that return char[] are null-terminated
- Add missing unsupported device enums to HIP adapter
- Reverted bindless image types back to size_t
- Moved ComponentDevices test to a separate PlatformTest derived fixture
- Added new ASSERT_QUERY_RETURNS_VALUE macro as a helper to validate simple numerical GetInfo queries
- Added new stringPropertyIsValid helper function to validate string queries
- Add missing unsupported device enums to HIP adapter

https://github.com/intel/llvm/pull/16746

### DIFF
--- a/scripts/core/exp-bindless-images.yml
+++ b/scripts/core/exp-bindless-images.yml
@@ -88,19 +88,19 @@ etors:
       desc: "[$x_bool_t] returns true if the device supports sampling cubemapped images across face boundaries"
     - name: BINDLESS_SAMPLED_IMAGE_FETCH_1D_USM_EXP
       value: "0x2012"
-      desc: "[$x_bool_t] returns true if the device is capable of fetching USM backed 1D sampled image data."
+      desc: "[$x_bool_t] returns true if the device supports fetching USM backed 1D sampled image data."
     - name: BINDLESS_SAMPLED_IMAGE_FETCH_1D_EXP
       value: "0x2013"
-      desc: "[$x_bool_t] returns true if the device is capable of fetching non-USM backed 1D sampled image data."
+      desc: "[$x_bool_t] returns true if the device supports fetching non-USM backed 1D sampled image data."
     - name: BINDLESS_SAMPLED_IMAGE_FETCH_2D_USM_EXP
       value: "0x2014"
-      desc: "[$x_bool_t] returns true if the device is capable of fetching USM backed 2D sampled image data."
+      desc: "[$x_bool_t] returns true if the device supports fetching USM backed 2D sampled image data."
     - name: BINDLESS_SAMPLED_IMAGE_FETCH_2D_EXP
       value: "0x2015"
-      desc: "[$x_bool_t] returns true if the device is capable of fetching non-USM backed 2D sampled image data."
+      desc: "[$x_bool_t] returns true if the device supports fetching non-USM backed 2D sampled image data."
     - name: BINDLESS_SAMPLED_IMAGE_FETCH_3D_EXP
       value: "0x2017"
-      desc: "[$x_bool_t] returns true if the device is capable of fetching non-USM backed 3D sampled image data."
+      desc: "[$x_bool_t] returns true if the device supports fetching non-USM backed 3D sampled image data."
     - name: IMAGE_ARRAY_SUPPORT_EXP
       value: "0x2019"
       desc: "[$x_bool_t] returns true if the device supports allocating and accessing image array resources."
@@ -109,10 +109,10 @@ etors:
       desc: "[$x_bool_t] returns true if the device supports unique addressing per dimension."
     - name: BINDLESS_SAMPLE_1D_USM_EXP
       value: "0x201B"
-      desc: "[$x_bool_t] returns true if the device is capable of sampling USM backed 1D sampled image data."
+      desc: "[$x_bool_t] returns true if the device supports sampling USM backed 1D sampled image data."
     - name: BINDLESS_SAMPLE_2D_USM_EXP
       value: "0x201C"
-      desc: "[$x_bool_t] returns true if the device is capable of sampling USM backed 2D sampled image data."
+      desc: "[$x_bool_t] returns true if the device supports sampling USM backed 2D sampled image data."
 --- #--------------------------------------------------------------------------
 type: enum
 extend: true
--- a/scripts/core/program.yml
+++ b/scripts/core/program.yml
@@ -381,11 +381,11 @@ etors:
           [$x_device_handle_t[]] Return list of devices associated with a program.
           This is either the list of devices associated with the context or a subset of those devices when the program is created using $xProgramCreateWithBinary.
     - name: IL
-      desc: "[char[]] Return program IL if the program was created with $xProgramCreateWithIL, otherwise return size will be set to 0 and nothing will be returned."
+      desc: "[char[]] Return program IL if the program was created with $xProgramCreateWithIL, otherwise return size will be set to 0 and nothing will be returned. This is not null-terminated."
     - name: BINARY_SIZES
       desc: "[size_t[]] Return program binary sizes for each device."
     - name: BINARIES
-      desc: "[unsigned char[]] Return program binaries for all devices for this Program."
+      desc: "[unsigned char[]] Return program binaries for all devices for this Program. These are not null-terminated."
     - name: NUM_KERNELS
       desc: "[size_t][optional-query] Number of kernels in Program, return type size_t."
     - name: KERNEL_NAMES
--- a/test/conformance/device/urDeviceGetInfo.cpp
+++ b/test/conformance/device/urDeviceGetInfo.cpp
@@ -10,335 +10,2558 @@
 #include <uur/fixtures.h>
 #include <uur/known_failure.h>
 
-static std::unordered_map<ur_device_info_t, size_t> device_info_size_map = {
-    {UR_DEVICE_INFO_TYPE, sizeof(ur_device_type_t)},
-    {UR_DEVICE_INFO_VENDOR_ID, sizeof(uint32_t)},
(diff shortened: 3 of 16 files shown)


scripts/core/program.yml
```diff
@@ -381,11 +381,11 @@ etors:
           [$x_device_handle_t[]] Return list of devices associated with a program.
           This is either the list of devices associated with the context or a subset of those devices when the program is created using $xProgramCreateWithBinary.
     - name: IL
-      desc: "[char[]] Return program IL if the program was created with $xProgramCreateWithIL, otherwise return size will be set to 0 and nothing will be returned."
+      desc: "[char[]] Return null-terminated program IL if the program was created with $xProgramCreateWithIL, otherwise return size will be set to 0 and nothing will be returned."
```
aarongreig, line 384: the program ones aren't null terminated, you can have a null byte in a program

scripts/core/exp-bindless-images.yml
```diff
@@ -55,13 +55,13 @@ etors:
...
       value: "0x2005"
-      desc: "[size_t] returns the maximum linear width allowed for images allocated using USM"
+      desc: "[uint32_t] returns the maximum linear width allowed for images allocated using USM"
     - name: MAX_IMAGE_LINEAR_HEIGHT_EXP
       value: "0x2006"
-      desc: "[size_t] returns the maximum linear height allowed for images allocated using USM"
+      desc: "[uint32_t] returns the maximum linear height allowed for images allocated using USM"
     - name: MAX_IMAGE_LINEAR_PITCH_EXP
       value: "0x2007"
-      desc: "[size_t] returns the maximum linear pitch allowed for images allocated using USM"
```
przemektmalon, line 64: These should be `size_t`, the types in the SYCL RT are incorrect and don't reflect the BI extension spec.

I've created a PR changing these in the RT here: https://github.com/intel/llvm/pull/16829

(assuming that the ABI exception for experimental features applies, and the above PR is merged, these changes can be reverted)

test/conformance/device/urDeviceGetInfo.cpp
```diff
@@ -10,392 +10,2645 @@
...
   ASSERT_SUCCESS(
-      urDeviceGetInfo(device, info_type, size, info_data.data(), nullptr));
+      urDeviceGet(platform, UR_DEVICE_TYPE_ALL, 0, nullptr, &device_count));
+  ASSERT_GT(device_count, 0);
+
+  std::vector<ur_device_handle_t> devices(device_count);
+  ASSERT_SUCCESS(urDeviceGet(platform, UR_DEVICE_TYPE_ALL, device_count,
+                             devices.data(), nullptr));
+
+  for (const ur_device_handle_t device : devices) {
```
aarongreig, line 1984: I think this should have its own `PlatformTest` derived fixture, right now it's effectively doing
```
for (const ur_device_handle_t device : devices) {
  for (const ur_device_handle_t device : devices) {
```

test/conformance/device/urDeviceGetInfo.cpp
```diff
@@ -10,392 +10,2645 @@
...
+
+  ASSERT_SUCCESS_OR_OPTIONAL_QUERY(
+      urDeviceGetInfo(device, property_name, 0, nullptr, &property_size),
+      property_name);
+  ASSERT_GT(property_size, 0);
+
+  std::vector<char> property_value(property_size, 'x');
+  ASSERT_SUCCESS(urDeviceGetInfo(device, property_name, property_size,
+                                 property_value.data(), nullptr));
+  EXPECT_NE(property_value[0], 'x');
```
aarongreig, line 1532: Could we do this by initializing the vector with null characters instead? unless there's a spec somewhere that says the first byte of a pci address can't be `'x'`. Same goes for other queries with this check

test/conformance/device/urDeviceGetInfo.cpp
```diff
@@ -10,392 +10,2645 @@
...
+  ASSERT_SUCCESS_OR_OPTIONAL_QUERY(
+      urDeviceGetInfo(device, property_name, 0, nullptr, &property_size),
+      property_name);
+  ASSERT_EQ(property_size, sizeof(size_t));
+
+  size_t property_value = 999;
+  ASSERT_SUCCESS(urDeviceGetInfo(device, property_name, property_size,
+                                 &property_value, nullptr));
+
+  ASSERT_NE(property_value, 999);
```
aarongreig, line 1247: Unless the queries in question have specifically defined valid ranges I don't think we should be doing checks like this. 999 is unlikely but not necessarily invalid. I realize having no certainty as to what value we're expecting makes it hard to test if the adapter is returning anything at all.. maybe we could do have a validation helper for these cases like this:

```
size_t property_value = 0;
(do query)
auto returned_value = property_value;
property_value = 1;
(do query again)
ASSERT_EQ(property_value, returned_value);
```
that way we're sure the adapter is returning a value without putting any undocumented restrictions on what the value is

scripts/core/exp-bindless-images.yml
przemektmalon: https://github.com/intel/llvm/pull/16829 has been merged.

This change can be reverted.

test/conformance/device/urDeviceGetInfo.cpp
martygrant: totally makes sense but might this be a false positive if the value returned really is 1? what about using `std::numeric_limits<size_t>::max()` as an initial default value?

aarongreig: it's set up so that the initial value doesn't matter, as long as it changes between the two queries e.g.:
```
size_t property_value = 0;
(do query) - property_value is now 1 - verifying that the adapter is returning something
auto returned_value = property_value; // returned_value is now 1
property_value = 1; // change default value  - if the adapter returned 0 the first time this second call will verify that it's returning a value when it overwrites
(do query again) - property_value remains 1
ASSERT_EQ(property_value, returned_value); // this ensures the return is consistent between the two calls
```

scripts/core/exp-bindless-images.yml
martygrant: I've reverted these back to `size_t`, thanks!

test/conformance/device/urDeviceGetInfo.cpp
martygrant: Added a new test suite for this, using `urAllDevicesTest`

test/conformance/device/urDeviceGetInfo.cpp
martygrant: Good idea, an issue was pointed out with testing char returns in https://github.com/oneapi-src/unified-runtime/pull/2626#discussion_r1935549342 and I've added a new helper function `stringPropertyIsValid` for this

test/conformance/device/urDeviceGetInfo.cpp
martygrant: Thanks for this, have added a new macro `ASSERT_GET_INFO` as a helper to validate these kind of queries

scripts/core/exp-bindless-images.yml
przemektmalon: Thanks! One last thing, is to update the PR description and remove the line:

> Update UR_DEVICE_INFO_MAX_IMAGE_LINEAR_WIDTH/HEIGHT/PITCH should be uint32_t, matching SYCL RT

Otherwise, bindless images changes LGTM

przemektmalon (approved):

test/conformance/testing/include/uur/checks.h
```diff
@@ -69,4 +110,15 @@ inline std::ostream &operator<<(std::ostream &out,
   return out;
 }
 
+#ifndef ASSERT_GET_INFO
+#define ASSERT_GET_INFO(CALL, TYPE)                                            \
+  do {                                                                         \
+    ASSERT_SUCCESS(CALL);                                                      \
+    TYPE returned_value = property_value;                                      \
```
aarongreig, line 117: Not sure about the assumption that we're going to have a variable called `property_value` declared before using the macro. Could we make it take CALL, PROPERTY_VALUE and then this line looks like
```suggestion
    auto returned_value = PROPERTY_VALUE;                                      \
```
I also would prefer a more specific name, maybe like `ASSERT_QUERY_RETURNS_VALUE`

martygrant: that makes sense thanks, I've changed the macro over now

aarongreig (approved):

martygrant:
@oneapi-src/unified-runtime-cuda-write @oneapi-src/unified-runtime-hip-write @oneapi-src/unified-runtime-level-zero-write @oneapi-src/unified-runtime-native-cpu-write friendly bump for a review please, the adapter changes are very minor

frasercrmck (approved):

coldav:
LGTM native cpu

coldav (approved):

pbalcer (approved):

martygrant:
## Information

## PR Migration

---

martygrant:
Migrated to https://github.com/intel/llvm/pull/16746

PR #2598, opened 2025-01-22
DO NOT MERGE: Improved lifetime handling and docs of loader

RossBrunton:
* The loader now checks that all handles have been released before
  dropping factories.
* Some arguments of functions have been marked as [alloc]. When the
  loader generates a handle from them, they will have an initial
  reference count that requires a single release.
* Handles that are assigned to non-alloc pointers must exist and their
  reference counts are not incremented.


RossBrunton:
Not needed with removal of loader handles entirely.

PR #2601, opened 2025-01-22
In docs, mark "allocating" functions as such

RossBrunton:
Mark all functions in the API who allocate handles (which must later be
free'd with `Release*` or similar with an `[alloc]` annotation.

----

I've marked the functions I feel should be free'd with the "Release" family of functions, mostly functions named "Create"
or that return events, but I'm not 100% sure I've not miscategorized any. Any feedback is appreciated.

### DIFF
--- a/scripts/core/exp-bindless-images.yml
+++ b/scripts/core/exp-bindless-images.yml
@@ -429,7 +429,7 @@ params:
       desc: "[in] pointer to image description"
     - type: $x_exp_image_mem_native_handle_t*
       name: phImageMem
-      desc: "[out] pointer to handle of image memory allocated"
+      desc: "[out][alloc] pointer to handle of image memory allocated"
 returns:
     - $X_RESULT_ERROR_INVALID_CONTEXT
     - $X_RESULT_ERROR_INVALID_VALUE
@@ -484,7 +484,7 @@ params:
       desc: "[in] pointer to image description"
     - type: $x_exp_image_native_handle_t*
       name: phImage
-      desc: "[out] pointer to handle of image object created"
+      desc: "[out][alloc] pointer to handle of image object created"
 returns:
     - $X_RESULT_ERROR_INVALID_CONTEXT
     - $X_RESULT_ERROR_INVALID_VALUE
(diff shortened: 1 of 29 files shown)


RossBrunton:
@oneapi-src/unified-runtime-command-buffer-write @oneapi-src/unified-runtime-bindless-images-write 

I've tagged some command buffer/bindless image functions as "alloc", which means the returned handles must be freed with a call to Release/Destroy/whatever. However, I'm not too familiar with everything, so could you give it a quick look and make sure I've got everything right for your projects?

konradkusiak97 (approved):
Looks good to me

scripts/core/exp-bindless-images.yml
```diff
@@ -648,7 +648,7 @@ params:
       desc: "[in] requested level of the mipmap"
     - type: "$x_exp_image_mem_native_handle_t*"
       name: phImageMem
-      desc: "[out] returning memory handle to the individual image"
+      desc: "[out][alloc] returning memory handle to the individual image"
```
przemektmalon, line 651:
```suggestion
      desc: "[out] returning memory handle to the individual image"
```

`MipmapGetLevelExp` doesn't allocate any handles that need to be freed, it returns a handle to a sub-resource, which itself is freed when the top-level resource is freed (it cannot be freed individually).

przemektmalon (changes requested):

scripts/core/exp-bindless-images.yml
RossBrunton: Thanks for checking; updated.

przemektmalon (approved):
Bindless Images changes LGTM.

martygrant (approved):
LGTM.

Just a thought: what about having one annotation called "counted" or "ref_counted" or something like that and merge the description of the new "alloc" and "retain" annotations? Just so the list of annotations is smaller.

RossBrunton:
> LGTM.
> 
> Just a thought: what about having one annotation called "counted" or "ref_counted" or something like that and merge the description of the new "alloc" and "retain" annotations? Just so the list of annotations is smaller.

I think `alloc` and `retain` have different meanings; `alloc` creates a new resource (in theory, anyway) whilst `retain` must be given a valid existing resource.

PR #1811, opened 2024-07-02
[UR][CUDA] Add tensor map APIs

hdelan:
Intended to target the APIs [here](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TENSOR__MEMORY.html)

PR #2610, opened 2025-01-23
Revert "[UR][CUDA] Add tensor map APIs"

npmiller:
This patch reverts the following PR and its follow up build fix:
* https://github.com/oneapi-src/unified-runtime/pull/1811
* https://github.com/oneapi-src/unified-runtime/pull/2502

After further consideration it was decided that the only known use case for this extension would be better handled through interop, as a result this extension is no longer needed. This extension is CUDA specific and has no other known use cases than the aforementioned extension, which has now been dropped, so hopefully it should be fine to just remove.


ldrumm (approved):

Naghasan (approved):
Related closed PR https://github.com/intel/llvm/pull/16247

kbenzie (approved):

kbenzie:
I've added the *ready to merge* label as the changes to Level Zero and Native CPU adapters is minimal. If either of @oneapi-src/unified-runtime-level-zero-write or @oneapi-src/unified-runtime-native-cpu-write object, please remove the label. Otherwise I'll aim to merge this later this week.

PR #2612, opened 2025-01-23
[L0]: fix missing destroy of event given enqueue wait out event

nrspruit:
- Added check during event handle release to perform an additional
  cleanup of an enqueue wait event once the user has released their
usage of the out event.

### DIFF
--- a/source/adapters/level_zero/event.cpp
+++ b/source/adapters/level_zero/event.cpp
@@ -887,7 +887,14 @@ ur_result_t
 
 urEventRelease(/** [in] handle of the event object */ ur_event_handle_t Event) {
   Event->RefCountExternal--;
+  bool isEventsWaitCompleted =
+      Event->CommandType == UR_COMMAND_EVENTS_WAIT && Event->Completed;
   UR_CALL(urEventReleaseInternal(Event));
+  // If this is a Completed Event Wait Out Event, then we need to cleanup the
+  // event at user release and not at the time of completion.
+  if (isEventsWaitCompleted) {
+    UR_CALL(CleanupCompletedEvent((Event), false, false));
+  }
 
   return UR_RESULT_SUCCESS;
 }


source/adapters/level_zero/event.cpp
```diff
@@ -882,6 +882,12 @@ urEventRelease(ur_event_handle_t Event ///< [in] handle of the event object
 ) {
   Event->RefCountExternal--;
   UR_CALL(urEventReleaseInternal(Event));
+  // If this is a Completed Event Wait Out Event, then we need to cleanup the
+  // event at user release and not at the time of completion.
+  if (Event->CommandType == UR_COMMAND_EVENTS_WAIT && Event->Completed) {
```
pbalcer, line 887: isn't this a potential use-after-free?

nrspruit: I retested including while disabling the event caching and I found that the additional release was not needed, just the cleanup call.

Please re-review, the updated changed still solves the issue, while ensuring no use after free. The fix ensures that the event is "cleaned up" after it has been released by the user in this case instead of having a floating event with no reuse or destroy. internally, the cleanup call calls release internal such that the event is either destroyed or is added to the cache now that the user is done with it.

winstonzhang-intel (approved):

source/adapters/level_zero/event.cpp
```diff
@@ -882,6 +882,12 @@ urEventRelease(ur_event_handle_t Event ///< [in] handle of the event object
 ) {
   Event->RefCountExternal--;
   UR_CALL(urEventReleaseInternal(Event));
+  // If this is a Completed Event Wait Out Event, then we need to cleanup the
+  // event at user release and not at the time of completion.
+  if (Event && Event->CommandType == UR_COMMAND_EVENTS_WAIT &&
```
pbalcer, line 887: Event is not taken by a reference by urEventReleaseInternal, so `Event` can't become NULL here. This doesn't protect us from use-after-free.
My primary concern is the event being deleted here:
https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/event.cpp#L1143

Which can occur e.g., with interop events where ownership is not passed over to UR.

nrspruit: Updated to save the details on the event before the event release to avoid a use after delete possible scenario.

PR #2614, opened 2025-01-24
Add UR_KERNEL_INFO_SPILL_MEM_SIZE kernel info prop

kurapov-peter:
This allows for querying the information about produced register spills during kernel compilation. The query is optional and is only implemented for L0 now. The direct counterpart for the query is [spillMemSize](https://spec.oneapi.io/level-zero/1.9.3/core/api.html#_CPPv4N22ze_kernel_properties_t12spillMemSizeE) property returned by `zeKernelGetProperties`.

### DIFF
--- a/source/adapters/cuda/kernel.cpp
+++ b/source/adapters/cuda/kernel.cpp
@@ -299,6 +299,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urKernelGetInfo(ur_kernel_handle_t hKernel,
                                       hKernel->get()));
     return ReturnValue(static_cast<uint32_t>(NumRegs));
   }
+  case UR_KERNEL_INFO_SPILL_MEM_SIZE:
   default:
     break;
   }
--- a/source/adapters/hip/kernel.cpp
+++ b/source/adapters/hip/kernel.cpp
@@ -234,6 +234,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urKernelGetInfo(ur_kernel_handle_t hKernel,
                                        hKernel->get()));
     return ReturnValue(static_cast<uint32_t>(NumRegs));
   }
+  case UR_KERNEL_INFO_SPILL_MEM_SIZE:
   default:
     break;
   }
--- a/source/adapters/level_zero/kernel.cpp
+++ b/source/adapters/level_zero/kernel.cpp
@@ -751,6 +751,8 @@ ur_result_t urKernelGetInfo(
   case UR_KERNEL_INFO_NUM_REGS:
   case UR_KERNEL_INFO_NUM_ARGS:
     return ReturnValue(uint32_t{Kernel->ZeKernelProperties->numKernelArgs});
+  case UR_KERNEL_INFO_SPILL_MEM_SIZE:
+    return ReturnValue(uint32_t{Kernel->ZeKernelProperties->spillMemSize});
   case UR_KERNEL_INFO_REFERENCE_COUNT:
     return ReturnValue(uint32_t{Kernel->RefCount.load()});
   case UR_KERNEL_INFO_ATTRIBUTES:
--- a/source/adapters/level_zero/v2/kernel.cpp
+++ b/source/adapters/level_zero/v2/kernel.cpp
@@ -624,6 +624,8 @@ ur_result_t urKernelGetInfo(ur_kernel_handle_t hKernel,
   case UR_KERNEL_INFO_NUM_REGS:
   case UR_KERNEL_INFO_NUM_ARGS:
     return ReturnValue(uint32_t{hKernel->getCommonProperties().numKernelArgs});
+  case UR_KERNEL_INFO_SPILL_MEM_SIZE:
+    return ReturnValue(uint32_t{hKernel->getProperties().spillMemSize});
   case UR_KERNEL_INFO_REFERENCE_COUNT:
     return ReturnValue(uint32_t{hKernel->RefCount.load()});
   case UR_KERNEL_INFO_ATTRIBUTES: {
--- a/test/conformance/kernel/urKernelGetInfo.cpp
+++ b/test/conformance/kernel/urKernelGetInfo.cpp
@@ -136,6 +136,22 @@ TEST_P(urKernelGetInfoTest, SuccessNumRegs) {
                                  property_value.data(), nullptr));
 }
 
+TEST_P(urKernelGetInfoTest, SuccessSpillMemSize) {
+  UUR_KNOWN_FAILURE_ON(uur::HIP{}, uur::OpenCL{});
+
+  ur_kernel_info_t property_name = UR_KERNEL_INFO_SPILL_MEM_SIZE;
+  size_t property_size = 0;
+
+  ASSERT_SUCCESS_OR_OPTIONAL_QUERY(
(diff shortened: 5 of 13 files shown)


source/adapters/cuda/kernel.cpp
```diff
@@ -299,6 +299,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urKernelGetInfo(ur_kernel_handle_t hKernel,
                                       hKernel->get()));
     return ReturnValue(static_cast<uint32_t>(NumRegs));
   }
+  case UR_KERNEL_INFO_SPILL_MEM_SIZE:
```
kbenzie, line 302: This should return `UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION` rather than `UR_RESULT_ERROR_INVALID_ENUMERATION`.

source/adapters/hip/kernel.cpp
```diff
@@ -234,6 +234,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urKernelGetInfo(ur_kernel_handle_t hKernel,
                                        hKernel->get()));
     return ReturnValue(static_cast<uint32_t>(NumRegs));
   }
+  case UR_KERNEL_INFO_SPILL_MEM_SIZE:
```
kbenzie, line 237: This should return `UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION` rather than `UR_RESULT_ERROR_INVALID_ENUMERATION`.

source/adapters/cuda/kernel.cpp
aarongreig, line 302: for here and in the equivalent file in the hip adapter you'll need to explicitly add a `return UR_RESULT_ERROR_UNSUPPORTED_ENUMERATION;` - right now it's falling through to `return UR_RESULT_ERROR_INVALID_ENUMERATION;` which isn't the expected return for an unsupported optional query

aarongreig (approved):

kbenzie (commented):
This should also update `source/adapters/native_cpu/kernel.cpp`.

kurapov-peter:
> This should also update `source/adapters/native_cpu/kernel.cpp`.

I see that in the native cpu adapter some other properties are also missing, e.g., `UR_KERNEL_INFO_NUM_REGS`. Should I add that too?

kbenzie:
> > This should also update `source/adapters/native_cpu/kernel.cpp`.
> 
> I see that in the native cpu adapter some other properties are also missing, e.g., `UR_KERNEL_INFO_NUM_REGS`. Should I add that too?

Just this one should be fine, it's WIP.

kurapov-peter:
@pbalcer, what would be a proper way of implementing the v2 portion? The API now doesn't include the device handle which is needed to get the property through `ur_kernel_handle_t_::getProperties(ur_device_handle_t hDevice)`. Returning a map for each device also doesn't seem like an option, as it breaks the interface.

pbalcer (approved):

source/adapters/cuda/kernel.cpp
kurapov-peter: done

kurapov-peter: done

source/adapters/level_zero/v2/kernel.cpp
```diff
@@ -624,6 +624,16 @@ ur_result_t urKernelGetInfo(ur_kernel_handle_t hKernel,
   case UR_KERNEL_INFO_NUM_REGS:
   case UR_KERNEL_INFO_NUM_ARGS:
     return ReturnValue(uint32_t{hKernel->getCommonProperties().numKernelArgs});
+  case UR_KERNEL_INFO_SPILL_MEM_SIZE: {
+    std::shared_lock<ur_shared_mutex> Guard(hKernel->getProgramHandle()->Mutex);
```
kurapov-peter, line 628: I took this from another similar implementation, assuming this protects the device retrieval call. Not sure if necessary, though.

source/adapters/level_zero/v2/kernel.cpp
```diff
@@ -624,6 +624,16 @@ ur_result_t urKernelGetInfo(ur_kernel_handle_t hKernel,
...
+    auto devices = hKernel->getProgramHandle()->AssociatedDevices;
+    std::vector<uint32_t> spills(devices.size());
+    for (size_t i = 0; i < spills.size(); ++i) {
+      spills[i] = uint32_t{hKernel->getProperties(devices[i]).spillMemSize};
+    }
+    return ReturnValue(static_cast<const uint32_t *>(spills.data()),
+                       spills.size());
```
kurapov-peter, line 635: This relies on the `ReturnValue` to preserve the memory for the vector.

kurapov-peter:
Looks like the only failure is a formatting issue, but I don't see any errors in the log. Is my yaml ill-formatted or something?

aarongreig:
> Looks like the only failure is a formatting issue, but I don't see any errors in the log. Is my yaml ill-formatted or something?

it looks like the description in the yml is out of sync with the one in `ur_api.h`, I think you just need to re-run the `generate` target

source/adapters/level_zero/kernel.cpp
```diff
@@ -751,6 +751,10 @@ ur_result_t urKernelGetInfo(
   case UR_KERNEL_INFO_NUM_REGS:
   case UR_KERNEL_INFO_NUM_ARGS:
     return ReturnValue(uint32_t{Kernel->ZeKernelProperties->numKernelArgs});
+  case UR_KERNEL_INFO_SPILL_MEM_SIZE:
+    return ReturnValue(static_cast<const uint32_t *>(
```
pbalcer, line 755: This should be similar to v2 implementation.

jchlanda (approved):
CUDA/HIP 👍

test/conformance/kernel/urKernelGetInfo.cpp
```diff
@@ -136,6 +136,22 @@ TEST_P(urKernelGetInfoTest, SuccessNumRegs) {
...
+
+  ur_kernel_info_t property_name = UR_KERNEL_INFO_SPILL_MEM_SIZE;
+  size_t property_size = 0;
+
+  ASSERT_SUCCESS_OR_OPTIONAL_QUERY(
+      urKernelGetInfo(kernel, property_name, 0, nullptr, &property_size),
+      property_name);
+  ASSERT_EQ(property_size, sizeof(uint32_t));
+
+  std::vector<char> property_value(property_size);
```
pbalcer, line 150:
```suggestion
  std::vector<uint32_t> property_value(property_size / sizeof(uint32_t));
```

pbalcer (approved):

pbalcer:
I removed the "ready to merge" label because the intel/llvm part of this PR still has required reviews. The process is that we wait until both PRs are fully reviewed.

kurapov-peter:
FYI: There are no changes in that part of the PR, only the cmake changes for CI to pick up my fork. The CI is green as well.

kbenzie:
> I removed the "ready to merge" label because the intel/llvm part of this PR still has required reviews. The process is that we wait until both PRs are fully reviewed.

I added it again, the intel/unified-runtime-reviewers team isn't required for approval until the UR tag is updated in the intel/llvm PR.

PR #2618, opened 2025-01-24
[L0] MAX_COMPUTE_UNITS using ze_eu_count_ext_t

winstonzhang-intel:
For some recovery SKUs, MAX_COMPUTE_COUNT calculation does not provide the correct number of EUs. Now we will  use ze_eu_count_t when available.

LLVM/SYCL: https://github.com/intel/llvm/pull/16818

### DIFF
--- a/source/adapters/level_zero/device.cpp
+++ b/source/adapters/level_zero/device.cpp
@@ -873,6 +873,18 @@ ur_result_t urDeviceGetInfo(
     return ReturnValue(int32_t(ZeDeviceNumIndices));
   } break;
   case UR_DEVICE_INFO_GPU_EU_COUNT: {
+    if (Device->Platform->ZeDriverEuCountExtensionFound) {
+      ze_device_properties_t DeviceProp = {};
+      DeviceProp.stype = ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES;
+      ze_eu_count_ext_t EuCountDesc = {};
+      EuCountDesc.stype = ZE_STRUCTURE_TYPE_EU_COUNT_EXT;
+      DeviceProp.pNext = (void *)&EuCountDesc;
+      ZE2UR_CALL(zeDeviceGetProperties, (ZeDevice, &DeviceProp));
+      if (EuCountDesc.numTotalEUs > 0) {
+        return ReturnValue(uint32_t{EuCountDesc.numTotalEUs});
+      }
+    }
+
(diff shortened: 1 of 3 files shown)


source/adapters/level_zero/device.cpp
```diff
@@ -867,6 +867,16 @@ ur_result_t urDeviceGetInfo(
...
   case UR_DEVICE_INFO_GPU_EU_COUNT: {
+    ze_device_properties_t DeviceProp = {};
+    DeviceProp.stype = ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES;
+    ze_eu_count_ext_t EuCountDesc = {};
+    EuCountDesc.stype = ZE_STRUCTURE_TYPE_EU_COUNT_EXT;
+    DeviceProp.pNext = (void *)&EuCountDesc;
+    ZE2UR_CALL(zeDeviceGetProperties, (ZeDevice, &DeviceProp));
+    if (DeviceProp.pNext &&
+        EuCountDesc.stype == ZE_STRUCTURE_TYPE_EU_COUNT_EXT) {
+      return ReturnValue(uint32_t{EuCountDesc.numTotalEUs});
```
nrspruit, line 878: should we check that this value is >0? I am not sure the driver is required to set the pNext to nullptr, this might always return 0.

nrspruit: You should also be checking the driver extensions for ZE_extension_eu_count. If the driver does not report it supports 1.0 of this extension, then this pNext is guaranteed to not be set.

source/adapters/level_zero/device.cpp
```diff
@@ -867,6 +867,16 @@ ur_result_t urDeviceGetInfo(
     return ReturnValue(int32_t(ZeDeviceNumIndices));
   } break;
   case UR_DEVICE_INFO_GPU_EU_COUNT: {
+    ze_device_properties_t DeviceProp = {};
```
nrspruit, line 870: Check the value of Device->Platform->ZeDriverEuCountExtensionFound before reading the device properties again.

winstonzhang-intel: good call, fixed.

nrspruit (approved):

nrspruit:
@winstonzhang-intel , please update the intel/llvm PR https://github.com/intel/llvm/pull/16818 with your latest changes since it is failing.

PR #2624, opened 2025-01-27
Assert that Device is valid for memory poisoning

RossBrunton:
This (hopefully) silences a Coverity issue.


source/loader/layers/sanitizer/msan/msan_interceptor.cpp
```diff
@@ -75,6 +75,7 @@ ur_result_t MsanInterceptor::allocateMemory(ur_context_handle_t Context,
   if (Type != AllocType::DEVICE_USM) {
     return UR_RESULT_SUCCESS;
   }
+  assert(Device);
```
aarongreig, line 78: I think Device is allowed to be null here since it's a helper that can be used for host allocations as well

RossBrunton: If that is the case, then Type should be `AllocType::HOST_USM` and should have already returned.

aarongreig (approved):

PR #2626, opened 2025-01-27
Reorganised urEventGetProfilingInfo test into separate tests instead of a switch

martygrant:
For https://github.com/oneapi-src/unified-runtime/issues/2290

Also done a housekeeping pass across all urXGetInfo tests, renaming and adding consts where appropriate, and adding extra checks to tests with vectors.

### DIFF
--- a/test/conformance/kernel/urKernelGetInfo.cpp
+++ b/test/conformance/kernel/urKernelGetInfo.cpp
@@ -12,20 +12,23 @@ using urKernelGetInfoTest = uur::urKernelTest;
 UUR_INSTANTIATE_DEVICE_TEST_SUITE(urKernelGetInfoTest);
 
 TEST_P(urKernelGetInfoTest, SuccessFunctionName) {
-  ur_kernel_info_t property_name = UR_KERNEL_INFO_FUNCTION_NAME;
+  const ur_kernel_info_t property_name = UR_KERNEL_INFO_FUNCTION_NAME;
   size_t property_size = 0;
 
   ASSERT_SUCCESS_OR_OPTIONAL_QUERY(
       urKernelGetInfo(kernel, property_name, 0, nullptr, &property_size),
       property_name);
+  ASSERT_GT(property_size, 0);
 
-  std::vector<char> property_value(property_size);
+  std::vector<char> property_value(property_size, '\0');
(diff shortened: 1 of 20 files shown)


test/conformance/kernel/urKernelGetInfo.cpp
```diff
@@ -12,100 +12,106 @@ using urKernelGetInfoTest = uur::urKernelTest;
...
       urKernelGetInfo(kernel, property_name, 0, nullptr, &property_size),
       property_name);
+  ASSERT_GT(property_size, 0);
 
-  std::vector<char> property_value(property_size);
+  std::vector<char> property_value(property_size, 'x');
   ASSERT_SUCCESS(urKernelGetInfo(kernel, property_name, property_size,
                                  property_value.data(), nullptr));
+
+  ASSERT_EQ(property_size, property_value.size());
```
RossBrunton, line 27: NIT: Isn't this just testing that the standard library's std::vector class works rather than UR?

RossBrunton (approved):

test/conformance/kernel/urKernelGetInfo.cpp
martygrant: so my intention was to test that the entry point populates `pPropValue` and that its size matches the size retrieved from the first call

kbenzie: Are you expecting the `std::vector`'s size to be changed by the call to `urKernelGetInfo()`? If so that's never going to happen, only the data owned by the `std::vector` will change not the container itself.

martygrant: discussed this offline and cleared up my misunderstanding around testing for this, will update the PR with a better check

martygrant: I've removed these size checks and added a helper function `uur::stringPropertyIsValid` to validate `char[]` returns

martygrant:
I've added some more changes to this PR, tidying up some more `char[]` return query tests, and added in a helper macro `ASSERT_GET_INFO` to validate numerical queries. Could use a re-review.

kbenzie (approved):

PR #2627, opened 2025-01-27
[NATIVECPU] NativeCPU with optional oneTBB backend

uwedolinsky:
Initial PR to add option for Native CPU to use oneTBB as its backend.

Corresponding DPC++ PR: https://github.com/intel/llvm/pull/16803


source/adapters/native_cpu/enqueue.cpp
```diff
@@ -26,8 +26,8 @@ struct NDRDescT {
   RangeT GlobalOffset;
   RangeT GlobalSize;
   RangeT LocalSize;
-  NDRDescT(uint32_t WorkDim, const size_t *GlobalWorkOffset,
-           const size_t *GlobalWorkSize, const size_t *LocalWorkSize)
+  inline NDRDescT(uint32_t WorkDim, const size_t *GlobalWorkOffset,
+                  const size_t *GlobalWorkSize, const size_t *LocalWorkSize)
```
hvdijk, line 30: This is implicitly `inline` already; in OCK, I think we went the other way and removed `inline` wherever it was redundant.

source/adapters/native_cpu/enqueue.cpp
```diff
@@ -215,34 +214,32 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
       auto numGroups = groups.size();
       auto groupsPerThread = numGroups / numParallelThreads;
       auto remainder = numGroups % numParallelThreads;
-      for (unsigned thread = 0; thread < numParallelThreads; thread++) {
-        futures.emplace_back(
-            tp.schedule_task([groups, thread, groupsPerThread,
+      for (unsigned thread = 0; groupsPerThread && thread < numParallelThreads;
```
hvdijk, line 217: I can see the logic behind `groupsPerThread &&` in avoiding running a loop where each loop body does nothing, but personally, I think it's clearer to have this as a `if (groupsPerThread) { ... }` around the loop.

source/adapters/native_cpu/threadpool.hpp
```diff
@@ -208,7 +208,80 @@ template <typename ThreadPoolT> class threadpool_interface {
...
 };
+using simple_threadpool_t = threadpool_interface<detail::simple_thread_pool>;
 
-using threadpool_t = threadpool_interface<detail::simple_thread_pool>;
+class TasksInfo_TP {
+  using FType = std::future<void>;
+  std::vector<FType> futures;
+public:
+  inline void schedule(FType &&f) { futures.emplace_back(std::move(f)); }
+  inline void wait() {
```
hvdijk, line 218: Based on other APIs (e.g. Boost), this might possibly be clearer to rename to something like `wait_all` to highlight the difference between it and `future::wait`?

source/adapters/native_cpu/threadpool.hpp
```diff
@@ -208,7 +208,80 @@ template <typename ThreadPoolT> class threadpool_interface {
...
+      auto thread_id = tbb::this_task_arena::current_thread_index();
+      assert(thread_id >= 0 &&
+             thread_id < oneapi::tbb::info::default_concurrency());
+      task(thread_id);
+    }));
+  }
+};
+
+using tasksinfo_t = TBB_TasksInfo;
+using threadpool_t = TBB_threadpool;
```
hvdijk, line 278: For easier debugging, would it make sense to have a way to toggle between implementations at runtime, or does that cause more problems than it solves?

source/adapters/native_cpu/enqueue.cpp
uwedolinsky: I've removed the `inline`

source/adapters/native_cpu/threadpool.hpp
hvdijk: That makes sense, thanks for the explanation, let's leave this as it is then.

source/adapters/native_cpu/CMakeLists.txt
```diff
@@ -62,3 +91,23 @@ target_link_libraries(${TARGET_NAME} PRIVATE
...
+    # todo: check if compiler actually supports these options
+    target_compile_options(tbb PRIVATE -Wno-cast-qual)
+    target_compile_options(tbbmalloc PRIVATE -Wno-cast-qual)
+  endif()
+
+  # Undefine _DEBUG option in release builds to find
+  # release tbbbind
+  if (NOT uppercase_CMAKE_BUILD_TYPE STREQUAL "DEBUG")
+    target_compile_options(tbb PRIVATE -U_DEBUG)
+  endif()
```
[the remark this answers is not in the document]
uwedolinsky: Probably, but this may require some more rework and retesting, and the current `FetchContent` integration is working and passing the [DPCPP CI](https://github.com/intel/llvm/pull/15979) so I'd recommend to revisit this in a follow-up PR if need be.  Also, the latest [Cmake "Using Dependencies Guide"](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html) doesn't seem to mention `ExternalProject`, so it's not clear how recommended it is.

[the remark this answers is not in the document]
hvdijk: True, but the "Using Dependencies Guide" only gives instructions on how to either use pre-built packages, or how to include an external project's sources as this project's sources. Neither works for us and the result is that we need workarounds to include oneTBB's sources as this project's sources, except make it mostly, but not entirely, as if they are a different project's sources, because our project's configuration is not what we want for oneTBB.

But if we do not have an easy way of using an alternative, if it would be too much work to use `ExternalProject_Add` or some other alternatives, then sure, we can delay that until later.

martygrant:
## Information

## PR Migration

---

martygrant:

---

PR #2630, opened 2025-01-28
Don't use inheritence for L0 V2 event handles

RossBrunton:
In the future, we are going to require that handle objects have no
vtable, so this change combines "native" and "pooled" event into one
class. A variant is used to determine whether the event is native or
pooled.

For consistency, parameter order in constructors have been changed to
always start with the context.

### DIFF
--- a/source/adapters/level_zero/v2/event.hpp
+++ b/source/adapters/level_zero/v2/event.hpp
@@ -47,15 +47,24 @@ private:
 
 struct ur_event_handle_t_ : _ur_object {
 public:
-  ur_event_handle_t_(ur_context_handle_t hContext, ze_event_handle_t hZeEvent,
-                     v2::event_flags_t flags);
+  // cache_borrowed_event is used for pooled events, whilst ze_event_handle_t is
+  // used for native events
+  using event_variant =
+      std::variant<v2::raii::cache_borrowed_event, v2::raii::ze_event_handle_t>;
+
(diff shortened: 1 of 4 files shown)


source/adapters/level_zero/v2/event.hpp
```diff
@@ -97,11 +106,16 @@ struct ur_event_handle_t_ : _ur_object {
...
+private:
+  ur_event_handle_t_(ur_context_handle_t hContext, event_variant hZeEvent,
+                     v2::event_flags_t flags, v2::event_pool *pool);
+
 protected:
   ur_context_handle_t hContext;
 
-  // non-owning handle to the L0 event
-  const ze_event_handle_t hZeEvent;
+  // Pool is used iff this is a pooled event
```
pbalcer, line 116:
```suggestion
  // Pool is used if this is a pooled event
```

pbalcer (approved):

source/adapters/level_zero/v2/event.hpp
RossBrunton: https://en.wikipedia.org/wiki/If_and_only_if

But probably worth typing it out in full.

pbalcer: Didn't know that one :) Thanks.

RossBrunton:
LLVM testing doesn't test L0 version 2, so I'll trust that the CI testing is sufficient and merge it.

PR #2631, opened 2025-01-28
Enable cmd-buf local memory update CTS tests on L0

EwanC:
The CTS tests for updating local memory in command-buffer kernel commands are currently skipped on Level-Zero. This feature is now available in L0 drivers so the tests can be enabled.

There is a failure in `LocalMemoryUpdateTest.UpdateParametersSmallerLocalSize` with incorrect results. Mark the individual test as failed as it needs further investigation.


reble (approved):
LGTM

PR #2635, opened 2025-01-29
Remove virtual from L0 V2 queues

RossBrunton:


pbalcer:
This looks good to me in general. @igchor ?

igchor (approved):

igchor:
Looks, good. FYI, In V2 we also use virtual methods for buffers (memory.hpp). I'm actually working on support for images: https://github.com/oneapi-src/unified-runtime/pull/2640 so I can address this (remove the virtual functions) on that PR.

pbalcer (approved):

source/adapters/level_zero/v2/command_list_manager.cpp
```diff
@@ -13,11 +13,12 @@
 #include "../ur_interface_loader.hpp"
 #include "context.hpp"
 #include "kernel.hpp"
+#include "queue_handle.hpp"
```
Bensuo, line 16: Why do we need to include this header if we're not using the `queue_handle_t_` type anymore?

RossBrunton: Good spot, removed it.

Bensuo (approved):

RossBrunton:
Merging without LLVM commit since (as far as I know) there's no V2 testing there yet.

PR #2636, opened 2025-01-29
[L0] Enable Copy engine support with in-order command lists

konradkusiak97:
Currently, copy engine is only available to use for certain commands in the out-of-order command-buffer path. This PR adds required synchronization between the 2 command lists, when both are in-order. Hence, it enables using the copy engine with in-order path. 

For the immediateAppend path, this only works for now if the environment variable `UR_L0_USE_DRIVER_COUNTER_BASED_EVENTS=0` is set. 

[intel/llvm PR](https://github.com/intel/llvm/pull/16830)

### DIFF
--- a/source/adapters/level_zero/command_buffer.cpp
+++ b/source/adapters/level_zero/command_buffer.cpp
@@ -175,7 +175,7 @@ ur_result_t getEventsFromSyncPoints(
  * command list (copy or compute), to indicate when it's finished executing.
  * @param[in] CommandType The type of the command.
  * @param[in] CommandBuffer The CommandBuffer where the command is appended.
- * @param[in] isFirstNode A boolean inidicating if the current node is at the
+ * @param[in] IsFirstNode A boolean inidicating if the current node is at the
  * beginning of the command list.
  * @param[in] isCopy A boolean indicating if the current command uses copy
  * engine.
@@ -185,14 +185,14 @@ ur_result_t getEventsFromSyncPoints(
  */
 ur_result_t createSyncPointBetweenCopyAndCompute(
     ur_command_t CommandType, ur_exp_command_buffer_handle_t CommandBuffer,
-    bool isFirstNode, bool isCopy,
+    bool IsFirstNode, bool IsCopy,
     ze_event_handle_t &ZeSignalPrevCommandEvent) {
-  if (isFirstNode) {
-    CommandBuffer->MCopyCommandListEmpty = !isCopy;
-    CommandBuffer->MComputeCommandListEmpty = isCopy;
-    CommandBuffer->WasPrevCopyCommandList = isCopy;
+  if (IsFirstNode) {
+    CommandBuffer->MCopyCommandListEmpty = !IsCopy;
+    CommandBuffer->MComputeCommandListEmpty = IsCopy;
+    CommandBuffer->MWasPrevCopyCommandList = IsCopy;
     return UR_RESULT_SUCCESS;
-  } else if (CommandBuffer->WasPrevCopyCommandList == isCopy) {
+  } else if (CommandBuffer->MWasPrevCopyCommandList == IsCopy) {
     return UR_RESULT_SUCCESS;
   }
 
@@ -210,14 +210,14 @@ ur_result_t createSyncPointBetweenCopyAndCompute(
    * finished executing.
    */
   ZeSignalPrevCommandEvent = SignalPrevCommandEvent->ZeEvent;
-  if (CommandBuffer->WasPrevCopyCommandList && !isCopy) {
+  if (CommandBuffer->MWasPrevCopyCommandList && !IsCopy) {
     ZE2UR_CALL(zeCommandListAppendSignalEvent,
                (CommandBuffer->ZeCopyCommandList, ZeSignalPrevCommandEvent));
-    CommandBuffer->WasPrevCopyCommandList = false;
+    CommandBuffer->MWasPrevCopyCommandList = false;
   } else {
     ZE2UR_CALL(zeCommandListAppendSignalEvent,
                (CommandBuffer->ZeComputeCommandList, ZeSignalPrevCommandEvent));
-    CommandBuffer->WasPrevCopyCommandList = true;
+    CommandBuffer->MWasPrevCopyCommandList = true;
   }
 
   // Get sync point and register the event with it.
@@ -283,7 +283,7 @@ ur_result_t createSyncPointAndGetZeEvents(
   return UR_RESULT_SUCCESS;
 }
 
-// Shared by all memory read/write/copy PI interfaces.
+// Shared by all memory read/write/copy UR interfaces.
 // Helper function for common code when enqueuing memory operations to a command
 // buffer.
 ur_result_t enqueueCommandBufferMemCopyHelper(
@@ -300,9 +300,9 @@ ur_result_t enqueueCommandBufferMemCopyHelper(
       false, RetSyncPoint, ZeEventList, ZeLaunchEvent));
 
   // The chooseCommandList() will modify MComputeCommandListEmpty and
-  // MCopyCommandListEmpty so isFirstNode needs to be created before that.
-  bool isFirstNode{CommandBuffer->MComputeCommandListEmpty &&
-                   CommandBuffer->MCopyCommandListEmpty};
+  // MCopyCommandListEmpty so isFirstNode needs to be called beforehand.
+  bool IsFirstNode{CommandBuffer->isFirstNode()};
+
   ze_command_list_handle_t ZeCommandList =
       CommandBuffer->chooseCommandList(PreferCopyEngine);
 
@@ -312,7 +312,7 @@ ur_result_t enqueueCommandBufferMemCopyHelper(
       ZeCommandList == CommandBuffer->ZeCopyCommandList) {
     ze_event_handle_t ZeSignalPrevCommandEvent = nullptr;
     UR_CALL(createSyncPointBetweenCopyAndCompute(CommandType, CommandBuffer,
-                                                 isFirstNode, true /*isCopy=*/,
+                                                 IsFirstNode, true /*isCopy=*/,
                                                  ZeSignalPrevCommandEvent));
     if (ZeSignalPrevCommandEvent) {
       ZeEventList.push_back(ZeSignalPrevCommandEvent);
@@ -374,9 +374,9 @@ ur_result_t enqueueCommandBufferMemCopyRectHelper(
       false, RetSyncPoint, ZeEventList, ZeLaunchEvent));
 
   // The chooseCommandList() will modify MComputeCommandListEmpty and
-  // MCopyCommandListEmpty so isFirstNode needs to be created before that.
-  bool isFirstNode{CommandBuffer->MComputeCommandListEmpty &&
-                   CommandBuffer->MCopyCommandListEmpty};
+  // MCopyCommandListEmpty so isFirstNode needs to be called beforehand.
+  bool IsFirstNode{CommandBuffer->isFirstNode()};
+
   ze_command_list_handle_t ZeCommandList =
       CommandBuffer->chooseCommandList(PreferCopyEngine);
 
@@ -386,7 +386,7 @@ ur_result_t enqueueCommandBufferMemCopyRectHelper(
       ZeCommandList == CommandBuffer->ZeCopyCommandList) {
     ze_event_handle_t ZeSignalPrevCommandEvent = nullptr;
     UR_CALL(createSyncPointBetweenCopyAndCompute(CommandType, CommandBuffer,
-                                                 isFirstNode, true /*isCopy=*/,
+                                                 IsFirstNode, true /*isCopy=*/,
                                                  ZeSignalPrevCommandEvent));
     if (ZeSignalPrevCommandEvent) {
       ZeEventList.push_back(ZeSignalPrevCommandEvent);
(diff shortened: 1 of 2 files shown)


source/adapters/level_zero/command_buffer.hpp
```diff
@@ -110,6 +110,11 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
   // This flag must be set to false if at least one copy command has been
   // added to `ZeCopyCommandList`
   bool MCopyCommandListEmpty = true;
+  // This flag must be set to false if at least one copy command has been
```
EwanC, line 113:
```suggestion
  // This flag must be set to false if at least one compute command has been
```

source/adapters/level_zero/command_buffer.hpp
```diff
@@ -110,6 +110,11 @@ struct ur_exp_command_buffer_handle_t_ : public _ur_object {
...
+  // added to `ZeComputeCommandList`
+  bool MComputeCommandListEmpty = true;
+  // This flag tracks if the previous node submission was compute or copy type.
+  bool WasPrevCopyCommandList = false;
```
EwanC, line 117:
```suggestion
  bool MWasPrevCopyCommandList = false;
```

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -489,11 +597,12 @@ void ur_exp_command_buffer_handle_t_::registerSyncPoint(
...
 ze_command_list_handle_t
 ur_exp_command_buffer_handle_t_::chooseCommandList(bool PreferCopyEngine) {
-  if (PreferCopyEngine && this->useCopyEngine() && !this->IsInOrderCmdList) {
+  if (PreferCopyEngine && this->useCopyEngine()) {
     // We indicate that ZeCopyCommandList contains commands to be submitted.
     this->MCopyCommandListEmpty = false;
     return this->ZeCopyCommandList;
   }
+  this->MComputeCommandListEmpty = false;
   return this->ZeComputeCommandList;
```
EwanC, line 606: A bit out of scope, but while we're touching this function can we get rid the `this->` pointer dereferences, they don't look necessary.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -241,9 +299,25 @@ ur_result_t enqueueCommandBufferMemCopyHelper(
       CommandType, CommandBuffer, NumSyncPointsInWaitList, SyncPointWaitList,
       false, RetSyncPoint, ZeEventList, ZeLaunchEvent));
 
+  // The chooseCommandList() will modify MComputeCommandListEmpty and
+  // MCopyCommandListEmpty so isFirstNode needs to be created before that.
+  bool isFirstNode{CommandBuffer->MComputeCommandListEmpty &&
+                   CommandBuffer->MCopyCommandListEmpty};
```
EwanC, line 305: rather than duplicating this variable N times across the different helpers, prefer to just have a method on the CommandBuffer class defined in the header so that it's inlined
```
 bool isFirstNode() { return MComputeCommandListEmpty && MCopyCommandListEmpty; }
```

Then at the `createSyncPointBetweenCopyAndCompute` callsites can do `isFirstNode()`

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -170,6 +170,64 @@ ur_result_t getEventsFromSyncPoints(
...
+ * beginning of the command list.
+ * @param[in] isCopy A boolean indicating if the current command uses copy
+ * engine.
+ * @param[out] ZeSignalPrevCommandEvent The event which signals when the
+ * previous command appended to a different command list is finished executing.
+ * @return UR_RESULT_SUCCESS or an error code on failure
+ */
+ur_result_t createSyncPointBetweenCopyAndCompute(
+    ur_command_t CommandType, ur_exp_command_buffer_handle_t CommandBuffer,
+    bool isFirstNode, bool isCopy,
```
EwanC, line 188: nitpicking for consistent variable naming (and update doxy comment)
```suggestion
    bool IsFirstNode, bool IsCopy,
```

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -241,9 +299,25 @@ ur_result_t enqueueCommandBufferMemCopyHelper(
       CommandType, CommandBuffer, NumSyncPointsInWaitList, SyncPointWaitList,
       false, RetSyncPoint, ZeEventList, ZeLaunchEvent));
 
+  // The chooseCommandList() will modify MComputeCommandListEmpty and
```
EwanC, line 302: Can't comment on the line itself, but in the comment for this function it says "PI", could you update it to "UR" since we're here

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -170,6 +170,64 @@ ur_result_t getEventsFromSyncPoints(
...
+  } else {
+    ZE2UR_CALL(zeCommandListAppendSignalEvent,
+               (CommandBuffer->ZeComputeCommandList, ZeSignalPrevCommandEvent));
+    CommandBuffer->WasPrevCopyCommandList = true;
+  }
+
+  // Get sync point and register the event with it.
+  ur_exp_command_buffer_sync_point_t SyncPoint =
+      CommandBuffer->getNextSyncPoint();
+  CommandBuffer->registerSyncPoint(SyncPoint, SignalPrevCommandEvent);
```
EwanC, line 226: Is every call to this function inside a check for `CommandBuffer->IsInOrderCmdList `? If so I don't think we need to care about sync points (and should also stick that in function comment) Also preference for renaming the function to remove "syncpoint" in that case.

source/adapters/level_zero/command_buffer.cpp
konradkusiak97: Yeah sounds better, I'll update that.

source/adapters/level_zero/command_buffer.cpp
konradkusiak97: I'm not sure I follow. `createSyncPointBetweenCopyAndCompute` is only ever called inside `CommandBuffer->IsInOrderCmdList` check to synchronize between the potential `in-orderCopy` vs `in-orderCompute` race. 

AFAIK Registering `SignalPrevCommandEvent` adds it to the event list which is later being reset. If we managed to use counter-based events, lines `223-226` wouldn't be necessary.

I might be misunderstanding the purpose of `getNextSyncPoint()` and `registerSyncPoint()`, maybe there is a more straightforward way to cache this signal even and later reset it but I thought that's the current way.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -170,6 +170,64 @@ ur_result_t getEventsFromSyncPoints(
   return UR_RESULT_SUCCESS;
 }
 
+/**
+ * If necessary, it creates a signal event and appends it to the previous
+ * command list (copy or compute), to indicate when it's finished executing.
+ * @param[in] CommandType The type of the command.
+ * @param[in] CommandBuffer The CommandBuffer where the command is appended.
+ * @param[in] IsFirstNode A boolean inidicating if the current node is at the
```
fabiomestre, line 178:
```suggestion
 * @param[in] IsFirstNode A boolean indicating if the current node is at the
```

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -241,9 +299,25 @@ ur_result_t enqueueCommandBufferMemCopyHelper(
...
+  // The chooseCommandList() will modify MComputeCommandListEmpty and
+  // MCopyCommandListEmpty so isFirstNode needs to be called beforehand.
+  bool IsFirstNode{CommandBuffer->isFirstNode()};
+
   ze_command_list_handle_t ZeCommandList =
       CommandBuffer->chooseCommandList(PreferCopyEngine);
 
+  // Create a sync point if both Copy and Compute in-order command lists are
+  // used.
+  if (CommandBuffer->IsInOrderCmdList &&
```
fabiomestre, line 311: This bit of code is copy pasted a lot. It can probably be refactored into its own function.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -170,6 +170,64 @@ ur_result_t getEventsFromSyncPoints(
...
+ * @param[out] ZeSignalPrevCommandEvent The event which signals when the
+ * previous command appended to a different command list is finished executing.
+ * @return UR_RESULT_SUCCESS or an error code on failure
+ */
+ur_result_t createSyncPointBetweenCopyAndCompute(
+    ur_command_t CommandType, ur_exp_command_buffer_handle_t CommandBuffer,
+    bool IsFirstNode, bool IsCopy,
+    ze_event_handle_t &ZeSignalPrevCommandEvent) {
+  if (IsFirstNode) {
+    CommandBuffer->MCopyCommandListEmpty = !IsCopy;
```
fabiomestre, line 191: Why is this bit of code needed? Isn't `chooseCommandList()` already  assigning a value to MCopyCommandListEmpty and MComputeCommandListEmtpy?

source/adapters/level_zero/command_buffer.cpp
EwanC: > I might be misunderstanding the purpose of getNextSyncPoint() and registerSyncPoint(), maybe there is a more straightforward way to cache this signal event and later reset it but I thought that's the current way.

This is more what I was getting at. As was trying to ask if we need this sync point if the UR user doesn't care about sync points. However, you've answered my question by saying that we need to do this to cache the L0 event and reset it later. So I guess my feedback is to update the code comment to append ", so that the event is reset between executions"
```

File: source/adapters/level_zero/command_buffer.cpp [hunk 11]
```diff
@@ -170,6 +170,64 @@ ur_result_t getEventsFromSyncPoints(
...
+ * engine.
+ * @param[out] ZeSignalPrevCommandEvent The event which signals when the
+ * previous command appended to a different command list is finished executing.
+ * @return UR_RESULT_SUCCESS or an error code on failure
+ */
+ur_result_t createSyncPointBetweenCopyAndCompute(
+    ur_command_t CommandType, ur_exp_command_buffer_handle_t CommandBuffer,
+    bool IsFirstNode, bool IsCopy,
+    ze_event_handle_t &ZeSignalPrevCommandEvent) {
+  if (IsFirstNode) {
```
fabiomestre, line 190: Can we avoid using `IsFirstNode` completely. It seems to only be used in this check and if I understand it correctly, initializing `MWasPrevCopyCommandList` to `true`  in the CommandBuffer constructor would do the same. I could be missing something though

konradkusiak97: Yes, I got rid of `IsFirstNode` completely.

source/adapters/level_zero/command_buffer.cpp
konradkusiak97: I've refactored this bit together with others.

source/adapters/level_zero/command_buffer.cpp
konradkusiak97: Yep, I've put some of this logic inside `createSyncPointBetweenCopyAndCompute()` and in `appendWaitBetweenCopyAndCompute()`. Let me know what you think.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -170,6 +170,95 @@ ur_result_t getEventsFromSyncPoints(
...
+ * @return UR_RESULT_SUCCESS or an error code on failure
+ */
+ur_result_t createSyncPointBetweenCopyAndCompute(
+    ur_exp_command_buffer_handle_t CommandBuffer,
+    ze_command_list_handle_t ZeCommandList, bool IsCopy,
+    std::vector<ze_event_handle_t> &WaitEventList) {
+
+  // Skip in cases when we're not using Copy engine or it's not in-order path.
+  if (!CommandBuffer->IsInOrderCmdList || !CommandBuffer->ZeCopyCommandList ||
+      (IsCopy && ZeCommandList != CommandBuffer->ZeCopyCommandList)) {
```
EwanC, line 191: > (IsCopy && ZeCommandList != CommandBuffer->ZeCopyCommandList)

Do we need a `IsCopy` function parameter? or could we deduce `IsCopy` inside this helper with 
`bool IsCopy = ZeCommandList == CommandBuffer->ZeCopyCommandList;`

konradkusiak97: Indeed, we can do that.

EwanC (approved):
LGTM, but would like Fábio to check this over as well before merging

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -1385,6 +1485,9 @@ ur_result_t urCommandBufferAppendUSMAdviseExp(
   ze_memory_advice_t ZeAdvice = static_cast<ze_memory_advice_t>(Value);
 
   if (CommandBuffer->IsInOrderCmdList) {
+    // Sync if copy engine is used in in-order path.
+    UR_CALL(appendWaitBetweenCopyAndCompute(CommandBuffer));
```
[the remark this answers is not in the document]
konradkusiak97: You mean the `zeCommandListAppendSignalEvent()` on line `1513` ? Yeah this call seems redundant to me.

source/adapters/level_zero/command_buffer.cpp
```diff
@@ -244,15 +329,18 @@ ur_result_t enqueueCommandBufferMemCopyHelper(
   ze_command_list_handle_t ZeCommandList =
       CommandBuffer->chooseCommandList(PreferCopyEngine);
 
+  UR_CALL(createSyncPointBetweenCopyAndCompute(CommandBuffer, ZeCommandList,
```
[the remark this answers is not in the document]
konradkusiak97: Indeed! Seems like this can be done. I've moved `createSyncPointBetweenCopyAndCompute()` to `createSyncPointAndGetZeEvents()`.

source/adapters/level_zero/command_buffer.cpp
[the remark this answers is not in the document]
konradkusiak97: I've simplified it as in your comment after moving the calls. I also removed the redundant `zeCommandListAppendSignalEvent()`, *I think* they're not necessary. That signal event is added to the `ZeEventsList` vector for later reset but otherwise I think its not used :thinking:

fabiomestre (approved):

martygrant:
## Information

## PR Migration

---

martygrant:

---

Issue #1463, opened 2024-03-22
[Bindless Image][Request] Create a sampled image with a single UR api call from both image and sampler descriptors


wenju-he:
Currently a bindless sampled is created in two steps as shown in https://github.com/intel/llvm/blob/c1e2957be8db95425f1c17df258a0830c83dcf47/sycl/source/detail/bindless_images.cpp#L436-L452
First a sampler object is created and then a sampled image is created using the sampler object.

The request is to combine the two calls into a single call.

Motivations:
- urSamplerCreate in CUDA/HIP adaptor is only setting up sampler properties and no backend object is created.
  Therefore, it is natural to combine the two calls for CUDA/HIP.
- urSamplerCreate in Level-zero adaptor is calling zeSamplerCreate API and a backend sampler object is created. We plan to update Level-zero bindless image spec to allow a sampled image to be created with a single zeImageCreate call with additional sampler descriptor information. Then a single 64-bit handle works for a Level-zero sampled image. Therefore, we need to combine the two calls to avoid redundant sampler being created by the first zeSamplerCreate call. POC of 64-bit handle is already producing correct results for bindless sampled image tests on DG2. 

Tagging @isaacault @przemektmalon @DBDuncan @Seanst98 @cppchedy @HoppeMateusz

przemektmalon:
Thanks for proposing this, it makes sense to add this API.

We've created an issue to track this internally. In the meantime, I don't believe this issue is of the highest priority right now, however, if it is causing problems, let us know and we can expedite work on this.

wenju-he:
Thank you @przemektmalon for tracking this.
> I don't believe this issue is of the highest priority right now

I agree

ProGTX:
This can now be closed because https://github.com/intel/llvm/pull/18384  has been merged

PR #2640, opened 2025-01-29
[L0 v2] add support for images

igchor:


source/adapters/level_zero/v2/kernel.cpp
```diff
@@ -290,9 +291,14 @@ ur_result_t ur_kernel_handle_t_::prepareForSubmission(
   for (auto &pending : pending_allocations) {
     void *zePtr = nullptr;
     if (pending.hMem) {
-      // NULL is a valid value
-      zePtr = pending.hMem->getDevicePtr(hDevice, pending.mode, 0,
-                                         pending.hMem->getSize(), migrate);
+      if (!pending.hMem->isImage()) {
```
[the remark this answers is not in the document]
igchor: Maybe, but it's not a great abstraction - we would need to pass 'migrate' which doesn't make any sense for images.

source/adapters/level_zero/sampler.cpp
```diff
@@ -95,7 +100,7 @@ ur_result_t urSamplerCreate(
     }
   }
 
-  ZE2UR_CALL(zeSamplerCreate, (Context->ZeContext, Device->ZeDevice,
```
JackAKirk, line 98: What is the status of this issue on the L0 side? https://github.com/oneapi-src/unified-runtime/issues/1463

Are you still planning to replace `zeSamplerCreate` with `zeImageCreate` which would allow urSamplerCreate (which is not officially in bindless-images UR spec) to be removed from bindless-image dpc++ calls,
or do we need to add a new version of the `urSamplerCreate` API that takes a device argument, so that we avoid setting a different device here than the one that the user passes to `create_image` here https://github.com/intel/llvm/blob/b0e975c88531125d077687549a2e40a78612507f/sycl/source/detail/bindless_images.cpp#L376
?

igchor: That's a good question. I don't actually know what is the status, perhaps @przemektmalon or @wenju-he can comment?

In any case, this PR doesn't really change any logic related to the sampler - it only enables it for v2. Any change should probably be done in a separate PR.

source/adapters/level_zero/v2/kernel.cpp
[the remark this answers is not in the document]
pbalcer: hm, ok, makes sense.

source/adapters/level_zero/sampler.cpp
wenju-he: > What is the status of this issue on the L0 side? https://github.com/oneapi-src/unified-runtime/issues/1463

Remaining work of #1463 should be done in UR for all adapters at the same time. urSamplerCreate should not be called for bindless image. So #1463 isn't done yet.

> Any change should probably be done in a separate PR.

Yes I think so.

JackAKirk: > > What is the status of this issue on the L0 side? #1463
> 
> Remaining work of #1463 should be done in UR for all adapters at the same time. urSamplerCreate should not be called for bindless image. So #1463 isn't done yet.
> 
> > Any change should probably be done in a separate PR.
> 
> Yes I think so.

So there is no blocker on the l0 side to removing the call to `urSamplerCreate` by either replacing fully the functionality of `zuSamplerCreate` by `zeImageCreate` (as indicated by the comment in https://github.com/oneapi-src/unified-runtime/issues/1463) or otherwise?

wenju-he: > So there is no blocker on the l0 side to removing the call to `urSamplerCreate`: because you've presumably either either replaced fully the functionality of `zuSamplerCreate` by `zeImageCreate` (as indicated by the comment in #1463) or otherwise?

There is no blocker, but `replaced fully the functionality of zuSamplerCreate by zeImageCreate` isn't done yet. I think it should be done for all adapters at the same time before we can remove the call to `urSamplerCreate` at https://github.com/intel/llvm/blob/d3568afe1d7142b6eab54403f8f1db8b8e1dcd27/sycl/source/detail/bindless_images.cpp#L416

urSamplerCreate in L0 adapter does two things: one is setting up sampler descriptor and another is creating a sampler object from the descriptor. The first part is still needed for bindless image.
urSamplerCreate in cuda and hip adapters only set up sampler descriptor.

My understanding is that the functionality of translating ur sampler description to backend sampler description can be moved into a utility function, which should be called in both urSamplerCreate and urBindlessImagesSampledImageCreateExp.

> Is the below now possible?
> 
> "We plan to update Level-zero bindless image spec to allow a sampled image to be created with a single zeImageCreate call with additional sampler descriptor information. Then a single 64-bit handle works for a Level-zero sampled image. "

Yes, this is already implemented in L0 adapter.

przemektmalon: > My understanding is that the functionality of translating ur sampler description to backend sampler description can be moved into a utility function, which should be called in both urSamplerCreate and urBindlessImagesSampledImageCreateExp.

That's correct.

The change to the Bindless Images UR spec would be to replace the `ur_sampler_handle_t` argument with a `ur_sampler_desc_t`, in the call to `urBindlessImagesSampledImageCreateExp`.

Then we could remove the call to `urSamplerCreate` from the SYCL RT. This is already possible, as you point out, on all of our backends, including L0 where currently we create an unnecessary `ur_sampler_handle_t`.

And as you note, the utility function to set up the sampler description would be made available to both `urBindlessImagesSampledImageCreateExp` and `urSamplerCreate`.

We still have the internal tracker for this, and I agree that this should be done as a separate PR. Our internal priority on this issue is still low for now, however, we are in the process of discussing planned priorities for all of our issues, and since this issue has existed for a while, with the fix being quite straightforward, we may be able to bump it.

igchor:
@oneapi-src/unified-runtime-bindless-images-write any concerns regarding changes in memory.cpp/image.cpp for the legacy adapter? I only moved common functionality to separate files (image_helpers).

It seems there were two functions called ur2zeImageDesc (one in memory.cpp and one in image.cpp) which are ALMOST the same. I left both of them for now, but we will probably want to unify them in a next PR.

source/adapters/level_zero/helpers/image_helpers.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+bool Is3ChannelOrder(ur_image_channel_order_t ChannelOrder);
+
+ur_result_t getImageRegionHelper(ze_image_desc_t ZeImageDesc,
+                                 ur_rect_offset_t *Origin,
+                                 ur_rect_region_t *Region,
+                                 ze_image_region_t &ZeRegion);
+
+std::pair<ze_image_format_type_t, size_t>
+getImageFormatTypeAndSize(const ur_image_format_t *ImageFormat);
```
pbalcer, line 38: missing newline

pbalcer (approved):
lgtm, just a tiny nit.

przemektmalon:
> @oneapi-src/unified-runtime-bindless-images-write any concerns regarding changes in memory.cpp/image.cpp for the legacy adapter? I only moved common functionality to separate files (image_helpers).
> 
> It seems there were two functions called ur2zeImageDesc (one in memory.cpp and one in image.cpp) which are ALMOST the same. I left both of them for now, but we will probably want to unify them in a next PR.

The Bindless Image changes LGTM.

The Bindless Images version of `ur2zeImageDesc` handles more channel orders and sets `ZeImageDesc.stype/pNext`, but it also makes the `ZeImageDesc.width/height` have minimum values of `1`. I'm not sure if this will be compatible with the implementation in `memory.cpp` which allows values of `0`. I don't know currently the reason why it was implemented this way, if we are planning on unifying this in a future PR this will have to be checked.

source/adapters/level_zero/helpers/image_helpers.hpp
igchor: Done.

PR #2642, opened 2025-01-30
Improvements to align CTS and Spec for USM

martygrant:
- Additional error case checking
- More validation for urUSMPoolRelease
- Refactored urUSMDevice/Shared/HostAlloc test structs into a shared parent struct
- Added ur_usm_advice_flags_t param to parameterised Alloc tests

### DIFF
--- a/test/conformance/usm/urUSMSharedAlloc.cpp
+++ b/test/conformance/usm/urUSMSharedAlloc.cpp
@@ -5,14 +5,13 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "helpers.h"
+#include <limits>
 #include <uur/fixtures.h>
 #include <uur/known_failure.h>
 
-struct urUSMSharedAllocTest
-    : uur::urQueueTestWithParam<uur::USMDeviceAllocParams> {
+struct urUSMSharedAllocTest : uur::urUSMAllocTest {
   void SetUp() override {
-    UUR_RETURN_ON_FATAL_FAILURE(
-        uur::urQueueTestWithParam<uur::USMDeviceAllocParams>::SetUp());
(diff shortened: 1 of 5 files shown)


test/conformance/usm/urUSMSharedAlloc.cpp
```diff
@@ -144,43 +126,68 @@ TEST_P(urUSMSharedAllocTest, InvalidNullPtrMem) {
...
-  ASSERT_EQ_RESULT(UR_RESULT_ERROR_INVALID_USM_SIZE,
-                   urUSMSharedAlloc(context, device, nullptr, pool, -1, &ptr));
+  size_t size = 0;
+
+  void *ptr = nullptr;
+  ASSERT_EQ_RESULT(
+      UR_RESULT_ERROR_INVALID_USM_SIZE,
+      urUSMSharedAlloc(context, device, nullptr, pool, size, &ptr));
+
+  size = std::numeric_limits<size_t>::max() + 1;
```
aarongreig, line 136: `std::numeric_limits<size_t>::max() + 1` is `0` due to the overflow

martygrant: very obvious mistake thanks for pointing this out. I've changed this to get `UR_DEVICE_INFO_MAX_MEM_ALLOC_SIZE` and pass that value + 1 to the entry point for the error case

aarongreig: Looks like it's caused a bunch of fails, probably due to known issues with implementing `UR_DEVICE_INFO_MAX_MEM_ALLOC_SIZE`. I think we might just need to omit testing the upper limit on this one and let the return code be a "possible" one rather than mandatory.

martygrant: fair enough I've removed it, added a TODO with a link to an issue so we can come back to it at a later date

aarongreig (approved):

PR #2649, opened 2025-01-30
fix missing cmake options for UMF benchmarks

EuphoricThinking:
Added more cmake options for UMF benchmarks, including an option for building libumf_proxy.so. The file was missing previously.


.github/workflows/benchmarks-reusable.yml
```diff
@@ -176,6 +176,16 @@ jobs:
         -B${{github.workspace}}/umf_build
         -DUMF_BUILD_BENCHMARKS=ON
         -DUMF_TESTS_FAIL_ON_SKIP=ON
```
lukaszstolarczuk, line 178: meant to add it here:

FYI, you don't need `UMF_TESTS_FAIL_ON_SKIP=ON` if you disabled tests 😉

pbalcer (approved):

PR #2651, opened 2025-01-31
[L0 v2] fix queue destruction

igchor:
after introducing variant to store the actual queue instance, urQueueRelease became incorrect. It called delete on 'this' and not on ur_queue_handle_t_ which is an UB.


igchor:
Even though the current behavior is technically UB, it isn't causing any problems right now. However, we might decide to add some data fields to ur_queue_handle_t_ in the future and then we'll have a memory leak (at best).

RossBrunton (approved):

pbalcer (approved):

PR #2653, opened 2025-01-31
Change MAX_MEMORY_BANDWIDTH device query to uint64

omarahmed1111:
The max value that could be assigned to `MemoryBandwidth` variable is (250 * 3200000 * 256) which will overflow the uint32 max capacity. so we should change that to be uint64 to fit the maximum value from this computation.

### DIFF
--- a/scripts/core/device.yml
+++ b/scripts/core/device.yml
@@ -393,7 +393,7 @@ etors:
     - name: GPU_HW_THREADS_PER_EU
       desc: "[uint32_t][optional-query] return Intel GPU number of threads per EU"
     - name: MAX_MEMORY_BANDWIDTH
-      desc: "[uint32_t][optional-query] return max memory bandwidth in Mb/s"
+      desc: "[uint64_t][optional-query] return max memory bandwidth in B/s"
     - name: IMAGE_SRGB
       desc: "[$x_bool_t] device supports sRGB images"
     - name: BUILD_ON_SUBDEVICE
--- a/source/adapters/cuda/device.cpp
+++ b/source/adapters/cuda/device.cpp
@@ -1008,8 +1008,12 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
           &MemoryBusWidth, CU_DEVICE_ATTRIBUTE_GLOBAL_MEMORY_BUS_WIDTH,
           hDevice->get()));
     }
+    // This is a collection of all constants mentioned in this computation:
+    // https://github.com/jeffhammond/HPCInfo/blob/aae05c733016cc8fbb91ee71fc8076f17fa7b912/cuda/gpu-detect.cu#L241
+    constexpr uint64_t MemoryBandwidthConstant = 250;
 
-    uint32_t MemoryBandwidth = MemoryClockKHz * MemoryBusWidth * 250;
+    uint64_t MemoryBandwidth =
+        MemoryBandwidthConstant * MemoryClockKHz * MemoryBusWidth;
 
     return ReturnValue(MemoryBandwidth);
   }
--- a/include/ur_api.h
+++ b/include/ur_api.h
@@ -2121,7 +2121,7 @@ typedef enum ur_device_info_t {
   UR_DEVICE_INFO_GPU_SUBSLICES_PER_SLICE = 94,
   /// [uint32_t][optional-query] return Intel GPU number of threads per EU
   UR_DEVICE_INFO_GPU_HW_THREADS_PER_EU = 95,
-  /// [uint32_t][optional-query] return max memory bandwidth in Mb/s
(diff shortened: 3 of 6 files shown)


source/adapters/cuda/device.cpp
```diff
@@ -1009,7 +1009,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
           hDevice->get()));
     }
 
-    uint32_t MemoryBandwidth = MemoryClockKHz * MemoryBusWidth * 250;
+    uint64_t MemoryBandwidth = (uint64_t)MemoryClockKHz * MemoryBusWidth * 250;
```
jchlanda, line 1012: I seem to remember that there was an effort to get rid of c-style casts, so maybe:
```suggestion
    uint64_t MemoryBandwidth = static_cast<uint64_t>(MemoryClockKHz) * MemoryBusWidth * 250;
```
instead.

jchlanda (approved):
👍

source/adapters/cuda/device.cpp
omarahmed1111: The `static_cast` does the calculation before casting so it still overflows, unlike the c-style cast like here: https://godbolt.org/z/hdsP4K419 . I think I will define one of the variables as `uint64_t` and that should be a simpler option for that.

jchlanda: That's only because you casted the result of calculation, not a single variable, see: https://godbolt.org/z/P8ocxPsar

omarahmed1111: Ah okay, make sense. 

I changed the magic number `250` to a variable on its own, felt that would be a clearer solution. I tried to see what `250` mean, but it seems a mix of multiple constants multiplied together.

jchlanda: I think you still need to flip the order of multiplication, to make sure that the 64-bit wide value is done first. The implicit conversions are done in order of evaluation of the expressions, so you need to make sure that all of the operations are done on widened, 64-bit, types. As it stands, it could still overflow.

https://godbolt.org/z/qn6K4Eh37

omarahmed1111: Oh, Good shout, changed that.

scripts/core/device.yml
```diff
@@ -393,7 +393,7 @@ etors:
     - name: GPU_HW_THREADS_PER_EU
       desc: "[uint32_t][optional-query] return Intel GPU number of threads per EU"
     - name: MAX_MEMORY_BANDWIDTH
-      desc: "[uint32_t][optional-query] return max memory bandwidth in Mb/s"
+      desc: "[uint64_t][optional-query] return max memory bandwidth in Mb/s"
```
aarongreig, line 396: The spec for this says it should be in Mb/s, which I'd be surprised if we're overflowing (uint32 max mb would be around 4000 terabytes/second). Is the sycl RT expecting mb/s? Should we update it to b/s?

omarahmed1111: I changed it to be b/s to match the sycl docs for this query in intel/llvm.

aarongreig (approved):

source/adapters/cuda/device.cpp
```diff
@@ -1008,8 +1008,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urDeviceGetInfo(ur_device_handle_t hDevice,
           &MemoryBusWidth, CU_DEVICE_ATTRIBUTE_GLOBAL_MEMORY_BUS_WIDTH,
           hDevice->get()));
     }
+    uint64_t MemoryBandwidthConstant = 250;
 
-    uint32_t MemoryBandwidth = MemoryClockKHz * MemoryBusWidth * 250;
+    uint64_t MemoryBandwidth =
```
ldrumm, line 1013: If this is bits per second, don't we need to multiply by 1000 to account for the kHz aspect?

ldrumm: Also where does the `MemoryBandwidthConstant` come from? I know it was already there, but have providence comments for magic numbers would be useful.

Also, since it's putatively a constant: `const` or `constexpr`, please

omarahmed1111: I think the original equation is meant to use it as KHz to get it as b/s. I actually changed the constant as I think 250 was reporting it as Byte/sec and it should be b/s as the sycl spec. I added a comment explaining how to get this constant from in the nvidia doc page. Do you think that should be enough or should I write the computation constants also in a comment?

scripts/core/device.yml
omarahmed1111: I updated this to be capital B/s as it should be byte per second.

martygrant:
## Information

## PR Migration

---

omarahmed1111:
Migrated to [16869](https://github.com/intel/llvm/pull/16869)

PR #2658, opened 2025-02-03
[L0 v2] extend USMFill implementation to support sizes which are not powers of 2

igchor:


source/adapters/level_zero/v2/command_list_manager.hpp
```diff
@@ -50,3 +50,8 @@ struct ur_command_list_manager : public _ur_object {
   ur_queue_t_ *queue;
   std::vector<ze_event_handle_t> waitList;
 };
+
+inline void clearWaitList(std::pair<ze_event_handle_t *, uint32_t> &waitList) {
```
pbalcer, line 54: This is so confusing... we are aliasing the name `waitList` in three different places. Here, in the queue, and in the append function. Better yet, they are two different types.

Please add a struct with a function instead of using a pair and this. And name it something other than `WaitList`.

source/adapters/level_zero/v2/command_list_manager.hpp
```diff
@@ -50,3 +50,8 @@ struct ur_command_list_manager : public _ur_object {
   ur_queue_t_ *queue;
   std::vector<ze_event_handle_t> waitList;
```
pbalcer, line 51: Is there a way we could make the `waitList` usable only inside of the `getWaitListView` function? Maybe we should store a class here with that member private whose only method is `getWaitListView` ?

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
```diff
@@ -582,17 +566,16 @@ ur_result_t ur_queue_immediate_in_order_t::enqueueMemBufferMap(
...
         ZE2UR_CALL_THROWS(zeCommandListAppendMemoryCopy,
                           (commandListManager.getZeCommandList(), dst, src,
                            size, nullptr, waitList.second, waitList.first));
-        memoryMigrated = true;
+        clearWaitList(waitList);
       }));
   *ppRetMap = pDst;
 
-  if (!memoryMigrated && waitList.second) {
+  if (!waitList.second) {
```
pbalcer, line 578: I had to scroll up to see what `second` is to understand that this isn't a null-check. Either use `!= 0` to make that explicit or use a struct with variable names.

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
```diff
@@ -654,31 +636,37 @@ ur_result_t ur_queue_immediate_in_order_t::enqueueGenericFillUnlocked(
...
-              patternSize, size, zeSignalEvent, waitList.second,
-              waitList.first));
+  if (isPowerOf2(patternSize)) {
+    ZE2UR_CALL(zeCommandListAppendMemoryFill,
+               (commandListManager.getZeCommandList(), pDst, pPattern,
+                patternSize, size, zeSignalEvent, waitList.second,
+                waitList.first));
+  } else {
+    // Copy pattern into every entry in memory array pointed by Ptr.
+    uint32_t numOfCopySteps = size / patternSize;
```
pbalcer, line 657: What if `size % patternSize != 0`? Is that behavior defined?

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
igchor: It doesn't really matter since you can check any of the member, but I changed it to be less confusing.

pbalcer (approved):

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
pbalcer: it matters  because if you were checking the pointer, i'd be `if (waitList.first)` without negation. I think that's an easy mistake to make (and what I initially thought was happening here).
Anyway, looks good now, thanks.

source/adapters/level_zero/v2/command_list_manager.hpp
pbalcer: Yes, but my concern was that the  variable `waitList` was aliased within the scope of a function in a class that already contained `waitList`. Making it inaccessible inside of anything but that function would make it impossible to make a mistake by using the wrong `waitList`.

source/adapters/level_zero/v2/queue_immediate_in_order.cpp
igchor: actually, I think I just made a mistake in the previous revision - there shouldn't be any negation here

PR #2675, opened 2025-02-07
Improvements to align CTS and Spec for Enqueue

martygrant:
- Add test for invalid null pointer for urEnqueueKernelLaunch and urEnqueueUSMFill
- Remove redundant sentence in urEnqueueMemBufferMap description
- Add advice flag parameters in test for urEnqueueUSMAdvise
- Added blocking/non-blocking success tests for urEnqueueDeviceGlobalVariableRead/Write
- Added all invalid size scenarios for urEnqueueMemBufferFill and urEnqueueMemBufferWrite/Read/CopyRect
- Added to description of urEnqueueMemAdvice to say some adapters will return UR_RESULT_ERROR_ADAPTER_SPECIFIC and allowing this return value in relevant tests

I've not attached an intel/llvm tag update as the spec change is very minor - just removing a sentence from an entry-point description.

### DIFF
--- a/test/conformance/enqueue/urEnqueueDeviceGlobalVariableRead.cpp
+++ b/test/conformance/enqueue/urEnqueueDeviceGlobalVariableRead.cpp
@@ -5,10 +5,16 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #include <uur/fixtures.h>
 
-using urEnqueueDeviceGetGlobalVariableReadTest = uur::urGlobalVariableTest;
-UUR_INSTANTIATE_DEVICE_TEST_SUITE(urEnqueueDeviceGetGlobalVariableReadTest);
+using urEnqueueDeviceGetGlobalVariableReadWithParamTest =
+    uur::urGlobalVariableWithParamTest<uur::BoolTestParam>;
 
-TEST_P(urEnqueueDeviceGetGlobalVariableReadTest, Success) {
+UUR_DEVICE_TEST_SUITE_WITH_PARAM(
+    urEnqueueDeviceGetGlobalVariableReadWithParamTest,
+    testing::ValuesIn(uur::BoolTestParam::makeBoolParam("Blocking")),
+    uur::deviceTestWithParamPrinter<uur::BoolTestParam>);
+
+TEST_P(urEnqueueDeviceGetGlobalVariableReadWithParamTest, Success) {
+  bool is_blocking = getParam().value;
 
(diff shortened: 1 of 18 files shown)


test/conformance/enqueue/urEnqueueDeviceGlobalVariableRead.cpp
```diff
@@ -29,10 +35,17 @@ TEST_P(urEnqueueDeviceGetGlobalVariableReadTest, Success) {
       queue, program, global_var.name.c_str(), true, sizeof(global_var.value),
       0, &global_var.value, 0, nullptr, nullptr));
 
+  if (!is_blocking) {
+    ASSERT_SUCCESS(urQueueFinish(queue));
+  }
+
   // kernel should increment value
```
isaacault, line 42: I don't really understand what's happening in this test. Could you help by explaining

1. Shouldn't the `true` on line 35 be changed to `is_blocking`? 
2. What is the value of waiting on the queue to finish, and why only when non-blocking?

martygrant: 1. yes it should, thanks!
2. it's asynchronous when non-blocking right? the waiting is just to make sure the read is completed before we assert its value, let me know if I've misunderstood

test/conformance/enqueue/urEnqueueDeviceGlobalVariableWrite.cpp
```diff
@@ -5,6 +5,46 @@
...
+  // execute the kernel
+  ASSERT_SUCCESS(urEnqueueKernelLaunch(queue, kernel, n_dimensions,
+                                       &global_offset, &global_size, nullptr, 0,
+                                       nullptr, nullptr));
+  ASSERT_SUCCESS(urQueueFinish(queue));
+
+  // read global var back to host
+  int return_value = 0;
+  ASSERT_SUCCESS(urEnqueueDeviceGlobalVariableRead(
+      queue, program, global_var.name.c_str(), is_blocking, sizeof(return_value), 0,
```
isaacault, line 41: I think here `is_blocking` should be `true`.

test/conformance/enqueue/urEnqueueDeviceGlobalVariableRead.cpp
isaacault: That makes sense! Thanks

isaacault (approved):

PR #2681, opened 2025-02-10
In-order path for OpenCL command-buffers

EwanC:
After the [spec bump of cl_khr_command_buffer to 0.9.7](https://github.com/KhronosGroup/OpenCL-Docs/), in the OpenCL adapter we no longer need to worry about the in-order/out-of-order property of the internal queue used on command-command-buffer creation matching the queue used to enqueue the command-buffer.

We can therefore take advantage of the in-order flag passed on UR command-buffer creation to use an in-order queue for command-buffer creation, and omit using sync points.

DPC++ PR https://github.com/intel/llvm/pull/16938

### DIFF
--- a/source/adapters/opencl/command_buffer.cpp
+++ b/source/adapters/opencl/command_buffer.cpp
@@ -34,7 +34,15 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferCreateExp(
     ur_exp_command_buffer_handle_t *phCommandBuffer) {
 
   ur_queue_handle_t Queue = nullptr;
-  UR_RETURN_ON_FAILURE(urQueueCreate(hContext, hDevice, nullptr, &Queue));
+  ur_queue_properties_t QueueProperties = {UR_STRUCTURE_TYPE_QUEUE_PROPERTIES,
+                                           nullptr, 0};
+  const bool IsInOrder =
+      pCommandBufferDesc ? pCommandBufferDesc->isInOrder : false;
+  if (!IsInOrder) {
+    QueueProperties.flags = UR_QUEUE_FLAG_OUT_OF_ORDER_EXEC_MODE_ENABLE;
+  }
+  UR_RETURN_ON_FAILURE(
+      urQueueCreate(hContext, hDevice, &QueueProperties, &Queue));
 
   cl_context CLContext = cl_adapter::cast<cl_context>(hContext);
   cl_ext::clCreateCommandBufferKHR_fn clCreateCommandBufferKHR = nullptr;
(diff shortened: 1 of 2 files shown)


source/adapters/opencl/command_buffer.cpp
```diff
@@ -34,7 +34,17 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferCreateExp(
...
   ur_queue_handle_t Queue = nullptr;
-  UR_RETURN_ON_FAILURE(urQueueCreate(hContext, hDevice, nullptr, &Queue));
+  ur_queue_properties_t QueueProperties = {UR_STRUCTURE_TYPE_QUEUE_PROPERTIES,
+                                           nullptr, 0};
+
+  const bool IsInOrder =
+      pCommandBufferDesc ? pCommandBufferDesc->isInOrder : false;
+  if (IsInOrder) {
+    QueueProperties.flags = UR_QUEUE_FLAG_OUT_OF_ORDER_EXEC_MODE_ENABLE;
+  }
```
Bensuo, line 44: This reads like it's doing the opposite of what's intended, if we requested in order we are enabling out of order execution? Or is the OpenCL flag just confusingly named?

source/adapters/opencl/command_buffer.cpp
```diff
@@ -148,11 +158,18 @@ UR_APIEXPORT ur_result_t UR_APICALL urCommandBufferAppendKernelLaunchExp(
 
   cl_command_properties_khr *Properties =
       hCommandBuffer->IsUpdatable ? UpdateProperties : nullptr;
+
+  cl_sync_point_khr *RetSyncPoint =
+      hCommandBuffer->IsInOrder ? nullptr : pSyncPoint;
+  const cl_sync_point_khr *SyncPointWaitList =
+      hCommandBuffer->IsInOrder ? nullptr : pSyncPointWaitList;
+  uint32_t WaitListSize =
+      hCommandBuffer->IsInOrder ? 0 : numSyncPointsInWaitList;
```
Bensuo, line 167: Nit but would prefer to store `hCommandBuffer->IsInOrder` in a local variable rather than calling that 3 times. Also applies to other functions in this file that use the same pattern.

source/adapters/opencl/command_buffer.hpp
```diff
@@ -46,6 +46,8 @@ struct ur_exp_command_buffer_handle_t_ {
   /// Set to true if the kernel commands in the command-buffer can be updated,
   /// false otherwise
   bool IsUpdatable;
+  /// Set to true if the command-buffer was created from an in-order queue.
+  bool IsInOrder;
```
Bensuo, line 50:
```suggestion
  /// Set to true if the command-buffer was created with an in-order queue.
  bool IsInOrder;
```
Extreme nit but I think `with` describes the relationship better than `from`.

Bensuo (approved):
LGTM just a few minor comments.

source/adapters/opencl/command_buffer.cpp
EwanC: This was a bug, thanks for catching it

EwanC:
Closing in favor of https://github.com/intel/llvm/pull/17056 after the UR repo move

PR #2698, opened 2025-02-12
Don't dynamically allocate OpenCL adapter

RossBrunton:
There is always only one, so there's no point in allocating it via
`new`. This fixes an issue where calling `urReleaseAdapter` (or any
other UR function) in an `atexit` handler could be called after the
adapter is deleted.


RossBrunton:
We want to take a different approach.

PR #2699, opened 2025-02-12
[NATIVECPU][perf] less copies and allocations 

uwedolinsky:
This PR removes some unnecessary copies and allocations from kernel launches in the NativeCPU adapter.

Corresponding DPC++ PR: https://github.com/intel/llvm/pull/16990


source/adapters/native_cpu/enqueue.cpp
```diff
@@ -247,11 +261,7 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
   if (phEvent) {
     *phEvent = event;
   }
-  event->set_callback([hKernel, event]() {
-    event->tick_end();
-    // TODO: avoid calling clear() here.
-    hKernel->_localArgInfo.clear();
-  });
+  event->set_callback([event]() { event->tick_end(); });
```
hvdijk, line 264: Why was the `_localArgInfo.clear();` here? Was it a workaround for another issue? Is whatever reason it had for being there being addressed in a different way after your PR?

source/adapters/native_cpu/enqueue.cpp
```diff
@@ -199,39 +216,36 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
-                              kernel = *hKernel](size_t threadId) {
-              for (unsigned i = 0; i < groupsPerThread; i++) {
-                auto index = thread * groupsPerThread + i;
-                groups[index](threadId, kernel);
-              }
-            }));
+      if (groupsPerThread) {
+        for (unsigned thread = 0; thread < numParallelThreads; thread++) {
+          futures.emplace_back(
+              tp.schedule_task([groups, thread, groupsPerThread,
```
hvdijk, line 232: While we're cleaning this up, it doesn't make sense that each task gets its own independent copy of `groups`, it causes the lambda to be more expensive than it should be. We should be able to create `groups` once, have each lambda maintain a pointer to it, and add its deletion to the cleanup code.

source/adapters/native_cpu/threadpool.hpp
```diff
@@ -24,7 +23,7 @@
 
 namespace native_cpu {
 
-using worker_task_t = std::function<void(size_t)>;
+using worker_task_t = std::shared_ptr<std::packaged_task<void(size_t)>>;
```
hvdijk, line 26: There is no need to wrap this in a `shared_ptr`, we can get rid of that entirely.

source/adapters/native_cpu/threadpool.hpp
```diff
@@ -201,10 +200,10 @@ template <typename ThreadPoolT> class threadpool_interface {
 
   threadpool_interface() : threadpool() {}
 
-  auto schedule_task(worker_task_t &&task) {
-    auto workerTask = std::make_shared<std::packaged_task<void(size_t)>>(
-        [task](auto &&PH1) { return task(std::forward<decltype(PH1)>(PH1)); });
-    threadpool.schedule([=](size_t threadId) { (*workerTask)(threadId); });
+  template <class T> auto schedule_task(T &&task) {
+    auto workerTask =
+        std::make_shared<std::packaged_task<void(size_t)>>(std::move(task));
```
hvdijk, line 205: This should forward, like before. If `T` is inferred as an lvalue reference, it's important to not `move` because the caller isn't going to be expecting it to be changed.

source/adapters/native_cpu/enqueue.cpp
uwedolinsky: This is a known issue and this PR was not going to address this. We are looking into removing the `groups` variable altogether, but this would be for subsequent PR.

source/adapters/native_cpu/enqueue.cpp
```diff
@@ -63,6 +63,23 @@ static native_cpu::state getResizedState(const native_cpu::NDRDescT &ndr,
 }
 #endif
 
+class LaunchInfoLocalArgs {
```
hvdijk, line 66: Naming wise, it might be better to have something like `LaunchInfoGroup`, that's closer to what it is, I think. The fact that it calls `handleLocalArgs()` is less what this class is for and more just something that happens to be needed to be done for each group.

source/adapters/native_cpu/enqueue.cpp
```diff
@@ -63,6 +63,23 @@ static native_cpu::state getResizedState(const native_cpu::NDRDescT &ndr,
...
+  native_cpu::state state;
+  const size_t numParallelThreads;
+
+public:
+  LaunchInfoLocalArgs(const native_cpu::state &state_, unsigned g0_,
```
hvdijk, line 71: No need for the `_` suffixes, there is no conflict between the parameter and field names, I think we don't do that elsewhere either? For instance, in `kernel.hpp`:
```c++
  ur_kernel_handle_t_(ur_program_handle_t hProgram, const char *name,
                      nativecpu_task_t subhandler)
      : hProgram(hProgram), _name{name}, _subhandler{std::move(subhandler)} {}
```

source/adapters/native_cpu/enqueue.cpp
```diff
@@ -108,10 +125,10 @@ UR_APIEXPORT ur_result_t UR_APICALL urEnqueueKernelLaunch(
...
   hKernel->updateMemPool(numParallelThreads);
   std::vector<std::future<void>> futures;
-  std::vector<std::function<void(size_t, ur_kernel_handle_t_)>> groups;
-  auto numWG0 = ndr.GlobalSize[0] / ndr.LocalSize[0];
-  auto numWG1 = ndr.GlobalSize[1] / ndr.LocalSize[1];
-  auto numWG2 = ndr.GlobalSize[2] / ndr.LocalSize[2];
+  std::vector<LaunchInfoLocalArgs> groups;
+  const auto numWG0 = ndr.GlobalSize[0] / ndr.LocalSize[0];
+  const auto numWG1 = ndr.GlobalSize[1] / ndr.LocalSize[1];
+  const auto numWG2 = ndr.GlobalSize[2] / ndr.LocalSize[2];
```
hvdijk, line 131: Not directly related to this PR, but do we require global size to be a multiple of local size (for each dimension)? If so, an assert would be useful. If not, this looks suspicious. Either way, no need to change that as part of this PR unless you want to, if changes are needed we can do that in a new PR.

source/adapters/native_cpu/threadpool.hpp
uwedolinsky: Done.

source/adapters/native_cpu/threadpool.hpp
uwedolinsky: No, this is intended to always move and all argument tasks should come through as rvalues. I've added an `enable_if` to this function to ensure it cannot be called if `T` deduces to an lvalue reference.

martygrant:
## Information

## PR Migration

---

martygrant:

---

PR #2700, opened 2025-02-12
[NativeCPU] Fix kernel argument passing.

hvdijk:
We were reading the kernel arguments at kernel execution time, but kernel arguments are allowed to change between enqueuing and executing. Make sure to create a copy of kernel arguments ahead of time.

### DIFF
--- a/source/adapters/native_cpu/kernel.hpp
+++ b/source/adapters/native_cpu/kernel.hpp
@@ -35,18 +35,9 @@ struct ur_kernel_handle_t_ : RefCounted {
   ur_kernel_handle_t_(const ur_kernel_handle_t_ &other)
       : Args(other.Args), hProgram(other.hProgram), _name(other._name),
         _subhandler(other._subhandler), _localArgInfo(other._localArgInfo),
-        _localMemPool(other._localMemPool),
-        _localMemPoolSize(other._localMemPoolSize),
-        ReqdWGSize(other.ReqdWGSize) {
-    incrementReferenceCount();
-  }
+        ReqdWGSize(other.ReqdWGSize) {}
 
-  ~ur_kernel_handle_t_() {
-    if (decrementReferenceCount() == 0) {
-      free(_localMemPool);
-      Args.deallocate();
-    }
-  }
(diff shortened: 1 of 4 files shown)


hvdijk:
Corresponding DPC++ PR: https://github.com/intel/llvm/pull/16995

source/adapters/native_cpu/kernel.hpp
```diff
@@ -64,27 +55,62 @@ struct ur_kernel_handle_t_ : RefCounted {
...
+      for (size_t Index = 0; Index < Indices.size(); Index++) {
+        if (!Other.OwnsMem[Index]) {
+          continue;
+        }
+        addArg(Index, ParamSizes[Index], Indices[Index]);
+      }
+    }
+
+    arguments(arguments &&Other) : arguments() {
+      std::swap(Indices, Other.Indices);
```
coldav, line 72: why swap here, isn't Other dead after the std::move?

coldav (approved):

source/adapters/native_cpu/kernel.hpp
hvdijk: We need to make sure that when `Other`'s destructor runs, it doesn't free any memory that we are still using. The easiest way that I could see of doing that is by swapping.

coldav: ok

source/adapters/native_cpu/kernel.hpp
```diff
@@ -64,27 +55,62 @@ struct ur_kernel_handle_t_ : RefCounted {
     std::vector<bool> OwnsMem;
     static constexpr size_t MaxAlign = 16 * sizeof(double);
 
+    arguments() = default;
+
+    arguments(const arguments &Other)
+        : Indices(Other.Indices), ParamSizes(Other.ParamSizes),
+          OwnsMem(Other.OwnsMem.size()) {
```
hvdijk, line 62: In C++03 this would have guaranteed `OwnsMem` is a vector of `false`, but in C++11 it does not, this probably needs to be `OwnsMem(Other.OwnsMem.size(), false)` instead.

PR #2701, opened 2025-02-12
[L0] Align L0 adapters DEPENDENTLOADFLAG with other adapters

againull:
Currently add_ur_library function which is called by each adapter automatically adds /DEPENDENTLOADFLAG:0x2000 linker flag. L0 adapter adds its own /DEPENDENTLOADFLAG:0x800 flag on top of that which is excessive. More over it looks like now L0 adapters depend on umf library as well (in addition to l0 loader) which is not necessarily in system directory, so /DEPENDENTLOADFLAG:0x2000 seems more suitable.

### DIFF
--- a/source/adapters/level_zero/CMakeLists.txt
+++ b/source/adapters/level_zero/CMakeLists.txt
@@ -98,11 +98,6 @@ if(UR_BUILD_ADAPTER_L0)
         SOVERSION "${PROJECT_VERSION_MAJOR}"
     )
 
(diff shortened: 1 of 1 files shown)


pbalcer:
We have to identical PRs. See #2703. I'm going to close this one since the other one has a green CI run already.

lukaszstolarczuk:
yikes, sorry @againull I missed your PR.

PR #2731, opened 2025-03-07
In mirror script, ensure it uses remote sycl

RossBrunton:
Previously, it used the currently checked out commit, causing desyncs
if the intel llvm directory isn't updated.


aarongreig (approved):

RossBrunton:
Merging without CI passes since this should have no effect on the CI whatsoever.

PR #2737, opened 2025-03-20
add BMG perf nightly job

pbalcer:


pbalcer:
Successful run: https://github.com/oneapi-src/unified-runtime/actions/runs/13965769290/job/39101141228

.github/workflows/benchmarks-nightly.yml
```diff
@@ -35,3 +35,32 @@ jobs:
...
+      bench_script_params: '--save Baseline_BMG_L0'
+      sycl_config_params: ''
+      sycl_repo: 'intel/llvm'
+      sycl_commit: ''
+
+  nightly4:
+    # we need to wait until previous job is done so that the html report
+    # contains both runs
+    needs: nightly3
+    runner: 'DSS-L0-BMG'
```
lukaszstolarczuk, line 56: this runner is put in the wrong place - should be put as a param to the reusable job

.github/workflows/benchmarks-nightly.yml
```diff
@@ -35,3 +35,32 @@ jobs:
         sycl_config_params: ''
         sycl_repo: 'intel/llvm'
         sycl_commit: ''
+
+  nightly3:
+    name: Compute Benchmarks Nightly level-zero
```
lukaszstolarczuk, line 40: nit: perhaps add `BMG` into jobs' names

lukaszstolarczuk (changes requested):

.github/workflows/benchmarks-nightly.yml
pbalcer: I don't get it. It's part of the runner reusable job? right here:
https://github.com/oneapi-src/unified-runtime/pull/2737/files#diff-ed216951def5b0a1b03955d3600bda8c6e7e4738f5e9279d558fa43535634804R37-R40

lukaszstolarczuk: I mean here. It should be under `with:`
If used as it is now it won't be passed to the reusable job

.github/workflows/benchmarks-nightly.yml
pbalcer: done.

.github/workflows/benchmarks-nightly.yml
pbalcer: aaa, ok, thanks :D done.

lukaszstolarczuk (approved):

PR #2746, opened 2025-03-27
Fixed bug with zero depth `ur_rect_region_t` which must be checked before calling funcs

GermanAizek:
@RossBrunton, it looks like a common typo when writing functions.

Fixed functions:
- urEnqueueMemBufferWriteRect()
- urEnqueueMemBufferReadRect()

### DIFF
--- a/source/loader/layers/validation/ur_valddi.cpp
+++ b/source/loader/layers/validation/ur_valddi.cpp
@@ -4807,7 +4807,7 @@ __urdlllocal ur_result_t UR_APICALL urEnqueueMemBufferReadRect(
     if (phEventWaitList != NULL && numEventsInWaitList == 0)
       return UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST;
 
-    if (region.width == 0 || region.height == 0 || region.width == 0)
+    if (region.width == 0 || region.height == 0 || region.depth == 0)
       return UR_RESULT_ERROR_INVALID_SIZE;
 
     if (bufferRowPitch != 0 && bufferRowPitch < region.width)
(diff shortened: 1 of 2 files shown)


source/loader/layers/validation/ur_valddi.cpp
```diff
@@ -4807,7 +4807,7 @@ __urdlllocal ur_result_t UR_APICALL urEnqueueMemBufferReadRect(
     if (phEventWaitList != NULL && numEventsInWaitList == 0)
       return UR_RESULT_ERROR_INVALID_EVENT_WAIT_LIST;
 
-    if (region.width == 0 || region.height == 0 || region.width == 0)
+    if (region.width == 0 || region.height == 0 || region.depth == 0)
```
aarongreig, line 4810: thanks for the fix, note this is a generated file so this change will need to be made [here](https://github.com/oneapi-src/unified-runtime/blob/9e7fe483c3eb9e61fee89f79e8cdd2aeaf950133/scripts/core/enqueue.yml#L331), and you'll need to then run the `generate` target to propagate the update to the headers + validation layer

GermanAizek: @aarongreig, Im fixed here: https://github.com/oneapi-src/unified-runtime/pull/2746/commits/5a5811a4f5f9b3c85bc8ff18c7e02e4f6d98e567

aarongreig: Cool, thanks. Sorry I should have also reiterated the [bot comment](https://github.com/oneapi-src/unified-runtime/pull/2746#issuecomment-2757985661): this repo isn't used for development any more, it's purely a mirror of https://github.com/intel/llvm/tree/sycl/unified-runtime so you'll need to migrate this PR to that repo

